PR #1, opened 2015-07-13
[ADD] mail_restrict_follower_selection

hbrunn:
# Restrict follower selection

This module was written to allow you to restrict the selection of possible followers. For example, if you use the social ERP functions only internally, it makes sense to filter possible followers for being employees. Otherwise, you'll get a quite crowded list of partners to choose from.
# Configuration

To configure this module, you need to go to `System parameters` and adjust `mail_restrict_follower_selection.domain` as you see fit. This restricts followers globally, if you want to restrict only the followers for a certain record type (or have different restrictions for different record types), create a parameter `mail_restrict_follower_selection.domain.$your_model`.

Note: This module won't change existing followers!


dreispt:
Do you consider also adding a filter to the `mail.compose.message` wizard?
This can be is opened when adding a new chatter message, and allows to pick Partners that will also be added as Followers.

I have a use case where In would like like to limit possible contacts to only those that have an email address.

hbrunn:
sounds good. does it make sense to have this in a different parameter or use the same one?

dreispt:
That's a good question. I say it's far from overwhelming to have two parameters, and it could be nice to have that flexibility.

mail_restrict_follower_selection/README.rst
```diff
@@ -0,0 +1,44 @@
...
+    :alt: License: AGPL-3
+Restrict follower selection
+===========================
+
+This module was written to allow you to restrict the selection of possible followers. For example, if you use the social ERP functions only internally, it makes sense to filter possible followers for being employees. Otherwise, you'll get a quite crowded list of partners to choose from.
+
+Configuration
+=============
+
+To configure this module, you need to go to `System parameters` and adjust `mail_restrict_follower_selection.domain` as you see fit. This restricts followers globally, if you want to restrict only the followers for a certain record type (or have different restrictions for different record types), create a parameter `mail_restrict_follower_selection.domain.$your_model`.
```
pedrobaeza, line 11: Please put an example of a valid domain.

mail_restrict_follower_selection/__init__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
```
pedrobaeza: You can remove this line or maybe adjust to new brand

mail_restrict_follower_selection/models/__init__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
```
pedrobaeza: The same

mail_restrict_follower_selection/README.rst
dreispt, line 11: @hbrunn The default parameter created has a key `mail_restrict_follower_selection.domain`. Does this work as a global value?

mail_restrict_follower_selection/models/mail_wizard_invite.py
```diff
@@ -0,0 +1,49 @@
...
+##############################################################################
+from lxml import etree
+from openerp import models, api
+
+
+class MailWizardInvite(models.TransientModel):
+    _inherit = 'mail.wizard.invite'
+
+    @api.model
+    def mail_restrict_follower_selection_get_domain(self):
```
dreispt: Add underscore prefix to make it private?

mail_restrict_follower_selection/README.rst
```diff
@@ -0,0 +1,44 @@
...
+
+Note: This module won't change existing followers!
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/crm/issues>`_.
```
adrienpeiffer: Replace `crm` by `social`

mail_restrict_follower_selection/README.rst
```diff
@@ -0,0 +1,44 @@
...
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/crm/issues/new?body=module:%20mail_restrict_follower_selection%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
adrienpeiffer: Same here

mail_restrict_follower_selection/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Restrict follower selection",
+    "version": "1.0",
```
adrienpeiffer: 8.0.1.0.0

mail_restrict_follower_selection/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+    "author": "Therp BV",
```
adrienpeiffer: Add OCA as author

mail_restrict_follower_selection/README.rst
hbrunn, line 11: @dreispt yes, that's what I want to express in "...adjust `mail_restrict_follower_selection.domain` as you see fit. This restricts followers globally" - unclear?

hbrunn:
@dreispt @adrienpeiffer @pedrobaeza thanks for your reviews, all suggestions taken & commits squashed

mail_restrict_follower_selection/README.rst
pedrobaeza, line 11: For me it's OK

pedrobaeza:
:+1:

rafaelbn:
Hi @hbrunn , tested in runbot.

By default runbot has:
- mail_restrict_follower_selection.domain      [('employee', '=', True)]
- runbot has 25 employee

When I try to add a follower to any object (employee, sale order, task, ...) I don't have any choice. Is this default parameter correct? It looks is not working.

dreispt:
Well, the module does not depend on `hr`, so having a default setting based on employees looks awkward...

hbrunn:
@rafaelbn @dreispt it's a field from base: https://github.com/OCA/OCB/blob/8.0/openerp/addons/base/res/res_partner.py#L261 - and you have to set that of course to find anything.

But considering the confusion, I changed the default to customer, that should give everyone some results.

pedrobaeza:
Employee option is not that bad (to avoid leaks to outside) if it's explained, because people don't know about that field in general.

hbrunn:
When I looked into that, I noticed that the standard UI doesn't allow to set this field, so I don't think it's a good choice for a default. And it's just a default anyways, I don't really think end users install this module, but integrators use it to set up whatever restrictions they see fit.

pedrobaeza:
Yeah, that's right

dreispt:
:+1: !

pedrobaeza:
@rafaelbn, give us your approval also and I'll merge

yajo:
Maybe adding to default & instructions `('opt_out', '=', False)` would be a good idea.

yajo:
Also add to instructions that `("user_ids","!=",False)` will enable only actual users in your database.

yajo:
I noticed that when you install the first module after you installed `mail_restrict_follower_selection` it leads the user to this strange page that shows the default value for its setting. (In runbot the url is http://3115521-1-2053bb.runbot1.odoo-community.org/web?db=3115521-1-2053bb-all#id=5&view_type=form&model=ir.config_parameter&action=110&active_id=268):

captura de pantalla de 2015-09-21 19-25-39

It's not something critical though.

rafaelbn:
Hi @hbrunn ,

With Sales, Project, HR installed I tried to restrict only for employees, globaly, but I can't. Maybe I'm missing something you say here https://github.com/OCA/social/pull/1#issuecomment-137678952

My param is this one again (with HR installed):

```
Key
    mail_restrict_follower_selection.domain
Value
    [('employee', '=', True)] 
```

I expect only employee in follower bit it doesn't works.

I have also tried to play with more parameters with 

`mail_restrict_follower_selection.domain.$your_model`

in project.task, sale.orders, etc.. but it doesn't works. For sure I don't understand something.

Thanks

hbrunn:
as written in https://github.com/OCA/social/pull/1#issuecomment-137678952, the employee flag does nothing. If you want to use it, you need to set it yourself

yajo:
What do you (and others) think about https://github.com/OCA/social/pull/1#issuecomment-142038065?

rafaelbn:
Hi @hbrunn could you check comments of @Yajo . They are logic and I think could be considered. After this we can finally merge it! Thanks!

hbrunn:
I'm happy to merge any PRs into this branch, but extra work on my side for a default is expected too much

rafaelbn:
:+1: on my site!

yajo:
OK then let's merge and I'll add it later. :+1:

rafaelbn:
Hi @Yajo please take care of pending PR about this one.

PR #2, opened 2015-08-03
[ADD] mail_compose_select_lang

pedrobaeza:
# Select language in mail compose window

This module allows to select the language for the mail content directly in
the mail compose window.

It also translates attachments that can be linked to the email template.
# Usage

By default, mail compose window will use corresponding language for showing the
contents, but you will be able to select another language from the dropdown
box "Force language" in the bottom right part.


pedrobaeza:
@rafaelbn, reviews are welcome

mail_compose_select_lang/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- encoding: utf-8 -*-
```
dreispt: s/encoding/coding/g

mail_compose_select_lang/__openerp__.py
```diff
@@ -0,0 +1,37 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Select language in mail compose window',
+    'version': '1.0',
+    'category': 'Marketing',
+    'author': 'Serv. Tecnolog. Avanzados - Pedro M. Baeza, '
+              'Antiun Ingeniería S.L.',
```
dreispt: Add OCA

mail_compose_select_lang/models/report.py
```diff
@@ -0,0 +1,18 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+# Copyright 2015 Pedro M. Baeza
+# For license notices, see __openerp__.py file in root directory
```
dreispt: Shouldn't you at least add the it's "Licensed under AGPL v3"?

dreispt:
Some nitpicks, but it's a :+1:

mail_compose_select_lang/models/report.py
```diff
@@ -0,0 +1,18 @@
...
+##############################################################################
+from openerp import models, api
+
+
+class Report(models.Model):
+    _inherit = 'report'
+
+    @api.model
+    def translate_doc(self, doc_id, model, lang_field, template, values):
```
nhomar, line 13: Did yopu test how this affect to the global translate_doc, for this feature are you overwritting entirely the qweb translate system.

A test for that should be good (copying original ones).

pedrobaeza:
Changes done

mail_compose_select_lang/models/report.py
pedrobaeza, line 13: As the change is only done over a context parameter, and this parameter is only affected when another one parameter is used, it's very unlinkely that there's any side effect on this code. A test would be good, but this one in special is complicated to be done and I don't have the time for now.

rafaelbn:
@pedrobaeza 
- I created mail template with to languages (English and Spanish)
- I set one partner to language English and one partner to language Spanish
- I sent a Partner Mass Maling to this 2 partners plus another one without language field set
- I check in setting - Technical - Email => [Messages & Emails]
- All e-mails are sent in the same language, so it looks it's not working

pedrobaeza:
@rafaelbn, this is not intended for mass mailing, but for mails sended via the compose mail (as the README states). For example, when you click on "Send by mail" in the sales order form view. Another hint for this is that mass mailing module is not a dependency for this module.

pedrobaeza:
@rafaelbn, anyways, maybe your problem is that the template you have made is not multi-lingual, so forces the language doesn't do anything at all because of this.

mail_compose_select_lang/models/report.py
nhomar, line 13: I am sorry dude, but without test it is not feasible for me, you will face horizontal dependencies issues here.

Look a simple:

``` shell
 $: rgrep translate_doc . --include=*.py
```

Returns this method overwritten once.

``` shell
./odoo/addons/report/models/abstract_report.py:62:        # Used in template translation (see translate_doc method from report model)
./odoo/addons/report/models/report.py:93:    def translate_doc(self, cr, uid, doc_id, model, lang_field, template, values, context=None):
./odoo/addons/report/models/report.py:97:        <t t-raw="translate_doc(doc_id, doc_model, 'partner_id.lang', account.report_invoice_document')"/>
./odoo/addons/report/models/report.py:134:        def translate_doc(doc_id, model, lang_field, template):
./odoo/addons/report/models/report.py:135:            return self.translate_doc(cr, uid, doc_id, model, lang_field, template, values, context=context)
./odoo/addons/report/models/report.py:145:            translate_doc=translate_doc,
./odoo/addons/website_report/report.py:29:    def translate_doc(self, cr, uid, doc_id, model, lang_field, template, values, context=None):
./odoo/addons/website_report/report.py:41:        return super(Report, self).translate_doc(cr, uid, doc_id, model, lang_field, template, values, context=context)
```

And you do not depend of any of them.

We have 3 options here:
1.- You can set in the meta data for the email the  lang with templates and not use this method.
2.- Make this change in an extra module which will be a technical one to extend this capability and depend of website_report.
3. Develop the post mortem test with python and ensure all work as expected with the normal tests on all reports.

On my experience touch this global templating method is too dangerous and bring a lot of unexpected results (I faced some problems already).

Sorry for teh inconvenience here.

rafaelbn:
@pedrobaeza I'm talking about mail.compose.message in sales where the field to select language is shown. Any user will become confused. I'm not talking about mail.mass_mailing.

pedrobaeza:
That's working as expected. I have just tried on runbot:

seleccion_004

nhomar:
> > @rafaelbn, anyways, maybe your problem is that the template you have made is not multi-lingual, so forces the language doesn't do anything at all because of this.

This is the kind of problems with horizontal dependency, you are overwritting a global method which is used in 100% of modules if the behavior will be specific, you should overwrite the variables on the template itself not the global method to expect a different result.

The main email template mentioned by @rafaelbn is one of the places where your logic will bring issues, that's a multilingual template already which take the lang from the partner.

@rafaelbn did you set in your template the lang? can you share your template configuration?

pedrobaeza:
Talking directly with @rafaelbn, we have seen that mass mailing module has itself some bugs that prevents the sending of mails in different languages according partner's lang, so it's not a problem of this module.

pedrobaeza:
@nhomar, this kind of changes is something you do everyday, overwriting write, create and other standard methods. There's no need for additional tests, because if I break something of the standard modules, standard tests will be broken, which is not the case.

nhomar:
I am absolutly sorry, but your last statement in my opinion is wrong, I am not talking about the write of an specific model, I am talking about the translation of 100% of elements rendered with this method.

Please rethink. (the feature is cool) but too dangeraous without a proper test.

pedrobaeza:
I insist, the change is very weak, so no side effect, but I don't know how to test what you want. Let's other tell their opinion, or a contribution for this test is also welcome.

nhomar:
@pedrobaeza that's a good question:

Takle a look with an rgrep in all odoo:

``` shell
[ nhomar@MBPNhomar ]-[ ~/Trabajo/odoo80 ]
 (8.0) $: rgrep translate_doc . --include=*.xml
./addons/account/views/report_invoice.xml:153:            <t t-raw="translate_doc(doc_id, doc_model, 'partner_id.lang', 'account.report_invoice_document')"/>
./addons/account/views/report_overdue.xml:113:            <t t-raw="translate_doc(doc_id, doc_model, 'lang', 'account.report_overdue_document')"/>
./addons/purchase/views/report_purchaseorder.xml:125:            <t t-raw="translate_doc(doc_id, doc_model, 'partner_id.lang', 'purchase.report_purchaseorder_document')"/>
./addons/purchase/views/report_purchasequotation.xml:67:            <t t-raw="translate_doc(doc_id, doc_model, 'partner_id.lang', 'purchase.report_purchasequotation_document')"/>
./addons/report_intrastat/views/report_intrastatinvoice.xml:144:            <t t-raw="translate_doc(doc_id, doc_model, 'partner_id.lang', 'report_intrastat.report_intrastatinvoice_document')"/>
./addons/sale/views/report_saleorder.xml:132:            <t t-raw="translate_doc(doc_id, doc_model, 'partner_id.lang', 'sale.report_saleorder_document')"/>
```

You can odoo develop a main header management which use such method, then if you add a test that send your email and print a report it will be ok.

Example of test to send emails (generic not using your wizard). 

odoo80/addons/mail/tests

Examples of test printing reports.

odoo80/addons/report/tests

As you can see in the case of reports they print all reports in the system one time all is installed:

``` python
@openerp.tests.common.at_install(False)
@openerp.tests.common.post_install(True)
```

But as you do not depend of module report and website_report you will not be able to use your method on this tests (at least being sure) and the method itself is not being tested for the feature it self. then you should add to the test the context to test your case.

I hope it helps.-

pedrobaeza:
We have 2 problems for what you ask, @nhomar:
- In Travis/Runbot, we cannot test mail, as there's no mail environment set.
- My time/tests skills are not big for this task.

So for now, I'm going to leave it as is, and if anyone wants to contribute, it's very welcome.

nhomar:
I like the feature.

One time we finished the oca/rma part I can ask to do such job, let's have it pause until we finish that, if we do not send a notice about "We start work on this" then somebody else can take the job and help us.

Regards,.

nhomar:
But in terms of functionality I give a +1

rafaelbn:
@pedrobaeza 

> "Talking directly with @rafaelbn, we have seen that mass mailing module has itself some bugs that prevents the sending of mails in different languages according partner's lang, so it's not a problem of this module"

We opened 2 issues:
- [8.0] mail templates in marketing never translate #7896 
  https://github.com/odoo/odoo/issues/7896
-  [8.0] When i edit email template and try change the language "500: Internal Server Error" #7898 
  https://github.com/odoo/odoo/issues/7898

Not sure if Odoo will answer.

rafaelbn:
As the problem is from Odoo this is a :+1:

dreispt:
@rafaelbn I can also make a PR to [OCB](https://github.com/OCA/OCB). You will surely get an answer there.

rafaelbn:
@dreispt , OK, Thank you!

rafaelbn:
@dreispt ! Something is happening :-)
-    [8.0] mail templates in marketing never translate #7896
  odoo/odoo#7896

"Stage changed
    • Stage: Technical & Platform → Functional
Denis Ledoux (dle) updated document • about 9 hours ago • like"
-    [8.0] When i edit email template and try change the language "500: Internal Server Error" #7898
  odoo/odoo#7898

"Stage changed
    • Stage: Technical & Platform → Bug Fixes
Denis Ledoux (dle) updated document • about 10 hours ago • like"

rafaelbn:
## @dreispt issue https://github.com/odoo/odoo/issues/7898 has been solved by Denis Ledoux.

This issue has been solved in Odoo 8.0 from the below revision:

## https://github.com/odoo/odoo/commit/a1328c9054c1907ce4996811bb7befc6ca031a7f

Still waiting for https://github.com/odoo/odoo/issues/7896

rafaelbn:
Hi all, about https://github.com/odoo/odoo/issues/7896 Odoo answers:

> Hello Rafael,
> We transmit your query to our usability team and they will look into your query.
> As of now we can't do anything else to change this behaviour, only we can send your request to usability team for improvement that's all we can do due to our stable policy.
> Thank you.
> Bindiya Patel (bip)

So resume:
- If you send a quote, sale order or an invoice, it will be sent in the language of the partners or if you have this module you can force language.
- If you send an e-mail to a Partner through the chatter in the partner's form it will send an e-mail with you signature and no matter if you choose a template translated, it will always send the e-mail in the language of the user an not of the partner
- If you select multiple partners and go to "more" -> mass mailing selected partners you will also send the e-mails in the language of the user an not of the parter
- Basically (big error of Odoo in my opinion), no matter you translate e-mails templates!! the will be sent in user's languages and not in partner's languages!

Surely this module on another one must fix this...

rafaelbn:
I give :+1:  as what I expected is not supported by core. So we must implement an additional module.

We have 3 :+1: from @nhomar @dreispt and me.

rafaelbn:
We have 3 :+1: from @nhomar @dreispt and me.

pedrobaeza:
OK, merging.

PR #12, opened 2015-09-14
Add module mail_forward

yajo:
Move mail_forward module from [odoo-grupoesoc-addons](https://github.com/grupoesoc/odoo-grupoesoc-addons).

This module allows the user to forward a message to another object of the database and/or to any contact. See README for details.


pedrobaeza:
@rafaelbn, you'll be interested in this one.

mail_forward/wizard/mail_forward.py
```diff
@@ -0,0 +1,120 @@
...
+    """Allow forwarding a message.
+
+    It duplicates the message and optionally attaches it to another object
+    of the database and sends it to another recipients than the original one.
+    """
+
+    _name = "mail_forward.compose.message"
+    _inherits = {"mail.compose.message": "original_wizard_id"}
+
+    _models = [
```
dreispt: Can we have this in a setup table instead? It would make it easier to customize.

yajo: I'll try.

pedrobaeza: Or maybe use res.request.link table or another one created for this purpose.

yajo: What does it do? Actually I was just adding a boolean to `ir.model` called `mail_forward_target`. Seems the easiest way.

pedrobaeza: That table is a list of models allowed to be referenced by default in reference fields.

dreispt: I agree with Pedro - that was the solution used in the core for similar needs, such as [`res.request`](https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=res.request) for "Reference" type fields.

yajo:
If you agree with the current implementation, I'll start with the test suite.

pedrobaeza:
I think it's better to use a new table instead adding a flag on existing one, because you can want to add a new forward target, but not having on referenceable models.

yajo:
But then I'd have to hard-code them again. I thought that was the motivation of the change...

pedrobaeza:
They're imported as data, and you can add any record you want via XML or by UI (if views are provided). It's not the same "hard-coding".

yajo:
But other modules will not add themselves to that table. The nice thing about `res.request.link` is that for example you install `sale` and it takes care of adding the correct models to that table.

That could not be done the way you say unless developers create `foo_model` and `foo_model_mail_forward` as auto-install module when `mail_forward` and `foo_model` are installed.

pedrobaeza:
But with your flag, you force to set that flag programatically or manually if you want to forward.

yajo:
No. [`default=True`](https://github.com/OCA/social/pull/12/files#diff-39017bdf8ca51ef1f2027ebe237a81ecR12).

pedrobaeza:
OK, @dreispt, what do you think?

rafaelbn:
Rebuliding for testing! @pedrobaeza @dreispt @Yajo any technical modification will be applied? Thanks

rafaelbn:
Tested in runbot, could you add subject and To when forwarding?
´´´
----------Forwarded message----------
From: Administrator
Date: Mon Oct 05 2015 19:39:26 GMT+0200
´´´

> To:
> Subject:

rafaelbn:
One doubt:

yajo:
> any technical modification will be applied?

I have to write some tests.

> Tested in runbot, could you and subject and To when forwarding?

I don't understand you, sorry.

> One doubt [...]

When you forward the e-mail, you need to select a target. It can be:
- One or more contacts.
- One database object.
- Both.

So when you select a contact, the database object is not required anymore, and vice versa.

yajo:
Tests written. I hope Coveralls is happy now.

mail_forward/README.rst
```diff
@@ -0,0 +1,101 @@
...
+To do it:
+
+* Go to *Settings > Technical > Database Structure > Referenceable Models*.
+
+* Any model there with *Mail forward target* enabled will appear in the list of
+  models that can get forwarded messages.
+
+  * If you want to *remove* a model from the list, it's usually better to just
+    *disable* its check box instead of deleting it, because that record might
+    be used by other modules.
```
dreispt, line 35: This design looks perfect to me.

mail_forward/models/compose_message.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        result = self.original_wizard_id.default_get(fields)
+
+        if "subject" in result and "default_subject" in self.env.context:
+            result["subject"] = self.env.context["default_subject"]
+
+        return result
+
+    @api.model
+    def models(self):
```
dreispt: Choose a better name here. Maybe `_get_model_selection`

mail_forward/views/compose_message.xml
```diff
@@ -0,0 +1,64 @@
...
+
+<record id="compose_form" model="ir.ui.view">
+    <field name="name">Forward Email Form</field>
+    <field name="model">mail_forward.compose_message</field>
+    <field name="inherit_id" ref="mail.email_compose_message_wizard_form"/>
+    <field name="arch" type="xml">
+        <data>
+            <!-- It's required to fill either `partner_ids` or
+                 `destination_object_id` -->
+            <xpath expr="//field[@name='partner_ids']" position="attributes">
```
dreispt, line 17: You can avoid an xpath here, and use a simpler `<field name="partner_ids" position="attributes">`

dreispt:
Congratulations, looks very good.

mail_forward/models/compose_message.py
yajo: Done.

mail_forward/views/compose_message.xml
yajo, line 17: I did not avoid xpath because I prefer it. It's a standard, it's more concise, better documented and it always works.

dreispt, line 17: It's unnecessarily complicated, and Odoo provides a convenient shortcut. Odoo wise, it's also standard, well documented and always works.

yajo, line 17: It does not seem so complicated to me. XPath can be used in any case, while the other method only for some. That's why I prefer XPath: one tool to any case. Sometimes there are more than one way to do the right thing. I prefer mine this time. If you don't like it, feel free to PR after merging.

dreispt, line 17: We disagree, but not a blocking issue either...

dreispt:
:+1:

mail_forward/views/compose_message.xml
pedrobaeza, line 17: I also avoid xpath whenever I can. Yes, you use the same tag (xpath), but the variety of forms that the xpath can take makes it even more difficult to handle. For example, novice users tend to switch these xpath expressions without knowing exactly the difference:

```
<xpath expr="field[@name='partner_ids']"
<xpath expr="/field[@name='partner_ids']"
<xpath expr="//field[@name='partner_ids']"
```

yajo, line 17: I always love standards. [XPath is one, backed by W3C](http://www.w3.org/TR/xpath/).

I also love using the right tool for the job. Quoting from that link before:

>  The primary purpose of XPath is to address parts of an XML document.

Which means that probably any developer that has needed to select a XML element in any app (not only Odoo) has met already XPath, which makes them less newbies in Odoo. I did not need to read any docs when I first saw this tag in an Odoo view, its purpose is obvious, but I needed to do it when using the shortcuts you say.

Any Odoo programmer will have to learn XPath sooner or later, since QWeb inheritance works only with it.

But the most important thing is:

hakuna-matata

rafaelbn:
Could you add this https://github.com/OCA/social/pull/12#issuecomment-145609508 ?

yajo:
Done.

hbrunn:
:+1:

PR #14, opened 2015-09-14
Do not ignore *.pot files.

yajo:


pedrobaeza:
Of course! I fast-track it because it's an obvious mistake

yvaucher:
:+1:

PR #17, opened 2015-09-29
[ADD] New addon mass_mailing_exclude_opt_out

yajo:
Installing this addon will fix https://github.com/odoo/odoo/issues/8463.

It enables by default the "Only available for mass mailing" filter. Less error prone.

CC @rafaelbn @janetrp.


pedrobaeza:
Why not a PR to Odoo/OCB?

pedrobaeza:
Not needed now

PR #18, opened 2015-10-05
[ADD] mail_mandrill addon

antespi:
This module logs Mandrill email messages events.
# Configuration

To configure this module, you need to:
- Define an STMP server in Odoo at Settings > Technical > Email > Outgoing Mail Servers
  using Mandrill credentials from Mandrill settings panel (Settings > SMTP & API Credentials)
- Define a webhook in Mandrill settings panel (Settings > Webhooks) with
  several triggers (Message Is Sent, Message Is Delayed, ...) and 'Post to URL'
  like `https://your_odoodomain.com/mandrill/event`
- Copy Webhook key and paste in your Odoo configuration file, in `[options]`
  section, using `mandrill_webhook_key` variable. This is optional, but recommended because it is
  used to validate Mandrill POST requests
# Usage

When any email message is sent via Mandrill SMTP server, Odoo will add
some metadata to email (Odoo DB, Odoo Model and Odoo Model record ID) using an
special SMTP header (X-MC-Metadata). More info at [Mandrill doc: Using Custom Message Metadata](https://mandrill.zendesk.com/hc/en-us/articles/205582417-Using-Custom-Message-Metadata)

Then when an event occurs related with that message (sent, open, click, bounce, ...)
Mandrill will trigger webhook configured and Odoo will log the message and the event.

In `Setting > Technical > Email > Mandrill emails` you can see all messages sent
using Mandrill. When clicking in one of them you'll see message details and events
related with it

In `Setting > Technical > Email > Mandrill events` you can see all Mandrill events
received


antespi:
Tests pending

antespi:
Sorry @hbrunn, what do you expect to finish in this addon?

hbrunn:
a green travis and coverage >= 0

antespi:
OK, travis is green right now, promising :smile: 
Now is time of testing

rafaelbn:
We are using this module in all our customers, in production environment :+1: tested!
Please remove label work in provess add new review, thanks!

mail_mandrill/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+"id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
+"access_mail_mandrill_message_group_system","mail_mandrill_message group_system","model_mail_mandrill_message","base.group_system",1,1,1,1
+"access_mail_mandrill_event_group_system","mail_mandrill_event group_system","model_mail_mandrill_event","base.group_system",1,1,1,1
```
houssine78, line 5: it would be better to have a dedicated group for this instead of the admin group

hbrunn:
@antespi could you add tests? then it's fit for merging I think

antespi:
@hbrunn Test added and some bugfixes btw :smile:

antespi:
But coveralls is down... coveralls is a misunderstood guy ;)

hbrunn:
@antespi cool! do you think there's something to do with the tests to increase coverage? If there are parts that can't be tested sensibly (need a service account, only for debugging, things like that), use `# pragma: no cover` on the appropriate line(s) or branch(es)

hbrunn:
but on the other hand, coveralls also has the habit to lie... :+1:

antespi:
Yes, that's right. We do not need a service account, but I only added unit test because test all flow is complex

antespi:
I did not test:
- Controller: https://github.com/OCA/social/pull/18/files#diff-a64bc9d4e8a5d7bacc136f6fcb28af79R20
- Send overwrite: https://github.com/OCA/social/pull/18/files#diff-e89dd92ee9127b67461dae6f961eb9a0R33
- Process reject event: https://github.com/OCA/social/pull/18/files#diff-494af699617ae3d2ed2095dabc1e8d6dR166
- Main event process: https://github.com/OCA/social/pull/18/files#diff-3875d1f959ae5a4d2d68dbeef15d52f9R90

How can I add those pragmas

hbrunn:
the pragmas can be added just in your code:

```
def myfunction:  # pragma: no cover  <- would ignore the whole function
    code
    code
    if something:
        code
        code
    else:  # pragma: no cover <- would ignore the else branch
        code
```

but those are not meant to exclude a whole bunch of code just because it's complicated to write the tests. Then better don't add any pragmas, and try to think of a way to have the code in question tested anyways

antespi:
Well, I don't know how to test a controller, in this case the controller receive a mandrill event, try to figure out for which database it is. Controller reads HTTP Request.

Send method, sents an email, but important code is the one that adds metadata headers  (already tested) 

Process reject event is a rare event, I don't have any real sample (other tests have real ones)

I can add a new test for testing 'Main event process', give me a try.

hbrunn:
@antespi for testing a controller, just instantiate it and call it. or probably the core or server-tools or website projects have invented smarter ways to do this. for the rest, just invent whatever data is necessary.
the tests are not here to give us all a bad time, they are there to avoid having a bad time later

mail_mandrill/security/ir.model.access.csv
antespi, line 5: This addon is a base addon for getting events from Mandrill. It only depends on mail addon and this is intentional.
There will be more addons that depends on this one and marketing (for instance) that creates security groups and menus in order to give other users access to this info.
BTW I've added read access to base.group_user (Employee), as they have access for reading partners email addresses.

antespi:
Two more tests added and read access to employees

antespi:
@hbrunn I think this addon is ready, I hope anyone review today (OCAReviewDay)

rafaelbn:
Awesome! ;-) :+1:

mail_mandrill/controllers/main.py
```diff
@@ -0,0 +1,130 @@
...
+        # Check database selected by mandrill event
+        if not db:
+            _logger.info('%s: No DB selected', info)
+            return False
+        try:
+            registry = openerp.registry(db)
+        except OperationalError:
+            _logger.info("%s: Selected BD '%s' not found", info, db)
+            return False
+        except:
```
lepistone, line 78: Do we need a catch-all here?

mail_mandrill/controllers/main.py
```diff
@@ -0,0 +1,130 @@
...
+        and setting this Post to URL:
+            https://your_odoodomain.com/mandrill/event
+        """
+        if not self._mandrill_validation(**kw):
+            _logger.info('Validation error, ignoring this request')
+            return 'NO_AUTH'
+        events = []
+        try:
+            events = json.loads(kw.get('mandrill_events', '[]'))
+        except:
```
lepistone, line 118: Here also, maybe we could catch something more specific?

lepistone:
Two minor remarks on exceptions, but this is great! :+1:

mail_mandrill/controllers/main.py
antespi, line 118: Here mandrill must include a `mandrill_events` parameter JSON compliance, so if any exception, no events could be processed. I that case events will be [] and `NO_EVENTS` appears.

What exception do you expect to be catched here?

mail_mandrill/controllers/main.py
antespi, line 78: I don't know if `openerp.registry(db)` could raise more exceptions, this is a high level call. Lots of calls to postgresql libs are done inside, so more exceptions could be raised. In that case, we log connection error and `ALL_EVENTS_FAILED` or `SOME_EVENTS_FAILED` appears.

lepistone, line 78: In cases where we don't know I have a preference for a crash than catching the unknown. But I see your point, I'm not blocking your PR for that. Thanks! (same on the other comment).

rafaelbn:
@hbrunn are you agree merging?

mail_mandrill/README.rst
```diff
@@ -0,0 +1,110 @@
...
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/social/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/social/issues/new?body=module:%20mail_mandrill%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+License
+=======
```
max3903: No need for this section. You have the AGPL badge on top of the README.

antespi: Fixed, thanks

max3903:
:+1:

pedrobaeza:
:+1:

PR #22, opened 2015-10-15
[ADD] mail_restrict_auto_follower

pedrobaeza:
# 
# Restrict automatic follower subscription

This module restricts the followers that are added automatically to a record.
For example, if you use the social ERP functions only internally, it makes
sense to filter possible followers for being employees. Otherwise, you'll get
a quite crowded list of partners to choose from.
# Configuration

To configure this module, you need to go to `System parameters` and adjust
`mail_restrict_auto_follower.domain` with the domain to apply to the partners
involved.

This will restrict followers globally. If you want to restrict only the
followers for a certain record type (or have different restrictions for
different record types), create a parameter
`mail_restrict_auto_follower.domain.$your_model`.

As an example, you may use `[('user_ids', '!=', False)]` to allow only
partners that are also users to be added automatically as followers. This is
also the default.

Note: This module won't change existing followers!
# Known issues / Roadmap
- Due to a problem in current API with inheritance, maybe some models ignore
  the restriction. In that cases, a glue module is needed that redeclare the
  the model this way:
  
  ``` python
  class Model(models.Model):
      _name = 'model'
      _inherit = ['mail.thread', 'model']
  ```
- For v9, this should be merged with `mail_restrict_follower_selection`
  module to have one module for both things that can be called
  `mail_restrict_follower`.


pedrobaeza:
@hbrunn, for your information

pedrobaeza:
This is red because a test fails due to https://github.com/odoo/odoo/issues/9084

pedrobaeza:
Tests solved thanks to the hint provided by Raphael Collet.

mail_restrict_auto_follower/models/mail_thread.py
```diff
@@ -0,0 +1,30 @@
...
+    def _mail_restrict_follower_get_domain(self):
+        parameter_name = 'mail_restrict_auto_follower.domain'
+        return self.env['ir.config_parameter'].get_param(
+            '%s.%s' % (parameter_name, self._name),
+            default=self.env['ir.config_parameter'].get_param(
+                parameter_name, default='[]'))
+
+    @api.multi
+    def message_subscribe(self, partner_ids, subtype_ids=None):
+        if self.env.context.get('allow_follower_subscription'):
```
hbrunn, line 21: I think you should override https://github.com/OCA/OCB/blob/8.0/addons/mail/mail_thread.py#L1874 to set a context key there, and react to it here. This we, the default behavior is what standard does, and only when called by `message_auto_subscribe`, we do the filtering. This is safer.

mail_restrict_auto_follower/models/mail_thread.py
```diff
@@ -0,0 +1,30 @@
...
+            new_partner_ids = partner_ids
+        else:
+            partner_model = self.env['res.partner']
+            domain = self._mail_restrict_follower_get_domain()
+            valid_partners = partner_model.search(safe_eval(domain))
```
hbrunn, line 26: add `[('id', 'in', partner_ids)]`, then you can remove the set operations below. Consider using `osv.expression.AND` and `osv.expression.normalize_domain`

mail_restrict_auto_follower/models/res_partner.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Serv. Tecnol. Avanzados - Pedro M. Baeza
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from openerp import models
+
+
+class ResPartner(models.Model):
+    """This is for avoid inheritance problem:
+    https://github.com/odoo/odoo/issues/9084"""
+    _name = 'res.partner'
```
hbrunn, line 10: this is specific to `res.partner` now. Can't you enumerate all models in `_register_hook` and patch the ones who need it?

mail_restrict_auto_follower/wizard/mail_wizard_invite.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from openerp import models, api
+
+
+class MailWizardInvite(models.TransientModel):
+    _inherit = 'mail.wizard.invite'
+
+    @api.multi
+    def add_followers(self):
+        obj = self.with_context(allow_follower_subscription=True)
```
hbrunn, line 12: if you implement https://github.com/OCA/social/pull/22/files#r45470005, this file can go away

mail_restrict_auto_follower/models/res_partner.py
hbrunn, line 10: I had the same problem, have a look at this solution: https://github.com/OCA/social/pull/28/files#diff-371a424a175e245428d526a9c14007bfR82

antoniocanovas:
:+1:  functional

ghost:
We just test it, it's working fine.

Good Job!

mail_restrict_auto_follower/models/res_partner.py
hbrunn, line 10: @pedrobaeza I started working on a general solution for this: https://github.com/hbrunn/social/blob/8.0-mail_follower_custom_notification/mail_follower_custom_notification/models/base_patch_models_mixin.py

Up to now, I didn't find any weirdnesses, when it's tested more, I'll propose this mixin as an own module for server-tools. But for the time being, you can simply grab a copy and see if it works for you too

pedrobaeza:
cc @Tecnativa

rafaelbn:
@pedrobaeza could you review @hbrunn comments please? Thanks

rafaelbn:
Hi @pedrobaeza do you think this module is still necessary? There is an alternative way to configure Odoo with subtypes.

@hbrunn comments are still missing answer

thanks

pedrobaeza:
Closing as we don't work anymore on v8.

PR #23, opened 2015-11-04
[ADD] event_registration_mass_mailing

JavierIniesta:
This module add actions to send mass mailings and add to mailing lists from
event registrations tree view.

PR #24, opened 2015-11-10
[ADD] mass_mailing_event_registration

JavierIniesta:
This module adds actions to send mass mailings and allows to add to mailing
lists from event registrations tree view.

### DIFF
--- a/mass_mailing_event_registration/README.rst
+++ b/mass_mailing_event_registration/README.rst
@@ -1,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========================
+Mass mailing from events
+========================
+
+This module adds actions to send mass mailings and allows to add to mailing
+lists from event registrations tree view.
+
+Usage
+=====
+
+Event registrations tree view:
+
+* Add 'Mass mailing to partners' button: with this button, you can send an
+  email to the selected registrations.
+
+* Add 'Add to mailing list' button: with this button, you can add the selected
+  registrations to a mailing list.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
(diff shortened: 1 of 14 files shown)


JavierIniesta:
Tests in progress

JavierIniesta:
cc @yajo @antespi @sergio-incaser @carlos-incaser @janetrp @rafaelbn @pedrobaeza  @dreispt

JavierIniesta:
[IMP] Tests added

dreispt:
This is related to another existing PR, correct?

JavierIniesta:
Related to OCA/social/pull/23

dreispt:
According to the discussion there, shouldn't this PR be in OCA/event?

janetrp:
@rafaelbn could you pls answer?

rafaelbn:
```
+    'depends': [
+        'event',
+        'mass_mailing'
```

@dreispt , @pedrobaeza thinks it must be in OCA/event, I answered:
- https://github.com/OCA/social/pull/23#issuecomment-155394095
- https://github.com/OCA/social/pull/23#issuecomment-155430252

and @pedrobaeza agree to maintain it in Social here:
- https://github.com/OCA/social/pull/23#issuecomment-155434071

I maintain my opinion of OCA/social, please let me know @pedrobaeza @dreispt  what do you think please

pedrobaeza:
It mixes both areas, so it can be either in one place or in another, but the name should reflect first the area where is hosted. If we put it in social, it must be _mass_mailing_event_registration_. If we host it in event, it must be _event_registration_mass_mailing_.

Which part is bigger/principal? I'm not really sure

dreispt:
Not an easy call since this is quite borderline. My reasoning is:
- It adds UI features on Event objects
- Event already depends on Social (`mail`), but Social does not require Event.

So, I would say that this is adding a Mail feature to Event, rather than adding an Event feature to Mail.

rafaelbn:
Ok, I agree as Pedro and Daniel make same reasoning. Then @JavierIniesta:
- name: event_registration_mass_mailing
- repo: OCA/Event

PR #28, opened 2015-11-25
[ADD] mail_follower_custom_notification

hbrunn:
# 
# Custom notification settings for followers

In standard Odoo, receiving mail notifications is an all or nothing affair.
This module allows you users to decide per followed record if they want to
receive emails or not. Further, they can choose to receive notification about
their own messages.

You can also set defaults for this settings on the subtype in question.
# Configuration

When followers open their subscriptions, they will be offered the choice to
override mail settings and to force being notified about their own messages.

You can add defaults per message sub type for this settings in Settings /
Technical / Email / Subtypes. Here, you also have the opportunity to apply
those defaults to existing subscriptions. Note that this overrides all
customizations your users already have done.
# Usage

To use this module, for example you need to:
- Go to Sales -> Sales -> Customers
- Go Inside any customer and check new functionality as described in screenshoot:


hbrunn:
@gurneyalex do you think we can install this on the oca instance and set force_own as default for our mailing lists? I still can't get used to it that I don't receive my own emails

gurneyalex:
:+1:

@hbrunn let's get this merged first.

rafaelbn:
Hi, it looks :+1: for me but please add section Usage in README in order to show quickly how to use module (go to ...). For a user which is not experimented with Odoo follower and notification it's not clear. Thanks

rafaelbn:
Could you check travis?

hbrunn:
@rafaelbn travis fixed. please provide a PR to https://github.com/hbrunn/social/tree/8.0-mail_follower_custom_notification on what you have in mind for the readme

rafaelbn:
Done, thanks!

PR #30, opened 2015-12-17
[ADD] mail_read_new_window addon

codingforfun:
Enhancement to mail message feed. Add button to open mail in new browser window. 
(A real new window, not in a modal one on top the current one) 

posteingang - odoo - mozilla firefox privater modus _061


pedrobaeza:
I'm afraid this already exists here: https://github.com/OCA/social/tree/8.0/mail_full_expand

codingforfun:
No it's not the same. I know your module but yours is opening the mail in a modal window on top (like all Odoo wizards) while mine is using an URL action to try to open a real new browser window.

pedrobaeza:
Well, then you should add this functionality in the existing module (via a key combination or a different icon). And the module is not mine :wink:

codingforfun:
@pedrobaeza 
Why should I add this into the same module instead of a different one? I only need this functionality and not the other one and I think it's good practice in this case to be as modular as possible. 

What do you mean by using a different icon? I already choosed another icon than 'mail_full_expand' at least to what I can tell from looking at it on runbot. See attached screenshot.
inbox - odoo - mozilla firefox_062

pedrobaeza:
This is for avoiding multiple modules that performs almost the same thing. For different icon I mean to have 2 icons in the same module, one opening in a popup, and the other opening in your way.

OSguard:
Using key (Shift + Strg) do not work on tablets

pedrobaeza:
You're right, then with 2 icons.

codingforfun:
If you install both modules this is already solved by having 2 icons. If I integrate both modules into one, then people who only need one of the both possible options can't choose anymore.

pedrobaeza:
Yes, they can choose clicking on one or the other. I prefer in this case to have it integrated in only one module (because they are very tiny), but I'm not the only one with voice. Let's see what others say.

@rafaelbn, @Yajo, @antespi, what do you think?

codingforfun:
Of course they can choose where to click, but this is not really the option I'm thinking of. I think space for unused icons on this bar isn't getting more, is it?

pedrobaeza:
I think there's still enough room for several icons more, so no issue on this one.

codingforfun:
I don't see why tests fail. Is there any way to trigger them again?

Btw: This is how it looks on runbot if both modules are installed.
inbox - odoo - mozilla firefox_069

pedrobaeza:
The warning in runbot seems unrelated:

```
2015-12-17 16:57:15,844 31718 WARNING 3137559-30-9f5fba-all openerp.addons.hw_blackbox_be.controllers.main: /dev/serial/by-id/ doesn't exist
```

elicoidal:
I see it confusing for the users to have 2 similar icons for a similar feature. 
I prefer either:
- one module with a set up part to choose
- 2 modules
  Could be nice to have a reference of each other modules in the README

codingforfun:
> one module with a set up part to choose

I thought of this based on the comments above, but I don't see the benefit of doing extra work of re-factoring both modules into one + configuration code if the configuration option can already be achieved easily by just installing one or the other or both modules

OSguard:
personal i think it is okay to add in the README of both modules the other option

yajo:
IMHO if you have 2 ways of doing the same thing with no clear benefit of one above the other, then choose the one you like most. In either case, be clear in README(s) as others commented, please.

codingforfun:
The motivation (or benefit) of our approach over mail_full_expand is, that in the case of our customer he wants his employees to manually copy/add information to sale orders and leads from customer emails. Opening the email on top of the objects view doesn't help in this case.

codingforfun:
I will add references to README files of both modules.

pedrobaeza:
@codingforfun, about that, it's very simple: open a new tab with Odoo, and you copy/paste between tabs, so there's no real problem with the other approach, but again, people choices.

codingforfun:
@pedrobaeza Of course there's many ways of skinning a cat, but end users expect something that is intuitive to them because they know it from other software like their mail client etc. In our case he likes it like this and since we think other people might too we like to propose to include this module into OCA/social.

yajo:
Using your module, window title is always "False". Does not seem very user-friendly. Please, fix that.

OSguard:
This is a problem of mail.message object as is has no name method so should be cover of an other module.

mail_read_new_window/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+#
+#   You should have received a copy of the GNU Affero General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+{
+    'name': 'Mail open in new window',
+    'summary': 'Open mail in new window',
+    'version': '1.0',
```
yajo: 8.0.1.0.0

mail_read_new_window/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+    "category": "Social Network",
+    'website': 'http://www.initos.com, https://odoo-community.org',
```
yajo: Only one please.

mail_read_new_window/i18n/mail_read_new_window.pot
```diff
@@ -0,0 +1,31 @@
+# Translation of Odoo Server.
```
yajo: pot not needed.

mail_read_new_window/static/src/js/mail_read_new_window.js
```diff
@@ -0,0 +1,27 @@
+
```
yajo, line 4: Please add AGPL header.

yajo:
Runbot affected by https://github.com/odoo/odoo/issues/10159

codingforfun:
I added cross references to README, however it doesn't look very nice. Is there some recommended  formatting (example) for cross referencing OCA  modules ?

mail_read_new_window/i18n/mail_read_new_window.pot
codingforfun: Why not? Where are 2 strings that need translation and I also supplied german translation for them in de.po.

pedrobaeza: See https://github.com/OCA/pos/pull/42#discussion_r48109507

codingforfun: Ok. Thx for the link.

codingforfun:
@dreispt You added the question label. Do you still have open questions or is it consensus now that it is OK if both modules reference each other.

OSguard:
:+1:

yajo:
:+1:

hbrunn:
:+1:

PR #31, opened 2015-12-21
[FIX] api decorator fixed

JavierIniesta:
@api.multi changed to @api.model in create method.


pedrobaeza:
:+1: 

I fast-track as this is a blocking issue.

PR #32, opened 2015-12-23
[FIX] mail_statistics_extras

cvinh:
When installing mail_statistics_extras, we cannot send standard emails anymore

There was a typo, here is the fix


pedrobaeza:
You are not working with final version, that doesn't contain s. Maybe the typo is also fixed in the final version.

cvinh:
hi pedro
There might be a bug with github because that commit has been merged here : https://github.com/Antiun/social/pull/2
and Antiun PR with OCA/8.0 has been made without the fix... ??

Regards

pedrobaeza:
In 8.0 branch, the final name was https://github.com/OCA/social/tree/8.0/mass_mailing_statistic_extra (without s), renamed after merge. Please update your branch and propose the fix if still present.

cvinh:
ok, I see... the bug has been fixed anyway (only my commit disappeared)
I close the PR

PR #38, opened 2016-01-25
[8.0][mass_mailing_unique] Avoid duplicates.

yajo:
This new module avoids duplicates in mass mailing lists.

Now you cannot have 2 lists with the same name, and you cannot have the same
email repeated in a list.

This will avoid sending the same mail to a contact repeated times, which will
disturb him and most probably cause you to be blocked as spam.


mass_mailing_unique/README.rst
```diff
@@ -0,0 +1,72 @@
...
+mass_mailing_unique%0Aversion:%20
+8.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
```
hbrunn: No need to give credit for this image, as this is an OCA module anyways

mass_mailing_unique/__openerp__.py
```diff
@@ -0,0 +1,23 @@
...
+    "summary": "Avoids duplicate mailing lists and contacts",
+    "version": "8.0.1.0.0",
+    "category": "Marketing",
+    "website": "https://grupoesoc.es",
+    "author": "Grupo ESOC Ingeniería de Servicios, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "pre_init_hook": "require_no_duplicates",
```
hbrunn: please follow OCA's guidelines about hooks: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#installation-hooks

mass_mailing_unique/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+            result)
+
+    # Search for duplicates in list's name
+    cr.execute("""SELECT name, COUNT(id) as duplicates
+                  FROM mail_mass_mailing_list
+                  GROUP BY name
+                  HAVING COUNT(id) > 1""")
+    for result in cr.dictfetchall():
+        errors.append(
+            _("there are %(duplicates)d lists with name %(name)s") % result)
```
hbrunn: how about emitting a SQL statement on the log that stupidly deletes duplicated ids? For big lists, your users will be thankful I guess

rafaelbn:
@Yajo this is already done in other OCA module, here:

https://github.com/OCA/crm/blob/8.0/mass_mailing_partner/models/mail_mass_mailing_contact.py#L13

mass_mailing_unique/__init__.py
yajo: If you have an email repeated 5 times with 4 different names, which is the good one?

If you have 2 lists with the same name, do you prefer to merge them in one or to rename one of them?

I prefer the current non-destructive implementation.

yajo:
@rafaelbn I did not notice that one, but it's different. I'm ensuring here that no duplicated emails are found, while that one links to partners and ensures no duplicated partners are found. If you have 2 separate partners with the same email, with `mail_mass_mailing_contact` you will end up sending 2 emails. You can install both modules and get both benefits.

mass_mailing_unique/hooks.py
```diff
@@ -0,0 +1,43 @@
...
+            result)
+
+    # Search for duplicates in list's name
+    cr.execute("""SELECT name, COUNT(id) as duplicates
+                  FROM mail_mass_mailing_list
+                  GROUP BY name
+                  HAVING COUNT(id) > 1""")
+    for result in cr.dictfetchall():
+        errors.append(
+            _("there are %(duplicates)d lists with name %(name)s") % result)
```
pedrobaeza: Only as reference, the format with dictionaries is avoided on translated strings, because people get confused if they have to translate the keyword, provoking errors on execution time. For example, they translate it in spanish as "hay %(duplicados)d listas con nombre %(nombre)s"

yajo: I tend to prefer them if possible because English tends to order stuff reversed to Spanish, so I guess that can happen in many other languages. All would be better with `.format()` OTOH.

pedrobaeza: Well, I prefer too, but the translators' mistakes are there, so I get back always to tradition '%s' or the new '{}' with 'format'

yajo: Let me change it.

hbrunn:
:+1:

rafaelbn:
OK, thanks! :+1:

PR #43, opened 2016-02-15
[8.0][FIX][website_snippet_bg_color][website_snippet_responsive] Allow to pick color for almost-background element.

yajo:
Right now only able to pick for the backest element in responsive snippets. This allows it for the previous one.

Also allow other elements than divs to be colored.

@dgzurita @rafaelbn


rafaelbn:
tested in local :+1: thanks!

pedrobaeza:
:+1:

PR #45, opened 2016-02-18
[8.0][mail_forward] Custom migration

yajo:
This is a little migration script I need because in [Grupo ESOC](https://grupoesoc.es) we had a slightly different version of this module before publishing to OCA, and now there's a conflict this solves.

This also serves as an example migration for https://github.com/OCA/server-tools/pull/369, since it uses that toolkit for doing it. But to avoid conflicts with others, as you can see, all is inside a `try...except` that will just allow anyone to run the migration without hassle.

Thanks for your understanding.

### DIFF
--- a/mail_forward/__openerp__.py
+++ b/mail_forward/__openerp__.py
@@ -5,7 +5,7 @@
 {
     "name": "Message Forward",
     "summary": "Add option to forward messages",
-    "version": "8.0.7.0.0",
+    "version": "8.0.7.0.1",
     "category": "Social Network",
     "website": "https://grupoesoc.es",
     "author": "Grupo ESOC Ingeniería de Servicios, "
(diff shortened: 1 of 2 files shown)


pedrobaeza:
As stated on https://github.com/OCA/server-tools/pull/369, this can be done with openupgradelib or simple migration scripts.

:-1: while this is not rebuild with current tools

yajo:
This is not harmful, even if migratk is never merged.

pedrobaeza:
You will get an error...

yajo:
Where?

pedrobaeza:
OK, I see that you have encapsulate all in a try. Well, too tricky...

yajo:
I did it suspecting somebody would ask me to rewrite using openupgrade :grin: or would tell me they will never use this, so since this is a 1-time 1-company migration I suspect nobody else will ever need (unless somebody was using my published dev version without my knowledge) I tried to harm the least possible.

pedrobaeza:
Why don't you simply run the upgrade in your DB and forget about this patch? If nobody has accessed previous version, there's no need of this script.

PR #48, opened 2016-03-01
[8.0][website_mail_snippet_tr_edit] Edit <tr> in your mails.

yajo:
With this module, you can now duplicate or delete <tr> elements in your table-based templates or snippets for emails.

@rafaelbn


pedrobaeza:
Can't you insert tables at your own?

Maybe the name should be `table` instead of `tr` to be more descriptive.

What about an icon of a table :wink:?

yajo:
> Can't you insert tables at your own?

I tried but some weird behavior inside the Odoo drag & drop engine does not allow that. It's in the known issues.

> Maybe the name should be table instead of tr to be more descriptive.
> What about an icon of a table :wink:?

Thanks, fixed.

pedrobaeza:
Oh, what a pity! There's no possibility then to insert tables in an empty template?

dgzurita:
Tested in runbot. Perfect! :+1:

yajo:
@pedrobaeza at least not to move them around

rafaelbn:
Thanks! snippets become useful! :+1: :smile:

pedrobaeza:
:+1: Thanks for the cool icon!

pedrobaeza:
One last question. Can't this be done for all the website builder instead only the mail part? Is not the same code? Or they are different snippets?

rafaelbn:
Let's @Yajo answer but as I talked with him, they are different snippet.

yajo:
They are different.

PR #49, opened 2016-03-02
[ADD] Module mass_mailing_no_unsubscribe 8.0

astirpe:
This module lets you configure whether you want to have the standard unsubscribe link or
add your own link to the mass mailing emails.


pedrobaeza:
This is the same as https://github.com/OCA/social/tree/8.0/mass_mailing_custom_unsubscribe

Have you check it?

astirpe:
I checked it, yes.
I think that the provided funcionality is different.
This module lets you disable the link at all.

pedrobaeza:
Then please include your functionality in the current module, instead of creating another one.

@antespi, didn't you include an option to disable the unsubscribe link?

astirpe:
I would keep this PR open for a while, in case someone wants to share an opinion on: merging the proposed functionality in the existing module or keeping that functionality in a separate module.
I would prefer to have the two modules separated and the two features not coupled.
But it's my opinion... If you decide the other way, I would be happy to contribute on that as well :+1:

pedrobaeza:
But they are very related. Anyway, please read the other module README: https://github.com/OCA/social/blob/8.0/mass_mailing_custom_unsubscribe/README.rst#configuration

This functionality is already present.

astirpe:
@pedrobaeza Yes I see, it's already present. Sorry, I didn't notice it.
Thank you!

PR #52, opened 2016-03-17
[FIX] Make sequence of view higher than the original view

lfreeke:
Add a higher priority to the view.


lfreeke:
Ping @hbrunn

pedrobaeza:
What is the reasoning behind this change?

hbrunn:
for whatever reasons, the mail module's view has priority 20: https://github.com/OCA/OCB/blob/8.0/addons/mail/mail_message_view.xml#L8

So without the patch, the custom view for the action wins in cases where no view is explicitly given, which is most places. This makes this module combine badly with others that modify the original view.

:+1: from my side

pedrobaeza:
:+1:

guewen:
:+1:

PR #54, opened 2016-04-13
[ADD] mail_notification_email_template

hbrunn:
# 
# Use email templates in notifications

This module was written to allow you to send notification rendered with an
email template.
# Configuration

To configure this module, you need to:
- go to Settings/Technical/Email/Subtypes
- open the subtype for which you want to change the notification
- select an email template
- within the email template, `object` refers to the notification,
  from there you have access to the message the notification is for via
  `${object.message_id}`, to the partner who is to be notified via
  `${object.partner_id}` and indirectly to the record the message is
  about via `${object.record}`. To generate links to the record, use
  `${object.record_access_link}`.

Also have a look at the demo template for the minimal fields you should set.


pedrobaeza:
cc @rafaelbn, you'll love this

@hbrunn, I think we are connected in our needs!

pedrobaeza:
Can we connect this one with the other for rendering mails with qweb? A bridge is needed, right?

hbrunn:
@pedrobaeza no, email_template_qweb templates will just work, as well as normal ones. It's up to you which one you use. Speaking of which, I still have to move this here. No access to my ssh key at the moment, will do that tomorrow or at the weekend.

andhit-r:
Hi @hbrunn i tested this PR on runbot. I would like put a notification rendered with template when a lead is created. So i:
1. Create new template. I follow your demo data
   template_configuration_1
   template_configuration_2
2. Then i use template above on message subtype
   message_subtype
3. I create a lead
   I assume that the notification when lead is created will rendered using template. But when i tested it, the notification is still the default one
   notification

Did i missed some step?

hbrunn:
what you describe sounds exactly like I'd expect it to work. Message subtypes are a bit intransparent, is it possible the created message is of type 'Discussion'? Attach your template to this subtype too, if this works as expected, it's a bug in `crm` for not using the message subtype we'd expect here.

andhit-r:
# Case 1
1. Create new subtype
   mail_notification_email_template_1
2. Create template for that subtypethat exact copy of your demo data
   mail_notification_email_template_2
   mail_notification_email_template_3
3. Then i send message from crm.lead open chatter

## Result
1. Email was sent to followers using template defined
   mail_notification_email_template_5

# Case 2
1. I add email template notification for crm.lead's subtype **State Change**
   mail_notification_email_template_6
2. I then adjust email template used according your demo data
3. I change one of the crm.lead state. I expect that it will trigger email like previous case since i modified **State Change** subtype

## Result

No email sent

@hbrunn it it expected behavior? The subtype should named Discussion?

hbrunn:
This plugin doesn't change anything concerning mailing. Check that there are partners subscribed to this subtype on the relevant lead, that they have an email address set, that they are not opted out.

Or better: First test without this plugin, if this does send a mail and doesn't with the plugin installed, you have a case. Otherwise, it's a misconfiguration that has nothing to do with the plugin.

andhit-r:
@hbrunn my mistake then. Forget about mail related issues that i mentioned before. I retried again case that i mentioned on https://github.com/OCA/social/pull/54#issuecomment-213382295. No effect on open chatter message. The message still not using template that i use on message subtype.

lasley:
I didn't realize I have been living without this until I saw it. Awesome - code/functional 👍

hbrunn:
@OCA/social-maintainers can we merge this?

pedrobaeza:
@hbrunn, please provide as data (not as demo) the template, as it can be used as example and as minimal layout.

hbrunn:
@pedrobaeza done

mail_notification_email_template/models/mail_message_subtype.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import fields, models
+
+
+class MailMessageSubtype(models.Model):
+    _inherit = 'mail.message.subtype'
+
+    template_id = fields.Many2one(
+        'email.template', string='Notification template',
+        domain=[('model_id.model', '=', 'mail.message.subtype')],
```
pedrobaeza, line 12: Will this domain prevent to select the default template?

hbrunn, line 12: no: https://github.com/OCA/social/pull/54/files/9ddfa894939db329a2dc6317fe09889151eb9376#diff-f0fcc36c6f0b6564154b101d60f9acc5R6

mail_notification_email_template/tests/test_mail_notification_email_template.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp.tests.common import TransactionCase
+
+
+class TestMailNotificationEmailTemplate(TransactionCase):
+    def test_mail_notification_email_template(self):
+        # demo data install a template for discussions, so we simply post
```
lmignon: the template is no more installed by demo data

hbrunn: thanks, done

lmignon:
:+1:

lmignon:
@hbrunn travis in red :-/

hbrunn:
@lmignon not any more :-)

pedrobaeza:
:+1:

lmignon:
:+1:

PR #55, opened 2016-04-14
[ADD] email_template_qweb

hbrunn:
# 
# QWeb for email templates

This module was written to allow you to write email templates in QWeb instead
of jinja2. The advantage here is that with QWeb, you can make use of
inheritance and the `call` statement, which allows you to reuse designs and
snippets in multiple templates, making your development process simpler.
# Usage

- Select `QWeb` in the field `Body templating engine`
- Select a QWeb view to be used to render the body field
- Apart from QWeb's standard variables, you also have access to `object` and
  `email_template`, which are browse records of the current object and the
  email template in use, respectively.

Demo data contains an example on how to separate corporate identity from a
template's content.


atchuthan:
:+1:

email_template_qweb/README.rst
```diff
@@ -0,0 +1,64 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+========================
+QWeb for email templates
+========================
+
+This module was written to allow you to write email templates in QWeb instead
```
pedrobaeza, line 8: In order to avoid the word write 2 times, you can put directly "This module allows you to write...", but it's not mandatory to change it.

pedrobaeza:
:+1: Thanks for this contribution

tremlin:
Code review :+1: , no test. Thanks!

PR #56, opened 2016-04-28
[ADD] new module base_mail_bcc 

tremlin:
This PR adds a new module to configure a e-mail addresses to which all outgoing e-mail are sent in BCC. There are various modules around for this purpose, but none of these is in the OCA.

Note that this is a v8 port of https://github.com/OCA/server-tools/pull/59  (v7) which seems to be unwanted in the other repository.

### DIFF
--- a/base_mail_bcc/README.rst
+++ b/base_mail_bcc/README.rst
@@ -6,7 +6,7 @@
 BCC all emails
 ==============
 
-This module extends the email meachnism to allow for sending a blind carbon copy (BCC) 
+This module extends the email mechanism to allow for sending a blind carbon copy (BCC)
 of all outgoing emails to configurable e-mail addresses.
 
 Configuration
@@ -15,7 +15,7 @@ Configuration
 To configure this module, you need to:
 
 * Go to Settings > Parameters > System Parameters
-* Create a new entry with key `mail.always_bcc_to` and set the desired e-mail addresses for BCC as value. This value must be comma-separated list of valid e-mail addresses.
+* Create a new entry with key `base_mail_bcc.bcc_to` and set the desired e-mail addresses for BCC as value. This value must be a comma-separated list of valid e-mail addresses.
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
--- a/base_mail_bcc/README.rst
+++ b/base_mail_bcc/README.rst
@@ -1,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==============
+BCC all emails
+==============
+
+This module extends the email meachnism to allow for sending a blind carbon copy (BCC) 
+of all outgoing emails to configurable e-mail addresses.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+* Go to Settings > Parameters > System Parameters
+* Create a new entry with key `mail.always_bcc_to` and set the desired e-mail addresses for BCC as value. This value must be comma-separated list of valid e-mail addresses.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/205/8.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/social/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Thomas Rehn <thomas.rehn@initos.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/base_mail_bcc/__openerp__.py
+++ b/base_mail_bcc/__openerp__.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# © 2014-2016 Thomas Rehn (initOS GmbH)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "BCC all emails",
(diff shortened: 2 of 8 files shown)


base_mail_bcc/README.rst
```diff
@@ -0,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==============
+BCC all emails
+==============
+
+This module extends the email meachnism to allow for sending a blind carbon copy (BCC)
```
hbrunn: mechAnism

base_mail_bcc/README.rst
```diff
@@ -0,0 +1,58 @@
...
+This module extends the email meachnism to allow for sending a blind carbon copy (BCC) 
+of all outgoing emails to configurable e-mail addresses.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+* Go to Settings > Parameters > System Parameters
+* Create a new entry with key `mail.always_bcc_to` and set the desired e-mail addresses for BCC as value. This value must be comma-separated list of valid e-mail addresses.
```
hbrunn: '[...] be A comma-separated [...]'

hbrunn: `mail.always_bcc_to` - this implies the parameter has something to do with the mail module, which is not completely the case. I propose to always prefix keys with your module name, so in your case, I'd call the key `base_mail_bcc.bcc_to`

base_mail_bcc/__openerp__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# © 2014-2016 Thomas Rehn (initOS GmbH)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "BCC all emails",
+    "version": "8.0.1.0.0",
+    "depends": ["base"],
+    'author': 'initOS GmbH, Odoo Community Association (OCA)',
+    "category": "Uncategorized",
```
hbrunn: maybe `Tools`?

base_mail_bcc/models/ir_mail_server.py
```diff
@@ -0,0 +1,40 @@
...
+            existing_bcc = []
+            if message['Bcc']:
+                existing_bcc.append(message['Bcc'])
+                del message['Bcc']
+            message['Bcc'] = COMMASPACE.join(
+                existing_bcc + config_email_bcc.split(',')
+            )
+
+        return super(IrMailServer, self)\
+            .send_email(cr, uid, message, mail_server_id, smtp_server,
```
hbrunn: pass everything that's a keyword argument in the original as keyword argument

base_mail_bcc/models/ir_mail_server.py
```diff
@@ -0,0 +1,40 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models
+from email.Utils import COMMASPACE
+
+
+class IrMailServer(models.Model):
+    _inherit = "ir.mail_server"
+
+    def send_email(self, cr, uid, message, mail_server_id=None,
```
hbrunn: any reason not to use new api?

hbrunn:
we're not far away from the half year of inactivity, after which I'll close this one

rafaelbn:
Hi @tremlin , are you going to finish this PR? If so, I will review it. Thanks

rafaelbn:
Well @hbrunn , as commented here: https://github.com/OCA/social/pull/93#issuecomment-251403628 plus this is an old one I closing now. If we get some answer we can reopen it.

tremlin:
I am working on the issues mentioned above and will push the results as soon as possible.

pedrobaeza:
Opening again then

tremlin:
I improved all issues mentioned in the review:
- typos in text
- renamed configuration parameter from `mail.always_bcc_to` to `base_mail_bcc.bcc_to`
- code clean up

@rafaelbn I would be happy about a new review. Regarding the rationale behind this module. We use in our installations for monitoring of outgoing e-mails.

pedrobaeza:
As stated by @rafaelbn in https://github.com/OCA/social/pull/93#issuecomment-251403628, no I'm doubting about the convenience of this module.

rafaelbn:
Hi @tremlin, sorry but I don't see any sense in this module.

> Regarding the rationale behind this module. We use in our installations for monitoring of outgoing e-mails.

You would like to use: https://github.com/OCA/social/tree/8.0/mail_tracking

Furthermore: you don't need this module. Odoo already send one single mail per recipient. So you can receive a much emails as you want. You will never receive an email from Odoo sent to 6 recipients with all the e-mail addresses in the "To" field.

Think about it please

tremlin:
@rafaelbn Sorry, I don't understand the argument. Some of our customers want a copy of every e-mail that was sent out by Odoo in a single mailbox which they can check in their e-mail client. I don't see how this is possible with other modules.

For computer-based monitoring and following the delivery path the module `mail_tracking` is indeed be the best choice.

pedrobaeza:
@tremlin, if you want that, just create a partner with the recipient address, and add it by default as follower to all objects. That's a better approach I think.

tremlin:
This will send out all notifications also to the new default follower partner but it doesn't allow to see the copies of the original e-mails.  I haven't tried this, but I think one limitation of this approach could be that it works only for notifications, but not for manually crafted `mail.mail`?

pedrobaeza:
Yeah, you're right about that point with manual mail.mail.

The module should be renamed then to _mail_big_brother_ :laughing: 

Let's continue then with the PR.

pedrobaeza:
Anyway, I would recommend you other mail tracking system less painful, like hiring or use the free layer from Mailgun (http://mailgun.com/), and then install mail_tracking and mail_tracking_mailgun. You will have one point to see all the outbound activity (in Mailgun portal), and you will also have status tracking for each email in Odoo.

base_mail_bcc/models/ir_mail_server.py
```diff
@@ -0,0 +1,34 @@
...
+
+from openerp import models, api
+from email.Utils import COMMASPACE
+
+
+class IrMailServer(models.Model):
+    _inherit = "ir.mail_server"
+
+    @api.model
+    def send_email(self, message, **kwargs):
```
hbrunn: say `*args, **kwargs` here, because otherwise, it will break if some code calls this function with positional arguments. Better is to exactly repeat the parameters as defined by super

hbrunn (changes requested):
I use this module now for a customer who wants to check if the mails going out are actually what they want to go out, and we can't just duplicate them at the mailserver level. So there's definitely a use case, but please don't break the API. If you fix this, and add a test for send_email, we can merge imho

base_mail_bcc/models/ir_mail_server.py
tremlin: I've now replicated all the parameters from the original `send_email` function.

tremlin:
Thanks for the review. I've improved the code as suggested and added a test case. I've not much experience with mock-testing, but I hope this is what you had in mind.

hbrunn (approved):
your test tests that it doesn't crash, but doesn't assert anything. That's a bad habit IMHO, but I'm not going to bug you about that after a year...

elicoidal (approved):

PR #58, opened 2016-05-04
[8.0][IMP][mass_mailing_custom_unsubscribe] Get reasons for unsubscription

yajo:
Now you can know why an unsubscription was made.

@Tecnativa.


yajo:
This one should be ready to review, please somebody label as such :blush:

mass_mailing_custom_unsubscribe/README.rst
```diff
@@ -8,10 +8,16 @@ Customizable unsubscription process on mass mailing emails
 With this module you can set a custom unsubscribe link appended at the bottom
 of mass mailing emails.
 
+It also displays a beautiful and simple unsubscription form when somebody
+unsubscribes, to let you know why he left; and then displays to him a beautiful
```
pedrobaeza: Don't use he/him. Better to use an inpersonal style. Remove that parts from the sentence. The meaning is the same.

mass_mailing_custom_unsubscribe/README.rst
```diff
@@ -28,21 +34,38 @@ default 'Click to unsubscribe' link will appear, with the advantage that it is
...
+#. Create / edit / remove / sort as usual.
+#. If *Details required* is enabled, they will have to fill a text area to
+   continue.
 
-* `Unsubscription successful </page/mass_mail_unsubscription_success>`_.
-* `Unsubscription failed </page/mass_mail_unsubscription_failure>`_.
+Unsubscription Goodbye Message
+------------------------------
 
+Your unsubscriptors will recieve a beautier goodbye page. You can customize it
```
pedrobaeza: s/recieve/receive

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -0,0 +1,160 @@
...
+        if email_fname and email:
+            domain.append((email_fname, "ilike", email))
+        else:
+            # Trying to unsubscribe without email? Bad boy...
+            raise exceptions.AccessDenied()
+
+        if record_ids._name == "mail.mass_mailing.contact":
+            domain.append(
+                ("list_id", "in",
+                 [l.id for l in mailing_id.contact_list_ids]))
```
pedrobaeza: Why not simply `mailing_id.contact_list_ids.ids`?

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -0,0 +1,160 @@
...
+            "crm.lead": {
+                "origin": "name",
+                "contact": "contact_name",
+            },
+            "hr.applicant": {
+                "related": "job_id",
+                "origin": "name",
+            },
+            # In case you install OCA's event_registration_mass_mailing
+            "event.registration": {
```
pedrobaeza: Can this be encapsulated in a method for making it inheritable? There's no way of auto-discover this?

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
```diff
@@ -0,0 +1,75 @@
...
+    mass_mailing_id = fields.Many2one(
+        "mail.mass_mailing",
+        "Mass mailing",
+        required=True,
+        help="Mass mailing from which he was unsubscribed.")
+    unsubscriber_id = fields.Reference(
+        lambda self: self._selection_unsubscriber_id(),
+        "Unsubscriber",
+        required=True,
+        help="Who he was unsubscribed.")
```
pedrobaeza: `Who was unsubscribed` (remove the `he`)

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
```diff
@@ -0,0 +1,75 @@
...
+    reason_id = fields.Many2one(
+        "mail.unsubscription.reason",
+        "Reason",
+        ondelete="restrict",
+        required=True,
+        help="Why he was unsubscribed.")
```
pedrobaeza: Impersonal: `Why the unsubscription was made`

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
```diff
@@ -0,0 +1,75 @@
...
+    details = fields.Char(
+        help="More details on why he was unsubscribed.")
```
pedrobaeza: `More details on why the unsubscription was made`

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
```diff
@@ -0,0 +1,75 @@
...
+    details_required = fields.Boolean(
+        related="reason_id.details_required")
+    success = fields.Boolean(
+        help="If this is unchecked, it indicates some failure happened in the "
+             "unsubscription process.")
+
+    @api.model
+    def _default_date(self):
+        return datetime.now()
```
pedrobaeza: `fields.Datetime.now`

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
```diff
@@ -0,0 +1,75 @@
...
+from .. import exceptions
+
+
+class MailUnsubscription(models.Model):
+    _name = "mail.unsubscription"
+    _inherit = "mail.thread"
+    _rec_name = "date"
+
+    date = fields.Datetime(
+        default=lambda self: self._default_date(),
```
pedrobaeza, line 15: Use `default=fields.Datetime.now`

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
pedrobaeza: (not needed if you change the default definition)

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
```diff
@@ -0,0 +1,75 @@
...
+                raise exceptions.DetailsRequiredError(
+                    _("This reason requires an explanation."))
+
+
+class MailUnsubscriptionReason(models.Model):
+    _name = "mail.unsubscription.reason"
+    _order = "sequence, name"
+
+    name = fields.Char(
+        index=True,
```
pedrobaeza, line 67: index on char fields is useless, as like/ilike (the operator used in Odoo) doesn't use the index.

mass_mailing_custom_unsubscribe/views/mail_unsubscription_view.xml
```diff
@@ -0,0 +1,103 @@
...
+        <form>
+            <sheet>
+                <group>
+                    <field name="date"/>
+                    <field name="mass_mailing_id"/>
+                    <field name="unsubscriber_id"/>
+                    <field name="email"/>
+                    <field name="success"/>
+                    <field name="reason_id"/>
+                    <field
```
pedrobaeza: Although there's no guideline about this, the way you indent XML files makes Odoo to treat this very weird. Some points:
- All that goes inside an attribute value should be in the same line, or it will be parsed it a lot of spaces (and not respecting the logical indentation you see in the file), so for example, please put `attrs="{'required': [('details_required', '=', True)]}"` 
- Most editors put the first attribute in the same line of the tag , and the second aligned with the first letter of this attribute. I like this convention because it saves one line as it's still very readable. For example:

```
<field name="details"
       attrs="{'req...
```

mass_mailing_custom_unsubscribe/README.rst
yajo: How?

mass_mailing_custom_unsubscribe/controllers/main.py
yajo: I thought in what you say already, but found that it makes no sense. After all, this method already returns a `dict()`, so you can already overwrite any keys on it depending on the model, no need to externalize this part.

BTW, you can just add more stuff there if you want, like the `event.registration` part, that will work in case you install `event_registration_mass_mailing` but will not harm otherwise.

mass_mailing_custom_unsubscribe/README.rst
pedrobaeza: `It also displays a beautiful and simple unsubscription form when somebody unsubscribes, to let you know why; and then displays a beautiful...`

mass_mailing_custom_unsubscribe/controllers/main.py
pedrobaeza: The difference is you need to provide also the code for the matching and all the logic. Having the method, you only need to define the keys `related` and `origin` (and maybe `contact`), and reuse the same logic.

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
yajo, line 15: Guidelines say to use lambdas: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#field

pedrobaeza, line 15: Well, that guideline is outdated (no new API), because you can overwrite the field only the default definition to get a new default, but OK, until we change it, follow the other comment.

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
yajo, line 67: As long as it is used in the order by clause it should provide a performance boost when selecting, AFAIK.

pedrobaeza, line 67: I remember a discussion about this where it can have even the contrary effect (because of partition), but I'm not sure. Only a profiling would say, but as this is not critical, we can let it as is.

mass_mailing_custom_unsubscribe/views/mail_unsubscription_view.xml
yajo: I tend to care less about parsers and more about humans, when we speak about whitespaces. Since the contents are Python, I wrote them as such, but nevermind I'll change it.

pedrobaeza: Yes, please, this time is for improving copy/paste operations an readiness inside Odoo views/HTML editor/debugger (it's the reason why I'm asking this).

yajo:
All should be ready now.

mass_mailing_custom_unsubscribe/controllers/main.py
yajo: Fixed.

mass_mailing_custom_unsubscribe/README.rst
```diff
@@ -65,6 +90,17 @@ In case of trouble, please check there if your issue has already been reported.
 If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
 `here <https://github.com/OCA/social/issues/new?body=module:%20mass_mailing_custom_unsubscribe%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
 
+Known Issues / Roadmap
+======================
+
+* This module adds a security hash for mass mailing unsubscription URLs, which
+  makes URLs of mass mailing messages sent before its installation not work
```
pedrobaeza: s/makes URLs.../makes to not work anymore URLs... (and remove it from the end).

mass_mailing_custom_unsubscribe/README.rst
```diff
@@ -65,6 +90,17 @@ In case of trouble, please check there if your issue has already been reported.
...
+  anymore. If you need backwards compatibility, disable this security feature
+  by emptying the ``mass_mailing_custom_unsubscribe.salt`` system parameter.
```
pedrobaeza: The key should be `mass_mailing.salt`

pedrobaeza:
Fixes done, so :+1:

rafaelbn:
:+1

hbrunn:
add tests for your code

pedrobaeza:
Sorry, but the budget for this module is out, so for now tests are out of the scope. It has been put in the "Known issues" section anyway.

pedrobaeza:
@hbrunn, it doesn't need fixing. It's correctly working, as Travis and runbot says, the red cross is about the drop in coverage.

hbrunn:
untested coded needs fixing in my book, and the conventions seem to agree: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#tests

I only make exceptions to this for OCB and repos where nobody cares about code coverage anyways.

pedrobaeza:
No, the conventions says as general rule and for fixing bugs, not for new modules.

hbrunn:
'New modules or addtions should ideally test all the functions defined'

pedrobaeza:
Yeah, you're right, but ideally is far from reality... As said, the budget is really exhausted, so we need to move forward in this PR or close it.

(however, I'm publishing now another commit with a bug we have discovered).

pedrobaeza:
@hbrunn, are you going to approve this one?

rafaelbn:
Hi, tested in a production customer instance :+1: 
Ready to merge functionally

rafaelbn:
@hbrunn could you give an opinion? Thanks

hbrunn:
I gave my opinion in https://github.com/OCA/social/pull/58#issuecomment-220967065

pedrobaeza:
@hbrunn, are you going to block this PR without tests (take into account that they will need to be Phantom JS ones, which they are cumbersome)?

hbrunn:
I'm not blocking, I'm simply not merging it.

Coveralls only measures python code by the way.

pedrobaeza:
Yeah, but that code is activated by website events (click on a button in a form for example). We can trick this simply calling these methods from a regular test. But it's better to be consequent and let this task to be made properly by other user or any other moment in time.

rafaelbn:
@Yajo @pedrobaeza which test do we need in order to get coverage green? (python code)

hbrunn:
one that covers the red areas here: https://coveralls.io/builds/6589095/source?filename=mass_mailing_custom_unsubscribe%2Fmodels%2Fmail_mail.py (some of the handler branches can be marked with `# pragma: no cover`)

yajo:
I prefer to not set the `no cover` trick. I agree it lacks tests, but I don't want to hide that. Just have no time. BTW these tests should be JS tours.

rafaelbn:
@Yajo @pedrobaeza maybe @kutyel could help here with test? ...

cc @lasley :smile:

lasley:
Seems like mocks might be the best way to test this. I can knock out if you would like.

lasley:
These should theoretically work. Will check on the Travis later tonight. Could use a bit of DRY too

https://github.com/laslabs/social/commit/b7b96d080ab0f61201e69e8cde7d9826f701faaf

lasley:
Yup that worked for the most part, just needed a quick fix. 

While I was there I noticed the controller name was misspelled so I fixed that, which made coveralls bitch about that file too - turns out it's not tested at all either.

I added a few tests there too, but I don't know if I'm up to testing this deep of a controller at the moment. IMO this logic has been fine for a while, so we'd probably be safe merging as is in that regard.

Branch: https://github.com/laslabs/social/tree/8.0-mass_mailing_custom_unsubscribe-reason
Travis: https://travis-ci.org/laslabs/social/builds/146553895
Coveralls: https://coveralls.io/github/laslabs/social?branch=8.0-mass_mailing_custom_unsubscribe-reason

Not sure the protocol here - am I supposed to make a new PR, or does @Yajo cherry pick from mine?

rafaelbn:
Hi @lasley you can make a PR in Tecnativa's branch: https://github.com/Tecnativa/social/tree/8.0-mass_mailing_custom_unsubscribe-reason, we will merge you work and then comes here to this PR.

lasley:
Done, thanks for the advice @rafaelbn 

https://github.com/Tecnativa/social/pull/1

rafaelbn:
WOW! :smile: from 76.615% to 80.708% !!! @hbrunn it is enough? Thanks a lot to @lasley !

rafaelbn:
WOW! :smile: from 76.615% to 80.708% !!! @hbrunn  is it enough? Thanks a lot to @lasley  !

lasley:
While I was there, code + functional :+1:

yajo:
I got late here, but absolutely thank you @lasley! :tada:

PR #63, opened 2016-05-25
[9.0] Migrate mail_optional_follower_notification

lmignon:


hbrunn:
please get the commits from 8.0 to allow simple cherry picking of further developments in 8.0: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#howto

rafaelbn:
Waiting for the fix in order to test it. Thanks

lmignon:
@hbrunn This branch is based on the one merged in 8.0. It seems that the original commits have been squashed when the branch has been merged. Why? I've asked @adrienpeiffer if he is aware of this squash and he isn't....  Now I've to redo my branch and I'm not happy with that.

hbrunn:
@lmignon looking at the commit log, it seems I'm the culprit. Then ignore my request, and I'll look into how to avoid those mistakes in the future. Apologies for the noise.

hbrunn:
everything else looks good btw, :+1:

lmignon:
@hbrunn Rebase done...

lmignon:
@hbrunn 

> @lmignon looking at the commit log, it seems I'm the culprit. Then ignore my request, and I'll look into how to avoid those mistakes in the future. Apologies for the noise.

I don't blame you, In this case, the 'rebase' was easy. (reset --hard oca/9.0 + cherry-pick commits on 8.0 + cherry-pick specific commits on 9.0 = 5 commands :smirk: )

Unfortunately, it seems that I've to fix my branch...

lmignon:
@hbrunn Can yout put the label WIP? I've to fix the tests but didn't have time now.

rafaelbn:
Hi @lmignon , which is the **difference** between this module and configure the same in Settings - Email - Subtypes?

I know that this is a migration, but in fact module in v8 was not necessary as you can configure it.

adrienpeiffer:
@rafaelbn What do you mean about `configure the same` ?
 We can use this module without a subtype. it's not the same feature. Here, the user can simply decide if he want to notify follower or not without any configuration.

mail_optional_follower_notification/models/res_partner.py
```diff
@@ -0,0 +1,18 @@
...
+    _inherit = 'res.partner'
+
+    @api.multi
+    def _notify(self, message, force_send=False, user_signature=True):
+        if self.env.context.get('force_partners_to_notify'):
+            partners_to_notify =\
+                self.env.context.get('force_partners_to_notify')
+            self = self.filtered(lambda p: p.id in partners_to_notify)
+        super(ResPartner, self)._notify(
+            message, force_send=False, user_signature=True)
```
adrienpeiffer: Don't propagate default value of named arguments

adrienpeiffer (changes requested):

dreispt:
I did not read the code, but from the module description, this seems to be the same feature as the standard "Log a message" - it writes a message on the Chatter without sending it to the Followers.

adrienpeiffer:
@dreispt You have the possibility to define some recipent but you can exclude followers.

dreispt (approved):
I suggest a better description:
"When sending a message in a document's mail chatter, all Followers are notified.
This feature allows to uncheck the notify followers checkbox, and have the notification sent only to selected people.
Beware that the message will still be visible to all followers when viewing the document."

But not blocking and implementation LGTM

lmignon:
rebased

mail_optional_follower_notification/tests/test_mail_optional_follower_notifications.py
```diff
@@ -0,0 +1,54 @@
...
+                                 (4, self.partner_03.id)]
+        compose_id = mail_compose.with_context(ctx).create(values)
+        compose_id.with_context(ctx).send_mail()
+        res = self.env["mail.message"].search(
+            [('model', '=', 'res.partner'),
+             ('res_id', '=', self.partner_01.id)])
+        self.assertEqual(len(res.ids), 1)
+        message = self.env['mail.message']
+        for record in res:
+            if record.notified_partner_ids.ids == [self.partner_03.id] and\
```
hbrunn: this should be `needaction_partner_ids` by now: https://github.com/OCA/OCB/blob/9.0/addons/mail/models/mail_message.py#L950

hbrunn: @lmignon can you fix the test? Then we can merge I think

lmignon:
@hbrunn I've edited the file on github If tests are OK I'll squash my last commits

hbrunn:
yay, worked. I'll count @adrienpeiffer as a +1 because his comment was addressed too, so I'll merge after the squashing

lmignon:
@hbrunn Commit squahed...

PR #66, opened 2016-05-30
[8.0][ADD][website_snippet_mass_mailing_name] Snippet + name

yajo:
Let you have the name field in the mass mailing subscription snippet.

When somebody tells his name, a partner is linked (if a match is found) or created.

Leave most logic to base module.

@Tecnativa relaunching https://github.com/OCA/website/pull/201, related to OCA/crm#96.


website_snippet_mass_mailing_partner/views/snippets.xml
```diff
@@ -0,0 +1,20 @@
...
+<data>
+
+<template id="mailing_list_subscribe"
+          inherit_id="mass_mailing.mailing_list_subscribe">
+    <xpath expr="//input[@class='js_subscribe_email form-control']"
+           position="before">
+        <input
+            name="name"
+            class="js_subscribe_name form-control"
+            placeholder="your name..."/>
```
pedrobaeza: This should be configurable at snippet level (to include or not to include)

yajo: I don't understand you. The only purpose of this module is to include this input. If you don't want that, don't install it, no?

pedrobaeza: Yes, but you must be able to select the name granularity by snippet. In some occasions, you want the snippet to serve only for logued users, in other, for all introducing the name. I have another doubt now. If you are logged, the name of the user is respected?

yajo: Not right now.

pedrobaeza: OK, so at least last point should be developed.

yajo: Fixed

website_snippet_mass_mailing_partner/controllers/main.py
```diff
@@ -0,0 +1,30 @@
...
+
+class MassMailingPartner(MassMailController):
+    @route()
+    def subscribe(self, list_id, email, **post):
+        """Handle name if provided."""
+        Partner = request.env["res.partner"].sudo()
+        name = post.get("name")
+
+        # Update partner's name, to make it get updated in contact list later
+        Partner.search([("email", "=", email)], limit=1).write({"name": name})
```
hbrunn: check if you actually got a name

pedrobaeza: AFAIK, it doesn't matter, as write will act as dummy in that case

website_snippet_mass_mailing_partner/static/src/js/website_snippet_mass_mailing_partner.js
```diff
@@ -0,0 +1,57 @@
...
+                    self.$target.find('.js_subscribe_name:visible')
+                    .val(data.name)
+                    .attr(
+                        "disabled",
+                        Boolean(data.is_subscriber && data.name.length)
+                    );
+                });
+            }
+        },
+        on_click: function() {
```
hbrunn: you break modularity with this override. `_super` is a mess that doesn't lend itself well for overriding, but still, don't break modularity. If necessary, write the name only after super is done with its stuff. This module anyways allows everyone to write every partner's name given the email address is known, which might not be a good idea in the first place.

website_snippet_mass_mailing_partner/controllers/main.py
yajo: no, @hbrunn is right. We could be removing a name from a partner.

website_snippet_mass_mailing_partner/static/src/js/website_snippet_mass_mailing_partner.js
pedrobaeza: You're right, @hbrunn. @Yajo, better to remove the name update line, and only present the button without the name when someone is logged.

website_snippet_mass_mailing_partner/controllers/main.py
pedrobaeza: Remove this line as explained later.

website_snippet_mass_mailing_partner/static/src/js/website_snippet_mass_mailing_partner.js
yajo: @hbrunn You are right, but if you review the code, I break it because there's no other way. `_super` would do its stuff inside a namespaced promise that is not returned, so there's absolutely no way to know if it succeeded or not.

OTOH I will remove the name update to avoid that case you say, thanks for the insight.

hbrunn: be creative. You can monkey patch `openerp.jsonRpc` to interject a call to `/website_mass_mailing/subscribe` and add a success handler. Or just inject your value then. Or use http://api.jquery.com/ajaxStop. Or write the name into the session before you call super and get this value in your controller

hbrunn: better link: http://api.jquery.com/ajaxSend

yajo:
Now that https://github.com/odoo/odoo/pull/12265 is merged this module will be dramatically simplified.

yajo:
Well, as said before, now the module is much simplified.

The JS calls are intercepted thanks to suggestions by @hbrunn.

No more dependency on mass_mailing_partner. Just install it if you want that extra functionality. Thus, it has been renamed accordingly.

lasley:
👍  Code + runbot

website_mass_mailing_name/__openerp__.py
```diff
@@ -0,0 +1,21 @@
...
+    "summary": "Ask for name when subscribing, and create and/or link partner",
+    "version": "8.0.1.0.0",
+    "category": "Website",
+    "website": "https://tecnativa.com/",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "mass_mailing",
```
pedrobaeza, line 15: website?

rafaelbn:
Tested in runbot :+1:

website_mass_mailing_name/__openerp__.py
yajo, line 15: mass_mailing depends on website.

carlosdauden:
:+1:

PR #67, opened 2016-06-03
[8.0][ADD] mail_tracking addon

antespi:
# Mail tracking

This module shows email notification tracking status for any messages in
mail thread (chatter). Each notified partner will have an intuitive icon just
right to his name.
# Configuration

If you're using a multi-database installation (with or without dbfilter option)
where /web/databse/selector returns a list of more than one database, then
you need to add `mail_tracking` addon to wide load addons list
(by default, only `web` addon), setting `--load` option.
For example, `--load=web,mail_tracking`
# Usage

When user sends a message in mail_thread (chatter), for instance in partner
form, then an email tracking is created for each email notification. Then a
status icon will appear just right to name of notified partner.

**NOTE**: It is not possible to test functionally this addon because runbot do not send any email. So no email tracking object are created too. 

@Tecnativa


mail_tracking/README.rst
```diff
@@ -0,0 +1,107 @@
...
+=============
+Mail tracking
+=============
+
+This module shows email notification tracking status for any messages in
+mail thread (chatter). Each notified partner will have an intuitive icon just
+right to his name.
+
+
+Configuration
```
pedrobaeza: This is installation, not configuration

mail_tracking/README.rst
```diff
@@ -0,0 +1,107 @@
...
+Usage
+=====
+
+When user sends a message in mail_thread (chatter), for instance in partner
+form, then an email tracking is created for each email notification. Then a
+status icon will appear just right to name of notified partner.
+
+These are all available status icons:
+
+.. |check| image:: http://image005.flaticon.com/25/png/16/25/25404.png
```
pedrobaeza: That images are persistent. If not, you should link to local resources.

mail_tracking/README.rst
```diff
@@ -0,0 +1,107 @@
...
+.. |dcheck| image:: http://image005.flaticon.com/1/png/16/88/88051.png
+   :width: 15px
+
+.. |error| image:: http://image005.flaticon.com/25/png/16/25/25298.png
+   :width: 10px
+
+.. |clock| image:: http://image005.flaticon.com/25/png/16/25/25692.png
+   :width: 10px
+
+.. |unknown| image:: http://image005.flaticon.com/25/png/16/25/25342.png
```
pedrobaeza: This image is also grey. Or is this that there's little difference? Are these icons the real ones?

antespi: This icons are not the real ones because I use FontAwesome (already included in bootstrap and Odoo core) but I didn't find how to insert a FontAwesome icons in a reStructuredText file.
But an image is better than thounsand words, isn't it?

pedrobaeza:
Remove .gitkeep file. What is this? If it's a regular file that can be generated by any editor, consider to put it on .gitignore (but it will be a problem through each repository).

mail_tracking/README.rst
antespi: OK, I can download them and call locally

mail_tracking/models/ir_mail_server.py
```diff
@@ -0,0 +1,93 @@
...
+            email_bcc=email_bcc, reply_to=reply_to, attachments=attachments,
+            message_id=message_id, references=references, object_id=object_id,
+            subtype=subtype, headers=headers,
+            body_alternative=body_alternative,
+            subtype_alternative=subtype_alternative)
+        return msg
+
+    def _tracking_email_get(self, message):
+        tracking_email_id = False
+        try:
```
pedrobaeza: Instead of try, you can use `message['X-Odoo-Tracking-ID'].isdigit()` to see if it's a number.

mail_tracking/README.rst
pedrobaeza: Of course, but then generate your own images through snapshots properly masked without background.

antespi:
.gitkeep is a convention. Git ignores empty directories, so if you add a non-ignored useless file in a directory, then the directory structure is kept.
It's like a note to remember to translate the addon, I'll do it BTW

mail_tracking/models/mail_tracking_email.py
```diff
@@ -0,0 +1,165 @@
...
+    sender = fields.Char(string='Sender email', readonly=True)
+    state = fields.Selection([
+        ('error', 'Error'),
+        ('deferred', 'Deferred'),
+        ('sent', 'Sent'),
+        ('delivered', 'Delivered'),
+        ('opened', 'Opened'),
+        ('rejected', 'Rejected'),
+        ('spam', 'Spam'),
+        ('unsub', 'Unsubscribed'),
```
pedrobaeza, line 45: This state is going to be used?

pedrobaeza:
Please, don't use that "trick"/convention. Put a "Known issues" instead if you're not going to do it, or do it :wink: !

mail_tracking/models/ir_mail_server.py
antespi: This try is protecting also to KeyError exception

mail_tracking/models/mail_tracking_email.py
antespi, line 45: Yes, not in this addon, but in other addon that inherits from it and implement this event. It is an standard event in mailing. There is an SMTP header for this also.

antespi:
All remarks answered, README fixed and squashed
@pedrobaeza thanks for your review

mail_tracking/models/ir_mail_server.py
lasley: I agree with Better to Ask Forgiveness Than Permission, but it would be best to catch both exceptions explicitly instead of everything. You never know when something can go wrong outside of what is expected.

pedrobaeza: If the key is not present always, then do this:

```
message.get('X-Odoo-Tracking-ID', '').isdigit()
```

mail_tracking/views/mail_tracking_email_view.xml
```diff
@@ -0,0 +1,119 @@
...
+        </tree>
+    </field>
+</record>
+
+<record model="ir.ui.view" id="view_mail_tracking_email_search">
+    <field name="name">mail.tracking.email.search</field>
+    <field name="model">mail.tracking.email</field>
+    <field name="arch" type="xml">
+        <search string="MailTracking email search">
+            <field name="sender" string="Email"
```
pedrobaeza, line 82: Include here also a direct filter for recipients

antespi, line 82: It is included in filter_domain attribute. This direct filter search in sender and recipient fields

pedrobaeza, line 82: Ah, OK, I see.

mail_tracking/models/ir_mail_server.py
antespi: OK, done

antespi:
Fixed, translated to spanish and squashed

rafaelbn:
This is a incredible module!!! :smiley: 
Tested in customer enviroment :+1:

antespi:
Description icon added

mail_tracking/controllers/main.py
```diff
@@ -0,0 +1,53 @@
...
+            'ip': request.remote_addr,
+            'user_agent': request.user_agent,
+            'os_family': request.user_agent.platform,
+            'ua_family': request.user_agent.browser,
+        }
+
+    @http.route('/mail/tracking/open/<string:db>'
+                '/<int:tracking_email_id>/blank.gif',
+                type='http', auth='none')
+    def mail_tracking_open(self, db, tracking_email_id, **kw):
```
yajo, line 28: Wouldn't it be better to overload [track_mail_open](https://github.com/odoo/odoo/blob/8.0/addons/mass_mailing/controllers/main.py#L10-L18)?

mail_tracking/models/ir_mail_server.py
```diff
@@ -0,0 +1,91 @@
...
+        headers = headers or {}
+        headers['X-Odoo-Tracking-ID'] = tracking_email_id
+        return headers
+
+    def _tracking_email_id_body_get(self, body):
+        body = body or ''
+        tracking_email_id = False
+        # https://regex101.com/r/lW4cB1/2
+        match = re.search(
+            r'<img [^>]* data-odoo-tracking-email=["\']([0-9]*)["\']', body)
```
yajo, line 23: I'd prefer usage of lxml + XPath. Not a blocker though.

mail_tracking/models/ir_mail_server.py
```diff
@@ -0,0 +1,91 @@
...
+                smtp_encryption=smtp_encryption, smtp_debug=smtp_debug)
+        except Exception as e:
+            if tracking_email:
+                tracking_email.smtp_error(self, smtp_server_used, e)
+            raise
+        if message_id and tracking_email:
+            vals = tracking_email._tracking_sent_prepare(
+                self, smtp_server_used, message, message_id)
+            if vals:
+                self.env['mail.tracking.event'].sudo().create(vals)
```
yajo, line 90: Is sudo really needed? I think it would be useful to record the mail.tracking.event object creator uid. Can this be replaced by an ACL?

mail_tracking/models/mail_mail.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @api.model
+    def send_get_email_dict(self, mail, partner=None):
+        email = super(MailMail, self).send_get_email_dict(
+            mail, partner=partner)
+        m_tracking = self.env['mail.tracking.email']
+        tracking_email = False
+        if mail:
+            vals = self._tracking_email_prepare(mail, partner, email)
+            tracking_email = m_tracking.sudo().create(vals)
```
yajo, line 40: same about sudo

mail_tracking/models/ir_mail_server.py
pedrobaeza, line 90: You can't put ACL, because it depends of which object creates the tracking. You will need a glue module for each of the possible installed module. I don't think it's needed either: you already have the object reference that creates it. You only need to query the creator of this object instead.

mail_tracking/models/mail_tracking_email.py
```diff
@@ -0,0 +1,165 @@
...
+    _name = "mail.tracking.email"
+    _order = 'time desc'
+    _rec_name = 'name'
+    _description = 'MailTracking email'
+
+    name = fields.Char(string="Subject", readonly=True)
+    timestamp = fields.Float(
+        string='UTC timestamp', readonly=True,
+        digits=dp.get_precision('MailTracking Timestamp'))
+    time = fields.Datetime(string="Time", readonly=True)
```
yajo, line 26: Please add `index=True` to fields most subjected to order/search, specially time that is the default order.

mail_tracking/models/mail_tracking_email.py
```diff
@@ -0,0 +1,165 @@
...
+        ('delivered', 'Delivered'),
+        ('opened', 'Opened'),
+        ('rejected', 'Rejected'),
+        ('spam', 'Spam'),
+        ('unsub', 'Unsubscribed'),
+        ('bounced', 'Bounced'),
+        ('soft-bounced', 'Soft bounced'),
+    ], string='State', index=True, readonly=True, default=False,
+        help=" * The 'Error' status indicates that there was an error "
+             "when try to sent the email, for example, 'No valid recipient'\n"
```
yajo: "trying"

mail_tracking/models/mail_tracking_email.py
```diff
@@ -0,0 +1,165 @@
...
+             " * The 'Sent' status indicates that message was succesfully "
+             "sent via outgoing email server (SMTP).\n"
+             " * The 'Delivered' status indicates that message was "
+             "succesfully delivered to recipient Mail Exchange (MX) server.\n"
+             " * The 'Opened' status indicates that message was opened or "
```
yajo: Not sure but [this shouldn't this be "Open"](http://dictionary.cambridge.org/es/gramatica/gramatica-britanica/open-or-opened)?

mail_tracking/models/mail_tracking_email.py
```diff
@@ -0,0 +1,165 @@
...
+             "by recipient Mail Exchange (MX) server.\n"
+             " * The 'Soft bounced' status indicates that message was soft "
+             "bounced by recipient Mail Exchange (MX) server.\n")
+    error_smtp_server = fields.Char(string='Error SMTP server', readonly=True)
+    error_type = fields.Char(string='Error type', readonly=True)
+    error_description = fields.Char(
+        string='Error description', readonly=True)
+    bounce_type = fields.Char(string='Bounce type', readonly=True)
+    bounce_description = fields.Char(
+        string='Bounce description', readonly=True)
```
yajo, line 75: Odoo automatically gives string attribute as the field name, uppercasing the first letter and converting `_` to spaces.

So, by default `bounce_description` gets string `Bounce description`. You can remove the `string=` here and in some fields above if you want.

mail_tracking/models/mail_tracking_email.py
```diff
@@ -0,0 +1,165 @@
...
+import openerp.addons.decimal_precision as dp
+
+_logger = logging.getLogger(__name__)
+
+
+class MailTrackingEmail(models.Model):
+    _name = "mail.tracking.email"
+    _order = 'time desc'
+    _rec_name = 'name'
+    _description = 'MailTracking email'
```
yajo, line 20: It would be nice to make this inherit from `mail.thread`, to allow a conversation to start out from a tracking status.

yajo, line 20: It could be used to log status changes too.

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -0,0 +1,60 @@
+/* © 2016 Antonio Espinosa - <antonio.espinosa@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). */
+
+if (typeof window.jQuery === 'undefined') { throw new Error('Requires jQuery'); }
```
yajo: This is implicit, you can remove it.

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -0,0 +1,60 @@
...
+            instance.mail.MessageCommon.include({
+                init: function (parent, datasets, options) {
+                    this._super(parent, datasets, options);
+                    this.partner_trackings = datasets.partner_trackings || [];
+                }
+            });
+            instance.mail.ThreadMessage.include({
+                bind_events: function () {
+                    this._super();
+                    this.$('.oe_mail_action_tracking').on('click', this.on_tracking_status_clicked);
```
yajo, line 22: I'm afraid AFAIK if you don't call `this.on_tracking_status_clicked` from inside a lambda, the `this` it gets is the `event.target`, not the `instance.mail.ThreadMessage` object.

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -0,0 +1,60 @@
...
+
+        if (instance.mail.MessageCommon === undefined) {
+            instance.mail = function(instance) {
+                instance.mail = mail_orig;
+                instance.mail(instance, instance.mail);
+                mail_inherit();
+            };
+        } else {
+            mail_inherit();
+        }
```
yajo, line 60: :open_mouth: What are you doing here?

mail_tracking/static/src/xml/mail_tracking.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- © 2016 Antonio Espinosa - <antonio.espinosa@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+<template>
```
yajo, line 4: [`<templates>`](https://www.odoo.com/documentation/9.0/reference/qweb.html#id3)

mail_tracking/models/ir_mail_server.py
yajo, line 90: ok

yajo:
I tested on runbot. I expected it to fail since no SMTP server is configured. But after I send the message I get the _unknown_ status:

captura de pantalla de 2016-06-10 10-34-54

When the status should be _error_:

captura de pantalla de 2016-06-10 10-35-23

OTOH, is it possible to make the status icon clickable and make it send you to the tracking document?

mail_tracking/README.rst
```diff
@@ -0,0 +1,112 @@
...
+
+These are all available status icons:
+
+.. |sent| image:: mail_tracking/static/src/img/sent.png
+   :width: 10px
+
+.. |delivered| image:: mail_tracking/static/src/img/delivered.png
+   :width: 15px
+
+.. |opened| image:: mail_tracking/static/src/img/opened.png
```
yajo, line 39: Are you sure these are not 3 checks?

pedrobaeza, line 39: No, 3 checks are when NSA has read your email :stuck_out_tongue_winking_eye:

mail_tracking/controllers/main.py
antespi, line 28: This controller is on mass_mailing addon, which is not a dependency.

mail_tracking/models/ir_mail_server.py
antespi, line 23: re.search is more efficient in this case and this img attribute is specific enougth, IMO, in order to guarantee  that there will be only one in a email body.

mail_tracking/models/mail_mail.py
antespi, line 40: Same @pedrobaeza's answer is a good explanation

mail_tracking/README.rst
yajo, line 39: [I know that feeling.](http://www.sephko.com/2013/07/112.html)

No, [seriously](https://github.com/OCA/social/pull/67/files/129ed1ba89cff54fcee0ab538f2c20093031cea5#diff-9b28e6f1adb58fc99a48502d5946a38dR34).

mail_tracking/models/mail_tracking_email.py
antespi, line 26: Cool

mail_tracking/controllers/main.py
yajo, line 28: Maybe it should... This way you could put tracking stuff under _Marketing_ menu and not under technical features, which would make more sense. And you could use it for tracking mass mailings too. Think about it.

mail_tracking/models/ir_mail_server.py
yajo, line 23: ok

mail_tracking/models/mail_mail.py
yajo, line 40: ok

mail_tracking/controllers/main.py
pedrobaeza, line 28: No, that was a design decision: to have a base module with minimal dependencies. Later, we will have a mass_mailing_tracking with the rest of the stuff (but I still don't think it's a good idea to allow marketing people to see all trackings).

mail_tracking/models/mail_tracking_email.py
antespi, line 75: I write it as a good practice

mail_tracking/models/mail_tracking_email.py
antespi, line 20: This would be a quite recursive, isn't it? Because each message notification sends an email :smile:

mail_tracking/controllers/main.py
yajo, line 28: OK, there's another way to do this. The menus could go into _Messaging_ for users belonging to _Mail tracking_ group*. Surely somebody will need to review this, and giving them admin & technical features privileges is too much for the job. They just need to see the reason of rejection, for instance to delete an email address or to call the customer to tell information manually.

And, besides, [right now all users have read access](https://github.com/OCA/social/pull/67/files/129ed1ba89cff54fcee0ab538f2c20093031cea5#diff-104f5a6b8a70f361300607b4137d4673R2).

mail_tracking/static/src/js/mail_tracking.js
antespi, line 22: That's why this call is not inside a lambda, I'm setting the click callback to this.on_tracking_status_clicked function, a normal assignment

mail_tracking/models/mail_tracking_email.py
yajo, line 75: Not writing what you don't need is a better one IMHO, but it's just recommendation anyway.

mail_tracking/static/src/js/mail_tracking.js
antespi: Why? This is a requirement, isn't it?

mail_tracking/models/mail_tracking_email.py
yajo, line 20: Well you're right. However I'm sure somebody would like a notification when an email delivery fails. Any ideas?

mail_tracking/static/src/js/mail_tracking.js
antespi, line 60: Ahh, this deserve a good comment. I'll do, sorry

mail_tracking/static/src/xml/mail_tracking.xml
antespi, line 4: i think are equivalent. In fact, I copied from Odoo core. Is this deprecated or something?

mail_tracking/static/src/js/mail_tracking.js
yajo: Yes, but it's implicit. I mean, you use jQuery when you call the function at the end, and if it is not defined the browser will throw that error by itself, so no need to do it manually.

yajo: And besides, jQuery _will_ be there.

mail_tracking/static/src/xml/mail_tracking.xml
yajo, line 4: I don't know, but it just does not appear in the docs AFAIK.

antespi:
The forbidden icon means that there is no tracking object, because there was no mail created.
Please @gurneyalex could you tell us how runbot avoid sending emails? Maybe I can do a workaround for runbot testing

mail_tracking/static/src/js/mail_tracking.js
yajo, line 22: But I'm sure you want [`this.do_action(action)`](https://github.com/OCA/social/pull/67/files/129ed1ba89cff54fcee0ab538f2c20093031cea5#diff-43f85605fe16954a8ada1881011a8bd1R44) to reach the method it's calling, and that's not in the DOM element but in the `ThreadMessage` object.

mail_tracking/controllers/main.py
pedrobaeza, line 28: Yeah, but the idea is that they don't need that menu. They already see the tracking in the chatter.

mail_tracking/static/src/xml/mail_tracking.xml
```diff
@@ -0,0 +1,61 @@
...
+            <i class="fa fa-check fa-inverse fa-stack-1x" style="margin-left:-2px;"></i>
+            <i class="fa fa-check fa-stack-1x" style="margin-left:-3px"></i>
+        </span>
+    </t>
+</t>
+
+<t t-extend="mail.thread.message">
+    <t t-jquery="span[t-attf-class='oe_partner_follower']" t-operation="append">
+        <t t-set='tracking' t-value='widget.partner_trackings[partner[0]]'/>
+        <t t-if="widget.options.show_link and tracking[1]">
```
yajo: It's sad because this would be very useful, but nobody adds this "show_link" option in core odoo. IMHO it should just check if the user has rights and then show the link if so, or show it always and let Odoo show the access error, or the option be "hide_link", or default "show_link" to True.

yajo:
But there's a MailDeliveryException. Is that not enough to know there was an error?

mail_tracking/controllers/main.py
yajo, line 28: If https://github.com/OCA/social/pull/67/files/129ed1ba89cff54fcee0ab538f2c20093031cea5#r66583200 is fixed, OK

mail_tracking/static/src/xml/mail_tracking.xml
pedrobaeza: I also think that we can omit `widget.options.show_link` check

yajo: That would be easiest.

mail_tracking/README.rst
antespi, line 39: I'm sure yes, You can check it out here: https://jsfiddle.net/qe74n1g9/1/

yajo, line 39: Ah sorry, I saw 3 elements and did not notice the 2nd was inverted. Ok then.

antespi:
This MaileliveryException is raised after mail creation, when Odoo tries to send the email. But in runbot mail is never created nor sent, but maybe @gurneyalex can tell us more about it.

yajo:
No, sorry, what I mean is that the status is not "unknown". It's "error".

Runbot should get "error", the same as would any unconfigured odoo instance do.

mail_tracking/static/src/xml/mail_tracking.xml
antespi: OK, I'll omit that check

mail_tracking/models/mail_tracking_email.py
antespi, line 20: I think that this is a proactive task. In other words, if you want to see what happened with one message you'll see status icon. I you want to see if any message is bounced, rejected, ... then go to settings and see las tracking emails in that state

mail_tracking/static/src/js/mail_tracking.js
antespi: OK, you're right.

mail_tracking/static/src/js/mail_tracking.js
antespi, line 22: I'm not following you.
This is a simple callback to be called when a `$('.oe_mail_action_tracking')` element is clicked. `this` is `instance.mail.ThreadMessage` in the callback assignment time.

mail_tracking/static/src/xml/mail_tracking.xml
antespi, line 4: You can see mail.xml https://github.com/odoo/odoo/blob/8.0/addons/mail/static/src/xml/mail.xml, for instance

antespi:
@Yajo, @pedrobaeza I think all remarks are done or answered.
Thanks for your review!

mail_tracking/models/mail_tracking_email.py
yajo, line 20: :ok_hand:

mail_tracking/static/src/js/mail_tracking.js
yajo, line 22: But not at executing time. http://api.jquery.com/on/#event-handler says:

> When jQuery calls a handler, the this keyword is a reference to the element where the event is being delivered; for directly bound events this is the element where the event was attached and for delegated events this is an element matching selector. (Note that this may not be equal to event.target if the event has bubbled from a descendant element.) To create a jQuery object from the element so that it can be used with jQuery methods, use $( this ).

mail_tracking/static/src/xml/mail_tracking.xml
yajo, line 4: ok

mail_tracking/static/src/js/mail_tracking.js
yajo, line 60: Better now, but is it not enough to put the asset after the mail one?

mail_tracking/static/src/xml/mail_tracking.xml
```diff
@@ -0,0 +1,61 @@
...
+            <i class="fa fa-check fa-stack-1x" style="margin-left:-3px"></i>
+        </span>
+    </t>
+</t>
+
+<t t-extend="mail.thread.message">
+    <t t-jquery="span[t-attf-class='oe_partner_follower']" t-operation="append">
+        <t t-set='tracking' t-value='widget.partner_trackings[partner[0]]'/>
+        <t t-if="tracking[1]">
+            <span class="mail_tracking oe_mail_action_tracking"
```
yajo, line 47: I still cannot see this in the runbot to test it.

mail_tracking/static/src/js/mail_tracking.js
antespi, line 22: Well, the better proof of what I'm saying is that it works.
If you were right, then it doesn't.

mail_tracking/models/ir_mail_server.py
```diff
@@ -0,0 +1,91 @@
...
+from openerp import models, api, tools
+
+
+class IrMailServer(models.Model):
+    _inherit = "ir.mail_server"
+
+    def _tracking_headers_add(self, tracking_email_id, headers):
+        """Allow other addons to add its own tracking SMTP headers"""
+        headers = headers or {}
+        headers['X-Odoo-Tracking-ID'] = tracking_email_id
```
lasley, line 15: Any idea what happens when there is an Odoo using this module in play on both ends of the email? I'll setup a test case if nobody does, but I'm wondering if there is going to be cross-talk with the tracking IDs.

mail_tracking/models/mail_tracking_email.py
lasley: In this scenario "Opened" is correct, as we are referring to the past-tense form of it - the message was Opened. Comparing this to account.invoice Open status, which is indicating that the record is currently Open

mail_tracking/models/mail_tracking_email.py
```diff
@@ -0,0 +1,165 @@
...
+    date = fields.Date(string="Date", readonly=True)
+    mail_message_id = fields.Many2one(
+        string="Message", comodel_name='mail.message', readonly=True)
+    mail_id = fields.Many2one(
+        string="Email", comodel_name='mail.mail', readonly=True)
+    partner_id = fields.Many2one(
+        string="Partner", comodel_name='res.partner', readonly=True)
+    recipient = fields.Char(string='Recipient email', readonly=True)
+    sender = fields.Char(string='Sender email', readonly=True)
+    state = fields.Selection([
```
lasley, line 37: I may be misunderstanding how this is working, because I don't see reading of the SMTP Diagnostic Code header in order to determine delivery status, but if we are actually inspecting this header it would also be nice to store the status codes header. 

As a mail admin, these are more granular and could help in troubleshooting outbound issues. There are a lot of RFCs for this, and all may not be implemented on each server, but they operate in a way similar to HTTP status codes. A fairly complete list can be seen here - http://www.iana.org/assignments/smtp-enhanced-status-codes/smtp-enhanced-status-codes.xhtml

mail_tracking/models/mail_tracking_event.py
```diff
@@ -0,0 +1,72 @@
...
+    _order = 'timestamp desc'
+    _rec_name = 'event_type'
+    _description = 'MailTracking event'
+
+    recipient = fields.Char(string="Recipient", readonly=True)
+    timestamp = fields.Float(
+        string='UTC timestamp', readonly=True,
+        digits=dp.get_precision('MailTracking Timestamp'))
+    time = fields.Datetime(string="Time", readonly=True)
+    date = fields.Date(string="Date", readonly=True)
```
lasley: Why the two fields for date and datetime? IIRC it's possible to use a date widget on a datetime field in the event that it needs to be displayed w/o the time.

lasley:
Thanks for the great work, I ❤️  it!

mail_tracking/models/ir_mail_server.py
pedrobaeza, line 15: There's no problem, because this header is injected to be recovered after in the Odoo code and immediately removed before being actually sent, so there's no trace of this "trick", even between Odoos.

This is the only way we found for passing the needed values (there's no possibility of using context).

lasley, line 15: Oh cool, that makes sense. I was equating this to a standard tracking ID, which typically makes its way into the MIME & is extracted later. Thanks for the explanation!

mail_tracking/models/mail_tracking_email.py
pedrobaeza, line 37: The opened status is got through a controller that is called through an image inserted in the mail. All tracking systems use this trick.

lasley, line 37: Yeah I was figuring that, but what about the `rejected`, `spam`, `unsub`? Or maybe this is the conversation going on slightly below where these are simply reserved for future use. My bad, this makes sense in hindsight

pedrobaeza, line 37: As commented below, that states can be used in other subsequent modules that can detect that states, for example, using a connector to Mandrill or ElasticMail.

rafaelbn:
What else needed for merging? @yajo @pedrobaeza @lasley do you approve it?

mail_tracking/models/mail_tracking_event.py
antespi: The date field is use to group messages by month in the search view. time field is used as a column in the tree view.

pedrobaeza: Antonio, I think it can be also done with the datetime field, or at least, make it a computed field to not be 2 regular fields.

mail_tracking/models/ir_mail_server.py
antespi, line 15: @lasley This header is only used just before email is sent, in order to recover the tracking object created when the email was created. As @pedrobaeza said, this trick is used because there is no way to do using context.

This header is not read when the email is fetched by the same Odoo using fetchmail, so there will not be cross-talk. 

BTW, it would be great if you test it and share the results.

lasley:
IMO the date and datetime column should be merged using Pedro's suggestion, but everything else is 👍

rafaelbn:
In mail.tracking.email.tree view I think state should be visible

antespi:
@rafaelbn State is shown via line color

mail_tracking/models/mail_tracking_email.py
yajo: Thanks for the clarification.

mail_tracking/models/mail_tracking_event.py
yajo: https://www.odoo.com/nl_NL/forum/help-1/question/how-to-group-by-day-25139#answer_93405

yajo:
There's still https://github.com/OCA/social/pull/67/files#r66598628. I cannot click on the icon to get to the tracking record.

mail_tracking/static/src/js/mail_tracking.js
yajo, line 22: It seems `this.$` magically converts `this` into the current `web.Class` object. Otherwise [`this.do_action`](https://github.com/OCA/social/pull/67/files#diff-43f85605fe16954a8ada1881011a8bd1R43) would fail. I was being tricked by the default jQuery behavior that you can check here: https://jsfiddle.net/Lt94yLc1/

So, after all, this woks fine. Thanks!

yajo:
:+1:

yajo:
I'd only add a little CSS that makes the mouse become a finger pointer when hovering the clickable status icon, and becomes a forbidden cursor when not available for some reason, like haappens in runbot

yajo:
:+1: :+1:

antespi:
@lasley and @pedrobaeza date field have been converted to a computed field (store=True)
@Yajo, now mouse is a pointer if status is clicable

Thank you so much for your reviewing

pedrobaeza:
:+1: when Travis finishes, we can merge this. It has been a great work and a great review!

PR #69, opened 2016-06-15
mail_print module

rami-wafaie:
This module makes pdf reports from the Emails, so the user will be able to print them

### DIFF
--- a/mail_print/README.rst
+++ b/mail_print/README.rst
@@ -1,0 +1,45 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+==========
+Mail Print
+==========
+
+This module makes pdf reports from the Emails, so the user will be able to print them.
+
+Usage
+=====
+
+* go to ...
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/social/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Rami Alwafaie <rami.alwafaie@initos.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/mail_print/README.rst
+++ b/mail_print/README.rst
@@ -11,6 +11,7 @@ Usage
 =====
 
 * go to Technical -> Email -> Messages -> select the Email -> Print
+* Emails can be also downloaded directly from The Messaging dashboard
 
 
 Bug Tracker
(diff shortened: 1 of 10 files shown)


mail_print/README.rst
```diff
@@ -0,0 +1,45 @@
...
+==========
+Mail Print
+==========
+
+This module makes pdf reports from the Emails, so the user will be able to print them.
+
+Usage
+=====
+
+* go to ...
```
rafaelbn: Where is the option to print?

rami-wafaie:
@rafaelbn I added to README file where to find the option to print

rafaelbn:
In this way, only admin usar can print an email. Have you think about user?

OSguard:
@rafaelbn it works if you install it with mail_read_new_window
but having there a button is a nice improvment

mail_print/views/mail_message_report.xml
```diff
@@ -0,0 +1,60 @@
...
+                            <style>
+                                .table1 {width: 50%}
+                                .table2 {width: 100%}
+                                .font {font-family: Helvetica, sans-serif;font-size: 10pt}
+                                .bold {font-weight: bold}
+                                .space {padding-right: 0.3em;}
+                            </style>
+                            <table class="table1 table-condensed">
+                                <tr>
+                                    <td class="font bold"><span>From:</span></td>
```
hbrunn: consider using the `th` tag is you want to signify some cell is a column/row header

mail_print/views/mail_message_report.xml
```diff
@@ -0,0 +1,60 @@
...
+
+        <template id="mail_message">
+            <t t-call="report.html_container">
+                <t t-foreach="docs" t-as="o">
+                    <t t-call="report.internal_layout">
+                        <div class="page">
+                            <style>
+                                .table1 {width: 50%}
+                                .table2 {width: 100%}
+                                .font {font-family: Helvetica, sans-serif;font-size: 10pt}
```
hbrunn: setting a fixed size here but not touching it for everything else can lead to weird effects depending on the browser settings. What is you rationale to pick a font and size in the first place?

mail_print/views/mail_message_report.xml
```diff
@@ -0,0 +1,60 @@
...
+                                <tr>
+                                    <td class="font bold"><span>To:</span></td>
+                                    <td class="font" t-foreach="o.partner_ids" t-as="p"><span t-field="p.name"/> <t t-if="p.email"> <u><span t-field="p.email"></span></u></t></td>
+                                </tr>
+                                <tr>
+                                    <td class="font bold"><span>Subject:</span></td>
+                                    <td class="font"><span t-field="o.subject"/></td>
+                                </tr>
+                                <tr>
+                                    <td class="font bold"><span>Date:</span></td>
```
hbrunn: make al of the bold cells `th`

hbrunn:
this is nice! But https://github.com/OCA/social/pull/69#issuecomment-226554160 would indeed be great. You can inject the button via qweb into https://github.com/OCA/OCB/blob/8.0/addons/mail/static/src/xml/mail.xml#L251, then override https://github.com/OCA/OCB/blob/8.0/addons/mail/static/src/js/mail.js#L930 to attach a handler, and call `this.do_action($your_report_action)` there. Or call https://github.com/OCA/OCB/blob/8.0/addons/web/controllers/main.py#L1236 with your action's xmlid and pass that to `do_action`

rami-wafaie:
@hbrunn Thanks for your comments, will be done in the next days

rami-wafaie:
Print option is added - Review comments are done, Please have a look

print-option

hbrunn:
:+1:

rafaelbn:
:+1:

PR #73, opened 2016-06-22
[FIX] Bad default model_id

kryskool:
When we create a new template email from the record, the default model_id is fill with wrong model (see domain on mail_message_subtype.py)


hbrunn:
actually, the domain is wrong and the context is right: https://github.com/OCA/social/blob/8.0/mail_notification_email_template/models/mail_notification.py#L47

hbrunn:
thanks, :+1:

kryskool:
Hi @hbrunn 

I have some tests yesterday that confirm the behavior,
Now it works fine

Regards,

rafaelbn:
:+1:

PR #75, opened 2016-06-23
Update .travis.yml and enable wkhtml2pdf

dreispt:


pedrobaeza:
Is the Transifex secret wrong?

dreispt:
Not a clue. I just copied the latest file from MQT and deleted comments.

pedrobaeza:
Then don't change it and leave it the other.

dreispt:
Why? The template should provide the most up to date version for the config.
If so we should fix the template.

.travis.yml
```diff
@@ -1,27 +1,41 @@
...
+before_install:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
 
 env:
   global:
   - VERSION="8.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
   - TRANSIFEX_USER='transbot@odoo-community.org'
-  - secure: "GRXFURohRzzBbzRovVHd+cObj1yBe9x9bddKqUo7jEGPo+MYHKfGo1M7disTDq6HPLgCwacbhWRsMavGORB3AEkoNewP7q31MKGGPiS8Com9wMmYUtDPs4ZiuUlakJjvfsmGlpwTWUe1Xv1YSZweXVGIESAn0SwhzKHrSKiTcSFAapTwdZXt0lKIsDcra/xOesJ8p9QjY4mzsgyeLrLUSq/Gpt/icAJeoktJ2c4Z6sMV4ozbaRgMFSMCxXKXQTNtLgFICfiJZ2PAw3BQTYo6iX0sdX3RekvaWVBiFr1iblrFjZleoZ9CnkOEKq0JL8EXHMeX7UBvkSJvQV4M3DPd5GdHA0docKDwgdJKUSP9oxBJ6ElYxG5/bbwOxWfY0x6mh/iJdSuNyrfaUqzaCfhQsckuEw6IgqcMGWLPoO5FSovzAtC+u0+GIqphyFsA25lku9B76eZB8KuqWEUJGQZswIOYUKc0i1A2S/ej6MjCYW0BrazhN5laP/6nu7RHmYs0sIrIaxeJJZG0AaVKWAXmTbcDplF07wzwQau4nk922pH5Bxmt2hB1BOzeOqCGnE3wTOAuacuDIfyJaIpTmNMALWeWk/EkMghDZXRRCEueZO5QR2LdcxJk86x5ANj6sxKVo+nkl+x93ZBsehqELn4rTHwjAaA+Ck3vWRkzfV8nJ3A="
+  - secure: PjP88tPSwimBv4tsgn3UcQAD1heK/wcuSaSfhi2xUt/jSrOaTmWzjaW2gH/eKM1ilxPXwlPGyAIShJ2JJdBiA97hQufOeiqxdkWDctnNVVEDx2Tk0BiG3PPYyhXPgUZ+FNOnjZFF3pNWvzXTQaB0Nvz8plqp93Ov/DEyhrCxHDs=
```
pedrobaeza: The template doesn't have any problem. The problem is that you have changed the Transifex secure line, which is specific for each project.

dreispt:
I now see the explanation at https://github.com/OCA/maintainer-quality-tools/issues/194
I wonder why the template has an uncommented `secure` line, I find that misleading.

.travis.yml
pedrobaeza: Can you keep the current secure then?

PR #76, opened 2016-06-24
[8.0][ADD] mail_opt_out_default addon

astirpe:
This module sets the default value of the field "Opt-Out" to True while creating a Partner.

### DIFF
--- a/mail_opt_out_default/README.rst
+++ b/mail_opt_out_default/README.rst
@@ -1,0 +1,76 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================================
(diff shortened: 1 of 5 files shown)


pedrobaeza:
This can be easily done with default values through standard Odoo, so I see no value added in this module, sorry.

:-1:

lasley:
I second the 👎  - if you need a tutorial on defaults - https://blog.laslabs.com/2015/09/odoo-setting-defaults-fields/

astirpe:
@pedrobaeza @lasley this module is evolved in the meanwhile:
https://github.com/onesteinbv/addons-onestein/tree/10.0/mail_opt_out_default
maybe you could take a look at it and let me know in case it could be of interest for OCA.
In particular, the configuration feature was reviewed and made per Company.
Thank you!

pedrobaeza:
Well, I'm still think similar to my previous comment on this one. Thanks for telling us.

PR #77, opened 2016-07-13
[8.0][IMP][mail_tracking] Email stars and generic webhook

antespi:
- Add a generic webhook for an easier inheritance
- Add a smartbutton in res.partner form to see all tracking emails related to its email address
- Add email score and email stars fields to res partner form and a default simple algorithm to calculate them

cc @Tecnativa

### DIFF
--- a/mail_tracking/models/ir_mail_server.py
+++ b/mail_tracking/models/ir_mail_server.py
@@ -24,7 +24,7 @@ class IrMailServer(models.Model):
         if match:
             try:
                 tracking_email_id = int(match.group(1))
-            except:
+            except:  # pragma: no cover
                 pass
         return tracking_email_id
 
@@ -61,7 +61,7 @@ class IrMailServer(models.Model):
             mail_server = mail_server_ids[0] if mail_server_ids else None
         if mail_server:
             smtp_server_used = mail_server.smtp_host
-        else:
+        else:  # pragma: no cover
             smtp_server_used = smtp_server or tools.config.get('smtp_server')
         return smtp_server_used
 
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -5,6 +5,7 @@
 import logging
 import urlparse
 import time
+import re
 from datetime import datetime
 
 from openerp import models, api, fields, tools
@@ -16,10 +17,13 @@ _logger = logging.getLogger(__name__)
 class MailTrackingEmail(models.Model):
     _name = "mail.tracking.email"
     _order = 'time desc'
-    _rec_name = 'name'
+    _rec_name = 'display_name'
     _description = 'MailTracking email'
 
     name = fields.Char(string="Subject", readonly=True, index=True)
+    display_name = fields.Char(
+        string="Display name", readonly=True, store=True,
+        compute="_compute_display_name")
     timestamp = fields.Float(
         string='UTC timestamp', readonly=True,
         digits=dp.get_precision('MailTracking Timestamp'))
@@ -33,6 +37,9 @@ class MailTrackingEmail(models.Model):
     partner_id = fields.Many2one(
         string="Partner", comodel_name='res.partner', readonly=True)
     recipient = fields.Char(string='Recipient email', readonly=True)
+    recipient_address = fields.Char(
+        string='Recipient email address', readonly=True, store=True,
+        compute='_compute_recipient_address')
     sender = fields.Char(string='Sender email', readonly=True)
     state = fields.Selection([
         ('error', 'Error'),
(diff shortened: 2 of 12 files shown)


mail_tracking/models/mail_tracking_email.py
```diff
@@ -77,13 +84,81 @@ class MailTrackingEmail(models.Model):
         string="Tracking events", comodel_name='mail.tracking.event',
         inverse_name='tracking_email_id', readonly=True)
 
+    @api.model
+    def email_score(self, trackings):
+        """Default email score algorimth"""
```
pedrobaeza: Why this method here? And why not @api.multi in case of belonging here?

mail_tracking/models/mail_tracking_email.py
```diff
@@ -77,13 +84,81 @@ class MailTrackingEmail(models.Model):
...
+            elif tracking.state in ('soft-bounced', 'unsub'):
+                score -= 1.0
+            elif tracking.state in ('delivered',):
+                score += 1.0
+            elif tracking.state in ('opened',):
+                score += 2.0
+        return score
+
+    @api.model
+    def email_stars(self, score):
```
pedrobaeza: The same as the previous question

mail_tracking/models/mail_tracking_email.py
```diff
@@ -33,6 +37,9 @@ class MailTrackingEmail(models.Model):
     partner_id = fields.Many2one(
         string="Partner", comodel_name='res.partner', readonly=True)
     recipient = fields.Char(string='Recipient email', readonly=True)
+    recipient_address = fields.Char(
+        string='Recipient email address', readonly=True, store=True,
```
pedrobaeza, line 41: Don't store this auxiliary field

mail_tracking/models/res_partner.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @api.one
+    @api.depends('tracking_email_ids')
+    def _compute_tracking_emails_count(self):
+        self.tracking_emails_count = len(self.tracking_email_ids)
+
+    @api.multi
+    def write(self, vals):
+        email = vals.get('email')
+        if email:
```
pedrobaeza: It should be `if email is not None:` and check if you are resetting the e-mail to False.

mail_tracking/views/res_partner_view.xml
```diff
@@ -0,0 +1,32 @@
...
+                    type="action"
+                    class="oe_stat_button oe_inline"
+                    icon="fa-envelope-o">
+                <field name="tracking_emails_count"
+                       widget="statinfo"
+                       string="Tracking emails"/>
+            </button>
+        </div>
+        <field name="email" position="after">
+            <field name="email_score" />
```
pedrobaeza: Why both fields?

mail_tracking/models/mail_tracking_email.py
antespi: This is for reusing this in other models (like crm.lead) and inherit only in one place

mail_tracking/models/mail_tracking_email.py
antespi: The same answer

mail_tracking/models/mail_tracking_email.py
antespi, line 41: I need to store, because I search it

mail_tracking/models/mail_tracking_email.py
pedrobaeza: OK, make it @api.multi then

mail_tracking/models/mail_tracking_email.py
pedrobaeza: The same reply then :stuck_out_tongue_winking_eye:

mail_tracking/models/mail_tracking_email.py
pedrobaeza, line 41: OK

mail_tracking/views/res_partner_view.xml
antespi: User can filter using stars (selection field) or score (float field). User need to know score values in order to figure out what value to use when filtering by score.

pedrobaeza: Then put it with a div to display both in one line.

rafaelbn:
@antespi please review runbot and travis

mail_tracking/models/mail_tracking_email.py
```diff
@@ -77,13 +84,99 @@ class MailTrackingEmail(models.Model):
...
+                score += 10.0
+        if score > 100.0:
+            score = 100.0
+        return score
+
+    @api.multi
+    @api.depends('recipient')
+    def _compute_recipient_address(self):
+        for email in self:
+            matches = re.search(r'<(.*@.*)>', email.recipient)
```
hbrunn, line 150: won't you be better off with https://github.com/OCA/OCB/blob/8.0/openerp/tools/mail.py#L606?

mail_tracking/models/res_partner.py
```diff
@@ -0,0 +1,40 @@
...
+    @api.depends('tracking_email_ids.state')
+    def _compute_email_score(self):
+        self.email_score = self.tracking_email_ids.email_score()
+
+    @api.one
+    @api.depends('tracking_email_ids')
+    def _compute_tracking_emails_count(self):
+        trackings = self.env['mail.tracking.email'].search([
+            ('recipient_address', '=ilike', self.email)
+        ])
```
hbrunn: use `count=True`

hbrunn: or https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L1616 to be sure

hbrunn:
it seems you didn't git add `hooks.py`

hbrunn:
superseded by #78 it seems

mail_tracking/models/mail_tracking_email.py
antespi, line 150: This match is not for checking email address, but for extrating it if recipient is in the form `Name <email@domain.com>`

antespi:
This PR only affects to `mail_tracking` addon
https://github.com/OCA/social/pull/78 PR includes these commits, because it depends on this.

They are separated in order to review them separately

Please, open this PR again, in order to continues reviewing process

pedrobaeza:
@hbrunn. They are different things. #78 includes also this one.

hbrunn:
okay thanks, got the point. I really miss lauchpad's depending branch feature

rafaelbn:
@antespi please review coveralls. This PR is ready to be merged :+1:

mail_tracking/models/mail_tracking_email.py
hbrunn, line 150: check your regex against `This <hello@world>>`

yajo, line 150: I think https://github.com/OCA/social/pull/77#discussion_r72045025 should still work for you.

yajo:
Any chance to add some tests? :innocent:

mail_tracking/models/mail_tracking_email.py
antespi, line 150: @hbrunn This only could happend if partner email is `hello@world>` because this form if built by Odoo automatically concatenating partner name and partner email.
This match is not for checking, just for extracting and easy matching by partner email afterwords

antespi:
All green :frog: , rebase and squashed for merging.
Testing controllers using mock. @Yajo, this trick is really awesome.

mail_tracking/models/ir_mail_server.py
```diff
@@ -79,7 +79,7 @@ def send_email(self, message, mail_server_id=None, smtp_server=None,
                 smtp_server=smtp_server, smtp_port=smtp_port,
                 smtp_user=smtp_user, smtp_password=smtp_password,
                 smtp_encryption=smtp_encryption, smtp_debug=smtp_debug)
-        except Exception as e:
+        except Exception as e:  # pragma: no cover
             if tracking_email:
                 tracking_email.smtp_error(self, smtp_server_used, e)
             raise
```
pedrobaeza, line 85: If you raise again the exception, isn't the previous DB operations rollbacked?

antespi, line 85: I have to raise the exception because i'm inheriting the method, so I have to propagate the result, even an exception

antespi, line 85: I don't known if the last operation is rollbacked. This is a cool test case.

antespi, line 85: I've added a new test case for testing SMTP error using mock. I think raising exception will not affect DB operations because is the standard behavior.

lasley, line 85: I took a look at your mocks, and I do not see a test case for an SMTP Exception. In order to test this case, you need to use `mock_object.side_effect = SMTPError`. I only see return_values being set in the mocks.

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -101,13 +99,29 @@ def test_mail_send(self):
...
         with mock.patch(mock_request) as mock_func:
             mock_func.return_value = type('obj', (object,), self.request)
-            res = controller.mail_tracking_open(db, tracking_email.id)
+            res = controller.mail_tracking_open(db, tracking.id)
             self.assertEqual(image, res.response[0])
 
+    def test_smtp_error(self):
+        with mock.patch(mock_send_email) as mock_func:
+            mock_func.return_value = None
+            mock_func.side_effect = Warning('Test error')
```
lasley, line 119: No need to init the error, side_effect works like return_val so simply `Warning` 😉

pedrobaeza:
@antespi, we decided to not squash intermediate commits when fixing comments in review to ease reviewers to see what you have done new instead of reviewing again the full diff, and only squash on merge time or when all is clear. Please don't do it next time.

For me, this is :+1:, but I'm going to wait to @lasley to give his approval before merging.

lasley:
I admittedly die a bit each time I look at that `except Exception`, but it's in the core too so it's only right to replicate here.

I digress- LGTM 👍  thanks for the submission!

mail_tracking/models/mail_tracking_email.py
yajo, line 150: I know I'm late, just letting you know that still the suggested regexp should work for extracting the email as you want. :wink:

PR #81, opened 2016-08-19
[9.0][ADD] mail_as_letter

ThomasBinsfeld:
**Mail as Letter**

This module allows to download an email from the mail composer as a PDF file. The generated file aims to be a formal letter and is based on a QWeb report.

**Usage**

A button added on the mail composer wizzard allows to download the PDF file. Only a email with only one recipient can be downloaded.


ThomasBinsfeld:
@sbidoul

sbidoul:
@ThomasBinsfeld I suggest renaming the button as "Download Letter" and put it next to the Send button, as it is logically a similar type of action (whereas Save a new template is a completely different thing).

image

sbidoul:
@ThomasBinsfeld The position of the address should be on the right similar to invoices:

image

image

mail_as_letter/static/description/icon.svg
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
hbrunn: is it possible you included the svg file (which is not necessary if it's just the oca logo unchanged) and did not include the png?

hbrunn:
:+1:

hbrunn:
how does this relate to #69?

ThomasBinsfeld:
Hello @hbrunn. `mail_as_letter` allows to print an email from the mail composer instead of sending it, in order to make a formal letter from the email. #69 seems to print received emails.

mail_as_letter/static/description/icon.svg
ThomasBinsfeld: I'm not sure to understand.

hbrunn: so my knowledge, odoo searches for .png or .jpg files in this folder, and my assumption was that you accidentally git added the .svg file instead of the png

ThomasBinsfeld: I see, right. I'll fix that, thanks !

sbidoul:
:+1: (code review and test on runbot)

rafaelbn:
Tested un runbot :+1:

PR #82, opened 2016-09-01
[8.0][IMP][mail_tracking] Speed installation time and discard concurrent events

antespi:
Several improvements after using mail_tracking_\* addons in production environment:
- Remove post_init_hook: When installing these addons there's no mail_tracking_mail objects in DB, so there's no point to relate them to any res.partner
- Add pre_init_hook: Create and set a value to compute + store columns (and email_score column), before installing this addon
- Remove compute and api.depends on email_score fields
- Do not save concurrent open or click events: 
  - If there's two concurrent open events (events received with a delta time less than 10 seconds) only the first one will be saved, the second one will be discarded
  - If there's two concurrent click events (events received with a delta time less than 5 seconds, and referred to the same URL) only the first one will be saved, the second one will be discarded.
- Propagate bounce event to mass_mailing stats
- Improvements on mail_tracking_mailgun proposed by @Yajo after merging

@Tecnativa

### DIFF
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -13,8 +13,8 @@ import openerp.addons.decimal_precision as dp
 
 _logger = logging.getLogger(__name__)
 
-EVENT_OPEN_DELTA = 10
-EVENT_CLICK_DELTA = 5
+EVENT_OPEN_DELTA = 10  # seconds
+EVENT_CLICK_DELTA = 5  # seconds
 
 
 class MailTrackingEmail(models.Model):
@@ -293,7 +293,7 @@ class MailTrackingEmail(models.Model):
                 if partners:
                     partners.email_score_calculate()
             else:
-                _logger.info("Concurrent event '%s' discarded", event_type)
+                _logger.debug("Concurrent event '%s' discarded", event_type)
         return event_ids
 
     @api.model
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -13,6 +13,9 @@ import openerp.addons.decimal_precision as dp
 
 _logger = logging.getLogger(__name__)
 
+EVENT_OPEN_DELTA = 10
+EVENT_CLICK_DELTA = 5
+
 
 class MailTrackingEmail(models.Model):
     _name = "mail.tracking.email"
@@ -104,7 +107,7 @@ class MailTrackingEmail(models.Model):
                     obj.write({
                         tracking_field: [(5, False, False)]
                     })
-        return True
+        return objects
 
     @api.model
     def _tracking_ids_to_write(self, email):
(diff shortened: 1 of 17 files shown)


rafaelbn:
Hi @antespi , runbot and travis fails

mail_tracking/models/mail_tracking_email.py
```diff
@@ -13,6 +13,9 @@
 
 _logger = logging.getLogger(__name__)
 
+EVENT_OPEN_DELTA = 10
```
pedrobaeza: Comment that this is in seconds

mail_tracking/models/mail_tracking_email.py
```diff
@@ -256,13 +259,41 @@ def _event_prepare(self, event_type, metadata):
...
+                vals = tracking_email._event_prepare(event_type, metadata)
+                if vals:
+                    event_ids += event_ids.sudo().create(vals)
+                partners = self.tracking_ids_recalculate(
+                    'res.partner', 'email', 'tracking_email_ids',
+                    tracking_email.recipient_address)
+                if partners:
+                    partners.email_score_calculate()
+            else:
+                _logger.info("Concurrent event '%s' discarded", event_type)
```
pedrobaeza: Please, all of these messages as debug, not info

mail_tracking/models/res_partner.py
```diff
@@ -15,13 +15,12 @@ class ResPartner(models.Model):
...
-        string="Email score",
-        compute="_compute_email_score", store=True, readonly=True)
+        string="Email score", readonly=True, default=50.0)
 
-    @api.one
-    @api.depends('tracking_email_ids.state')
-    def _compute_email_score(self):
-        self.email_score = self.tracking_email_ids.email_score()
+    @api.multi
+    def email_score_calculate(self):
```
pedrobaeza, line 21: Why do you remove the compute method? This way, you have the risk of not having updated the value.

mail_tracking_mass_mailing/hooks.py
```diff
@@ -3,26 +3,14 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import logging
-from openerp import api, SUPERUSER_ID
+from openerp.addons.mail_tracking.hooks import column_add_with_value
```
pedrobaeza: Protect the import

mail_tracking/models/res_partner.py
antespi: This is causing a inter-block in PostgreSQL, because it's trying to read tracking email table inside the write of field state.

Now email score calculation is made in event_create method

mail_tracking_mass_mailing/hooks.py
antespi: This addon depends on mail_tracking. I think this is not necessary.

antespi:
@pedrobaeza Comments fixed or answered, thanks

mail_tracking_mass_mailing/hooks.py
pedrobaeza: Yes, it is. This causes server doesn't start if you don't have both modules accessible.

antespi: But then should applies to every `openerp.addons.xxx` import. I see lots of them in odoo and OCA and none protected

mail_tracking/models/res_partner.py
antespi: Comments added to remember why we changed compute method

mail_tracking_mass_mailing/hooks.py
pedrobaeza: Yes, that's the question. Remember @Yajo's PRs for protecting them.

pedrobaeza: Please change this

yajo: I'm thinking about adding the protection [directly in the `__init__.py` file of the addon in the template](https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py), to skip these errors in the future more automatically.

antespi: Could you give me one sample as reference?

pedrobaeza: You have to put:

```
try:
    from openerp.addons.mail_tracking.hooks import column_add_with_value
except ImportError:
    column_add_with_value = False
```

antespi: Done

pedrobaeza:
:+1:

carlosdauden:
:+1:

PR #85, opened 2016-09-08
[8.0][ADD] mass_mailing_sending_queue addon

antespi:
# Mass mailing sending queue

This module adds a sending queue when mass mailing 'Send to All' button is
clicked. This is an aditional queue, there is another one for sending emails.
# Usage

When 'Send to All' button is clicked al mass mailing form, all 'mail.mail'
and 'mail.mail.statistics' objects are created. This process shall take
so much time if recipient list is 10k+ and the famous
"Take a minute to get a coffee, because it's loading..." text appears.

With this new queue user will see that mass mailing is in 'Sending' state until
all emails are sent or failed. Screen will return to mass mailing form
immediately after 'Send to All' button is clicked.

In 'Settings > Technical > Email > Mass mailing sending' admin can monitorize
all running mass mailing sending objects and see
- Recipients pending: How many recipients have no email created yet
- Emails sending: How many emails are created and not sent yet
- Emails sent: How many emails have been sent successfully
- Emails failed: How many emails couldn't be sent because an error occurred

@Tecnativa


antespi:
image

All green :frog:

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Usage
+=====
+
+When 'Send to All' button is clicked al mass mailing form, all 'mail.mail'
+and 'mail.mail.statistics' objects are created. This process shall take
+so much time if recipient list is 10k+ and the famous
+"Take a minute to get a coffee, because it's loading..." text appears.
+
+With this new queue user will see that mass mailing is in 'Sending' state until
```
hbrunn: 'a' or 'the' user

mass_mailing_sending_queue/models/mail_mail_statistics.py
```diff
@@ -0,0 +1,21 @@
...
+    _inherit = 'mail.mail.statistics'
+
+    mass_mailing_sending_id = fields.Many2one(
+        comodel_name='mail.mass_mailing.sending',
+        string="Mass mailing sending", readonly=True)
+
+    @api.model
+    def create(self, vals):
+        res = super(MailMailStatistics, self).create(vals)
+        sending_id = self.env.context.get('mass_mailing_sending_id', False)
```
hbrunn: any reason not to overwrite `value`'s `mass_mailing_sending_id` with this?

mass_mailing_sending_queue/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import api, fields, models, _
+from openerp.exceptions import Warning as UserError
+
+
+class MailMassMailingSending(models.Model):
```
hbrunn: class name should be `MailMassMailing`

mass_mailing_sending_queue/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,43 @@
...
+
+    state = fields.Selection(selection_add=[
+        ('sending', "Sending"),
+    ])
+    pending = fields.Integer(string="Pending", compute='_compute_pending')
+
+    @api.multi
+    def send_mail(self):
+        # Override 'Send to All' button completly.
+        # NOTE: Please, notice that this breaks inheritance chain
```
hbrunn: Another approach would be to have a separate function where you do the queuing. Then you could adapt the UI in a way that it's either the user's choice which button to click or maybe steered by group membership who does what.
Then you could also use this function in your cronjob and profit from whatever other modules do here

mass_mailing_sending_queue/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_mail_mass_mailing_sending_admin,mail_mass_mailing_sending group_system,model_mail_mass_mailing_sending,base.group_system,1,1,1,1
```
yajo: Is there a reason for admins to be the only ones that can see this? Shouldn't this be `base.group_user`?
1. Log in runbot as demo user.
2. Go to Marketing > Mass mailing.
3. Click on "First Newsletter".

```
AccessError

Sorry, you are not allowed to access this document. Only users with the following access level are currently allowed to do that:
- Administration/Settings

(Document model: mail.mass_mailing.sending) 
```

mass_mailing_sending_queue/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<openerp>
+<data noupdate="1">
+
+<record id="sending_batch_size" model="ir.config_parameter" forcecreate="True">
+    <field name="key">mail.mass_mailing.sending.batch_size</field>
+    <field name="value">500</field>
```
yajo, line 10: Would this addon then close https://github.com/odoo/odoo/pull/6386?

mass_mailing_sending_queue/models/mail_mass_mailing.py
yajo: [Looking at the code, I guess the part that gets stuck is `mail.compose.message.send_mail`](https://github.com/odoo/odoo/blob/8.0/addons/mass_mailing/models/mass_mailing.py#L614).

I imagine you could simply call super with a context like `.with_context(quick_queue=True)` and then overload `mail.compose.message.send_mail` and change behavior when its context contains that flag. Is that possible?

hbrunn: sounds good. I make it a habit to namespace my context keys with the addon's name to avoid semantic clashes, not sure if that's necessary here

antespi:
Thanks all for your reviews. We've got some other feedbacks from customers, too. I'm refactoring to fix all comments. I hope I'm finishing today.

mass_mailing_sending_queue/models/mail_mass_mailing.py
yajo: Also, context could be added in the view button instead, and leave this method almost as it is now, but overriding behavior only under that context.

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==========================
+Mass mailing sending queue
+==========================
+
+This module adds a sending queue when mass mailing 'Send to All' button is
```
pedrobaeza: `This module adds a queue for generating mail records when mass mailing...`

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+clicked. This is an aditional queue, there is another one for sending emails.
+
+Configuration
+=============
+
+There is a system parameter to control how many emails are created in each
+cron iteration: 'mail.mass_mailing.sending.batch_size'. By default, 500
```
pedrobaeza: dot at the end.

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==========================
+Mass mailing sending queue
+==========================
+
+This module adds a sending queue when mass mailing 'Send to All' button is
+clicked. This is an aditional queue, there is another one for sending emails.
```
pedrobaeza: `This is an additional queue, apart from the existing one for doing the actual sending.`

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+=============
+
+There is a system parameter to control how many emails are created in each
+cron iteration: 'mail.mass_mailing.sending.batch_size'. By default, 500
+
+
+Usage
+=====
+
+When 'Send to All' button is clicked at mass mailing form, all 'mail.mail'
```
pedrobaeza: Start with: `Without this module, when you click in..`

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+and 'mail.mail.statistics' objects are created. This process shall take
+so much time if recipient list is 10k+ and the famous
+"Take a minute to get a coffee, because it's loading..." text appears.
+
+With this new queue the user will see that mass mailing is in 'Sending' state
```
pedrobaeza: comma after queue

pedrobaeza: Is this state real? Or are you talking about `Generating mails`, but `Sent` is apart from the control you make on the queue?

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+until all emails are sent or failed. Screen will return to mass mailing form
+quickly after 'Send to All' button is clicked.
+
+In 'Mass mailing' form there is a new tab 'Sending tasks' for checking
+sendings history.
```
pedrobaeza: s/sendings/sending

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+
+In 'Settings > Technical > Email > Mass mailing sending' admin can monitorize
```
pedrobaeza: s/admin/allowed users

pedrobaeza: s/monitorize/track

mass_mailing_sending_queue/data/ir_config_parameter.xml
pedrobaeza: Not the same, but it can be considered as a workaround. It depends on how you program your scheduled tasks.

mass_mailing_sending_queue/README.rst
pedrobaeza: Include here the reference to the scheduled task that generates on batch the mails

mass_mailing_sending_queue/__openerp__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Mass mailing sending queue",
+    "summary": "A new queue for sending mass mailing",
+    "version": "8.0.1.0.0",
+    "category": "Social Network",
```
pedrobaeza: Marketing (the same as mass_mailing)

mass_mailing_sending_queue/i18n/es.po
```diff
@@ -0,0 +1,156 @@
...
+
+#. module: mass_mailing_sending_queue
+#: model:ir.model,name:mass_mailing_sending_queue.model_mail_mail_statistics
+msgid "Email Statistics"
+msgstr "Estadísticas de correo electrónico"
+
+#. module: mass_mailing_sending_queue
+#: field:mail.mass_mailing.sending,failed:0
+msgid "Emails failed"
+msgstr "Emails fallidos"
```
pedrobaeza: Change "Emails" by "Correos electrónicos" (don't use anglicisms)

mass_mailing_sending_queue/i18n/es.po
```diff
@@ -0,0 +1,156 @@
...
+
+#. module: mass_mailing_sending_queue
+#: field:mail.mass_mailing.sending,create_date:0
+msgid "Created on"
+msgstr "Creado en"
+
+#. module: mass_mailing_sending_queue
+#: field:mail.mass_mailing.sending,display_name:0
+msgid "Display Name"
+msgstr "Nombre a mostrar"
```
pedrobaeza: I prefer "Nombre mostrado"

mass_mailing_sending_queue/i18n/es.po
```diff
@@ -0,0 +1,156 @@
...
+
+#. module: mass_mailing_sending_queue
+#: view:mail.mass_mailing:mass_mailing_sending_queue.view_mail_mass_mailing_form
+msgid "recipients pending for creating email."
+msgstr "destinatarios pendientes de crear el correo electrónico."
+
+#. module: mass_mailing_sending_queue
+#: view:mail.mass_mailing:mass_mailing_sending_queue.view_mail_mass_mailing_form
+msgid "{'invisible': [('scheduled', '=', 0), ('pending', '=', 0)]}"
+msgstr ""
```
pedrobaeza: Don't leave this empty, because then the translation will be considered. Put the same as source of remove the entry.

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Antonio Espinosa <antonio.espinosa@tecnativa.com>
```
pedrobaeza, line 77: May I deserve to be here due to the design?

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
```diff
@@ -0,0 +1,179 @@
...
+        string="Emails sent", compute='_compute_sent')
+    failed = fields.Integer(
+        string="Emails failed", compute='_compute_failed')
+    error = fields.Char(string="Error message")
+    date_start = fields.Datetime(
+        string="Date start", default=fields.Datetime.now())
+    date_end = fields.Datetime(string="Date end")
+
+    def _batch_size_default(self):
+        return 500
```
pedrobaeza: Why this method? I prefer to hardcode it in the other one. Inheratibility is not a problem here, because you can easily overwrite XML if you want another value in your child module.

mass_mailing_sending_queue/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,94 @@
...
+    def _sendings_get(self):
+        self.ensure_one()
+        return self.env['mail.mass_mailing.sending'].search([
+            ('mass_mailing_id', '=', self.id),
+            ('state', 'in', ('enqueued', 'sending')),
+        ])
+
+    @api.multi
+    def send_mail(self):
+        if self.env.context.get('sending_queue_enabled', False):
```
pedrobaeza: Instead of adding one indentation and one level to the cyclomatic complexity, its' better to put:

```
if self.env.context.get('sending_queue_enabled'):
    return super(MailMassMailing, self).send_mail()
```

and repeat after the same super call

mass_mailing_sending_queue/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,94 @@
...
+            for mailing in self:
+                m_sending = self.env['mail.mass_mailing.sending']
+                sendings = mailing._sendings_get()
+                if sendings:
+                    raise UserError(_(
+                        "There is another sending task running. "
+                        "Please, be pacient. You can see all sending tasks in "
```
pedrobaeza: s/pacient/patient

pedrobaeza: s/all sending tasks/all the sending tasks

mass_mailing_sending_queue/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,94 @@
...
+                })
+                sending_state = 'enqueued'
+                if len(res_ids) < (batch_size / 2):
+                    mailing.with_context(
+                        mass_mailing_sending_id=sending.id,
+                        sending_queue_enabled=False).send_mail()
+                    sending_state = 'sending'
+                sending.state = sending_state
+                mailing.write({
+                    'sent_date': fields.datetime.now(),
```
pedrobaeza: Shouldn't be `fields.Datetime.now()`?

mass_mailing_sending_queue/models/mail_mail.py
```diff
@@ -0,0 +1,21 @@
...
+    _inherit = 'mail.mail'
+
+    def _postprocess_sent_message(self, cr, uid, mail, context=None,
+                                  mail_sent=True):
+        # Read before super, because mail will be removed if sent successfully
+        stats = mail.statistics_ids
+        res = super(MailMail, self)._postprocess_sent_message(
+            cr, uid, mail, context=context, mail_sent=mail_sent)
+        if stats:
+            for stat in stats:
```
pedrobaeza: Use `filtered` here for avoiding an if sentence and one indentation

mass_mailing_sending_queue/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,94 @@
...
+            sending = self.env['mail.mass_mailing.sending'].browse(sending_id)
+            res_ids = sending.get_recipient_batch(res_ids)
+        return res_ids
+
+    @api.multi
+    def _compute_pending(self):
+        for mailing in self:
+            sendings = mailing._sendings_get()
+            self.pending = (
+                sum(sendings.mapped('pending')) +
```
pedrobaeza: Use `len`, not `sum`

mass_mailing_sending_queue/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,94 @@
...
+import logging
+
+
+class MailMassMailing(models.Model):
+    _inherit = 'mail.mass_mailing'
+
+    state = fields.Selection(selection_add=[
+        ('sending', "Sending"),
+    ])
+    pending = fields.Integer(string="Pending", compute='_compute_pending')
```
pedrobaeza: This seems to be more a boolean than an integer. Use the suffix `_count` for noting the number

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
```diff
@@ -0,0 +1,179 @@
...
+    state = fields.Selection([
+        ('draft', "Draft"),
+        ('enqueued', "Enqueued"),
+        ('sending', "Sending"),
+        ('sent', "Sent"),
+        ('error', "Error"),
+    ], string="State", required=True, copy=False, default='enqueued')
+    mass_mailing_id = fields.Many2one(
+        string="Mass mailing", comodel_name='mail.mass_mailing', readonly=True)
+    pending = fields.Integer(
```
pedrobaeza: This seems to be more a boolean than an integer. Use the suffix `_count` for noting the number

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
```diff
@@ -0,0 +1,179 @@
...
+        string="Recipients pending", compute='_compute_pending')
+    sending = fields.Integer(
```
pedrobaeza: This seems to be more a boolean than an integer. Use the suffix `_count` for noting the number

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
```diff
@@ -0,0 +1,179 @@
...
+        string="Emails sending", compute='_compute_sending')
+    sent = fields.Integer(
```
pedrobaeza: This seems to be more a boolean than an integer. Use the suffix `_count` for noting the number

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
```diff
@@ -0,0 +1,179 @@
...
+        string="Emails sent", compute='_compute_sent')
+    failed = fields.Integer(
```
pedrobaeza: This seems to be more a boolean than an integer. Use the suffix `_count` for noting the number

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
```diff
@@ -0,0 +1,179 @@
...
+            _logger.info("Sending [%d] mass mailing [%d] '%s' (%s)",
+                         sending.id, sending.mass_mailing_id.id,
+                         sending.mass_mailing_id.name, sending.state)
+            sending._process()
+        return True
+
+    @api.multi
+    def _compute_pending(self):
+        for sending in self:
+            sending.pending = 0
```
pedrobaeza: This is not needed, as all is initialized to its default value. Simply do:

```
for sending in self.filtered(lambda x: x.state == 'enqueued'):
    sending.pending = len(sending.pending_recipients())
```

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
```diff
@@ -0,0 +1,179 @@
...
+            if sending.state == 'enqueued':
+                sending.pending = len(sending.pending_recipients())
+
+    @api.multi
+    def _compute_sending(self):
+        m_stats = self.env['mail.mail.statistics']
+        for sending in self:
+            sending.sending = 0
```
pedrobaeza: The same about 0

mass_mailing_sending_queue/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_mail_mass_mailing_sending_admin,mail_mass_mailing_sending group_marketing_user,model_mail_mass_mailing_sending,base.group_user,1,1,1,0
```
pedrobaeza, line 2: Regular user shouldn't create this, isn't it? Can't you replicate the same permissions as mass mailing object? Or do they come from several modules?

mass_mailing_sending_queue/views/mass_mailing_view.xml
```diff
@@ -0,0 +1,58 @@
...
+                        <field name="sent"/>
+                        <field name="failed"/>
+                    </tree>
+                </field>
+            </page>
+        </notebook>
+    </field>
+</record>
+
+<!--record model="ir.ui.view" id="view_mail_mass_mailing_kanban">
```
pedrobaeza: Remove this

pedrobaeza (changes requested):

pedrobaeza:
A better icon would be also appreciated

mass_mailing_sending_queue/models/mail_mass_mailing.py
```diff
@@ -2,42 +2,93 @@
...
     state = fields.Selection(selection_add=[
         ('sending', "Sending"),
     ])
     pending = fields.Integer(string="Pending", compute='_compute_pending')
+    mass_mailing_sending_ids = fields.One2many(
+        comodel_name='mail.mass_mailing.sending', readonly=True,
+        inverse_name='mass_mailing_id', string="Sending tasks")
+
+    def read_group(self, cr, uid, domain, fields, groupby, **kwargs):
+        res = super(MailMassMailing, self).read_group(
```
yajo, line 25: I'd appreciate some comment on why this is needed.

mass_mailing_sending_queue/README.rst
yajo: I think he means sendings as messages that have been sent. Not sure if it is correct, [it seems like yes](https://en.wiktionary.org/wiki/sending#Noun), but our favorite human spell checker @elicoidal can help here for sure :wink:

Maybe "deliveries" would be better, not sure...

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
```diff
@@ -40,54 +47,51 @@ def _batch_size_get(self):
         return batch_size
 
     @api.multi
-    def pending_recipients(self):
+    def pending_emails(self):
+        self.ensure_one()
+        return self.env['mail.mail.statistics'].search([
+            ('mass_mailing_sending_id', '=', self.id),
```
yajo: Not sure if what I'm going to say makes sense at all, but if you use `'in', self.ids`, you can remove the `self.ensure_one()`

mass_mailing_sending_queue/views/mass_mailing_view.xml
```diff
@@ -10,15 +10,49 @@
...
+    <field name="name">Show Sending state</field>
+    <field name="model">mail.mass_mailing</field>
+    <field name="inherit_id" ref="mass_mailing.view_mail_mass_mailing_kanban"/>
+    <field name="arch" type="xml">
+        <kanban position="attributes">
+            <attribute name="default_group_by">state2</attribute>
+        </kanban>
+    </field>
+</record-->
```
yajo: Clean this please

mass_mailing_sending_queue/README.rst
elicoidal: "cron iteration: 'mail.mass_mailing.sending.batch_size' (Default value is 500)."

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+
+There is a system parameter to control how many emails are created in each
+cron iteration: 'mail.mass_mailing.sending.batch_size'. By default, 500
+
+
+Usage
+=====
+
+When 'Send to All' button is clicked at mass mailing form, all 'mail.mail'
+and 'mail.mail.statistics' objects are created. This process shall take
```
elicoidal: "This process might take
a long time if the recipient list is 10k+ and the famous
"Take a minute to get a coffee, because it's loading..." text might appear."

mass_mailing_sending_queue/README.rst
elicoidal: With this new queue,mass mailing will appear in 'Sending' state to the user...

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+so much time if recipient list is 10k+ and the famous
+"Take a minute to get a coffee, because it's loading..." text appears.
+
+With this new queue the user will see that mass mailing is in 'Sending' state
+until all emails are sent or failed. Screen will return to mass mailing form
```
elicoidal: After 'Send to All' button is clicked, the user will land to the mass mailing form.

mass_mailing_sending_queue/README.rst
elicoidal: @Yajo lol: I couldnot compete with @pedrobaeza for the python part so I specialized myself :p 

"In 'Mass mailing' form, a new tab "Sent mail" has been added where the user can check the Sent mails history."

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+quickly after 'Send to All' button is clicked.
+
+In 'Mass mailing' form there is a new tab 'Sending tasks' for checking
+sendings history.
+
+In 'Settings > Technical > Email > Mass mailing sending' admin can monitorize
+all running mass mailing sending objects and see
+
+* Recipients pending: How many recipients have no email created yet
```
elicoidal: Pending Recipients: Number of recipients for which the email is not yet created.

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+* Start date: Date when user press 'Send to All' button
+* Emails sending: How many emails are created and not sent yet
```
elicoidal: Mails to be sent: number of emails waiting to be sent.

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+* Emails sent: How many emails have been sent successfully
```
elicoidal: Sent emails: number of emails successfully sent.

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+* Emails failed: How many emails couldn't be sent because an error occurred
```
elicoidal: Failed mails: number of unsent emails due to error.

mass_mailing_sending_queue/README.rst
```diff
@@ -0,0 +1,89 @@
...
+
+NOTE: User will not be able to send the same mass mailing again if there is
```
elicoidal: if there is one sending running
=> if another one is ongoing. An UserError exception is raised in this case.

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
```diff
@@ -0,0 +1,179 @@
...
+
+    state = fields.Selection([
+        ('draft', "Draft"),
+        ('enqueued', "Enqueued"),
+        ('sending', "Sending"),
+        ('sent', "Sent"),
+        ('error', "Error"),
+    ], string="State", required=True, copy=False, default='enqueued')
+    mass_mailing_id = fields.Many2one(
+        string="Mass mailing", comodel_name='mail.mass_mailing', readonly=True)
```
elicoidal: Please review naming according to my comments above

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
elicoidal: why not a property?

elicoidal (changes requested):

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
yajo: Yes, actually given this is the default only for cases where the ir.config_parameter is unset, you can hardcode it elsewhere. Just set a parameter and the hardcoded default goes away.

mass_mailing_sending_queue/data/ir_config_parameter.xml
antespi: This is the number of emails created in each cronjob iteration. Those emails are add to the standard emails queue (mail.mail.process_email_queue), processed by other cronjob. These two queues are independent.

odoo/odoo#6386 PR fix `process_email_queue`, so I think that this addon doesn't close it.

mass_mailing_sending_queue/README.rst
antespi: That's it,  and mass mailing will go to `Sent` state when all emails are created and sent.

mass_mailing_sending_queue/data/ir_config_parameter.xml
yajo: But at least, if you install this, then that bug would not affect you, right?

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
antespi: But I have to set to pending_count = 0 the sending objects not in 'enqueued' state

antespi: This if here is to avoid calling `pending_recipients()` method if state is not `enqueued`

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
antespi: Same answer

mass_mailing_sending_queue/security/ir.model.access.csv
antespi: These permissions are the same as `mail.mass_mailing` and group is the same also

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
antespi: I think so

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
pedrobaeza: Please see my code and test it. It works the same and more efficiently.

mass_mailing_sending_queue/i18n/es.po
yajo: AFAIK it's the same.

mass_mailing_sending_queue/security/ir.model.access.csv
yajo: When you push send emails, you create queue records, so I guess you need those create perms. Another thing is about editing, not sure on that.

antespi:
Fixes done and tests reviewed

pedrobaeza:
Check Travis status

hbrunn (approved):

antespi:
All green. Thanks @hbrunn for your approval.
Thansk all for your reviews

antespi:
Some compute improvements added

pedrobaeza (approved):

pedrobaeza:
We have to squash on merge

yajo (approved):

mass_mailing_sending_queue/models/mail_mail_statistics.py
```diff
@@ -0,0 +1,20 @@
...
+
+    mass_mailing_sending_id = fields.Many2one(
+        comodel_name='mail.mass_mailing.sending',
+        string="Mass mailing sending", readonly=True)
+
+    @api.model
+    def create(self, vals):
+        res = super(MailMailStatistics, self).create(vals)
+        res.mass_mailing_sending_id = \
+            self.env.context.get('mass_mailing_sending_id', False)
```
yajo, line 19: Wouldn't this be automatic if you used `default_mass_mailing_sending_id` context key?

pedrobaeza:
Incorrect rebase

mass_mailing_sending_queue/models/mail_mass_mailing_sending.py
```diff
@@ -146,7 +146,9 @@ def cron(self):
                          sending.id, sending.mass_mailing_id.id,
                          sending.mass_mailing_id.name, sending.state)
             # Process sending using user who created it
-            sending.sudo(user=sending.create_uid.id)._process()
+            sending = sending.sudo(user=sending.create_uid.id)
```
pedrobaeza, line 149: Do you need both calls or it's enough with the last one?

antespi: I need to change user and set his context (lang and tz). Then it is equivalent to standard click on "Send to all" button, but executed by cronjob.

pedrobaeza (approved):
OK, then please squash and I'll merge at last

antespi:
Rebased and squashed

pedrobaeza:
As @elicoidal's remarks has been attended already, I merge.

PR #86, opened 2016-09-08
[FIX][website_mail_snippet_table_edit] Compatibilize with website_less.

yajo:
Without this patch, `website_less` disables this addon.

With this patch, it improves it.

Win-win! :blush:

@Tecnativa @rafaelbn @pedrobaeza


antespi:
:+1: Tested in production environment with website_less installed

pedrobaeza:
:+1: 

Merging as it's a needed patch

PR #91, opened 2016-09-20
[FIX] cope with partner without users, fixes #84

### LINKED ISSUE #84: Unable to send email on new template
module: mail_notification_email_template
version: 8.0

**Steps to reproduce**
- Install Sales module and activate technical settings
  -configure and test one outgoing server
- Create a new template for Quotation sent ( can keep most values the same as per the Discussions example) , although my final goal is to utilize fields from the sales Quotation
- Create a new draft quotation , add new followers (internal users and one customer)
- make sure that followers is part of the Quotation sent sub type (eg they will receive emails)
- Print the quotation

**Current behavior**
- Emails fails most of the time to all followers, although i have had cases where it succeeds in sending to one.
  -remainder of emails is stuck showing failed status.
  Log file shows :
  **odoo attributeerror: 'mail.notification' object has no attribute '_get_access_link'**

**Expected behavior**
-All emails are supposed to go out using the configured template.

(i have had success where i manually configure the to Email field with email addresses, thus the issue is possibly with the formatting of the to(partners) field, which is currently
${object.partner_id.ids|join(',')}

I currently don't have access to my server logs, but will attach clean logs asap. 
Is there any verbose mode i might need to activate?  (there is debugging on the outgoing servers, but it doesn't seem to generate more data)

hbrunn:


pedrobaeza (approved):

rafaelbn:
👍

rafaelbn (approved):

adrienpeiffer (approved):

PR #93, opened 2016-09-28
add mail_bcc module

tpalan:
Add a module for specifying a BCC address in mail templates. All emails sent with this template are also sent to the specified address.


tpalan:
I get many "E501 line too long (x > 79 characters)" errors from travis in code I copied from the original odoo mail_mail.py and mail_template.py.
Also a"[E8102(invalid-commit), MailMail.send] Use of cr.commit() directly" from the original code, I did not add the commit.
What should be done about that?

pedrobaeza:
First of all, if you are copying/pasting code, there's a great chance of that you're not doing right the thing, as the inheritance avoids all of this. If it can't be avoided, then you should solve these errors and also replace (if possible) the cr.commit() by ORM operations. Odoo code can be considered perfect and it has lower quality checks than OCA's one.

rafaelbn:
Why do you need this module? Odoo already sent one single mail per recipient.

tpalan:
you are right, I did not see that. I'm closing this pull request, sorry to bother

Issue #90, opened 2016-09-18
No mail_footer_notified_partners after changing the api.one to api.model


sysadminmatmoz:
It seems that mail_footer_notified_partners does not function properly since the last module update. No errors in log, but the list of the notified partners is missing since the update (tried on several servers, databases and instances).

sterbhai:
Confirmed: After finding this Issue reversed the changes in V8 and it is working again.

PR #96, opened 2016-10-10
Return with context, solves #90.

sysadminmatmoz:
In response to issue #90 .
With api.model the footer didn't get populated with followers as it was supposed to. This commit restores the functionality, but using @api.multi to avoid the obsolete @api.one.

### DIFF
--- a/mail_footer_notified_partners/models/mail_followers.py
+++ b/mail_footer_notified_partners/models/mail_followers.py
@@ -57,7 +57,7 @@ class MailNotification(models.Model):
             ctx.update({
                 'partners_to_notify': partners_to_notify,
             })
-        return super(MailNotification, self)._notify(
+        return super(MailNotification, self.with_context(ctx))._notify(
             message_id,
             partners_to_notify=partners_to_notify,
-            force_send=force_send, user_signature=user_signature, context=ctx)
+            force_send=force_send, user_signature=user_signature)


sysadminmatmoz:
PS: I strongly doubt the Travis fails because of this specific commit.

mail_footer_notified_partners/models/mail_followers.py
```diff
@@ -48,16 +48,17 @@ def get_signature_footer(
 
         return footer
 
-    @api.model
+    @api.multi
```
hbrunn: this function cannot be `multi` because `message_id` is not an id of a notification, but of a message

mail_footer_notified_partners/models/mail_followers.py
```diff
@@ -48,16 +48,17 @@ def get_signature_footer(
...
             user_signature=True):
+        self.ensure_one()
         ctx = self.env.context.copy()
         if not self.env.context.get('mail_notify_noemail'):
             ctx.update({
                 'partners_to_notify': partners_to_notify,
             })
-        return super(MailNotification, self)._notify(
-            message_id,
+        return super(MailNotification, self._model)._notify(
```
hbrunn: why not `self`?

mail_footer_notified_partners/models/mail_followers.py
```diff
@@ -48,16 +48,17 @@ def get_signature_footer(
...
+            self.env.cr, self.env.uid, self.id,
```
hbrunn: why do you call `super` v7 style?

hbrunn (changes requested):
https://travis-ci.org/OCA/social/jobs/166323807#L2053 looks very much like being caused by your commit

mail_footer_notified_partners/models/mail_followers.py
sysadminmatmoz: It's actually api.multi with ensure one

mail_footer_notified_partners/models/mail_followers.py
sysadminmatmoz: It was just a revert to the last working state... just to make it work, since a client of mine stressed about the missing partners in footer since the last update of the module - with api.model it simply didn't work.

mail_footer_notified_partners/models/mail_followers.py
hbrunn: I repeat: this cannot be multi or one because `message_id` is from a different model than `mail.notification`

sysadminmatmoz:
The purpose of this commit was just to speed things up, since it's months that the module doesn't function at all. A PR, even not a great one, gets an attention quicker than just an issue report.

If anyone has a better way to repair it... please, do... :smile: All I want is an issue that I opened - finally closed.

With the code as it is at least the module does the job.

mail_footer_notified_partners/models/mail_followers.py
hbrunn: I rather think the problem is that the original code calls super awkwardly. Try the original code, but instead of passing the context, say `super(MailNotification, self.with_context(ctx)._notify(message_id, partners_to_notify=partners_to_notify, force_send=force_send, user_signature=user_signature)`

sysadminmatmoz: I'll look into it :smile:

sysadminmatmoz:
@hbrunn  Done and tested (600 users DB, so quick response).

mail_footer_notified_partners/models/mail_followers.py
```diff
@@ -57,7 +57,7 @@ def _notify(
             ctx.update({
                 'partners_to_notify': partners_to_notify,
             })
-        return super(MailNotification, self)._notify(
+        return super(MailNotification, self.with_context(ctx))._notify(
             message_id,
             partners_to_notify=partners_to_notify,
             force_send=force_send, user_signature=user_signature, context=ctx)
```
hbrunn: remove the second context here, and possibly squash your commits. If you also add 'fixes #NNN' in the commit message, the issue will be closed automatically when this is merged

sysadminmatmoz: Without that context we get back to where we were ... the notified partners aren't added/returned. That's actually the only change I made.

hbrunn: I talk about `, context=ctx`, I wonder why this doesn't raise in the first place, but it's definitely not correct

hbrunn (approved):

sysadminmatmoz:
@pedrobaeza @jbeficent  Can one of you review and merge it? It's not really something, but the module doesn't work without it.

Thanks.

pedrobaeza (approved):

PR #97, opened 2016-10-11
[8.0][IMP] Performance issues in mail_tracking addons

antespi:
This PR fix some performance issues:
- [Fixed] In some cases `self.env.cr` is not closed properly.
- Add `index=True` to `time` and `recipient_address` fields of `mail.tracking.email` model
- Many2many relation between partners (and mail contacts) and mail tracking emails deleted
- `email_score` is now a non-stored computed field

Also includes some new features for processing bounced emails easier:
- Adds a menu for listing mass mailing emails and events
- Add a boolean field `email_bounced` to partners and mail contacts. To easy filter them and do manual operations: call the partner, remove the partner, opt-out, ... Also prepare easy inherited methods for programming automatic operations in customer addon.

### DIFF
--- a/mail_tracking/controllers/main.py
+++ b/mail_tracking/controllers/main.py
@@ -11,17 +11,37 @@ _logger = logging.getLogger(__name__)
 BLANK = 'R0lGODlhAQABAIAAANvf7wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw=='
 
 
-def _env_get(db):
+def _env_get(db, callback, tracking_id, event_type, **kw):
+    res = 'NOT FOUND'
     reg = False
-    try:
-        reg = registry(db)
-    except OperationalError:
-        _logger.warning("Selected BD '%s' not found", db)
-    except:  # pragma: no cover
-        _logger.warning("Selected BD '%s' connection error", db)
-    if reg:
-        return api.Environment(reg.cursor(), SUPERUSER_ID, {})
-    return False
+    current = http.request.db and db == http.request.db
+    env = current and http.request.env
+    if not env:
+        try:
+            reg = registry(db)
+        except OperationalError:
+            _logger.warning("Selected BD '%s' not found", db)
+        except:  # pragma: no cover
+            _logger.warning("Selected BD '%s' connection error", db)
+        if reg:
+            _logger.info("Creating a new environment for database '%s'", db)
+            env = api.Environment(reg.cursor(), SUPERUSER_ID, {})
+    else:
+        # make sudo when reusing environment
+        env = env(user=SUPERUSER_ID)
+    if env:
+        try:
+            res = callback(env, tracking_id, event_type, **kw)
+            if not current:
+                env.cr.commit()
+        except Exception as e:
+            if not current:
+                env.cr.rollback()
+            raise e
+        finally:
+            if not current:
+                env.cr.close()
+    return res
 
 
 class MailTrackingController(http.Controller):
@@ -35,49 +55,37 @@ class MailTrackingController(http.Controller):
             'ua_family': request.user_agent.browser or False,
         }
 
+    def _tracking_open(self, env, tracking_id, event_type, **kw):
+        tracking_email = env['mail.tracking.email'].search([
+            ('id', '=', tracking_id),
+        ])
+        if tracking_email:
+            metadata = self._request_metadata()
+            tracking_email.event_create('open', metadata)
+        else:
+            _logger.warning(
+                "MailTracking email '%s' not found", tracking_id)
+
+    def _tracking_event(self, env, tracking_id, event_type, **kw):
+        metadata = self._request_metadata()
+        return env['mail.tracking.email'].event_process(
+            http.request, kw, metadata, event_type=event_type)
+
     @http.route('/mail/tracking/all/<string:db>',
                 type='http', auth='none')
     def mail_tracking_all(self, db, **kw):
-        env = _env_get(db)
-        if not env:
-            return 'NOT FOUND'
-        metadata = self._request_metadata()
-        response = env['mail.tracking.email'].event_process(
-            http.request, kw, metadata)
-        env.cr.commit()
-        env.cr.close()
-        return response
+        return _env_get(db, self._tracking_event, None, None, **kw)
 
     @http.route('/mail/tracking/event/<string:db>/<string:event_type>',
                 type='http', auth='none')
     def mail_tracking_event(self, db, event_type, **kw):
-        env = _env_get(db)
-        if not env:
-            return 'NOT FOUND'
-        metadata = self._request_metadata()
-        response = env['mail.tracking.email'].event_process(
-            http.request, kw, metadata, event_type=event_type)
-        env.cr.commit()
-        env.cr.close()
-        return response
+        return _env_get(db, self._tracking_event, None, event_type, **kw)
 
     @http.route('/mail/tracking/open/<string:db>'
                 '/<int:tracking_email_id>/blank.gif',
                 type='http', auth='none')
     def mail_tracking_open(self, db, tracking_email_id, **kw):
-        env = _env_get(db)
-        if env:
-            tracking_email = env['mail.tracking.email'].search([
-                ('id', '=', tracking_email_id),
-            ])
-            if tracking_email:
-                metadata = self._request_metadata()
-                tracking_email.event_create('open', metadata)
-            else:
-                _logger.warning(
-                    "MailTracking email '%s' not found", tracking_email_id)
-            env.cr.commit()
-            env.cr.close()
+        _env_get(db, self._tracking_open, tracking_email_id, None, **kw)
 
         # Always return GIF blank image
         response = werkzeug.wrappers.Response()
--- a/mail_tracking/controllers/main.py
+++ b/mail_tracking/controllers/main.py
@@ -30,6 +30,9 @@ def _env_get(db, callback, tracking_id, event_type, **kw):
         # make sudo when reusing environment
         env = env(user=SUPERUSER_ID)
     if env:
+        # This try-except-finally is equivalent to a with statement
+        # We assure that new created cursor si commit/rollback/close
+        # in any case
         try:
             res = callback(env, tracking_id, event_type, **kw)
             if not current:
--- a/mail_tracking/controllers/main.py
+++ b/mail_tracking/controllers/main.py
@@ -17,33 +17,23 @@ def _env_get(db, callback, tracking_id, event_type, **kw):
     current = http.request.db and db == http.request.db
     env = current and http.request.env
     if not env:
-        try:
-            reg = registry(db)
-        except OperationalError:
-            _logger.warning("Selected BD '%s' not found", db)
-        except:  # pragma: no cover
-            _logger.warning("Selected BD '%s' connection error", db)
-        if reg:
-            _logger.info("Creating a new environment for database '%s'", db)
-            env = api.Environment(reg.cursor(), SUPERUSER_ID, {})
+        with api.Environment.manage():
+            try:
+                reg = registry(db)
+            except OperationalError:
+                _logger.warning("Selected BD '%s' not found", db)
+            except:  # pragma: no cover
+                _logger.warning("Selected BD '%s' connection error", db)
+            if reg:
+                _logger.info("New environment for database '%s'", db)
+                with reg.cursor() as new_cr:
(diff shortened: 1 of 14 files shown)


mail_tracking/models/mail_tracking_email.py
```diff
@@ -294,11 +294,11 @@ def event_create(self, event_type, metadata):
                 vals = tracking_email._event_prepare(event_type, metadata)
                 if vals:
                     event_ids += event_ids.sudo().create(vals)
-                partners = self.tracking_ids_recalculate(
-                    'res.partner', 'email', 'tracking_email_ids',
-                    tracking_email.recipient_address)
-                if partners:
-                    partners.email_score_calculate()
+                # partners = self.tracking_ids_recalculate(
```
pedrobaeza: Why not make it non stored?

antespi:
We are running this in one production instance, if works well, then I'll add new tests to improve codecov

mail_tracking/controllers/main.py
```diff
@@ -11,17 +11,37 @@
...
+            res = callback(env, tracking_id, event_type, **kw)
+            if not current:
+                env.cr.commit()
+        except Exception as e:
+            if not current:
+                env.cr.rollback()
+            raise e
+        finally:
+            if not current:
+                env.cr.close()
```
yajo: Not sure if you can use [`with env.cr:` or `with env.cr.savepoint():`](https://github.com/odoo/odoo/blob/9.0/openerp/sql_db.py#L385-L414)

yajo:
Could you please add some docstrings and comments, specially in the performance-critical parts?

mail_tracking/controllers/main.py
antespi: This pattern is equivalent to `with`, but in this case if more explicit and easier to debug if we need it in future. See: http://effbot.org/zone/python-with-statement.htm

lasley: I don't see a functional difference though? This pattern [fails Lint](https://travis-ci.org/OCA/social/jobs/167296604#L333) & I am not sure that debugging is a good enough reason to keep it.

mail_tracking/models/mail_tracking_email.py
```diff
@@ -23,14 +24,19 @@ class MailTrackingEmail(models.Model):
     _rec_name = 'display_name'
     _description = 'MailTracking email'
 
+    # This table is going to growth fast and to infinite, so we index:
```
lasley: s/growth/grow

mail_tracking/models/mail_tracking_email.py
```diff
@@ -87,70 +93,49 @@ class MailTrackingEmail(models.Model):
...
-                if trackings:
-                    obj.write({
-                        tracking_field: [(6, False, trackings.ids)]
-                    })
-                else:
-                    obj.write({
-                        tracking_field: [(5, False, False)]
-                    })
-        return objects
+    def _email_score_tracking_filter(self, domain, order='time desc',
```
lasley, line 96: `@api.model`

mail_tracking/models/mail_tracking_email.py
```diff
@@ -87,70 +93,49 @@ class MailTrackingEmail(models.Model):
...
                 score -= 10.0
             elif tracking.state in ('delivered',):
-                score += 5.0
+                score += 1.0
             elif tracking.state in ('opened',):
-                score += 10.0
+                score += 5.0
         if score > 100.0:
             score = 100.0
+        if score < 0.0:
```
lasley: `elif` - save some resources - if it already matched the 100, it definitely won't be less than 0

mail_tracking/models/mail_tracking_email.py
```diff
@@ -87,70 +93,49 @@ class MailTrackingEmail(models.Model):
...
+        Must return a value beetwen 0.0 and 100.0
+        - Bad reputation: Value between 0 and 50.0
+        - Unknown reputation: Value 50.0
+        - Good reputation: Value between 50.0 and 100.0
+        """
         score = 50.0
-        trackings = self._email_score_tracking_filter()
-        for tracking in trackings:
+        for tracking in self:
             if tracking.state in ('error',):
```
lasley: This would be cleaner & more efficient in a dictionary:

```
states = {
    'error': -50.0,
    'opened': 5.0,
   ...
}
score += states.get(tracking.state, 0.0)
```

mail_tracking/models/mail_tracking_email.py
```diff
@@ -287,20 +272,21 @@ def _concurrent_events(self, event_type, metadata):
...
                 if vals:
                     event_ids += event_ids.sudo().create(vals)
-                partners = self.tracking_ids_recalculate(
-                    'res.partner', 'email', 'tracking_email_ids',
-                    tracking_email.recipient_address)
-                if partners:
-                    partners.email_score_calculate()
+                    # Commit to DB to release exclusive lock
+                    if not testing:
+                        self.env.cr.commit()  # pragma: no cover
```
lasley: Why can we not use savepoint here?

mail_tracking/models/res_partner.py
```diff
@@ -8,39 +8,45 @@
...
-    tracking_email_ids = fields.Many2many(
-        string="Tracking emails", comodel_name="mail.tracking.email",
-        readonly=True)
+    # tracking_emails_count and email_score are non-store fields in order
+    # to improve performance
+    # email_bounced is store=True and index=True field in order to filter
+    # in tree view for processing bounces easier
     tracking_emails_count = fields.Integer(
-        string="Tracking emails count", store=True, readonly=True,
+        string="Tracking emails count", readonly=True, store=False,
```
lasley: AFAIK computed, non-stored attrs without an inverse are implicitly read only

mail_tracking_mass_mailing/hooks.py
```diff
@@ -5,7 +5,7 @@
 import logging
 try:
     from openerp.addons.mail_tracking.hooks import column_add_with_value
-except ImportError:
+except ImportError:  # pragma: no cover
```
lasley: Please remove this pragma ~~& vote on OCA/maintainer-quality-tools#372~~

lasley (changes requested):

mail_tracking/models/mail_tracking_email.py
antespi: You're right

mail_tracking/controllers/main.py
antespi: Thanks @lasley, I didn't read transaction related guidelines before. In a multi-db instance we need  to get a new env (and cursor) because ORM can initialize it (several databases, none selected). But I can make it better following guidelines. I'm doing it right now

mail_tracking_mass_mailing/hooks.py
antespi: Already merged, thanks :smile:

mail_tracking/models/mail_tracking_email.py
antespi: when using savepoint, if everything is going well (no exceptions), then transaction is committed?
And, if we are in testing mode, transaction is committed too?

mail_tracking/models/res_partner.py
antespi: ok

mail_tracking/models/mail_tracking_email.py
lasley: Hmmm ok looking at the intent here, we may not want a Savepoint. It's not immediately committing the transaction, and will roll back entirely if another transaction afterwards fails (assuming that transaction is also not within a savepoint). 

Think of it like a save-guard - if the transaction within the Savepoint fails, it will rollback just within that savepoint. If the transaction fails outside of a savepoint, all mutations & savepoints in that transaction will fail.

Note this is a Postgres feature, so it's significantly easier to play with in direct SQL if you're curious. Surprisingly, the [Postgres docs on Savepoint](https://www.postgresql.org/docs/8.1/static/sql-savepoint.html) are pretty nice & [the Odoo implementation](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/odoo/sql_db.py#L396) matches well.

Where is the exclusive lock that is mentioned in the comments actually created?

antespi: I can see this Odoo code: https://github.com/odoo/odoo/blob/d81258ab8e0dcb6c0df3a47ab8084ec99b7d3e24/openerp/sql_db.py#L390
And this PostgreSQL doc: https://www.postgresql.org/docs/9.1/static/sql-release-savepoint.html

According to that, RELEASE SAVEPOINT just remove the savepoint. But in this case, I need to have the event (and mail state) visible to other isolated cursors and release exclusivelock to mail_tracking_email and mail_tracking_event tables. With this commit, then other writes (for instance to res_partner table) can be done after reading mail_tracking_email or mail_tracking_event tables.

antespi: After thinking twice, I remove this commit. Now the addon is not reading mail_tracking_email nor email_tracking_event tables after create/write on them, so I don't need this.

antespi:
@lasley Thanks for your review. Changes done

mail_tracking/controllers/main.py
```diff
@@ -17,33 +17,23 @@ def _env_get(db, callback, tracking_id, event_type, **kw):
...
+            except OperationalError:
+                _logger.warning("Selected BD '%s' not found", db)
+            except:  # pragma: no cover
+                _logger.warning("Selected BD '%s' connection error", db)
+            if reg:
+                _logger.info("New environment for database '%s'", db)
+                with reg.cursor() as new_cr:
+                    new_env = api.Environment(new_cr, SUPERUSER_ID, {})
+                    res = callback(env, tracking_id, event_type, **kw)
+                    new_env.cr.commit()
```
lasley, line 32: Seeing as we've confirmed that we're awesome and doing things right, we'll need to disable lint on this line (I'm assuming, wait for Travis to find out for sure):

```
new_env.cr.commit()  #pylint: disable=E8102
```

lasley (approved):
One note on pylint, but we'll wait for Travis for confirmation

antespi:
pylint is cool :smile:

mail_tracking/controllers/main.py
antespi: Confirmed

pedrobaeza: This check is for avoiding bad commit() calls without control. If you are sure about the use, then we can put the pylint disable sentence.

antespi:
These improvements have been running well in a production instance today, with one mass mailing sending and Mailgun webhooks been received and 6 workers.

rafaelbn (approved):
OK, thanks

mail_tracking/controllers/main.py
```diff
@@ -35,49 +48,37 @@ def _request_metadata(self):
             'ua_family': request.user_agent.browser or False,
         }
 
+    def _tracking_open(self, env, tracking_id, event_type, **kw):
+        tracking_email = env['mail.tracking.email'].search([
+            ('id', '=', tracking_id),
```
pedrobaeza, line 53: Use `search_count` for optimizing this (see the tip in https://www.odoo.com/documentation/8.0/reference/orm.html#common-orm-methods)

mail_tracking/models/res_partner.py
```diff
@@ -8,39 +8,45 @@
...
-        readonly=True)
+    # tracking_emails_count and email_score are non-store fields in order
+    # to improve performance
+    # email_bounced is store=True and index=True field in order to filter
+    # in tree view for processing bounces easier
     tracking_emails_count = fields.Integer(
-        string="Tracking emails count", store=True, readonly=True,
+        string="Tracking emails count", readonly=True,
         compute="_compute_tracking_emails_count")
+    email_bounced = fields.Boolean(string="Email bounced", index=True)
```
pedrobaeza: Don't put string, as it's automatically inferred from the variable name.

mail_tracking/models/res_partner.py
```diff
@@ -8,39 +8,45 @@
...
     email_score = fields.Float(
-        string="Email score", readonly=True, default=50.0)
+        string="Email score", readonly=True,
```
pedrobaeza: The same about the string attribute

pedrobaeza (changes requested):

mail_tracking/controllers/main.py
antespi: I use tracking_email recordset after that

pedrobaeza (approved):

PR #99, opened 2016-10-11
[MIG] mail_attach_existing_attachment: Migrated to 10.0

bouvyd:


sbidoul:
@dbo-odoo Thanks for the migration. Works fine on runbot. The view is a bit disrupted but it was probably the case on 9 too and I don't see how to do better.

image

cc/ @adrienpeiffer

sbidoul (approved):

pedrobaeza (approved):

sbidoul:
@dbo-odoo can you update the branch in the runbot url in README.rst?

bouvyd:
@sbidoul View is indeed weird, but I kept it as-is since this is a migration (tbh, idk why someone decided to put a group with col=4 in this view; having two groups instead would have prevented weird reflows... but well, such is life)
readme updated

adrienpeiffer (approved):
:+1:

PR #101, opened 2016-10-11
[MIG] mail_compose_select_lang: Migrated to 10.0

bouvyd:
Note this migration comes with a feature loss as translated reports
is no longer possible via a python override (translations are done
in the report templates themselves; therefore every report would
need an inherited version).

This module was migrated from 8.0 to 10.0 directly, which explains
the consequent diff.


pedrobaeza:
Uhm, the lost feature is very important, because we want to have all the sent mail in the same language (not attachment in one lang, and body in another). Do you know any possible hack/override to restore the functionality?

I have also formed a new company called `Tecnativa`, so it would be good to switch the author name to it, but I can do it after the merge. You should deserve also to be included at least in the contributor list in the README, because the work you have made is more than simple renaming namespaces.

rafaelbn:
Hi @dbo-odoo , thanks for your initiative!

This module was created in v8 because Odoo functionality was not correct one when  a Company (like Odoo) work with customers from different countries. We discussed it here: https://github.com/odoo/odoo/issues/7896

Odoo has 2 customers:
- Tecnativa which speaks Spanish
- Akretion which speaks French

When Odoo send an email, sale order, invoice, mass mailing, etc... The correct form is to send it in Spanish to Tecnativa and in French to Akretion (res.partner language), but in Odoo before this module it was impossible (always with res.user language).

We need to resolve the problem of attachments because send an email in Spanish with an attachment in English has no sense. Same for french o any language.

In fact the best could be if we don't need this module and Odoo do this automatically 😏 

cc @alexis-via

bouvyd:
@rafaelbn I'm not going to solve issues in Odoo standard here as it is not the place nor the time to discuss it; I already do that internally and in support during my work time, I'd rather not do it on my free time as well :sleeping:  :smile: 

I clearly understand that the feature loss is a bummer and kills 50% of the module's usefulness, however I've been thinking about it for some time and cannot really find a technical solution that isn't fugly:
- kinda maybe somehow possibly could work: override the with_context function to support the force_lang key, but holy cow no please don't do that
- change the value of the partner's `lang` field and reset it after the rendering?
  I certainly won't commit anything of this kind under my name, tbh.

I honestly have no other idea at the moment... If anyone can find of something clever, I'm all ears!

rafaelbn:
Of course @dbo-odoo ! 😄  but that info must be in this PR to aware about it. I someone install in a new DB is OK, but migration DB from v8 to v10 should be solved as if not customers could have a problem sending attachments in a different language that they sent. This is could be named as a functional stable policy 😉 . I cannot read this cases in https://github.com/odoo/odoo/wiki/Contributing#what-does-stable-mean but I suppose that if Odoo migrates their customers from v5-v6-v7-v8-v9-v10 in all this changes customers don't loose data or functionality. Maybe it changes but not loose.

hbrunn:

PR #102, opened 2016-10-12
[8.0][ADD] Sendgrid modules for transactional e-mails

ecino:
Two modules for connecting e-mails and mass_mailing with [Sendgrid](http://sendgrid.net)

This enables the use of Sendgrid transactional e-mails with templates and Sendgrid Event Notifications hook for tracking all e-mails sent.

Note: we don't have any tests because it requires an API Key (cannot be used without).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ virtualenv:
   system_site_packages: true
 
 install:
+  - pip install sendgrid
   - git clone --depth=1 https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
--- a/mail_sendgrid/README.rst
+++ b/mail_sendgrid/README.rst
@@ -1,0 +1,103 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+========
+SendGrid
+========
+
+This module integrates
+`SendGrid <https://sendgrid.com/>`_ with Odoo. It can send transactional emails
+through SendGrid, using templates defined on the
+`SendGrid web interface <https://sendgrid.com/templates>`_. It also supports
+substitution of placeholder variables in these templates. The list of available
+templates can be fetched automatically.
+E-mails sent through SendGrid will be tracked using Sendgrid Webhook Events.
+
+Installation
+============
+You need to install python-sendgrid v3 API in order to install the module.
+
+Configuration
+=============
+
+You can add the following system parameters to configure the usage of SendGrid:
+
+* ``mail_sendgrid.substitution_prefix`` Any symbol or character used as a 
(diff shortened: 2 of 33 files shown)


ecino:
I cannot find where I should add the `sendgrid` library dependency so that it is installed in TRAVIS with PIP. Can someone help?

hbrunn:
in https://github.com/OCA/social/blob/10.0/.travis.yml#L29, just add another line there.
For the tests: This amount of untested code is not acceptable. If there's no way to obtain a demo/testing api key there, you'll have to use mock or the like to simulate this service's responses.

ecino:
I won't have time to write tests now.

There are no demo API keys. Do you think I could create a free account and use it for that ? But this may expose it for people to use it for sending e-mails. How then could I set the api key in the configuration of odoo (if possible encrypted).

Otherwise I don't see how we can test the code if we cannot make any real call of the sendgrid api (a lot of methods use it). I'm not familiar with mock.

hbrunn:
no tests = no merge
Travis allows to encrypt variables: https://docs.travis-ci.com/user/environment-variables I'm not so sure if a free account is a good idea, because it's probably subject to rate limiting and the like, tests can be run very often. You're better off with simulating the services responses, which is where mock comes into play.
About mock: It's a good thing to be familiar with anyways: https://mock.readthedocs.io/en/latest - the idea would be to replace the relevant function(s) in the library you use by an object that returns whatever you need it to return in order to run successful tests.

ecino:
Thanks for the tips, I'll have a look at it when I have time ! I'll be back with the pull request when I can manage to implement the tests.

In the meantime if someone is interested in those modules and would like to help with the tests, please contact me.

hbrunn:
I reopened your PR and set it to work in progress. This way it's still visible, but nobody is going to bug you about this until you request to remove the wip label

pedrobaeza:
You can mock your tests the same we make for mail_tracking_mailgun module. You can also use that "framework" to build on top of it your module instead of repeating everything.

ecino:
Ok I will have a look at this.

rafaelbn:
Hi @ecino , thanks!

No one commented by please make to diferents pull requiest for:

1 - mail_sendgrid
2 - mail_sendgrid_mass_mailing

We will review (1) and after mergin it, we review (2).

Thanks

Maybe @antespi you could make a review here, please

pedrobaeza:
@rafaelbn, my comment is that they should use mail_tracking framework and with this, they only need mail_tracking_sendgrid. Mass mailing part is automatically done.

ecino:
Like I said, I won't have time now to do all those requests. I'm not sure if mass_mailing can be removed. When I wrote the module the mass tracking was poorly done and didn't work with sendgrid. I guess it changed in between but this will take time to adapt.

If someone is interested, I'd be happy to have some help (PR are welcome).

pedrobaeza:
mass_mailing_tracking provides the needed framework also for having full capacities on mass mailing without the need of making anything more specific for Sendgrid. That's why I say it's worth to take a look, @ecino, you will have to make little effort to adapt it to this system as you know already Sendgrid, but obviously it's your choice.

As is, the module can't enter OCA right now.

ecino:
New PR for version 9.0

PR #104, opened 2016-10-13
[FIX][mail_tracking] display_name of tracking_emails with no subject

ecino:
There was a bug preventing a mail_tracking object to be displayed if the related e-mail has no subject.


pedrobaeza (approved):

antespi:
This is already included in https://github.com/OCA/social/pull/97/files#diff-20ecd59e476b4a5bf788df1261025614R147

pedrobaeza:
Then let's close it in favor of #97

Issue #94, opened 2016-10-06
[8.0] mail_notification_email_template: E-mails created with template are created with admin rights


ecino:
When using the module mail_notification_email_template, when sending a message from a chatter box, the generated e-mail is from user Administrator, and it creates user access rights problems when displaying the message.

When the module is not installed (or disable template), the generated e-mail is from the user sending the message. I think that this line is the problem :

https://github.com/OCA/social/blob/8.0/mail_notification_email_template/models/mail_notification.py#L47

But I don't see how it results of an administrator record creation.

hbrunn:
the difference should rather be that the email generated has a different object it is attached to. Can you check this?

ecino:
This is the e-mail generated when clicking "send message" from the chatter box below a res.partner object :
![mail1](https://cloud.githubusercontent.com/assets/8435180/19228588/136f1890-8ec7-11e6-82cd-deaa3a202301.PNG)

This is the same e-mail generated when the module mail_notification_email_template is installed:
![mail2](https://cloud.githubusercontent.com/assets/8435180/19228669/ab37f87c-8ec7-11e6-9887-c0b890f12b0a.PNG)

When I try to access it with the user that sent the message I get an access error.

I looked at the messages object and as you suggested there is a slight difference. Without module installed, it creates a comment type message attached to the res partner with "notified partners" holding followers of the document.

With the mail_notification_email_template module installed, it creates the same message, but it creates as well another message of type email attached to the mail.notification model. This record is created by administrator and may be the one that creates my access rights problems.

First: I don't know why it needs two mail.message records ?
Second: The second created record, if necessary, should as well be created by user instead of administrator, right?

hbrunn:
First: One email to be sent out, the message to be shown under the object in question
Second: It should. I'm not aware of any sudoing involved here, so I'd suggest to debug through to see what's going on there.

hbrunn:
you can probably circumvent your problem by setting auto delete on the email template

ecino:
First: but why without the module, it creates only one mail.message record ? It still generate an e-mail, plus the message is in the mail thread in the object. 

Auto delete was set on the template, so this does not change the behaviour. Maybe I have another module that does this sudo? I will try to debug and see if I can find the issue. Could you confirm if you have the same behaviour?

ecino:
Even with debugging I cannot explain this behaviour, so any help would be appreciated. Here are a few other observations I made:
- email generated with the module have not the "is notification" checkbox
[...]

PR #105, opened 2016-10-13
[FIX] Author of mail.notification when using notification templates

ecino:
This fixes issue #94 where notifications created with e-mail templates were from user Administrator.

With this fix, the mail.message created when sending the e-mail is from the user that created the message instead of Administrator.

### DIFF
--- a/mail_notification_email_template/models/mail_notification.py
+++ b/mail_notification_email_template/models/mail_notification.py
@@ -33,6 +33,7 @@ class MailNotification(models.Model):
                 continue
             custom_values = {
                 'references': message.parent_id.message_id,
+                'author_id': message.author_id.id
             }
             if message.res_id and hasattr(
                 self.env[message.model], 'message_get_email_values'


hbrunn (approved):
thanks. Next time, you can use one of the issue keywords in your commit message to have merging your PR close the issue automatically: https://help.github.com/articles/closing-issues-via-commit-messages/ (no need to change anything now)

hbrunn:
the only thing I ask myself: In a standard setting, this code is supposed to run as the user who posted the message. Are you sure there's no other code around it that causes the method to be called in a sudo context? Otherwise, I can't really explain myself how admin comes in there

ecino:
Sure, the core of Odoo calls the sudo user, but I don't know how to circumvent this in another way (it shouldn't call sudo in my opinion but there are maybe some cases where user rights would prevent to create notifications).

Here is the context, where the _notify_email function is called with sudo:
https://github.com/odoo/odoo/blob/8.0/addons/mail/mail_followers.py#L261

hbrunn:
thanks for doing my homework, this one slipped me. Then we definitely need your patch

pedrobaeza (approved):

lasley (approved):

PR #106, opened 2016-10-13
[8.0][FIX][website_mail_snippet_fixed] Clean before saving.

yajo:
Before this patch, when a user dragged a fixed snippet on the email body, the result had some attributes that are internally used in Odoo to know if you are replacing a part of a view, and which.

This made that editing the placeholders changed directly the snippet template itself.

On other cases, when this addon was removed and then reinstalled, it made that previous templates raise a MissingError when saving any changes on them.

Chances are that if you have (maybe unnoticedly) hit this bug, you have some snippet XMLID from this addon marked as `noupdate=1`, so the recommended upgrade path is to uninstall this addon and reinstall it.

@Tecnativa

### DIFF
--- a/website_mail_snippet_fixed/__openerp__.py
+++ b/website_mail_snippet_fixed/__openerp__.py
@@ -4,10 +4,12 @@
 {
     "name": "Fixed-Width Layout Snippets for Writing Emails",
     "summary": "560px width extra building blocks",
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):

rafaelbn (approved):
Thanks!

antespi:
:+1:

PR #107, opened 2016-10-14
[8.0][IMP][mail_tracking] Allow to define a custom mail_tracking base url

antespi:
This minor change allows to change the base url used for mail_tracking image, but if not defined it will fallbacks to `web.base.url`.
In some cases, this base URL must be different to `web.base.url`

### DIFF
--- a/mail_tracking/__openerp__.py
+++ b/mail_tracking/__openerp__.py
@@ -5,7 +5,7 @@
 {
     "name": "Email tracking",
     "summary": "Email tracking system for all mails sent",
-    "version": "8.0.3.0.0",
+    "version": "8.0.3.0.1",
     "category": "Social Network",
     "website": "http://www.tecnativa.com",
     "author": "Tecnativa, "
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -171,7 +171,9 @@ class MailTrackingEmail(models.Model):
                 fields.Date.from_string(email.time))
 
     def _get_mail_tracking_img(self):
-        base_url = self.env['ir.config_parameter'].get_param('web.base.url')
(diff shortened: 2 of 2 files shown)


mail_tracking/__openerp__.py
```diff
@@ -5,7 +5,7 @@
 {
     "name": "Email tracking",
     "summary": "Email tracking system for all mails sent",
-    "version": "8.0.3.0.0",
+    "version": "8.0.3.0.1",
```
yajo, line 8: Since this is a new feature, and not a fix, this should be `8.0.3.1.0` IMHO.

yajo (changes requested):

mail_tracking/__openerp__.py
antespi: No update is needed, and if you do not change anything in ir.config_parameter, then your addon will run exactly in the same way. That's why I use the last number.

yajo: http://semver.org/#summary

yajo (approved):
Although I think versioning is incorrect, I'm not blocking this for that.

mail_tracking/__openerp__.py
pedrobaeza: @Yajo, we use this syntax: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

pedrobaeza (approved):

mail_tracking/__openerp__.py
yajo: Thanks for the pointer, although again it says that backwards-compatible new features increment `y`. This is the case.

However I'm not going to insist any more on this, it's not worth the effort, just do as you please :wink:

pedrobaeza:
Fast-tracking as basic change

PR #114, opened 2016-10-17
[FIX] email_template_qweb: Decode body when encoded

pedrobaeza:
If the result is of type unicode, render method encodes it in utf-8.
We need to decode it in that case so that the rendering results correct.

### DIFF
--- a/email_template_qweb/models/mail_template.py
+++ b/email_template_qweb/models/mail_template.py
@@ -33,6 +33,8 @@ class MailTemplate(models.Model):
                             'email_template': this,
                         })
                     )
+                    # Some wizards, like when sending a sales order, need this
+                    # fix to display accents correctly
                     if isinstance(result[record_id]['body_html'], unicode):
                         result[record_id]['body_html'] = (
                             result[record_id]['body_html'].decode('utf-8')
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please review @hbrunn @lasley @Yajo 

cc @Tecnativa

lasley (approved):
LGTM

Care to add a test covering this edge case?

hbrunn (approved):
yes, the extra test would be great

pedrobaeza:
It's very difficult to test this, and more without having more dependencies (I detected it sending a sales order), so I prefer not to spend time on it if you don't mind.

yajo:
But... I don't understand... What was exactly the problem with plain utf8? Aren't all fields stored as such?

pedrobaeza:
No, they are passed directly to the wizard and accents appears incorrect

email_template_qweb/models/mail_template.py
```diff
@@ -33,6 +33,10 @@ def generate_email(self, res_ids, fields=None):
                             'email_template': this,
                         })
                     )
+                    if isinstance(result[record_id]['body_html'], unicode):
```
yajo, line 38: Please add a comment above this, explaining why it is needed

yajo (changes requested):

yajo (approved):
I took the freedom to patch it myself. Let's merge :blush:

PR #117, opened 2016-10-24
Add access to portal/public user to his own email trackings

antespi:
This access rule is necessary when a portal user write a message in chatter

### DIFF
--- a/mail_tracking/security/mail_tracking_email_security.xml
+++ b/mail_tracking/security/mail_tracking_email_security.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- © 2016 Antonio Espinosa - <antonio.espinosa@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+<odoo>
+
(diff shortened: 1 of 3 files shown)


mail_tracking/security/mail_tracking_email_security.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- © 2016 Antonio Espinosa - <antonio.espinosa@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+<odoo>
+
+<record model="ir.rule" id="mail_tracking_email_portal_public_rule">
+    <field name="name">mail_tracking_email: portal/public: read access on my email trackings</field>
+    <field name="model_id" ref="base.model_mail_tracking_email"/>
+    <field name="domain_force">[('recipient_address', '=', user.partner_id.email)]</field>
```
pedrobaeza: This means a security hole. Anyone can change his email address to match the one desire to spy, and then see the communications from that email. Please use another rule.

pedrobaeza (approved):
Fast-tracking as this is a needed patch if you use portal.

Issue #122, opened 2016-11-03
[?][8.0][mail_tracking] How to track user clicks


ivantodorovich:
@pedrobaeza @antespi, I ping you two because you are the contributors of this module.

I'm trying to figure out how to track user clicks on email links.
I thought it was possible, because the `event_type:` `click` exists and its tested.

However, after looking at the controller's code, it seems there's no route to handle this event (IMO: it should track the event and then redirect to the desired url through headers)

Am I missing something big on how to use this, or is this functionality indeed not yet implemented?

antespi:
This functionality is not implemented. mail_tracking is also a base addon for other ones that integrate event webhooks, for instance, mail_tracking_mailgun.

This could be a good improve to mail_tracking. In order to have a todo list, we have to:
- Create a new table in order to save the relation of link to an unique token. 
- Replace all links in email body with the link to a new controller (this new link will include only the database name, the unique token and the tracking_email_id). 
- Controller will look up the table (using the unique token) and return a redirect to the real link. 
- Also controller will create a click tracking_email_event with all the available metadata associated with the HTTP request (IP, user_agent, timestamp, etc ...)

With this improvement, an Odoo instance that use a simple SMTP service (in localhost, for instance) will have sent, open and click events out-of-the-box.

ivantodorovich:
Yeah, that will be really good.

I already solved my pressing issue using Google Analytics, but I'll make a PR with this improvements as soon as I have the time. Having all the tracking data in Odoo would be really awesome.

pedrobaeza:
Don't close this then until it's done. I assign you in the issue.

pedrobaeza:
In Odoo v9, it must use the existing link tracker from Odoo.

ivantodorovich:
~~I'm looking into this.
In order to replace all the original links in the email, I'm thinking of taking advantage of the beautifulsoup library.
Would that be ok, or should I try to do it using regexp?~~

Nevermind. I think I got it working with just regexp.

PR #123, opened 2016-11-04
[8.0][IMP][mail_tracking] Track click events.

ivantodorovich:
Implements tracking click events on links, as discussed in https://github.com/OCA/social/issues/122

@antespi I took you suggestion, except for this one:

> Create a new table in order to save the relation of link to an unique token.

Instead, it replaces `href` values from anchor tags like this: `tracking_email_url/?url=original_href`
Then it routes those requests and redirects to the original URL, saving that event.
Inspired by how google does it for search result links

### DIFF
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -4,6 +4,7 @@
 
 import logging
 import urlparse
+import urllib
 import time
 import re
 from datetime import datetime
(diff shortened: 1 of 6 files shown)


mail_tracking/models/mail_tracking_email.py
```diff
@@ -170,23 +171,67 @@ def _compute_date(self):
...
+                    g.group(2)),
+                email['body']
+            )
+        return email
+
+    # Should I leave it as an alias? TODO: Remove later on
+    # @api.multi
+    # def tracking_img_add(self, email):
+    #     self.ensure_one()
+    #     return self._tracking_img_add(email)
```
ivantodorovich: Should I? Or should I not?

ivantodorovich:
I'm wondering if there should be an on/off switch for this functionality.
It completely breaks links if the odoo instance is not exposed to the internet.

ivantodorovich:
Ping @pedrobaeza @antespi

antespi:
@ivantodorovich Thanks for your contribution. You should add a test case for click event.

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -126,6 +126,13 @@ def test_mail_send(self):
         self.assertEqual(image, res.response[0])
         # Two events again because no tracking_email_id found for False
         self.assertEqual(2, len(tracking.tracking_event_ids))
+        # Click event on links should redirect
+        http.request.httprequest.url = \
+            'http://www.odoo.com/?url=http://www.google.com.ar'
+        res = controller.mail_tracking_click(db, tracking.id)
+        self.assertEqual('301', res.status)
+        self.assertEqual('http://www.google.com.ar', res.headers['Location'])
+        self.assertEqual(3, len(tracking.tracking_event_ids))
```
ivantodorovich, line 135: @antespi something like this is enough? I'm sorry I don't have much experience writing tests 😢

ivantodorovich:
@antespi please check now and please, if its ok to you, remove the `work in progress` label so others can review.

ivantodorovich:
After working with this changes in my local instance, I'm thinking there should be a way to enable/disable links tracking, because it's a very intrusive feature: it routes all links through our server, and that might not be what you want all the time.

I'm thinking of doing this by reading some attribute in the anchor element. For instance:

``` html
<a href="http://www.google.com" tracking-disabled>Click here</a>
```

or

``` html
<a href="http://www.google.com" tracking-enabled>Click here</a>
```

Then, it will only replace urls to route through our server if its required to do so.

**The question is: default enabled or default disabled?**
I'm more inclined to the latter, because of how intrusive this feature is.

What do you think?

pedrobaeza:
I think a better strategy is the one taken in v9 in the link tracker. You create a link (like the shorteners goo.gl tc.o and similars), for your final link, and you paste in your mail this link, not the original one. Obviously, a way to do this from the mail itself would be more comfortable.

ivantodorovich:
So, basically, no automatic link replacement?
User has to manually create the tracking link and replace it in his email?

I can live with that, but I think maybe offering a way to do it automatically depending on some customized attribute might be better for the user.

pedrobaeza:
The best option would be to have a new mark in the link wizard to put "Track this link". In this case, you switch the provided link with this new link. In v8, directly the needed address for the controller with the corresponding token. In v9, creating a link in "Link tracker" and providing the result URL.

What do you think about this?

ivantodorovich:
I'm not sure I understand your idea.

Let me explain what I understood to see if we are talking about the same thing.

**In v8:**
Track links that have `odoo-track-this-link` attribute, like this:

``` html
<a href="http://www.google.com" odoo-track-this-link>Click here</a>
```

Those links will be automatically replaced with the address of the controller that will track and redirect.

**In v9:**
Also track links that have `track-this-link` attribute. exactly like v8.

``` html
<a href="http://www.google.com" odoo-track-this-link>Click here</a>
```

But instead of defining a new controller, it will create a native 'tracking link' and use that url. 
It will also, somehow, catch the tracking event to be added in `mail_tracking`

`odoo-track-this-link` can be something else.

---

So, basically, this module will provide a faster way to generate tracking links (in v9) and also report click events as it does for mail open events.

rafaelbn:
I change label to need review as all test are green @antespi

rafaelbn:
@antespi @carlosdauden could you please take a review here? Thanks!

PR #125, opened 2016-11-05
[FIX] Save and search recipient_address in lowercase

antespi:
```recipient_address``` is an indexed field, but in all searchings we've been using ```=ilike```, then PostgreSQL can't use index for searching.
With this PR we save ```recipient_address``` in lowercase and search it using ```=```
This improve performance when calculating email_score, for instance.

Backport from https://github.com/OCA/social/pull/124

### DIFF
--- a/mail_tracking/__openerp__.py
+++ b/mail_tracking/__openerp__.py
@@ -5,7 +5,7 @@
 {
     "name": "Email tracking",
     "summary": "Email tracking system for all mails sent",
-    "version": "8.0.3.0.1",
+    "version": "8.0.4.0.0",
     "category": "Social Network",
     "website": "http://www.tecnativa.com",
     "author": "Tecnativa, "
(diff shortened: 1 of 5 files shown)


pedrobaeza (changes requested):
You need to include a migration script that lower all existing emails and override partner write to make sure also that manual emails are in lower.

pedrobaeza (approved):
Thanks

pedrobaeza:
cc @Tecnativa

susport:
You need too override partner create to make sure also  that manual emails are in lower in the save.

antespi:
@susport It's done here: https://github.com/OCA/social/pull/125/files#diff-9bd607644e98dcc593ad45a5376183b8R48

rafaelbn (approved):

pedrobaeza:
@antespi, that's write method, not create method.

PR #127, opened 2016-11-07
[ADD] mass_mailing_event

antespi:
Mass mailing event
==================

This module links mass mailing with event in order to exclude recipients that
are already registered, confirmed, cancelled, attended, or a combination of
these states, when the mass mailing is sent.

Usage
=====

In a mass mailing, user can set an event related and exclude some selected
recipients. Those who has an email address that have already registered in
that event.

This is useful in this scenario:

1. Create a mass mailing for telling to 1000 partners that a new event is
available.
2. During a week some of them (50) have been registered in the event
3. Then, a week after the first mass mailing, duplicate it to get a second
mass mailing. Relate this one to the event and exclude the registered emails.
Change the message body to remember that early bird period is going to expire
soon.
4. Send the second mass mailing and registered emails are automatically excluded,
So it's been only sent to 950 partners the ones who are not registered in the
event yet.


rafaelbn (commented):
Please review travis.

Tested in runbot, it work fine! Thanks!

antespi:
Travis fixed. Thanks for your review

rafaelbn (approved):
Tested functionally in runbot

mass_mailing_event/models/event_state.py
```diff
@@ -0,0 +1,12 @@
...
+# Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, fields
+
+
+class EventState(models.Model):
+    _name = 'event.state'
+
+    name = fields.Char(required=True)
```
pedrobaeza: Why do you need this? If required, you need to include a pre-init hook that fills the fields that are empty.

mass_mailing_event/models/event_registration.py
```diff
@@ -0,0 +1,20 @@
...
+
+from openerp import models, api
+from .mail_mass_mailing import event_filtered_ids
+
+
+class EventRegistration(models.Model):
+    _inherit = 'event.registration'
+
+    @api.model
+    def search_count(self, domain):
```
pedrobaeza, line 13: Don't you need to overwrite also `search` method?

mass_mailing_event/models/event_state.py
lasley: This looks to be a new model though, why is a hook needed?

mass_mailing_event/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,54 @@
...
+        apply_domain.append((field, 'not in', exclude_emails))
+    rows = model.search(apply_domain)
+    return rows.ids
+
+
+class MailMassMailing(models.Model):
+    _inherit = 'mail.mass_mailing'
+
+    def _default_exclude_event_state_ids(self):
+        return self.env['event.state'].search([])
```
lasley: Exclude everything by default?

antespi: Yes, user only have to remove the states he doesn't want to exclude. Remove action is more simple than add action.

mass_mailing_event/models/event_state.py
antespi: Yes, this is new model

mass_mailing_event/models/event_registration.py
antespi: I don't need to overwrite `search`. There is only two actions to hook:
1. search_count: In order to show in the form the right number of recipients.
2. get_recipients: In order to exclude the ones that are registered in event.

`get_recipients` calls to `search` internally but I've preferred to overwrite `get_recipients` because there is only one and it is more specific.

mass_mailing_event/models/event_state.py
pedrobaeza: Oh, that raises then my following question (why a new model?)

antespi: Because selection of event states for excluding is a many2many field, similar to partner tags. None can create/update/delete items in this model except admin or via XML

pedrobaeza: OK, I get it. Thanks for the explanation. I'm auto-populated list would be desirable but there's no such tools in the framework.

pedrobaeza (approved):

lasley (approved):
Thanks @antespi 👍 

Gotta love it when the review is just a series of questions instead of changes 😄

mass_mailing_event/models/event_state.py
hbrunn: we actually can do this: Create an `AbstractModel`, make `search` there return the values of the selection, and `read` return a dict/list of dicts crafted from the selection's display values. Last time I tried this, nobody cared about the type of the ids, if this changed, you might have to translate the values to ints first and retranslate on read. This should be a nice exercise in bending the ORM a bit to our will.
Then as there are no database records, you need to save the values in a serialized fields, and wrap a function field around this that reads/writes to this serialized field.

This way, you're resilient against states added/removed from the registration model.

The model stands for event registration states, so I think `event.registration.state` whould be a better name

mass_mailing_event/README.rst
```diff
@@ -0,0 +1,73 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==================
+Mass mailing event
+==================
+
+This module links mass mailing with event in order to exclude recipients that
```
hbrunn: either use the module names (`mass_mailing`, `event`) or a plural in event

mass_mailing_event/README.rst
```diff
@@ -0,0 +1,73 @@
...
+are already registered, confirmed, cancelled, attended, or a combination of
+these states, when the mass mailing is sent.
+
+Usage
+=====
+
+In a mass mailing, user can set an event related and exclude some selected
```
hbrunn: a user/the user/users

mass_mailing_event/README.rst
```diff
@@ -0,0 +1,73 @@
...
+recipients. Those who has an email address that have already registered in
```
hbrunn: have

mass_mailing_event/README.rst
```diff
@@ -0,0 +1,73 @@
...
+that event.
```
hbrunn: this sentence sounds weird

hbrunn (changes requested):

mass_mailing_event/models/event_state.py
antespi: Thanks for your explanation but I think this is simpler and event registration states are not expected to change. And in that case, admin could add the new state to the table via XML.

I'll change model name, thanks.

mass_mailing_event/README.rst
antespi: Please check the alternative, thanks

antespi:
@hbrunn Changes done, thanks

hbrunn:
the texts look fine, but it seems the rename wasn't done on all places

hbrunn (approved):

antespi:
Thanks all @rafabn, @pedrobaeza, @lasley and @hbrunn for your reviews

PR #130, opened 2016-11-15
Improve default e-mail notification template

ecino:
I propose some modifications for default e-mail notification template :

1. When posting directly in a thread, genereated messages have no subject. Therefore, instead of letting the subject empty, we take the first message's subject in the thread (new ancestor field) and use it for subject (RE: ancestor_subject)
2. When sending a notification to a partner that don't have access to Odoo (not a User), we hide the record access link and only display the message body.
3. If no record_access_link is available, display base URL with "Your Odoo Inbox" (happens when a message is directly sent to a user and not attached to a record)


ecino:
If this is not reviewed, I will close it because it's an old one and the module is not ported on new versions of Odoo.

PR #133, opened 2016-11-24
[8.0][FIX][mail_tracking] Use event recipient address to find partners and contacts to bounce

antespi:
There are use cases when Odoo send an email to a recipient (To:) and copy to others (Cc:). In these cases, events like bounces indicates the recipient address who is actually bounce. And sometimes it is not the original one (To:)

Current implementation:
- Mark as bounce partners with original email recipient address (To:)

Desire implementation:
- Mark as bounce the partner with the recipient address indicated in bounce event (if available)

This PR affects to ```mail_tracking``` and ```mail_tracking_mass_mailing``` addons

### DIFF
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -195,11 +195,16 @@ class MailTrackingEmail(models.Model):
             })
 
     @api.multi
-    def _partners_email_bounced_set(self, reason):
-        for tracking_email in self:
+    def _partners_email_bounced_set(self, reason, event=None):
+        recipients = []
+        if event and event.recipient_address:
+            recipients.append(event.recipient_address)
+        else:
+            recipients = list(filter(None, self.mapped('recipient_address')))
+        for recipient in recipients:
             self.env['res.partner'].search([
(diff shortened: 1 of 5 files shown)


mail_tracking/models/mail_tracking_email.py
```diff
@@ -293,11 +298,13 @@ def event_create(self, event_type, metadata):
             if not other_ids:
                 vals = tracking_email._event_prepare(event_type, metadata)
                 if vals:
-                    event_ids += event_ids.sudo().create(vals)
+                    event = event_ids.sudo().create(vals)
+                    if event_type in {'hard_bounce', 'spam', 'reject'}:
+                        self.sudo()._partners_email_bounced_set(
+                            event_type, event=event)
+                    event_ids += event
```
pedrobaeza: Rename variable to `events`

pedrobaeza (approved):
Little thing, but :+1:

rafaelbn (approved):
Thanks!

susport (approved):
👍

rafaelbn:
ping @carlosdauden

PR #135, opened 2016-11-28
[9.0][FIX][mail_tracking] Use event recipient address to find partners and contacts to bounce

antespi:
Forward porting of https://github.com/OCA/social/pull/133


rafaelbn (approved):
👍

pedrobaeza (approved):
Fast-tracking as there is a proved bug

PR #138, opened 2016-12-14
[FIX][9.0] partner buttons view on enterprise

jjscarafia:
Class 'class="oe_stat_button oe_inline"' is no more used on buttons of partner view, only 'class="oe_stat_button"' (no more oe_inline). This cause an error on enterprise view when too many buttons (find image attached)
seleccion_004

Where normal behaviour is
seleccion_005


Emiq2 (approved):

pedrobaeza (approved):
OK, tried on runbot and it looks the same on community. Merging and cherry-picking in v10.

PR #142, opened 2016-12-30
[8.0] [ADD] mail message name search

serpentcs-dev1:
Mail message name search
========================

This module adds the capability to search for mail messages by subject or
body of the message. This will be useful in models that make intense use of
messages, like project issues or helpdesk tickets.

This module will add dynamically the message_ids to the search view of 
any model that inherits from the mail.thread and will incorporate the
capability to search for content in the mail messages.

Installation
============

This module depends on the module 'base_search_fuzzy' to ensure that
searches on emails are based on indexes. Please read carefully the install
instructions https://github.com/OCA/server-tools/blob/8.0/base_search_fuzzy/README.rst

This module installs by default the indexes that are required to
perform the searches on mail messages.


serpentcs-dev1:
@jbeficent Could you please review?

JordiBForgeFlow:
Please add oca-dependencies.txt to the repo (e.g. https://github.com/OCA/sale-workflow/blob/9.0/oca_dependencies.txt), including 'server-tools' as dependency

oca_dependencies.txt
```diff
@@ -1 +1,7 @@
 crm
+stock-logistics-workflow
```
JordiBForgeFlow: Why so many dependencies? Only add server-tools

serpentcs-dev1:
Fixed the travis

JordiBForgeFlow:
@hbrunn Can you help us here? We're struggling. The intention is to be able to search by the content of the messages in any model that inherits 'mail.thread'. 

With the current approach it does not work.

Steps to reproduce in runbot
1. Add a message to Lead "Need Quotation for 100 PC and 100 Keyboards" (ID 12). E.g. "Hello World".
2. Search for "Hello World" in customers. 

Expected results: No customer should be found
Actual results: Customer  "Bank Wealthy and sons" is found (ID 12). But this customer does not have a message logged!.

hbrunn:
and in your debugger you verified the changed functions are actually called? In https://github.com/OCA/social/blob/8.0/mail_follower_custom_notification/models/base_patch_models_mixin.py, I jump through a lot of hoops to make patches to `BaseModel` applied to all already existing models too. If this mixin helps you too, you could take the opportunity to make this its own module in `server-tools`, then everyone can use the same mechanism.

The code looks like you should do the patching in `_register_hook` btw.

JordiBForgeFlow:
The code is all at debug time. However the strange thing is that the occurs if I manually add the search for message_ids to a specific object manually. See https://github.com/OCA/server-tools/pull/576.

serpentcs-dev1:
> @hbrunn Can you help us here? We're struggling. The intention is to be able to search by the content of the messages in any model that inherits 'mail.thread'.
> 
> With the current approach it does not work.
> 
> Steps to reproduce in runbot
> 
> Add a message to Lead "Need Quotation for 100 PC and 100 Keyboards" (ID 12). E.g. "Hello World".
> Search for "Hello World" in customers.
> Expected results: No customer should be found
> Actual results: Customer "Bank Wealthy and sons" is found (ID 12). But this customer does not have a message logged!.

@jbeficent Have fixed the message_ids field type issue, now it works with the above case in local.

JordiBForgeFlow:
See this PR:
https://github.com/darshan-serpent/social/pull/1/

Fixes the issue found searching for any model by mail content.

FYI, the previous approach did not work neither on my local nor in runbot. If you recorded a message on a model, you would find results for the same message searching on another model that did not have that message.

The proposed implementation resolves this issue. However there's a small issue still. The new field 'message_content' is not available in model 'res.partner'. But strangely enough it IS visible in all other tested models.

serpentcs-dev1:
@hbrunn @pedrobaeza Can you guys please suggest what module name can fit to this?

Current Module name: mail_message_name_search (but as now the name_search approach is not used)

@jbeficent suggested base_search_mail_content

JordiBForgeFlow:
@darshan-serpent Please change to `base_search_mail_content`

dreispt:
This feature sounds great! 
But I'm puzzled and I miss a "user story": how and where is the user supposed to use this? When searching by "name" it also searches by mail content? Are are you supposed to afterwards add filters to the documents where you want this?

JordiBForgeFlow:
@dreispt The user story can be a helpdesk department that records tickets on a daily basis on issues or questions associated to their products. On each ticket there's a back and forth of messages between the person that raised the ticket and the user who attends to it. All the conversation history is tracked in the chatter.

After some time, the user wants to retrieve the helpdesk tickets associated to a past discussion, e.g. "Galaxy on fire". Same as you would do when you want to search for past emails. Why do you need to search based on content? Well, because you don't exactly remember in what ticket/s did the "Galaxy on fire" took place.

At this moment all the conversations are just stored, but cannot be searched.

JordiBForgeFlow:
@dreispt and, storing the content in the name of the ticket/project issue/lead, does not guarantee that it will contain the information that you're looking for, because it was discussed during the conversations.

dreispt:
Thanks Jordi, that sounds good. It would be nice for you to reuse that explanation in the README.

On the design side, the perfect solution for me would be to have a Search option to filter on "Mails Content".
This means adding a searchable computed field and add the corresponding `<filter>` option to the search views.

If you agree with this, it might be interesting to look and take ideas from the "smart search" feature proposed in https://github.com/OCA/server-tools/pull/683

JordiBForgeFlow:
@dreispt But this smart search option is there! You have the field 'mail content' available in any model that inherits from 'mail.thread'.

JordiBForgeFlow:
@dreispt The module does exactly what you ask for. Add a searchable computed field, and dynamically adds it to the search views.

dreispt:
Good! Thanks for explaining.

serpentcs-dev1:
@jbeficent The above commit will help to perform an advanced search for the operators in the Negative expression.

rafaelbn:
Hi @jbeficent could you review this PR? as module is finished in v9 i guess this PR is also ready to review? or not?

hbrunn:

PR #147, opened 2017-01-18
[mail_attach_existing_attachment][IMP] Move domain in python in order to allow inheritance

adrienpeiffer:


pedrobaeza (commented):
The domain is only evaluated at the beginning, so if you install a module changing the domain you will need to restart the service. Are you aware of that?

I haven't tried this possible workaround in the child module:
```
object_attachment_ids = fields.Many2many(
    domain=lambda self: self._get_object_attachment_domain()
)
```
If it works, it deserves to be put as a comment.

hbrunn:
to my knowledge, those domains are evaluated at view load. Does 10 do something different here?

rafaelbn:
Hi @adrienpeiffer could you please clarify question? Thanks!

hbrunn:

sbidoul:
@benwillig do you use this PR on your projects? If yes can you followup?

benwillig:
Yes this PR is used in some of my projects and I confirm it's working. @pedrobaeza If you uninstall/install a module which override ```_get_object_attachment_domain```, you won't have to restart the server. This PR should be merge IMHO.

benwillig (approved):

sbidoul (approved):

pedrobaeza:
@benwillig please read again my comment. I don't think uninstalling the module will evaluate again the domain.

benwillig:
Yes it does. When using lambda, it's evaluated on the fly, and not when the server starts.

pedrobaeza:
OK, if you are sure about that, I merge.

PR #129, opened 2016-11-11
[9.0][MIG][mass_mailing_custom_unsubscribe] Migrate

yajo:
- Imported last updates from v8.
- Adapted to v9.
- Added a saner default to `mass_mailing.salt` configuration parameter by
  reusing `database.secret` if available, hoping that some day
  odoo/odoo#12040 gets merged.
- Updated README.

@Tecnativa

PR #149, opened 2017-01-24
[FIX][mass_mailing_custom_unsubscribe] Reasons noupdate

yajo:
After this fix, when you update the addon, you will not lose your customized reasons.

Had to move from csv to XML because there's no noupdate=1 in csv.

@Tecnativa

### DIFF
--- a/mass_mailing_custom_unsubscribe/__openerp__.py
+++ b/mass_mailing_custom_unsubscribe/__openerp__.py
@@ -23,7 +23,7 @@
 {
     'name': "Customizable unsubscription process on mass mailing emails",
     'category': 'Marketing',
-    'version': '8.0.2.0.0',
+    'version': '8.0.2.1.0',
     'depends': [
         'mass_mailing',
         'website_crm',
@@ -31,7 +31,7 @@
     'data': [
         'security/ir.model.access.csv',
         'data/install_salt.xml',
-        'data/mail.unsubscription.reason.csv',
+        'data/mail_unsubscription_reason.xml',
         'views/assets.xml',
         'views/mail_unsubscription_reason_view.xml',
         'views/mail_mass_mailing_list_view.xml',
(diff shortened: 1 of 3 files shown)


mass_mailing_custom_unsubscribe/__openerp__.py
```diff
@@ -23,15 +23,15 @@
 {
     'name': "Customizable unsubscription process on mass mailing emails",
     'category': 'Marketing',
-    'version': '8.0.2.0.0',
+    'version': '8.0.2.0.1',
```
pedrobaeza: You need to update, so 8.0.2.1.0

pedrobaeza (changes requested):
Translation doesn't change?

This is needed also for 9.0

yajo:
No translation changed.

yajo:
~BTW this addon needs a full forward migration for v9, not just this patch~

v9 migration is in https://github.com/OCA/social/pull/129, I'll add the patch there

pedrobaeza:
Yeah, it's on https://github.com/OCA/social/pull/129, and it's yours! You don't even remember your own PRs :stuck_out_tongue:

rafaelbn:
@Yajo travis fails

yajo:
I've got 142 open PRs, indeed I don't remember all of them 🤣 

It turns out the branch is broken, I'll try to fix it.

yajo:
Nice, it's Odoo that is broken. See https://github.com/odoo/odoo/pull/15199. Now I have 143 😆.

In short, you can ignore Travis error because it's not related.

yajo:
Now that it's fixed, I squashed and rebased, and bots went 💚 😊

pedrobaeza (approved):

yajo:
Interesting, if you remove the records and update the addon, they get back in. It seems like noupdate != noreinstall. I'll investigate a little bit more.

pedrobaeza:
I thought you have already put `forcecreate="0"`. That's the way for avoiding recreate them on update.

yajo:
Yay thanks! TIL that. Squashed. Ready to merge then.

rafaelbn (approved):
Tested

PR #155, opened 2017-02-28
[MIG] 9.0 mail debrand

serpentcs-dev1:
Mail debrand
==========

This module removes the Odoo branding from the mail templates.

To use this functionality try to send an email using any email template.
 
Like : 
* Go to Sale / Sale Orders and click 'Send by Email' after selecting any sale order.


mail_debrand/data/mail_data.xml
```diff
@@ -0,0 +1,116 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 Eficent Business and IT Consulting Services S.L.
+     Serpent Consulting Services Pvt. Ltd.
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3.0) -->
+<openerp>
+    <data>
+        <!--Default Notification Email template -->
+        <record id="mail.mail_template_data_notification_email_default" model="mail.template">
+            <field name="name">Notification Email</field>
```
JordiBForgeFlow: Remove the <fields> that are not changing. Only "body_html" changes

mail_debrand/data/mail_data.xml
```diff
@@ -0,0 +1,116 @@
...
+                <strong>Sent by</strong>
+                % if ctx.get('website_url'):
+                <a href="${ctx['website_url']}" style="text-decoration:none; color: #a24689;">
+                % endif
+                <strong>${ctx.get('company_name')}</strong>
+                % if ctx.get('website_url'):
+                </a>
+                % endif
+                <strong></strong>
+                <a href="https://www.odoo.com" style="text-decoration:none; color: #a24689;"><strong></strong></a>
```
JordiBForgeFlow: Remove this

serpentcs-dev1:
@jbeficent Have made the requested changes, can you please review??

LoisRForgeFlow:
Which is the proper place to check this? " Sale Orders and click 'Send by Email' after selecting any sale order" is using <a href="https://github.com/OCA/OCB/blob/9.0/addons/sale/sale.py#L374">another template</a>, not the one modified in this PR.

serpentcs-dev1:
@lreficent Yes, that is the right, or you can also test by sending the email from the PO. It doesn't matter what email_template is passed in Sale, the notification email template of the mail message is handled [here](https://github.com/odoo/odoo/blob/9.0/addons/mail/models/res_partner.py#L154).

So in short, the body_html content of the mail notification template used in the PR is appended every time the mail message is created. Ref : https://github.com/odoo/odoo/blob/9.0/addons/mail/models/res_partner.py#L176

JordiBForgeFlow (approved):
+1

LoisRForgeFlow:
Ok, Tested. Thank You @darshan-serpent !

JordiBForgeFlow:
@darshan-serpent Can you squash your commits?

LoisRForgeFlow (approved):

JordiBForgeFlow:
@Yajo Do you agree with this approach?

mail_debrand/data/mail_data.xml
```diff
@@ -0,0 +1,110 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 Eficent Business and IT Consulting Services S.L.
+     Serpent Consulting Services Pvt. Ltd.
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3.0) -->
+<openerp>
+    <data>
```
yajo: Add `noupdate='1'`

yajo (commented):
Well, all you are doing is to replace a template, isn't it? Why? In v9 now they don't compute the signature? I haven't checked this.

hbrunn:
@darshan-serpent I'd like to merge this as soon as @Yajo's comment is addressed

yajo (approved):

mail_debrand/data/mail_data.xml
```diff
@@ -0,0 +1,110 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 Eficent Business and IT Consulting Services S.L.
+     Serpent Consulting Services Pvt. Ltd.
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3.0) -->
+<openerp>
```
pedrobaeza: Use `<odoo noupdate="1">` instead of `<openerp>` and remove `<data>`

pedrobaeza (changes requested):
Original mail template is noupdate? If so, this won't work. If not, why you put noupdate?

JordiBForgeFlow:
@pedrobaeza Original template is NOT noupdate=1. See https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/addons/mail/data/mail_data.xml#L63

serpentcs-dev1:
@pedrobaeza Original template does not have noupdate.

I did because @Yajo suggested it. Anyways I have removed it now.

cc @jbeficent

pedrobaeza:
Anyway, does this remove Odoo's brand on sent quotations for example? This is only for notifications, not for the rest of the mails.

mail_debrand/data/mail_data.xml
pedrobaeza: As stated, this shouldn't be changed in the inheritance chain. Although is interesting to modify the template by hand, this would break other possible inherited modules.

mail_debrand/data/mail_data.xml
pedrobaeza: With latest comments, this should be converted to just `<odoo>` without `<data>`

JordiBForgeFlow:
@pedrobaeza On mails sent from sales orders or purchase orders the Odoo branding is not included.

See https://github.com/odoo/odoo/blob/9.0/addons/purchase/data/mail_template_data.xml

serpentcs-dev1:
> Anyway, does this remove Odoo's brand on sent quotations for example? This is only for notifications, not for the rest of the mails.

@pedrobaeza Can you please check my above comment addressed to 'ireficent' which answers the same.

pedrobaeza:
Previous version includes a footer with "Sent by <company> using Odoo". That was the main purpose of the module in version 8 to remove that part.

serpentcs-dev1:
@pedrobaeza  Yes, that is the right, but now the notification email template of the mail message is handle [here.](https://github.com/odoo/odoo/blob/9.0/addons/mail/models/res_partner.py#L154)

So in short, the body_html content of the mail notification template used in the PR is appended every time the mail message is created. Ref : https://github.com/odoo/odoo/blob/9.0/addons/mail/models/res_partner.py#L176

serpentcs-dev1:
@pedrobaeza Now without this module, it will add the branding "Sent by using Odoo" in the header.

pedrobaeza:
OK, understood. Thanks for the explanations.

The only remaining task is to squash together "OCA Transbot" commits together and your fourth last commits in one.

serpentcs-dev1:
@pedrobaeza squashed.

pedrobaeza (approved):

pedrobaeza:
I have update copyright of our company and include yours. Tell me if it's correct.

serpentcs-dev1:
Yeah I think it's correct.

PR #157, opened 2017-03-14
[mail_tracking_mailgun] Add Contributors

rafaelbn:
Both Carlos and me have work a lot in this module, maybe not coding but much testing and definition in usability


pedrobaeza (approved):
Note: Carlos actually do on coding. Rafael on testing + design.

PR #158, opened 2017-03-24
[add] mail_digest

simahawk:
This module allows users/partners to:

- select "digest" mode in their notification settings
- with digest mode on select a frequency: "daily" or "weekly"
- configure specific rules per message subtype (enabled/disabled)

and receive or not receive any email notification for a given subtype.

This module is needed by its CMS integration module [cms_notifications](https://github.com/simahawk/website-cms/tree/add-cms_notifications-PR/cms_notifications)


simahawk:
TODO: check / fix failing tests w/ other social modules.

simahawk:
@antespi @pedrobaeza I'm checking failing tests. I see 

```bash
2017-03-24 08:42:43,626 5399 ERROR openerp_test openerp.addons.mail_tracking_mailgun.models.mail_tracking_email: Mailgun: Database 'bad_db' is not the current database
[...]
2017-03-24 08:42:44,481 5399 ERROR openerp_test openerp.addons.mail_tracking_mailgun.models.mail_tracking_email: Mailgun: event type 'bad_event' not supported
```

and other similar errors, both on travis and on my local setup (with just mail_tracking tests running).

Actually I found the same errors in the original PR https://travis-ci.org/OCA/social/jobs/161414564

Is there any particular setting to run tests for mail_tracking or can I just ignore them?

pedrobaeza:
AFAIK, these are the intended behavior of the tests, proving a wrong DB in them to see if it fails, so this is not problem for the status (and indeed Travis is green).

simahawk:
@pedrobaeza is green since I fixed my stuff :) I wanted to make sure I didn't break anything else before debugging. Tnx anyway.

mail_digest/models/res_partner.py
```diff
@@ -0,0 +1,209 @@
...
+        # notify_by_chat
+        self._notify_by_chat(message)
+        return True
+
+    @api.multi
+    def _notify_by_digest(self, message):
+        message_sudo = message.sudo()
+        if not message_sudo.message_type == 'email':
+            return
+        # TODO: what to do when a partner updates frequency?
```
dreispt: > what to do when a partner updates frequency?

Frequency tells how often we will try to send a digest to the Partner.
Whenever we hit that time, all unsent messages will be collected into that digest, and marked as sent.
It you look at it this way, the answer is trivial.
With this design, the frequency is irrelevant for the `mail.digest` queue; it only matters for the digest mail process.

simahawk: yep, you are right here. In your opinion, is there any other place where this should be taken into account?

dreispt: Can't remember any. Frequency should only be relevant to trigger the "cron job".

simahawk:
@pedrobaeza @dreispt anything else to do here but rebase?

guewen (approved):

simahawk:
rebased!

simahawk:
@pedrobaeza @rafaelbn @dreispt any stopper here?

mail_digest/__openerp__.py
```diff
@@ -0,0 +1,23 @@
...
+    'summary': """Basic digest mail handling.""",
+    'version': '9.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'Camptocamp,Odoo Community Association (OCA)',
+    'depends': [
+        'mail',
+    ],
+    'data': [
+        'data/ir_cron.xml',
+        # TODO: fix permissions (only admin now)
```
pedrobaeza: Is still this TODO?

mail_digest/README.rst
```diff
@@ -0,0 +1,90 @@
...
+Features
+--------
+
+This module allows users/partners to:
+
+* select "digest" mode in their notification settings
+* with digest mode on select a frequency: "daily" or "weekly"
+* configure specific rules per message subtype (enabled/disabled)
+
+to receive or not receive any email notification for a given subtype.
```
pedrobaeza: "to receive or to not receive"

pedrobaeza (changes requested):
Little changes

simahawk:
@pedrobaeza updated. Should be ready for merge ;)

pedrobaeza (approved):

PR #159, opened 2017-03-24
[ADD][fetchmail_thread_default] Default thread for incoming mails

yajo:
This module extends the functionality of mail fetching to support choosing a
mail thread that acts as a mail sink and gathers all mail messages that Odoo
does not know where to put.

Dangling emails are really a problem because if you do not care about them,
SPAM can enter your inbox and keep increasing fetchmail process network quota
because Odoo would gather them every time it runs the fetchmail process.

Before this, your only choice was to create a new record for those unbounded
emails. That could be useful under some circumstances, like creating a
``crm.lead`` for them, but what happens if you do not want to have lots of
spammy leads? Or if you do not need Odoo's CRM at all?

Here we come to the rescue. This simple addons adds almost none dependencies
and allows you to direct those mails somewhere you can handle or ignore at
wish.

@Tecnativa


rafaelbn:
Thanks @Yajo , travis fails. Could you check please?

yajo:
Travis should be happy now.

chienandalu:
Following these steps:

1. Settings > General Settings > Incoming Mail Servers > New
2. Configure mail server information.
3. Set "Default mail thread" as "Email thread".
4. Then click in the adjacent dropdown and this error comes up:

`Odoo Server Error
  File "/.repo_requirements/odoo/openerp/http.py", line 648, in _handle_exception
  File "/.repo_requirements/odoo/openerp/http.py", line 685, in dispatch
  File "/.repo_requirements/odoo/openerp/http.py", line 321, in _call_function
  File "/.repo_requirements/odoo/openerp/service/model.py", line 118, in wrapper
  File "/.repo_requirements/odoo/openerp/http.py", line 314, in checked_call
  File "/.repo_requirements/odoo/openerp/http.py", line 964, in __call__
  File "/.repo_requirements/odoo/openerp/http.py", line 514, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo-9.0/addons/web/controllers/main.py", line 888, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/odoo/odoo-9.0/addons/web/controllers/main.py", line 880, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/api.py", line 250, in wrapper
    return old_api(self, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/api.py", line 354, in old_api
    result = method(recs, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/models.py", line 1749, in name_search
    return self._name_search(name, args, operator, limit=limit)
  File "/.repo_requirements/odoo/openerp/api.py", line 248, in wrapper
    return new_api(self, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/api.py", line 490, in new_api
    result = method(self._model, cr, uid, *args, **old_kwargs)
  File "/.repo_requirements/odoo/openerp/models.py", line 1761, in _name_search
    ids = self._search(cr, user, args, limit=limit, context=context, access_rights_uid=access_rights_uid)
  File "/.repo_requirements/odoo/openerp/api.py", line 250, in wrapper
    return old_api(self, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/models.py", line 4811, in _search
    cr.execute(query_str, where_clause_params)
  File "/.repo_requirements/odoo/openerp/sql_db.py", line 141, in wrapper
    return f(self, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/sql_db.py", line 220, in execute
    res = self._obj.execute(query, params)
ProgrammingError: relation "mail_thread" does not exist
LINE 1: SELECT "mail_thread".id FROM "mail_thread" ORDER BY "mail_th...`

yajo:
Of course, that one is an abstract model. You are expected to select a real model that inherits from mail.thread. [Try with a `mail.channel`.](https://github.com/OCA/social/pull/159/files#diff-cf35413b676c902d9f05d72e6388ad00R36)

pedrobaeza:
Maybe you can restrict the list to the models that inherits the model and are not abstract through a computed many2many field that acts as domain for the other field.

yajo:
I thought on that, but the same happens with Odoo's default field that creates a new record.

After some thinking about it, it seems legit as they do it. This way there's less code to maintain. And after all, this addon is intended for sysadmins, and they are expected to read READMEs when installing addons, and to interpret exceptions.

Don't you think?

pedrobaeza:
Well, sometimes sysadmins are the worst users... If it's not too much complicated, I prefer to have that filtered. If not, a big warning on the README.

yajo:
Oh, it turns out [there's the check](https://github.com/OCA/social/pull/159/files#diff-566aa1a6e8b0d66b110d4146300eb997R26). It's as simple as possible. However, let me add a new check that excludes abstract models.

rafaelbn:
Hi @Yajo any update in this PR please? Thanks!

yajo:
Should be fixed now.

pedrobaeza (approved):

rafaelbn (approved):
Long time tested in production 👍

PR #161, opened 2017-03-28
[9.0] [ADD] mail_auto_follower_notify

LoisRForgeFlow:
Mail Auto Follower Notify
====================

(renamed from: Message Auto Subscribe Notify Force Send)

This module extends the functionality of mail by sending an email notification
to new followers that are system users (i.e. it won't sent email to customers).

Usage
=====

* Go to some record that uses chatter and write/edit some field related to a
  '''res.user'''.
* Save. A email notifications has been sent to the users that have been added
  as new followers as a consequence of your previous action.


rafaelbn:
Hi @lreficent could you explain it with a use case lika a sale order, invoices or project task please? Thanks!

LoisRForgeFlow:
Hello @rafaelbn here you  have an example with a crm.lead:

1. A lead created from a email is created (standard behavior after configure properly the incoming mail server):
image

2. I assign it to a salesperson. 
image

3. And due to this module the salesperson is notified by mail:
image

So, this happens because when you assign a salesperson he's added as a follower and this module force the notification when some action autosubcribe someone to a record.

MiquelRForgeFlow (approved):
LGTM :+1:

rafaelbn:
Hi, thanks for explanation. I understand now. But the name of the module is not correct IMHO as it says a different thing. In resume this module create an event (notification by email) to the user who has been added as follower by Odoo automatically.

I think is better mail_follower_notify or mail_auto_follower_notify or similar. Could you please @pedrobaeza give us and idea for the module. There is a "rule" for names of module.

message_auto_subscribe_notify_force_send/models/mail_thread.py
```diff
@@ -0,0 +1,28 @@
...
+            return
+        for record_id in self.ids:
+            messages = self.env['mail.message'].sudo().search([
+                ('model', '=', self._name),
+                ('res_id', '=', record_id),
+                ('subtype_id', '!=', False),
+                ('subtype_id.internal', '=', False)], limit=1)
+            if messages:
+                self.env['res.partner'].browse(partner_ids)._notify(
+                    messages, force_send=True)
```
LoisRForgeFlow: @rafaelbn This is the main reason for the name of the module, if kind of a technical name referring to this. But, yes maybe your options could be good as well.

simahawk (changes requested):
I agree w/ Rafael: the name does not match the goal.

rafaelbn:
Hi @lreficent , please change module name to ahead with this PR please! 😄

LoisRForgeFlow:
Sure, help me to decide: [poll](https://twitter.com/lreficent_erp/status/900376138052063234). :stuck_out_tongue_closed_eyes:

cc @rafaelbn @simahawk

LoisRForgeFlow:
We have a winner! :1st_place_medal: 

Module renamed to **mail_auto_follower_notify**

cc @rafaelbn @simahawk

simahawk (approved):

hbrunn (changes requested):
this absolutely needs tests:
36.36% of diff hit (target 96.54%)

LoisRForgeFlow:
For some reason, the code added by this module is not being called when running the tests. The method `_message_auto_subscribe_notify` is called but not the inherit done in this module.

Any idea?

hbrunn:
this will never work for models that existed before you did your override which are not re-overriden with mail.thread afterwards. Which will be quite a practical problem for your users too btw.
The proper fix for this is somewhere in the lines of https://github.com/OCA/social/blob/8.0/mail_follower_custom_notification/models/base_patch_models_mixin.py, but I suspect this won't play nice if multiple modules are using it. If you get this fixed, that would be optimal.
For just making the test work, you can call the function on `mail.thread` of course too.

mail_auto_follower_notify/tests/test_auto_follower_notify.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# Copyright <YEAR(S)> <AUTHOR(S)>
```
MiquelRForgeFlow: 🙃

LoisRForgeFlow:
@hbrunn I just tested the module in local with your mixin and the problem hasn't been fixed. But maybe I'm applying it wrong. What I did was:

1. Add `base.patch.models.mixin` to my module.
2. Inherit `mail.thread` with this syntax:
 ```
    class MailThread(models.AbstractModel):
        _name = 'mail.thread'
        _inherit = ['base.patch.models.mixin', 'mail.thread']
```

**Comment by hbrunn** [comment]
if you use the mixin, you'll have to call `_register_hook` manually, as that's where the magic happens and this function is only called after the tests: https://github.com/OCA/social/blob/8.0/mail_follower_custom_notification/tests/test_mail_follower_custom_notification.py#L9

**Comment by LoisRForgeFlow** [comment]
@hbrunn oh yes, sure. So do you think a module including just `base.patch.models.mixin` would be useful? so I can depend on it.

**Comment by hbrunn** [comment]
it would be utterly useful. The reason I didn't publish this already is that when multiple modules use this on the same model, weird stuff happens, and that in a hard to debug manner. But I think you're right, publish this and then we fix the problems it causes.

The reason I never really followed up on this was that starting from 10, we don't need it any more. Compare https://github.com/odoo/odoo/blob/10.0/odoo/models.py#L528 with https://github.com/OCA/OCB/blob/9.0/openerp/models.py#L596: in 9 and below, a new model is basically a copy of the classes involved, while >= 10 creates links to the classes, which behaves the way programmers would expect

File: mail_auto_follower_notify/tests/test_auto_follower_notify.py [hunk 3]
```diff
@@ -12,17 +13,24 @@ def setUp(self):
...
-            'login': 'test_login',
-        })
+             'name': 'Test user',
+             'login': 'test_login',
+             'password': 'demo',
+             'email': 'test@yourcompany.com',
+            })
 
+    @at_install(False)
+    @post_install(True)
```
LoisRForgeFlow: @hbrunn FYI finally the best approach was to run the test after the installation.

hbrunn: good one!

LoisRForgeFlow:
* [ ] : Depends on https://github.com/OCA/server-tools/pull/962

Adding it as a dependency to run tests.

JordiBForgeFlow (approved):
Technical and functional review :+1:

hbrunn (approved):

hbrunn:
(approved, but I set the label work in progress as reminder that we can't merge just yet)

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
I was just testing. Do you know what triggers oca-bot approval tag?

simahawk:
shall we merge this? :)

hbrunn:
hmm, no, the dependency never made it. Is this still relevant or should we close the PRs?

JordiBForgeFlow:
I'm closing it, as it is no longer relevant

PR #163, opened 2017-04-14
[ADD] website_mail_qweb

hbrunn:
Edit QWeb email templates
=========================

This module was written to make Odoo's WYSIWYG email template editor usable for email templates using the QWeb rendering engine.

Usage
=====

- click the 'Edit template' button on an email template
- note that the editor supports designing and translating your templates, but for variable substitutions et al, you'll need to edit the source code of the template

Known issues / Roadmap
======================

* the editor has problems with complex template inheritance, this seems rather a problem in Odoo generally than with this specific module


lasley (approved):
Nice! LGTM. 

Some tests would be nice for `UnquoteObject`, but I won't block for it.

hbrunn:
@lasley I borrowed this from myself (cf https://github.com/OCA/server-tools/blob/8.0/base_view_inheritance_extension/models/ir_ui_view.py#L8), but this is a somewhat different use case. `UnquoteRecordset` has tests now

lasley (approved):
Woot thanks @hbrunn!

rafaelbn:
I tried to test it but admin/admin doesn't work. I rebuild runbot but it looks doesn't work. I will try again next week

jholze:
Hi @hbrunn, Cesar already talked to you about the report designer we still have in progress. Now, we try to research a bit to find the best implementation way which is not so easy, because we have custom features our big customer needs. But we know that it would be the best to talk to you guys, because of your experience and we want to build something which could be a nice improvement for odoo in general. The complexity we have to solve is something like a complex history. Like Diff handling in a wiki server. The second than is how we display it, but i think this is not the most complex part. Do you think this PR is a good place to talk about this or what ? Because i would like to invite more people to join this topic, because we will save a lot of capacity for it. So it should be a useful integration not only for our project. 

Looking forward to your answer, best Jens

hbrunn:
I think you can best do what @lasley does a lot: Write an issue on some matching project with your [RFC], and then ping people of which you'd like some ideas/comments.
You guys got me curious, and diffing arbitrary (xml) trees in some tractable fashion seems to be a hard problem. Why don't you save the whole thing, and only diff on display? http://prettydiff.com looks nice, you'll have to test this with seriously big files of course.
And when you hijack my PR you might just as well review it too so that we can merge it ;-)

lasley:
It's actually quite astonishing how well the RFC in an issue thing works too - we have an awesome community that tends to want the same thing in lots of scenarios 😄

PR #164, opened 2017-04-16
[ADD] updates from maintainer-quality-tools

hbrunn:


pedrobaeza (commented):
codeclimate is not being finally used (we should remove it from MQT...), and comment lines are only for reference in the template, but you don't need to put them here.

hbrunn:
my main intention was to get coverage checks, so this is not the way to do that?

pedrobaeza:
Well, the only thing needed for having coverage is to include this line, which is already there: https://github.com/OCA/social/blob/ad5f0599f2de28ce0a539fa58982c65eb9ae016f/.travis.yml#L52

hbrunn:
thanks for clarifying, then this PR doesn't add anything useful

PR #166, opened 2017-04-20
[9.0][Add] Sendgrid modules

ecino:
Adds Sendgrid integration for Odoo.

- mail_sendgrid : new module for sending and tracking e-mails with sendgrid
- mail_sendgrid_mass_mailing : use mass_mailing with sendgrid


pedrobaeza:
The name should be mail_tracking_sendgrid, and according to the mail_tracking system, mail_sendgrid_mass_mailing shouldn't be needed...

ecino:
Well for the tracking part indeed, but the module does more by replacing or adding a new way of sending e-mails instead of the SMTP server. Morevover, a templating system enables to design e-mail templates in Sendgrid and use them in Odoo. The name mail_tracking_sendrid sounds to me like only the tracking feature is added. But if needed, I can still change the name.

For the mass mailing part, it is needed to add templating functionalities of Sendgrid

pedrobaeza:
OK, thanks for the explanation. Well, I have to check deeper to evaluate then.

rafaelbn (approved):
This king of modules are hard to test as you have create a sendfrig account, configure, etc... This test are not only related to runbot. So in this case for me is a optimistic merge as I guess @ecino is using them in production enviroment: isn't it? Thanks

rafaelbn:
Maybe @chienandalu would like to test it deeper 😄

ecino:
@rafaelbn Yes I use this module in production since many months.

ecino:
Any people willing to review this?

ecino:
This was improved on branch 10.0. This version was not maintained so I close it.

ecino:
Hi @rafaelbn , the PR is already merged at #185

PR #167, opened 2017-05-08
[8.0][ADD] mail_message_auto_expand

Kiplangatdan:


mail_message_auto_expand/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- encoding: utf-8 -*-
+
+import mail
```
hbrunn: use relative imports, and models belong into `/models`

mail_message_auto_expand/__openerp__.py
```diff
@@ -0,0 +1,13 @@
+﻿{
+    'name': 'Mail Message Auto Expand',
+    'version': '8.0',
```
hbrunn: use OCA's version convention: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

mail_message_auto_expand/__openerp__.py
```diff
@@ -0,0 +1,13 @@
...
+    'summary': 'Expands the Read More Automatically',
+    'description': '',
```
hbrunn: remove this key and add a readme based on OCA's template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

mail_message_auto_expand/__openerp__.py
```diff
@@ -0,0 +1,13 @@
...
+    'author': 'Sunflower IT',
```
hbrunn: always name OCA as author too: https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L10 (consider using this template folder to scaffold OCA modules anyways)

mail_message_auto_expand/mail.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# © 2017 Sunflower IT (http://sunflowerweb.nl)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
+from openerp import _, api, models, tools
+from openerp.tools import html_email_clean
+from openerp.tools.translate import _
+from HTMLParser import HTMLParser
+from openerp import SUPERUSER_ID, api
+_logger = logging.getLogger(__name__)
```
hbrunn: remove unused imports and variables

mail_message_auto_expand/__init__.py
thomaspaulb: Add a newline here

mail_message_auto_expand/mail.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class MailMessage(models.Model):
+    _inherit = 'mail.message'
+
+    def _message_read_dict(self, cr, uid, message, parent_id=False,
```
hbrunn: never use old api if not absolutely necessary

mail_message_auto_expand/__openerp__.py
thomaspaulb: Either a description, or the [README.rst template of OCA](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst)

mail_message_auto_expand/mail.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# © 2017 Sunflower IT (http://sunflowerweb.nl)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
+from openerp import _, api, models, tools
+from openerp.tools import html_email_clean
+from openerp.tools.translate import _
+from HTMLParser import HTMLParser
+from openerp import SUPERUSER_ID, api
```
thomaspaulb: A blank line here

mail_message_auto_expand/mail.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# © 2017 Sunflower IT (http://sunflowerweb.nl)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
thomaspaulb: and here

mail_message_auto_expand/mail.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class MailMessage(models.Model):
+    _inherit = 'mail.message'
+
+    def _message_read_dict(self, cr, uid, message, parent_id=False,
+                           context=None):
+        max_length = 10000 # Increase the max lenght
+        body_short = html_email_clean(message.body, remove=False, shorten=False,
+                                      max_length=max_length) # Disable shortening
```
hbrunn: why would you need this if you set `shorten` to `False` anyways?

mail_message_auto_expand/mail.py
```diff
@@ -0,0 +1,28 @@
...
+from openerp import SUPERUSER_ID, api
+_logger = logging.getLogger(__name__)
+
+
+class MailMessage(models.Model):
+    _inherit = 'mail.message'
+
+    def _message_read_dict(self, cr, uid, message, parent_id=False,
+                           context=None):
+        max_length = 10000 # Increase the max lenght
```
thomaspaulb: spelling error

mail_message_auto_expand/mail.py
```diff
@@ -0,0 +1,28 @@
...
+        body_short = html_email_clean(message.body, remove=False, shorten=False,
+                                      max_length=max_length) # Disable shortening
+        ret = super(MailMessage, self)._message_read_dict(cr, uid,
+                                                           message,
+                                                           parent_id=parent_id,
+                                                           context=context)
+        ret['body_short'] = body_short
+        return ret
+
```
thomaspaulb: remove extra lines here

thomaspaulb (changes requested):

thomaspaulb:
Tested this, works well. @hbrunn what do you think

mail_message_auto_expand/mail.py
```diff
@@ -0,0 +1,28 @@
...
+    def _message_read_dict(self, cr, uid, message, parent_id=False,
+                           context=None):
+        max_length = 10000 # Increase the max lenght
+        body_short = html_email_clean(message.body, remove=False, shorten=False,
+                                      max_length=max_length) # Disable shortening
+        ret = super(MailMessage, self)._message_read_dict(cr, uid,
+                                                           message,
+                                                           parent_id=parent_id,
+                                                           context=context)
+        ret['body_short'] = body_short
```
hbrunn: this will do a lot more than you want, because now every `body_short` is long. This might (or might not) cause problems in code that expects this key to be short indeed. Why don't you only manipulate the view layer, as in https://github.com/OCA/OCB/blob/8.0/addons/mail/static/src/xml/mail.xml#L271-L277 or intercept the call in https://github.com/OCA/OCB/blob/8.0/addons/mail/static/src/js/mail.js#L1422 or override the constructor in https://github.com/OCA/OCB/blob/8.0/addons/mail/static/src/js/mail.js#L178 - all of those cause only the display to change, and no internal side effect

hbrunn (changes requested):
I think the last comment is crucial.

For the icon: Did you draw this yourself? If not, mention the author/source in the README. If this is an unknown license or no permissive open source license, consider changing it so some with one. I myself often use fontawesome icons or from http://picol.org

mail_message_auto_expand/mail.py
thomaspaulb: We had thought the easiest way of achieving the result was to empty up `body_short`, so that the view layer will always pick the long version. You are right that there _might be_ some view somewhere that uses the `body_short` and that we need to keep that way..... if we are to prepare for that case, I think modifying the XML is best.

rafaelbn:
Hi @Kiplangatdan , thanks for this contribution. Please are you going to attend reviews? Let us know, thanks

thomaspaulb (approved):

hbrunn (changes requested):
thanks, now it's just a few style issues left: https://travis-ci.org/OCA/social/jobs/329886554#L607-L608 (search for your module name in the whole log, there are a few others)

thomaspaulb:
@Kiplangatdan Could you finish this?

PR #168, opened 2017-05-15
[10.0][MIG] mail_debrand

pedrobaeza:
There's no commits preservation, as the module doesn't serve as is on v9.

cc @Tecnativa


mail_debrand/README.rst
```diff
@@ -0,0 +1,73 @@
...
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/205/10.0
+
+
+Known issues / Roadmap
+======================
+
+* This module relies on the translation of the strings here in this module that
+  must match the mail notification template translation on Odoo core,
+  specifically the words "using" and "Odoo".
```
dreispt, line 32: I feel that it would be better to have these as System parameters - at least the Odoo word.
I worry the the words will creep in again in case an email is generated for a language we didn't edit the corresponding translation,.
Are you sure the translation trick also works for base english?

dreispt:
Some templates also include the "Odoo" brand, such as the example below. Ideally, we should also tweak them in a mail_debrand feature:

image

dreispt (dismissed):

dreispt (commented):
(Question about the translation approach)

pedrobaeza:
@dreispt please check code that is a regexp expression, so no possible to be confused.

mail_debrand/README.rst
pedrobaeza: No, it can't be a parameter, as you can translate your template to any installed language, and each one should have this translated to the correct words.

pedrobaeza:
Travis error is not coming from my module and it's supposed to not fail (it's in mail_tracking). Maybe something has changed on the way Travis parses the errors?

chienandalu (approved):
tested on runbot in base language and in Spanish

mail_debrand/README.rst
dreispt: I still worry about the case where no translation string is provided. Can't we use a parameter string in that case?

pedrobaeza: No need for the parameter string. If no translation is found, the used terms are "using" and "Odoo". If you want to care about the translation, just synchronize translations in your Odoo instance, and translate the corresponding term.

dreispt: > If no translation is found, the used terms are "using" and "Odoo"

That's exactly my problem...

pedrobaeza: But the solution is also at your hands. Please read above.

dreispt: The solution is to ensure the many dozens of languages supported by Odoo are custom translated?
And what about base english? In the past there were problems with translations on it.

pedrobaeza: But plain English works as is...

dreispt: Sorry, I don't understand what you mean.

pedrobaeza: OK, let me try to make you understand:

- With English language (with no translations), the matched expression that is removed is `using...<a lot of characters between>...Odoo` (not the isolated words).
- If you translate "using" and "Odoo" words to your language in both places: this module and mail template. For example, in Spanish, the matched expression will be `usando...Odoo`.
- If the mail template is translated, but not this module, you just need to go to the translation and translate the words.
- If none is translated, it will be correctly replaced.

dreispt: OK, I think I  get it now: the RE is supposed to remove the the "using Odoo" words in a translation-resistant way.
Sorry for all the trouble.

pedrobaeza: OK, no problem. Take also into account that you can send mails in a lot of languages inside same Odoo instance (if your customer is English, French, Portuguese...), so you need a translation mechanism, not a static parameter, for replacing this in all of them.

pedrobaeza:
@dreispt can you please give your blessing?

dreispt (approved):
LGTM

PR #172, opened 2017-06-16
[10.0][MIG] mass_mailing_security_group

luismontalba:
Mass mailing security group
-----------------------------------
This module adds these security features:
* ACL for allowing marketing user group to remove mass.mailing objects
* Security rule to allow marketing user to delete only his unsent
  mail.mass_mailing objects
* Security rule to allow marketing manager to delete all unsent
  mail.mass_mailing objects

For sent mail.mass_mailing objects, only a user in Administration / Settings
group could remove them, like standard Odoo defines.

cc @Tecnativa


yajo:
Please check the bots.

pedrobaeza (changes requested):
I think this module is not needed:
* In this version, there's a group mass mailing user that allows to unlink mass mailings (https://github.com/odoo/odoo/blob/74918dfa374e6d0468eccd881cb418d923da72ca/addons/mass_mailing/security/ir.model.access.csv#L8),
* There are no restriction for unlinking records. Maybe we want this restriction...
* ondelete='cascade' is already set on the TransientModel (https://github.com/odoo/odoo/blob/74918dfa374e6d0468eccd881cb418d923da72ca/addons/mass_mailing/wizard/test_mailing.py#L13)

What do you think?

yajo:
> There are no restriction for unlinking records. Maybe we want this restriction...

I think this was added as a *nice to have*, but maybe this addon would never exist had this been the situation before, so 👍 for removing it.

pedrobaeza:
Closing then as decided.

PR #176, opened 2017-06-29
[FIX] Remove _ from get_signature_footer function definition

### LINKED ISSUE #160: [8.0] mail_debrand not working
I tried to use mail debrand locally and it did not work for me, the last line mentioning about Odoo was still appearing.

I looked at the code and I did not understand how it could work.

mail_debrand seems to override the function `get_signature_footer` from `odoo/addons/mail`, but the function has a wrong name (extra `_` as first character) to override: `_get_signature_footer` instead of `get_signature_footer`.

Related source code:
 - https://github.com/odoo/odoo/blob/8.0/addons/mail/mail_followers.py#L138
 - https://github.com/OCA/social/blob/8.0/mail_debrand/models/mail.py#L12

I can submit a PR for this tiny fix if somebody could confirm that their is actually a problem (I did not succeed to perform a meaningful test to reproduce on Runbot).

diggy128:
Fixes #160

Upstream function does not have underscore.

### DIFF
--- a/mail_debrand/models/mail.py
+++ b/mail_debrand/models/mail.py
@@ -9,7 +9,7 @@ class MailNotification(models.Model):
     _inherit = "mail.notification"
 
     @api.model
-    def _get_signature_footer(self, user_id, res_model=None, res_id=None,
+    def get_signature_footer(self, user_id, res_model=None, res_id=None,
                               user_signature=True):
         """Generate signature footer only with the chosen parts.
 


pedrobaeza (approved):

yajo (approved):

mail_debrand/models/mail.py
```diff
@@ -9,7 +9,7 @@ class MailNotification(models.Model):
     _inherit = "mail.notification"
 
     @api.model
-    def _get_signature_footer(self, user_id, res_model=None, res_id=None,
+    def get_signature_footer(self, user_id, res_model=None, res_id=None,
                               user_signature=True):
```
pedrobaeza, line 13: Please fix this indentation:
```
mail_debrand/models/mail.py:13:31: E127 continuation line over-indented for visual indent
```

pedrobaeza (changes requested):
You need to repair tests also

StefanRijnhart:
Came across the same problem and fixed it independently of this PR. It adapts the test as well. https://github.com/OCA/social/pull/209

PR #180, opened 2017-07-10
[10.0][MIG] base_search_mail_content

MiquelRForgeFlow:
Standard migration.


pedrobaeza (approved):

base_search_mail_content/models/mail_thread.py
```diff
@@ -0,0 +1,79 @@
...
+                  ('email_from', operator, value),
+                  ('reply_to', operator, value)]
+        recs = self.env['mail.message'].search(domain)
+        return [('id', main_operator, recs.mapped('res_id'))]
+
+    @api.multi
+    def _compute_message_content(self):
+        """ We don't really need to show any content. This field is to be
+        used only by searches"""
+        return ''
```
yajo: Compute fields don't return, they just set `self.field=something`. Just `pass` please.

base_search_mail_content/models/mail_thread.py
```diff
@@ -0,0 +1,79 @@
...
+            elem = etree.Element('field', {
+                'name': 'message_content',
+            })
+            setup_modifiers(elem)
+            node.addnext(elem)
+            res['arch'] = etree.tostring(doc)
+    return res
+
+
+models.BaseModel.fields_view_get = _custom_fields_view_get
```
yajo: This monkey-patching should not get accepted now on v10. Override it as usual with `_inherit = "base"` instead (it is a new v10 feature).

base_search_mail_content/models/res_partner.py
```diff
@@ -0,0 +1,13 @@
...
+# © 2016 Serpent Consulting Services Pvt. Ltd. (<http://www.serpentcs.com>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class Partner(models.Model):
+
+    _name = 'res.partner'
+    _inherit = ['res.partner', 'mail.thread']
```
yajo: Why this file? `res.partner` already inherits from `mail.thread` by default... Isn't it?

yajo (changes requested):

base_search_mail_content/models/res_partner.py
MiquelRForgeFlow: I don't think so.

pedrobaeza: It already inherits or you wouldn't have chatter on partners, so this is not needed.

base_search_mail_content/models/mail_thread.py
MiquelRForgeFlow: Done :+1:

base_search_mail_content/models/res_partner.py
MiquelRForgeFlow: Ok. Let's see if it works now without the file :+1:

base_search_mail_content/models/mail_thread.py
```diff
@@ -0,0 +1,77 @@
...
+        pass
+
+    message_content = fields.Text(
+        string='Message Content',
+        help='Message content, to be used only in searches',
+        compute="_compute_message_content",
+        search='_search_message_content')
+
+
+class BaseModel(object):
```
yajo: I bet this shouldn't inherit from `object`

base_search_mail_content/models/mail_thread.py
```diff
@@ -0,0 +1,77 @@
...
+
+    _inherit = 'base'
```
yajo: Shouldn't this be `mail.thread` instead? So it should be merged with the class above... 🤔

base_search_mail_content/models/mail_thread.py
```diff
@@ -0,0 +1,77 @@
...
+    def _fields_view_get(self, view_id=None, view_type='form', toolbar=False,
+                         submenu=False):
+        """
+        Override to add message_ids field in all the objects
+        that inherits mail.thread
+        """
+        res = super(BaseModel, self)._fields_view_get(
+            view_id=view_id, view_type=view_type, toolbar=toolbar,
+            submenu=submenu)
+        if view_type == 'search' and self._fields.get('message_content'):
```
yajo: If you add the method in `mail.thread`, then you don't need to check `and self._fields.get('message_content')`, since it would *always* be true.

base_search_mail_content/models/mail_thread.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _search_message_content(self, operator, value):
+
+        main_operator = 'in'
+        if operator in expression.NEGATIVE_TERM_OPERATORS:
+            main_operator = 'not in'
+            operators = {'!=': '=', 'not like': 'like',
+                         'not ilike': 'ilike', 'not in': 'in'}
+            operator = operators[operator]
+        domain = [('model', '=', self._name), '|', '|', '|', '|',
```
yajo: It seems to me that it would be much cleaner and more efficient to do something like:

```python
domain = [('model', '=', self._name)]
if operator not in expression.NEGATIVE_TERM_OPERATORS:
    domain += ["|"] * 4
domain += [('record_name', operator, value),
           ('subject', operator, value), 
           ('body', operator, value),
           ('email_from', operator, value),
           ('reply_to', operator, value)]
```

base_search_mail_content/models/mail_thread.py
```diff
@@ -0,0 +1,77 @@
...
+    @api.multi
+    def _compute_message_content(self):
+        """ We don't really need to show any content. This field is to be
+        used only by searches"""
+        pass
+
+    message_content = fields.Text(
+        string='Message Content',
+        help='Message content, to be used only in searches',
+        compute="_compute_message_content",
```
yajo: Maybe you can `compute=lambda self: False` instead. Just a suggestion.

yajo (changes requested):

base_search_mail_content/models/mail_thread.py
MiquelRForgeFlow: Can you explain a bit more? You said before that we should override it with `_inherit = "base"`. 🤔

yajo: Yeah, I know 😆 

Sorry for confusing. I said that before because it's the new way to avoid monkey patches, but after seeing this now, I realized that you only need this code for addons that inherit from `mail.thread`, so possibly adding the method there should be better.

base_search_mail_content/models/mail_thread.py
```diff
@@ -22,42 +22,34 @@ def _search_message_content(self, operator, value):
             operators = {'!=': '=', 'not like': 'like',
                          'not ilike': 'ilike', 'not in': 'in'}
             operator = operators[operator]
```
yajo: Delete all above this in the method.

base_search_mail_content/models/mail_thread.py
```diff
@@ -22,42 +22,34 @@ def _search_message_content(self, operator, value):
...
+        domain = [('model', '=', self._name)]
+        if operator not in expression.NEGATIVE_TERM_OPERATORS:
+            domain += ["|"] * 4
+        domain += [('record_name', operator, value),
+                   ('subject', operator, value),
+                   ('body', operator, value),
+                   ('email_from', operator, value),
+                   ('reply_to', operator, value)]
         recs = self.env['mail.message'].search(domain)
         return [('id', main_operator, recs.mapped('res_id'))]
```
yajo: Delete these 2 lines and `return domain`

yajo (changes requested):

base_search_mail_content/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "version": "10.0.1.0.0",
+    "author": "Eficent,"
+              "SerpentCS,"
+              "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/social",
+    "category": "Social",
+    "data": ["data/trgm_index_data.xml",
+             "views/trgm_index_view.xml"],
+    "depends": ["mail",
+                "base_search_fuzzy"],
```
yajo: Is this really a dependency?

yajo (changes requested):

base_search_mail_content/__manifest__.py
MiquelRForgeFlow: Yes.
- `mail` is because the inherited `mail.thread` class
- `base_search_fuzzy`, a module in server-tools repository, is because the inherited `trgm_index_view.xml` view

base_search_mail_content/models/mail_thread.py
```diff
@@ -0,0 +1,62 @@
...
+                'message_content': {
+                    'type': 'char',
+                    'string': 'Message content',
+                }
+            })
+
+            for node in doc.xpath("//field[1]"):
+                # Add message_content in search view
+                elem = etree.Element('field', {
+                    'name': 'message_content',
```
yajo: This does not seem to be working, maybe yes it was required to inherit from `base` instead.

base_search_mail_content/models/mail_thread.py
```diff
@@ -0,0 +1,62 @@
...
+from odoo.osv.orm import setup_modifiers
+
+
+class MailThread(models.AbstractModel):
+
+    _inherit = 'mail.thread'
+
+    def _search_message_content(self, operator, value):
+
+        domain = [('model', '=', self._name)]
```
yajo: This seems to break the UI:

```
Odoo Server Error

  File "/.repo_requirements/odoo/odoo/http.py", line 640, in _handle_exception
  File "/.repo_requirements/odoo/odoo/http.py", line 677, in dispatch
  File "/.repo_requirements/odoo/odoo/http.py", line 333, in _call_function
  File "/.repo_requirements/odoo/odoo/service/model.py", line 101, in wrapper
  File "/.repo_requirements/odoo/odoo/http.py", line 326, in checked_call
  File "/.repo_requirements/odoo/odoo/http.py", line 935, in __call__
  File "/.repo_requirements/odoo/odoo/http.py", line 506, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 827, in search_read
    return self.do_search_read(model, fields, offset, limit, domain, sort)
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 849, in do_search_read
    offset=offset or 0, limit=limit or False, order=sort or False)
  File "/.repo_requirements/odoo/odoo/models.py", line 4678, in search_read
    records = self.search(domain or [], offset=offset, limit=limit, order=order)
  File "/.repo_requirements/odoo/odoo/models.py", line 1518, in search
    res = self._search(args, offset=offset, limit=limit, order=order, count=count)
  File "/.repo_requirements/odoo/odoo/addons/base/res/res_partner.py", line 633, in _search
    count=count, access_rights_uid=access_rights_uid)
  File "/.repo_requirements/odoo/odoo/models.py", line 4220, in _search
    query = self._where_calc(args)
  File "/.repo_requirements/odoo/odoo/models.py", line 4019, in _where_calc
    e = expression.expression(domain, self)
  File "/.repo_requirements/odoo/odoo/osv/expression.py", line 643, in __init__
    self.parse()
  File "/.repo_requirements/odoo/odoo/osv/expression.py", line 821, in parse
    raise ValueError("Invalid field %r in leaf %r" % (left, str(leaf)))
ValueError: Invalid field 'model' in leaf "<osv.ExtendedLeaf: ('model', '=', 'res.partner') on res_partner (ctx: )>"
```

yajo (changes requested):
Code looks good, but UI fails.

base_search_mail_content/models/mail_thread.py
MiquelRForgeFlow: And what do you propose? 😮

base_search_mail_content/models/mail_thread.py
MiquelRForgeFlow: So... return again to:
```
class BaseModel(object):
    _inherit = 'base'
```
🤔?

base_search_mail_content/models/mail_thread.py
yajo: Possibly changing this to `domain=[]` fixes it.

base_search_mail_content/models/mail_thread.py
yajo: So it seems... Remember that reviewers can be wrong, you should test what you do and let it as it was if new patch does not work 😉

MiquelRForgeFlow:
@Yajo , could you please update your review?

yajo:
I feel current code will not work, have you tested it?

MiquelRForgeFlow:
I have just tested, but it has some issues that should be fixed

rafaelbn:
Please @mreficent ping me when module is ready for a functional test! thanks

MiquelRForgeFlow:
@Yajo @rafaelbn Issues solved! Now it's ready functional! :green_apple:

rafaelbn:
Hi @mreficent thank you!!! you should improve test as it has 42.85% of diff hit (target 98.81%)

I will wait until @Yajo finish his review ;-) for testing

base_search_mail_content/README.rst
```diff
@@ -0,0 +1,88 @@
...
+The current search capabilities include searching into the subject, body,
+email from, reply to and record name.
+
+Installation
+============
+
+This module depends on the module 'base_search_fuzzy' to ensure that
+searches on emails are based on indexes. Please read carefully the install
+instructions:
+https://github.com/OCA/server-tools/blob/10.0/base_search_fuzzy/README.rst
```
elicoidal: Maybe better:
```
`install instructions <https://github.com/OCA/server-tools/blob/10.0/base_search_fuzzy/README.rst>`_.

File: base_search_mail_content/README.rst [hunk 15]
```diff
@@ -0,0 +1,88 @@
...
+
+Contributors
+------------
+
+* Jordi Ballester Alomar <jordi.ballester@eficent.com>
+* Serpent Consulting Services Pvt. Ltd. <support@serpentcs.com>
+* Lois Rilo Antelo <lois.rilo@eficent.com>
+* Aaron Henriquez <ahenriquez@eficent.com>
+
+
```
elicoidal: remove line

elicoidal (approved):
non-blocking suggestions

base_search_mail_content/models/mail_thread.py
```diff
@@ -0,0 +1,71 @@
...
+        return [('id', 'in', recs.mapped('res_id'))]
+
+    message_content = fields.Text(
+        string='Message Content',
+        help='Message content, to be used only in searches',
+        compute=lambda self: False,
+        search='_search_message_content')
+
+
+class BaseModel(object):
```
yajo: I can't seem to do quick search of message contents. Most likely this should inherit from `models.AbstractModel` instead, although I think you already fixed this before... 🤔

yajo (changes requested):

base_search_mail_content/models/mail_thread.py
MiquelRForgeFlow: Postgresql decides when is best to use trigram indexes. You will need to install this module into a DB with a rather large number of emails. 

To check if the trigram indexes are being used, you must:

1. start odoo with --log-level=debug_sql
2. Do the search that you want to analyze - e.g. search for emails from the leads. 
3. Copy the query and run in command line `EXPLAIN ANALYZE (my_query)`:
`
EXPLAIN ANALYZE (SELECT "mail_message".id FROM "mail_message" WHERE (("mail_message"."model" = 'res.partner')  AND  ((((("mail_message"."record_name"::text ilike 'dear')  OR  ("mail_message"."subject"::text ilike 'dear'))  OR  ("mail_message"."body"::text ilike 'dear'))  OR  ("mail_message"."email_from"::text ilike 'dear'))  OR  ("mail_message"."reply_to"::text ilike 'dear'))) ORDER BY "mail_message"."id" DESC);`

This will tell you if trigram indexes are being used.

yajo: what?

mreficent:
now travis fails but due to other module, not this one

JordiBForgeFlow:
the field `message_content` was not being added to the search view. I have fixed that, and now the module is :star2: :star2: :star2:

mreficent:
:green_apple: :green_apple:

MiquelRForgeFlow:
@Yajo @rafaelbn update review please :green_apple:

yajo (approved):
Excelent!

Please do the final squash

MiquelRForgeFlow:
Squash done

JordiBForgeFlow:
@rafaelbn can you review?

pedrobaeza:
Is there the possibility of improving test coverage?

rafaelbn (commented):
Tested functionaly 👍 but  codecov/patch — 38.46% of diff hit (target 98.81%) must be increased please! This is a great funcionality

ping @mreficent @jbeficent

mreficent:
Yes. Let's wait a bit :stuck_out_tongue_winking_eye:

base_search_mail_content/tests/test_base_search_mail_content.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo.tests.common import TransactionCase
+
+
+class TestBaseSearchMailContent(TransactionCase):
+    def setUp(self):
+        super(TestBaseSearchMailContent, self).setUp()
+        self.thread_obj = self.env["mail.thread"]
+
+    def test_base_search_mail_content(self):
+        res = self.thread_obj._search_message_content('not ilike', 'xxxyyyzzz')
```
yajo: Actually the test should be `self.thread_obj.search([("message_content", ...`

base_search_mail_content/tests/test_base_search_mail_content.py
```diff
@@ -0,0 +1,15 @@
...
+        self.assertEqual(res[0][2], [], "You have a message with xxxyyyzzz :O")
```
yajo: and then here assertFalse

yajo (changes requested):

MiquelRForgeFlow:
There is not codecov/patch now :joy:

rafaelbn (approved):
Tested functionaly 👍

rafaelbn:
@mreficent please could answer @Yajo 's comments ? I wish tomorrow we can merge this one! Great work!

MiquelRForgeFlow:
Which comments? All @Yajo's comments have been attended I think.

yajo (approved):

yajo:
Cool! could you squash the migration commits to merge please? 😊

MiquelRForgeFlow:
Done! codecov/patch at 100%

PR #181, opened 2017-07-14
[IMP] Shut mock exceptions up

chienandalu:
Mock tests exceptions make Travis fail on every dependence hanged on this module. That makes impossible to fully complete tests on those PRs. This hopefully solves it.

@Tecnativa

### DIFF
--- a/mail_tracking/tests/test_mail_tracking.py
+++ b/mail_tracking/tests/test_mail_tracking.py
@@ -3,6 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import mock
+from odoo.tools import mute_logger
 import base64
 import time
 from odoo import http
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please use this trick instead: https://github.com/OCA/server-tools/pull/873#discussion_r125333008

chienandalu:
Yes, @pedrobaeza, I tried something like `@mute_logger(mock_send_email)` with no success :disappointed:

pedrobaeza:
Check the example and you have to find the function that emits the log. It's nothing to do with mock, but the patched function instead.

chienandalu:
Well, `mock_send_email` is a global for `'odoo.addons.base.ir.ir_mail_server.IrMailServer.send_email'` wich is the functions we are trying to silence, isnt't it?

pedrobaeza:
Well, I haven't deep enough right now. Make several tries, and if not, at least reduce the scope of the mute to only the line that gives the error.

chienandalu:
@pedrobaeza I tried some things but still no success with mute_logger :( I reduced the scope of the logger shut though.

chienandalu:
@pedrobaeza Got it!

chienandalu:
Travis now fails on mail_tracking_mailgun. Similar issue and fix: https://github.com/OCA/social/pull/182

pedrobaeza:
Everything green, so I fast-track

PR #184, opened 2017-07-19
[9.0][FIX] fetchmail_thread_default: check model in env for update module

sergio-teruel:
@Yajo Hi, when you update the module all models are not be loaded in environment
cc @Tecnativa


pedrobaeza (approved):
Well, this is a workaround for a bad removed model, but it's harmless

yajo (approved):

pedrobaeza:
Merging as obvious.

Issue #175, opened 2017-06-23
[RFC] mail_thread_recursive: Multiple threads per record


lasley:
This module will provide the ability for true threaded conversations on Odoo models. In the current implementation of `mail.thread`, a record is considered to have one thread & is therefore considered one discussion. In some contexts, however, multiple independent discussions may be occurring - all with their own context and followers. 

An example of this is a blog, such as [this article](https://blog.laslabs.com/2013/01/pxe-server-ubuntu/), which contains numerous independent conversations regarding different aspects of the article. Each person that commented on the article could care less about the other comments on the article, but they also want to receive notifications for replies to their comment.

To be honest I have absolutely no idea where to begin here, so for a first time ever - I am submitting a blank RFC & am hoping that someone has an idea. I'll keep pondering as well, and update this if I come up with something substantial. 

Following are the high level requirements:
* Creating a new message in a record should create an entirely new `mail.thread` associated to the record
* All `mail.threads` associated to the record should be displayed in reverse chronological order (of last message received)
  * All messages of the `mail.thread` should be grouped together (think of email conversation view)
  * `mail.thread` views should be collapsible. 
* Following the record will follow all `mail.thread`s for the record
* `mail.thread` operation should not be changed
* Logic is required to catch the calls to all message functions at the record level & perform recursive actions instead

I'm thinking we can accomplish this view-side by simply iterating the `mail.thread` widgets. The real stumpers are:
* What to encapsulate the `mail.thread`s in
* How to override the `mail.thread` functionality from models already implementing that functionality, providing both the recursive and the regular threads (a regular thread would just be treated in the same way as the recursive I'm thinking)

So yeah as I said, this is a totally rough sketch. Anyone have some pointers or existing code?

lukebranch:
I've come across this need a number of times, it's definitely something that would be very useful for Odoo.

I've seen a rough implementation of this concept in Slack:

<img src="https://user-images.githubusercontent.com/2337666/27507040-9b5855c6-58f9-11e7-8f9f-25909047fcb3.png" />

I don't have anything useful to add to the discussion at this point in time, however i'll comment back if I come up with some ideas.

hbrunn:
[...]

PR #190, opened 2017-08-18
[ADD] mail_message_thread: Threaded messages

lasley:
This is an interestingly simple PoC for #175, implemented by inheriting `mail.thread` inside of `mail.message`, which makes it so that each message can be replied to independently. 

There is currently no UI because I'm still in conceptual. Below is the proof that it works though:

```
>>> env.user.partner_id.message_post('Test Message')
mail.message(678,)
>>> env.user.partner_id.message_ids
mail.message(678,)
>>> env.user.partner_id.message_ids[0].message_post('Test Reply')
mail.message(680,)
>>> env.user.partner_id.message_ids[0].message_ids
mail.message(680, 679)
>>> env.user.partner_id.message_ids[0].message_ids[0].body
u'<p>Test Reply</p>'
>>> env.user.partner_id.message_ids[0].message_ids[1].body
u'<p>Mail Message (Threaded) created</p>'
```

You'll notice in the above that there are actually two messages in the reply thread, one of which is a log being generated [here](https://github.com/odoo/odoo/blob/10.0/addons/mail/models/mail_thread.py#L231):

```
u'<p>Mail Message (Threaded) created</p>'
```

Does anyone know why this is necessary? This seems incredibly redundant to me; why is this information useful when we already have a create date and uid on the message itself? I see no other use of this logging.

cc @hbrunn


hbrunn:
I can't look into the minds of Odoo devs, but I think the rationale here is that this message creates a notification, and in some contexts, you want to get this as mail.
Unfortunately, the `context` field parameter doesn't work the way for many2many fields as it does for one2many ones, so you'll have to set this manually in mail.message's create function

hbrunn:
closing as @lasley probably won't finish this PR

PR #191, opened 2017-08-31
[MIG] mail_debrand: Migration to 10.0

lucode:
and adding DE language file


pedrobaeza:
There's already https://github.com/OCA/social/pull/168 that migrates this module. Please review that one instead.

PR #192, opened 2017-09-07
ADD module mail_inline_css

bealdav:
When you send HTML emails you can't use style tags but instead you have
to put inline ``style`` attributes on every element.

This module use premailer library to do this

### DIFF
--- a/mail_inline_css/README.rst
+++ b/mail_inline_css/README.rst
@@ -55,6 +55,14 @@ Just use any mail template as Odoo standard feature
    :target: https://runbot.odoo-community.org/runbot/205/10
 
 
+Note:
+
+  Odoo with module web_editor already implements this feature on the client side (js).
+  This module brings this server side feature for cases without js part. It could the more stable way over the Odoo versions with a stable api in a dedicated library 
+  with adhoc python unit tests.
+
+
+
 Bug Tracker
 ===========
 
--- a/mail_inline_css/README.rst
+++ b/mail_inline_css/README.rst
@@ -1,0 +1,101 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+===============
+Mail Inline CSS
+===============
+
+When you send HTML emails you can't use style tags but instead you have
+to put inline ``style`` attributes on every element. So from this:
+
+.. code:: html
+
+    <html>
+    <style type="text/css">
+    h1 { border:1px solid black }
+    p { color:red;}
+    </style>
+    <h1 style="font-weight:bolder">Peter</h1>
+    <p>Hej</p>
+    </html>
+
+You want this:
+
+.. code:: html
+
+    <html>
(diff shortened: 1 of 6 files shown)


yajo:
Did you know that v10 mass mailing editor has a method that automatically converts classes into inline styles when you send them, and the other way around when you edit them?

bealdav:
No, I don't know.
This method is in Odoo or OCA ?
Is mailing editor is website dependent ? 
Sorry for my bad knowledge on these topics :-(

Thanks for your indications

yajo:
It's core odoo and it's depending from web_editor.

The relevant source code is here: https://github.com/odoo/odoo/blob/10.0/addons/web_editor/static/src/js/transcoder.js

bealdav:
@Yajo thanks to point this. Nice addition.

On our side we don't want for now use this editor to better control produced code (maintenability). Odoo evolve a lot in .js .css .html sides in v11 in surely in v12 too. Then we want to use solution away of this stack and python library seems to be the more stable way. I think other module consummers could share our thinkings.

Do you consider this module could be find a place here ?

Thanks for your answer.

rafaelbn:
> On our side we don't want for now use this editor to better control produced code (maintenability)

This doesn't sound crazy 😄 . In my experience with Odoo It has been a huge headache how Odoo spits the the mail code. As for example is imposible to make a mail compatible with MS Outlook (desktop)...

yajo:
Current design in v10 is much better than prior versions IMHO: You can use themes for mails, and those themes can be only CSS, which is great. Upon sending, Odoo will convert those classes into inline styles, so you have good code while developing, and ugly code only while sending (which is good because the mail client landscape is full of 💩-y html renderers, with Outlook leading the way to 💩💩💩).

The only real benefit then of having vanilla `<style>` tags would be the ability to have media queries, which would be awesome because we could then start using some of the widely available pseudoresponsive email frameworks, but then the problem is is https://github.com/odoo/odoo/issues/10954, by which we don't have `<body>` or `<head>` tags. If you put a `<style>` outside a `<head>`, it will simply get ignored by any html engine, so unless your addon is fixing that part, I don't think it makes much sense honestly... 😕

sebastienbeau:
Hi Yajo, maybe this is a misunderstood.
Indeed in 10 the mass mailing use the widget "web_editor". In our case this module only convert html to inline css for email_template not the mass mailing.
Indeed the email_template do not use web_editor, you can edit the html using the basic html editor or by editing the html code.
As email_template is not something that change a lot eidting directly the html is a good option.
What do you think?

yajo:
I was just advising you, just in case you were reinventing the transcoder or you were going the hard-to-workaround https://github.com/odoo/odoo/issues/10954. If this is still not your use case, or you found a way to workaround https://github.com/odoo/odoo/issues/10954 or it doesn't affect you for any reason, then go on please, it's up to you 😊

sylvainc:
:+1:

hparfr (approved):

rafaelbn:
Please @bealdav review travis, it's failing . Thanks! 😄

bealdav:
@rafaelbn fixed

sebastienbeau (approved):
Code review, of for me

Issue #121, opened 2016-11-03
RFC: Message Anti-Spam Module


lasley:
We are currently planning a new module to combat spam messages in Odoo. The plan at the moment is to implement a basic backpropagation algorithm [similar to this](http://radimrehurek.com/data_science_python/).

The algorithm will be trained initially using some open source datasets like [Enron](http://www.aueb.gr/users/ion/data/enron-spam/), [SpamAssassin](https://spamassassin.apache.org/publiccorpus/), [LingSpam](http://csmining.org/index.php/ling-spam-datasets.html), etc. 

It will create a private Junk channel for each user, automatically filtering identified spam to it. Any item moved out of junk would be marked as a false positive, and anything moved into junk would be considered a false negative. TBD how this will play into the learning algorithm.

Messages would be evaluated in the `create` method of `mail.message`.

Spam scanning will need to be strictly opt-in. I'm thinking it makes sense to put this at the model level; something similar to the [`ir.model.website_form_access`](https://github.com/odoo/odoo/blob/10.0/addons/website_form/models/models.py#L19) used for forms, with the `mail.message.res_model` dictating the model.

Challenges I currently foresee:
* The obvious - training the algorithm
* Support of multiple languages
* Handling of spam not directed to specific users (projects, help desk, blog, crm)
* Storage and subsequent search of data. [Looks like pickle may work though](http://scikit-learn.org/stable/modules/model_persistence.html), but with some challenges in terms of upgrades - we will need to add versioning, re-training, and a/b testing.

Nice to haves:
* Model training on a user level (Cindy may like spam from `CompanyA`, but John might not) - this will require training the SVM in an incremental manor, instead of batch.
  * This is not so easy in `sklearn`, and we will be limited to linear case using Stochastic Gradient Descent (SGD). Would need to use something like [`pagasos`](https://github.com/ejlb/pegasos) instead
* Client/Server model, allowing for centralized & offloaded training

Additional reading for anyone interested:
* http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.455.7614&rep=rep1&type=pdf
* http://trec.nist.gov/pubs/trec15/papers/hit.spam.final.final.pdf
* http://ats.cs.ut.ee/u/kt/hw/spam/spam.pdf
* http://cse-wiki.unl.edu/wiki/index.php/Machine_Learning_for_Email_Spam_Filtering
* http://airccse.org/journal/jcsit/0211ijcsit12.pdf
[...]

PR #193, opened 2017-09-08
[ADD] mail_anti_spam: Anti-SPAM filtering

lasley:
This is a WIP of the anti-spam module described in #121.

SPAM filtering is implemented using Bayesian classification, powered by [LasLabs/python-reverend](https://github.com/LasLabs/python-reverend).

Bayesian logic is one of the standard ways that emails are classified in spam filters, such as SpamAssassin (described more in https://github.com/OCA/social/pull/193#issuecomment-329803134).

The current big question is how this is going to perform in regards to false positives, and with other languages - as somewhat described in https://github.com/OCA/social/pull/193#issuecomment-329543021.

There are a few other combiners (algorithms) that I can use here too, so the other question is exactly how well each performs. I'm working up some test cases on all that, hopefully with automatable benchmarks that can be obtained through our CI.


lasley:
Does anyone have some SPAM samples in non-English that they'd be willing to donate? I'd like to test this for non-English, which is typically where services like this are weak.

yajo:
Today I feel magnanimous and I will donate 1 Spanish SPAM message to the cause:

```
(saved and removed from comment)
```

lasley:
Thanks @Yajo - this would seriously be a good one in English too. Wow, this is seriously good SPAM.

Kind of funny though - they were going for replicating a thread, but didn't MIME encapsulate each message part correctly. Oh you spammers 😆

lasley:
Alright so this is doing absolutely horrible on blog spam, and the Spanish spam. 

I think I'm going to take another stab at the Bayesian using [redisbayes](https://github.com/jart/redisbayes) and OCA/server-tools#978

lasley:
Okkkkk so more and more testing, and I have now taken over maintenance of what looks to be the 
de facto (yet previously unmaintained) Bayesian library for Python - [reverend](https://github.com/LasLabs/python-reverend). I Pythonified it and am prepping for Py-3 overhaul.

It's now integrated here using the Pythonified version, with a training wizard that runs based off of [this Spam Archive](http://untroubled.org/spam/) by default, but can easily be adapted for whatever.

Still a bunch of testing to be done, but this looks promising.

mail_spam/models/reverend_thomas.py
```diff
@@ -0,0 +1,152 @@
...
+            messages (MailMessage): ``mail.message`` recordset to train with.
+            spam_or_ham (str): Value to train as (either self.SPAM or self.HAM)
+        """
+        for record in self:
+            for message in messages:
+                message_string = self._parse_message(message)
+                record.client.train(message_string, spam_or_ham)
+        self._save_db()
+
+    @api.model_cr_context
```
yajo: Weird decorator. Why not model or multi? .. or nothing?

mail_spam/models/reverend_thomas.py
```diff
@@ -0,0 +1,152 @@
...
+    def _get_client(self):
+        """Return a Bayes client."""
+        return Bayes()
+
+    @api.model_cr_context
```
yajo: Same here

mail_spam/models/reverend_thomas.py
```diff
@@ -0,0 +1,152 @@
...
+        """Parse a ``mail.message`` object"""
+        message.ensure_one()
+        author = message.author_id
+        from_header = author and author.email or message.email_from
+        return OrderedDict([
+            ('from', 'FROM: %s' % from_header),
+            ('body', message.body),
+        ])
+
+    @api.model_cr_context
```
yajo: Same

mail_spam/wizards/reverend_thomas_training.py
```diff
@@ -0,0 +1,112 @@
...
+            with self._temp_dir() as temp_dir:
+                command = self.unarchive_command % {
+                    'directory': temp_dir,
+                    'file_name': temp_file.name,
+                }
+                subprocess.check_call(command.split(' '))
+                return temp_dir
+
+    @contextmanager
+    @api.model_cr_context
```
yajo: same

yajo (changes requested):
I've got not much idea on spam filtering, honestly, but just some simple thoughts:

- Name the module `mail_antispam`. `mail_spam` is exactly the opposite of what this does!
- I feel that the training wizard is good for development, but I'd personally expect this one to work out of the box, so it should execute the wizard after installation, and possibly once a month (because it seems the spam source you use gets updated monthly).

Good job BTW!

mail_spam/models/reverend_thomas.py
lasley: This is new in v10 and is meant to decorate something that can be called from both a multi and a model api without issues.

The purpose of these methods is to be as useful as possible. Using either `multi`, or `model` limits that usefulness by requiring the method be called from within that type of decorator.

That said, the best way to go about this is to circumvent Odoo model api entirely and use a combination of class and static methods here.

lasley:
> I've got not much idea on spam filtering, honestly, but just some simple thoughts:

Hah me neither. Seems I'm now the maintainer of a pretty badass Bayesian classification library though, so I guess I've become an expert all of a sudden 😆 

> I feel that the training wizard is good for development, but I'd personally expect this one to work out of the box, so it should execute the wizard after installation, and possibly once a month (because it seems the spam source you use gets updated monthly).

I'm thinking along similar lines, although there are a few problems:

* The training wizard currently classifies all existing messages as "ham". This could be incredibly bad on databases that are long-lived and contain many pre-existing spam messages. While training is certainly required before usage, a wise admin may be able to make some better selections instead.
* My source (and all other spam sources) are all English. This won't really play nicely with anyone in other languages
  * Turns out other languages really get the short end of the stick here. Nearly all anti-SPAM research (and most Machine Learning in general) is English based. This made things really quite difficult, such as in the instances where I was trying true machine learning to filter things out instead of a naive Bayesian. ML requires data-points, which in the instance of strings, are created primarily by turning true messages into a structure of [lemmas](https://en.wikipedia.org/wiki/Lemma_(morphology)) instead (called a [Bag of Words](https://en.wikipedia.org/wiki/Bag-of-words_model)). There is literally no lemma dictionary for non-english that I could find.

yajo:
I guess for a 1.0 approach, English filtering is good. After all, everybody gets English SPAM. And most Spanish SPAM, at least for me, is google-translated.

You can keep a note in the roadmap about being able to get pre-trained in other languages, and that in the mean time you'll have to train it by yourself.

The problem also is that, for example, if I'm Spanish and get English messages, most likely they will be SPAM... All of this is tricky, man! 😨 

Isn't it possible to filter incoming messages through a spamassassin/clamav/whatever external pro antispam service and delegate all that specific knowledge in the hands of those who have it? I think that would be less scary to maintain...

lasley:
> Isn't it possible to filter incoming messages through a spamassassin/clamav/whatever external pro antispam service and delegate all that specific knowledge in the hands of those who have it? I think that would be less scary to maintain...

I don't really know of any with an API. 

Also I'm not sure about your implementation, but my Spamassassin uses Bayesian via [`sa-learn`](https://spamassassin.apache.org/full/3.1.x/doc/sa-learn.html). Outside of the Bayesian filter, everything else is based on things like the static rule sets, DKIM, SPF, etc. which aren't relevant in most cases here because we're not email based. 

Keep in mind that email based solutions are trained for emails, which include the MIME headers. A significant chunk of our data does not look like this which will skew results (this is also why Pyzor failed here when I tried it).

Furthermore, all of the filters I know of are trained only on English out of the box and will start (about) as dumb as this one. I can't find Spamassassin's training set right now, but it was last updated in like 2002 and is actually rather small (although does contain ham, which we don't have).

With all that said - I think the appropriate plan is to test how this works when completely untrained, and how the functionality changes over time as it is trained. My hope is that it will start dumb then just eventually find its way, with not many false positives. My fear is that right at the beginning we're going to have a bunch of false positives - which will be lost emails basically and isn't acceptable.

mail_anti_spam/README.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Roadmap
+-------
+
+* Add button to mark as spam in thread
+* Add a virtual channel to view SPAM messages and allow to mark as ham
+* Make sure that notifications do not go out to users for SPAM messages
+* Create user action definitions that can be used as training (such as
+  a user replying [ham] or unlinking [spam])
+
```
lasley, line 39: Add forget

mail_anti_spam/models/reverend_thomas.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class ReverendThomas(models.Model):
+    """This model acts as a proxy for Reverend Thomas Bayes operations.
+
+    All actual Bayesian classifications happen in this model, and are stored
+    to the internal database file in the ``database`` field.
+    """
+
+    _name = 'reverend.thomas'
```
lasley, line 27: Implement interface/adapter. The reverend should implement the interface

hbrunn:
closing as @lasley probably won't finish this PR

sotolucas:
Sorry for reopening, but [issue](https://github.com/OCA/social/issues/121) is closed too and it has a link to this PR.
As this repository has the approach of use Mailgun, maybe we could work on a solution that use X-Mailgun-SFlag and X-Mailgun-SScore headers (that are sent if we force to deliver spam anyway), and also it would be easier to develop.
Thougts @hbrunn @lasley @Yajo ?

yajo:
What we need mostly is a motivated developer to do it. @lasley is out of the game since long ago, and this subject is complex enough for most of us. :man_shrugging:

Issue #194, opened 2017-09-12
[RFC] mail_outbound_static: Email FROM Headers


lasley:
This module will allow you to configure the FROM header for any outbound email server, overriding the default operation of using the partner's email address in the FROM header.

The reasoning behind this is because manipulation of the FROM header is exactly what spammers do, and is called [email spoofing](http://searchsecurity.techtarget.com/definition/email-spoofing). This is why all of my emails to Odoo Community mailing lists go to SPAM, and this is why Odoo is not compatible with Amazon SES.

For anyone interested in email spoofing, particularly those of you who have email on your own domain and are not going to people's spam, I recommend you read [this article](https://www.digitalocean.com/community/tutorials/how-to-use-an-spf-record-to-prevent-spoofing-improve-e-mail-reliability) and implement.

Simple enough of a goal here - was just wondering if anyone knows of something pre-existing that we can base on for this?

lasley:
cc @gdgellatly

gdgellatly:
@lasley my own solution is to use my personal email address rather than
work ones.

One possibility i can think of is a secondary partner domain. Either a
subdomain or completely different which is automatically used when an email
address is not the main domain.

But more simply, why not just use the reply to as the from when domain
doesn't match and just manipulate the extra parts of the from header., E.g <
list@odoo.com> Odoo list (on behalf of x@x)

gdgellatly:
Or actually, what GitHub does is perfect

lasley:
> my own solution is to use my personal email address rather than
work ones.

Even my personal emails are on my own domain, so I'm screwed here.

> One possibility i can think of is a secondary partner domain. Either a
subdomain or completely different which is automatically used when an email
address is not the main domain.
> 
> But more simply, why not just use the reply to as the from when domain
doesn't match and just manipulate the extra parts of the from header., E.g <
list@odoo.com> Odoo list (on behalf of x@x)

With SES, you have to verify each outbound email address individually. Any sort of FROM header manipulation outside of a standardized email causes major scaling issues.

I'm not sure why Odoo chose header manipulation here, but this module will basically imitate the way that other true Enterprise products handle the email. The FROM Header will be defined in a configuration, the REPLY-TO as well. The only thing that should change is the Canonical From name in the header.

> Or actually, what GitHub does is perfect


Yup exactly. This is what I'm shooting for.

gdgellatly:
There are 2 cases I think. Mail from domain users. We handle this by using
[...]

PR #195, opened 2017-09-12
[ADD] mail_outbound_static: Allow configuration of from email header

lasley:
* Allow for the configuration of the FROM header for outbound emails, and move the existing from header to Sender as per RFC-2822

I still need to do some battle testing here, but the headers all look solid. 

REF #194 & odoo/odoo#3347


lasley:
Closing for https://github.com/odoo/odoo/pull/20911

PR #196, opened 2017-09-20
[FIX] mass_mailing_partner: permissions

chienandalu:
- Oddo changed `mass_mailing_contact`permissions from Employee to Mass Mailing User.
- That raises exceptions for users not in that group in partner view whenever accessing views or methods in wich such field is implied.

cc @Tecnativa

### DIFF
--- a/mass_mailing_partner/models/res_partner.py
+++ b/mass_mailing_partner/models/res_partner.py
@@ -19,7 +19,8 @@ class ResPartner(models.Model):
         comodel_name='mail.mass_mailing.contact', inverse_name='partner_id')
     mass_mailing_contacts_count = fields.Integer(
         string='Mailing list number',
-        compute='_compute_mass_mailing_contacts_count', store=True)
+        compute='_compute_mass_mailing_contacts_count', store=True,
+        compute_sudo=True)
     mass_mailing_stats = fields.One2many(
         string="Mass mailing stats",
         comodel_name='mail.mail.statistics', inverse_name='partner_id')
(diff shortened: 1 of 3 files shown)


mass_mailing_partner/models/res_partner.py
```diff
@@ -40,7 +40,7 @@ def _check_email_mass_mailing_contacts(self):
     def _compute_mass_mailing_contacts_count(self):
         for reg in self:
             reg.mass_mailing_contacts_count = len(
-                reg.mass_mailing_contact_ids)
+                reg.sudo().mass_mailing_contact_ids)
```
pedrobaeza: This one shouldn't be needed

pedrobaeza:
@cubells @luismontalba please review

cubells (approved):
:+1: 

LGTM

code review

mass_mailing_partner/models/res_partner.py
chienandalu: Yes, but permissions were raising without it :/

luismontalba (approved):

pedrobaeza:
Please change version number squashing

yajo:
I think that if a user doesn't have access to the records, he shouldn't have access to its `len()` also.

I think it would be better that you just `try:` to get it, and on `except AccessDenied`, you return `0`.

chienandalu:
@Yajo Ok, I changed to behave as you suggested. Anyway I just pass on the AccessError because the compute field is stored.

yajo:
If it is stored, then you should add `@api.depends`. Otherwise, it will get recomputed always, so no much sense on getting it stored.

pedrobaeza:
You can add the attribute `compute_sudo=True` on the field for avoiding permission problems. I think it's the best option, or it will put incorrect values depending on which user makes the computation to happen.

yajo:
Yeah, I didn't remember that...

chienandalu:
@pedrobaeza @Yajo Changes done

yajo:
There's still https://github.com/OCA/social/pull/196#issuecomment-331122391

chienandalu:
@Yajo But it already had the decorator: https://github.com/OCA/social/pull/196/files#diff-7173efc033fbb9673276a0ff5ad30422R39

yajo (approved):
Yay! My mistake, I thought you were editing the compute method, and it's the constraint one. Forget it, it's OK.

PR #199, opened 2017-09-22
[9.0][FIX] Mail tracking bugfixes

antespi:
This PR fix two bugs found in mail_tracking:
* When an email notification is an answer of another email notification (for example, in a invoice mail thread), then mail_tracking addon add several tracking email IDs in body. This fix assure that there is only one ID (and the last one).
* Support for multi-company. Mail tracking status of partners from other companies raise an access error exception. This fix get status as sudo.

### DIFF
--- a/mail_tracking/models/ir_mail_server.py
+++ b/mail_tracking/models/ir_mail_server.py
@@ -23,7 +23,7 @@ class IrMailServer(models.Model):
         tracking_email_id = False
         # https://regex101.com/r/lW4cB1/2
         match = re.search(
-            r'<img [^>]* data-odoo-tracking-email=["\']([0-9]*)["\']', body)
+            r'<img[^>]*data-odoo-tracking-email=["\']([0-9]*)["\']', body)
         if match:
             try:
                 tracking_email_id = int(match.group(1))
--- a/mail_tracking/__openerp__.py
+++ b/mail_tracking/__openerp__.py
@@ -5,7 +5,7 @@
 {
     "name": "Email tracking",
     "summary": "Email tracking system for all mails sent",
-    "version": "9.0.3.0.1",
+    "version": "9.0.3.0.2",
     "category": "Social Network",
     "website": "http://www.tecnativa.com",
     "author": "Tecnativa, "
--- a/mail_tracking/models/mail_message.py
+++ b/mail_tracking/models/mail_message.py
@@ -38,7 +38,7 @@ class MailMessage(models.Model):
             partner_trackings = []
             partners_already = self.env['res.partner']
             partners = self.env['res.partner']
-            trackings = self.env['mail.tracking.email'].search([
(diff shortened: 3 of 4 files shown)


mail_tracking/models/ir_mail_server.py
```diff
@@ -23,7 +23,7 @@ def _tracking_email_id_body_get(self, body):
         tracking_email_id = False
         # https://regex101.com/r/lW4cB1/2
         match = re.search(
-            r'<img [^>]* data-odoo-tracking-email=["\']([0-9]*)["\']', body)
+            r'<img[^>]*data-odoo-tracking-email=["\']([0-9]*)["\']', body)
```
yajo, line 26: I find all of this would be much easier to maintain and understand by using `lxml.html` to parse the email body, and then XPath to find/remove the node. This regexp for instance would produce false positives with i.e. `"/>`, and given those come from the outside world, it's a good attack vector.

yajo (changes requested):

mail_tracking/models/ir_mail_server.py
antespi: Yes, this is another way to do it.
I think regex is faster and consume less memory. BTW, that vector is not matching the regex, see https://regex101.com/r/lW4cB1/3

yajo: Indeed, I was wrong, but see:

- ``
- `<imgdata-odoo-tracking-email="0'/>`
- ``
- ``
- `" data-odoo-tracking-email="1999"/>`

want more? 😏

antespi: Well, in those cases lxml will break, isn't it?

This fix first remove any tag that match with this regex, then inject the tag and after read the tag, so if there is some broken tag then it's not a security problem. 

Using lxml, what would the code have to do if no valid HTML is read?

yajo: I think it should except and pass in such case, don't you think?

antespi: I think that using lxml and then except and pass when invalid tags in the email will miss a tracking for some emails. In my approach (btw, the current one) those corner cases will be tracked without any problem.

So, I prefer to keep the current implementation because:
* It's more resilience
* It has a better CPU & RAM performance
* It's simpler
* It's the current implementation, less code changes
* It hasn't any security issue

Any other opinions here?

yajo (approved):

rafaelbn (approved):
Thanks! 👍

pedrobaeza:
Please change module version number.  Does this apply to v10? If so, please forward port it.

yajo:
Done for him, no problem on final squashing, could you review @pedrobaeza?

pedrobaeza:
Merged squashing

PR #202, opened 2017-10-03
10.0 mail store outgoing

agent-z28:
we suggest to include our lovely outgoind mail store module. it will help users to see which mails have been sent by odoo.

improvements are welcome!!!!


agent-z28:
Roelt stop laughing :)

mail_store_outgoing/i18n/de.po
```diff
@@ -0,0 +1,132 @@
...
+
+#. module: outgoing_mail_save
+#: view:ir.mail_server:outgoing_mail_save.ir_mail_server_form
+msgid "Test Imap Connection"
+msgstr "Test Imap Connection"
+
+#. module: outgoing_mail_save
+#: field:ir.mail_server,has_separate_imap_server:0
+msgid "Use Separate Imap Server"
+msgstr "Use Separate Imap Server"
```
bealdav: Not sure this file is translated ?

agent-z28: I will add it :D

mail_store_outgoing/models/ir_mail_server.py
```diff
@@ -0,0 +1,118 @@
...
+                flags, delimiter, mailbox_name = self.parse_list_response(line)
+                res = {
+                    'server_id': self.id,
+                    'name': mailbox_name,
+                }
+                imap_pool.create(res)
+            self.write({'imap_mailbox_verified': True})
+        except Exception, e:
+            raise ValidationError(
+                _("Connection Test Failed!"),
```
RoelAdriaans: ValidationError only needs one parameter. 
See https://github.com/odoo/odoo/blob/10.0/odoo/exceptions.py#L73

RoelAdriaans (changes requested):

mail_store_outgoing/models/ir_mail_server.py
agent-z28: Okay i will fix it today

agent-z28:
Hello OCA MERGE Masters :) What else is needed to finally merge my pull request?

yajo:
Hi, you need at least 2 approvals, and also you have to fix (or explain why not) that ❌ in the CI. 😊

mail_store_outgoing/security/ir.model.access.csv
```diff
@@ -0,0 +1,4 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_ir_mail_imap_folder,access_ir_mail_imap_folder,model_ir_mail_imap_folder,,1,0,1,0
+
+
```
bealdav: not blocking for me but why these 2 extra lines ?

mail_store_outgoing/models/ir_mail_server.py
```diff
@@ -0,0 +1,118 @@
...
+            maillib.logout()
+        return True
+
+    imap_mailbox_folder = fields.Many2one('ir.mail.imap.folder',
+                                          'Imap Folders',)
+    imap_mailbox_verified = fields.Boolean('IMAP Connection Verified', )
+    store_outgoing_mail = fields.Boolean('Store Outgoing Mail', )
+    has_separate_imap_server = fields.Boolean('Use Separate Imap Server', )
+    separate_imap_server = fields.Char('Separate Imap Server', )
+    active = fields.Boolean('Active', )
```
bealdav: By convention, fields are on top of the class (not blocking for me)

agent-z28: I moved them to them top now. but the convention is strange. As soon as you have computed fields, your methods have to be declared before the fields.

agent-z28:
Jaho, as you can see in the discussion, this  :x: seems to be bcs of missing unit tests?
Here it doesnt make sense for a unit test.

mail_store_outgoing/models/ir_mail_server.py
yajo: You can put the name of the compute/inverse/search function as a string instead.
If using a computed domain/context, you should still use a lambda or the method cannot be inherited, so in any case it's better to stick to the standard.

mail_store_outgoing/models/ir_mail_server.py
```diff
@@ -0,0 +1,115 @@
...
+                folder.unlink()
+            for line in mBoxes:
+                flags, delimiter, mailbox_name = self.parse_list_response(line)
+                res = {'server_id': self.id, 'name': mailbox_name, }
+                imap_pool.create(res)
+            self.write({'imap_mailbox_verified': True})
+        except Exception, e:
+            raise ValidationError(
+                _("Connection Test Failed!"
+                    "Here is what we got instead:\n %s") % tools.ustr(e))
```
gurneyalex, line 60: add a space or you'll get "Failed!Here" in the message

mail_store_outgoing/models/ir_mail_server.py
```diff
@@ -0,0 +1,115 @@
...
+                pass
+
+    @api.model
+    def send_email(self, message, mail_server_id=None,
+                   smtp_server=None, smtp_port=None, smtp_user=None,
+                   smtp_password=None, smtp_encryption=None, smtp_debug=False):
+        super(IrMailServer, self).send_email(
+            message, mail_server_id, smtp_server, smtp_port,
+            smtp_user, smtp_password, smtp_encryption, smtp_debug)
+        return self._save_sent_message_to_sentbox(message, mail_server_id)
```
gurneyalex: this method can raise an exception which is bad : an email has been sent out and the transaction is going to be rollbacked. And potentially the same mail will be resent later when we retry.

gurneyalex (changes requested):

mail_store_outgoing/models/ir_mail_server.py
agenterp: Good Point! i will fix this!

mail_store_outgoing/models/ir_mail_server.py
agent-z28: thx, i did it

mail_store_outgoing/models/ir_mail_server.py
agent-z28: i changed the code

agent-z28:
TIME TO MERGE AMIGOS.
AFTER THIS INITIAL COMMIT MANY WILL FOLLOW!

agent-z28:
Slow!!

PR #207, opened 2017-10-20
[FIX] mail_optional_follower_notification: Check force_partners_to_notify instead of notify_followers.

adrienpeiffer:
Without this commit, followers are NEVER notified when a message is directly posted on the tracker (without click to open the mail compose message form).

My bad ...


yajo (approved):

pedrobaeza (approved):
Please change module version number

adrienpeiffer:
@pedrobaeza it's done :smile:

pedrobaeza (approved):

PR #209, opened 2017-11-06
[FIX] mail_debrand: name of overridden method

StefanRijnhart:
As per https://github.com/odoo/odoo/blob/8.0/addons/mail/mail_followers.py#L138.

@Yajo, this one is a bit of a mystery. I could not find any trace of this method name with the preceding underscore.

### DIFF
--- a/mail_debrand/__openerp__.py
+++ b/mail_debrand/__openerp__.py
@@ -4,7 +4,7 @@
 {
     "name": "Mail Debrand",
     "summary": "Remove Odoo branding from email footers",
-    "version": "8.0.1.0.0",
+    "version": "8.0.1.0.1",
     "category": "Social Network",
     "website": "http://www.antiun.com",
     "author": "Antiun Ingeniería S.L., Odoo Community Association (OCA)",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
There was a previous attempt at #176, but it misses the tests fix.

pedrobaeza:
Please change module version number

yajo (changes requested):
Indeed the code fix is quite obvious, but please make Travis happy 😊

StefanRijnhart:
@Yajo I fixed the lint error related to this change but the odoo/odoo build fails just like on the main branch (and I don't know how to fix timeouts in http tests anyway)

pedrobaeza:
Agh, same problem everywhere out of Odoo itself. There's a general problem with Travis + wkhtmltopdf, which reproduces the same way. Don't know if this one is the same.

yajo (approved):
Besides, [there's an error in another addon](https://travis-ci.org/OCA/social/jobs/298681131#L1476-L1513).

OK, not blocking then

hbrunn (approved):

PR #210, opened 2017-11-07
[ADD] mass_mailing_resend

pedrobaeza:
Resend mass mailings
====================

A frequent need for users of mass mailings is to resend one mailing that has already been sent in the past to new recipients that haven't received yet that mail. But the problem is to know which are the applicable ones.

Odoo already includes a method in its mass mailing logic that avoids to resend the same mail 2 times for one mass mailing, and for v9, there was a trick that allows to modify the state of a mass mailing from kanban view, covering the need.

But now on v10 both status bar in form view and dragging between states in kanban are not allowed.

This module introduces a button to restart a mass mailing to draft state, allowing you to reevaluate the sending domain or list for performing again the mailing.

Usage
=====

* Go to *Mass mailing > Mailings > Mass Mailings*.
* Click on one record that is done or create a new one and send it.
* You will see a button called "Resend".
* If you click on it, mass mailing will be set to Draft again.

Known issues / Roadmap
======================

* Add an indicator / filter for knowing resent mailings.
* Include information on the number of new recipients to be sent on the
  resending (through `get_remaining_recipients` method).

@Tecnativa


mass_mailing_resend/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
+# Copyright 2016 Tecnativa - Vicent Cubells
```
yajo: Drop copyright from `__init__` files

mass_mailing_resend/models/mass_mailing.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
+# Copyright 2016 Tecnativa - Vicent Cubells
```
yajo: Update copyright (it seems you copied it wrongly from elsewhere)

mass_mailing_resend/models/mass_mailing.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import _, exceptions, models
+
+
+class MailMassMailingList(models.Model):
+    _inherit = "mail.mass_mailing"
+
+    def button_draft(self):
+        """Return to draft state for resending the mass mailing."""
+        if any(self.mapped(lambda x: x.state != 'done')):
```
yajo, line 14: `if self.filtered(...):`, or use a normal `for` loop, which is more performant since it would stop at 1st match.

mass_mailing_resend/views/mass_mailing_views.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record model="ir.ui.view" id="view_mail_mass_mailing_form">
+        <field name="model">mail.mass_mailing</field>
+        <field name="inherit_id" ref="mass_mailing.view_mail_mass_mailing_form"/>
+        <field name="arch" type="xml">
+            <field name="state" position="before">
+                <button string="Resend"
```
yajo, line 9: Don't you think "resend" is a bit misleading? Because you are not actually "resending" it, just setting back as draft. 🤔

yajo (approved):
Minor improvements to code. Functional test OK.

mass_mailing_resend/views/mass_mailing_views.xml
chienandalu: A `help`would also be useful here noting that the mail will only be sent to new recipients.

chienandalu (commented):
It'd be nice also to have always access to the stats button box when the mailing has been sent at least one time. That's useful also to be able to delete some mails from the stats so it's possible to resend the mail to them.

image

mass_mailing_resend/models/mass_mailing.py
pedrobaeza: In this case, as the list is going to be very short, I prefer this formula to avoid one cyclomatic level more.

mass_mailing_resend/views/mass_mailing_views.xml
pedrobaeza: I prefer resend word that is more user friendly, because setting back to draft is for resending purpose, although I'm going to add a text when clicked explaining about the new recipients.

pedrobaeza:
Now there's a explanatory text and the buttons are back, so I think it's good enough

resend

chienandalu (approved):
Great!

mass_mailing_resend/views/mass_mailing_views.xml
```diff
@@ -12,6 +12,16 @@
                         type="object"
                 />
             </field>
+            <header position="after">
+                <div class="alert alert-warning oe_text_center" attrs="{'invisible': ['|', ('state', '!=', 'draft'), ('sent', '=', '0')]}">
```
chienandalu: Following the flow: shouldn't this be more like:
```python
{'invisible': ['|', ('state', '=', 'done'), ('state', '!=', 'done'), ('sent', '=', '0')]}
```
So it's only visible when a mailing is reset to draft and it has sent mails. Otherwise it's visible at the first composition.

pedrobaeza: No, it's correct as it's now: it's only visible when on draft and sent > 0.

chienandalu: Isn't it when draft **or** sent = 0

pedrobaeza: I don't get you. Please re-read my comment

chienandalu: Ok, yeah. I got it wrong because the second condition doesn't seem to be working fine. When you create a new draft mailing the bubble is visible (maybe is False until the field is computed by the first time?)

mass_mailing_resend/models/mass_mailing.py
yajo: In such case, use `if self.filtered(...)`

mass_mailing_resend/views/mass_mailing_views.xml
chienandalu: Ok, it was more subtle than that :smile: The quotes in `('sent', '=', '0')` are avoiding proper evaluation. So, this should work:
```python
{'invisible': ['|', ('state', '!=', 'draft'), ('sent', '=', 0)]}
```

pedrobaeza: Agh...

yajo:
@pedrobaeza Can I squash and merge?

pedrobaeza:
Yes, please

PR #211, opened 2017-11-10
[FIX+IMP] mass_mailing_list_dynamic: tests, icons, filters...

pedrobaeza:
* Brand new icon
* Added feature of loading an existing filter as criteria
* Tests as SavepointCase for optimizing times
* Tests in post-install for avoiding errors on res.partner not null constraints
  when several modules added them.
* Updated documentation.
* Fix mock in test for not commiting test data.

@Tecnativa


mass_mailing_list_dynamic/tests/test_dynamic_lists.py
```diff
@@ -82,9 +85,22 @@ def test_sync_when_sending_mail(self):
             "category_id": [(4, self.tag.id, False)],
             "email": "extra@example.com",
         })
-        # Before sending the mail, the list is updated
-        with patch("odoo.addons.base.ir.ir_mail_server"
-                   ".IrMailServer.send_email") as send_email:
+        # Mock sending low level method, because an auto-commit happens there
+        with patch("odoo.addons.mail.models.mail_mail.MailMail.send"):
             self.mail.send_mail()
-            self.assertEqual(6, send_email.call_count)
```
yajo: Please keep this check, it was added on purpose because there was a bug in my initial implementation that sent many mails per recipient.

mass_mailing_list_dynamic/models/mail_mass_mailing_contact.py
```diff
@@ -11,7 +11,7 @@ class MassMailingContact(models.Model):
 
     @api.constrains("partner_id", "list_id", "name", "email")
     def _check_no_manual_edits_on_fully_synced_lists(self):
-        if self.env.context.get("syncing"):
+        if self.env.context.get("syncing"):  # pragma: no cover
```
yajo: Don't do this, it's already being tested: https://github.com/OCA/social/pull/211/files#diff-3c3cb2de40288fdbce8b7e58014c6028L60

yajo (changes requested):
I added a fix to another bug we were hitting: when the sync process had to create contacts in a fully synchronized list, there was an exception:

captura de pantalla de 2017-11-13 09-56-53

Besides that, I see a bit confusing the new button, UX POV. Now there are 2 ways of defining a domain instead of just one, so I think the UI should be a little bit more self-explanatory on when to use one or the other. Also, the new button should be aligned to the criteria field, the same way the sync now button is aligned to the sync method.

To finish, there are a couple of minor things that this PR is un-improving, commented below. Please keep those as they were 😉 

Thanks! Good job 😊

pedrobaeza:
@cubells what are you doing with this PR?

mass_mailing_list_dynamic/tests/test_dynamic_lists.py
pedrobaeza: It doesn't make sense, as I have changed the mocked call (because yours was incorrect as was causing a commit to be performed). Any way, I let the assert with the count=1, that is the one needed.

pedrobaeza:
@Yajo ready to merge

@cubells @rafaelbn please review

mass_mailing_list_dynamic/i18n/es.po
```diff
@@ -20,37 +22,49 @@ msgstr ""
...
-"<i class=\"fa fa-info-circle\"/> You cannot make manual editions of contacts"
-" in fully synchronized lists."
-msgstr ""
+msgid "<i class=\"fa fa-info-circle\"/> You cannot make manual editions of contacts in fully synchronized lists."
+msgstr "<i class=\"fa fa-info-circle\"/> No puede editar manualmente los contactos en las listas que están completamente sincronizadas."
 
 #. module: mass_mailing_list_dynamic
 #: selection:mail.mass_mailing.list,sync_method:0
 msgid "Add and remove records as needed"
 msgstr ""
```
cubells, line 36: Are not translated, It isn't?

Spanish file is incomplete.

mass_mailing_list_dynamic/wizards/mail_mass_mailing_load_filter_views.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 Tecnativa - Pedro M. Baea
```
cubells, line 2: Your last name is incorrect.

cubells (changes requested):
Minor changes ready to approve.

cubells (changes requested):
README file style is broken because a blank line is needed before second indentation. But is an error from previous versions.

mass_mailing_list_dynamic/tests/test_dynamic_lists.py
```diff
@@ -73,7 +80,7 @@ def test_list_sync(self):
             contact1.partner_id = self.partners[0]
 
     def test_sync_when_sending_mail(self):
-        """Dynamic list is synced before mailing to it."""
+        """Check that list in synced when sending a mass mailing."""
```
yajo, line 83: "is"

yajo (approved):

rafaelbn (approved):
👍 using it in production

rafaelbn:
@cubells please review again and unblock 😄

rafaelbn:
Hi @pedrobaeza ,

Have you deleted the option of filtering? The only way now is onle loading filter?



Thanks

rafaelbn (commented):
❔

pedrobaeza:
Because web_advanced_search_x2x is not installed here.

rafaelbn:
@cubells please review, thanks

cubells (approved):
👍 

LGTM

PR #214, opened 2017-11-21
[10.0][ADD] bus_presence_override: Add module.

### LINKED ISSUE #186: [RFC] bus_presence_override: Chose availability states
Currently Odoo presence (IM availability) states are based purely on a hard coded [AWAY_TIMER](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/addons/bus/models/bus_presence.py#L11), which is set at 30 minutes of inactivity. `bus_presence_override` module will improve upon this logic by providing the ability for users to select their current availability state - as you would in a standard IM system.

It will accomplish this by adding a two buttons (red, yellow, and green) in the header of the [MessagingMenu](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/addons/mail/static/src/js/systray.js#L53):

image

Clicking the red will make you `offline`, the yellow `away`, the green `online`. 

The technical side of this looks fairly simple. We will add an `im_status_custom` to `res.partner` that will be used to overload the [default `_compute_im_status`](https://github.com/odoo/odoo/blob/44276dfdc3b69cd9b834bfa7037671558e6d0078/addons/bus/models/res_partner.py#L14). I think we'll need to break the inheritance chain here in order to be efficient & just rewrite the direct SQL, but I don't think that's too big of a deal in this instance. Alternatively, we could re-iterate the partners after calling super in order to append any manual changes - but I think that's going to do more harm than good.

Does anyone know of anything doing something similar to this? It seems weird to me that we can't control these states manually, so I honestly feel like I'm missing something.

woodbrettm:
Hi there,

Resolves #186 

This module adds the ability to manually select a user's online status while keeping the disconnect and away timers intact.

online_status

#### TODO
- [x] Write to users model also in JS (for im_status_custom)(Irrelevant)
- [x] Check if session.partner_id actually exists first in JS before querying. (perhaps just query and write to users table?) (Irrelevant)
- [x] Refactor JS
- [x] Refactor LESS
- [x] Write new python tests
- [x] systray should be renamed to bus_presence_selection (renamed to bus_presence_systray) 

Thanks!
Brett

@lasley


lasley:
> When changing the status manually in the UI, it will take up to 50 seconds for other people to see your status change in their UI. I think this is a limitation because longpolling happens every 50 or so seconds (which is when im_status is checked). Perhaps there is a way to force a poll after the status is manually changed?

This is just the limitation of poll refreshing. Maybe put it in the known issues, but not worth fussing over - especially for a v1

> In a few select cases, I need to automatically update the im_status in the image above but can't quite figure out how. For example, in the mail module, the status icons will update automatically when needed after every poll.

So basically you need to update the state in Object A (the polling JS) from Object B (the mail JS)?

In JavaScript, you do this through events. You set an event listener on an object for "change_status" and have it accept an argument (the status). In any code, you then trigger the event on the object in the DOM and viola. 

In Odoo, we don't have to attach the event to an element on the DOM, but instead can use the `core.bus` mechanism:
* https://odoo-development.readthedocs.io/en/latest/dev/js/core.bus.html
* https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=%22core.bus%22&type=

bus_presence_override/models/res_partner.py
```diff
@@ -0,0 +1,83 @@
...
+            (AWAY, 'Away'),
+            (OFFLINE, 'Offline'),
+        ],
+        string='Status',
+        default=ONLINE,
+    )
+
+    @api.model
+    def _get_partners_presence(self):
+        """ This method gets any relevant partners' im status.
```
lasley: How do you define the partner? Is it just all of them?

lasley: Looking further, this should be an `api.multi` method

bus_presence_override/models/res_partner.py
```diff
@@ -0,0 +1,83 @@
...
+                        WHEN age(now() AT TIME ZONE 'UTC', B.last_presence)
+                            > interval %s THEN 'away'
+                        ELSE 'online'
+                    END as status
+                FROM bus_presence B
+                    JOIN res_users U ON B.user_id = U.id
+                WHERE U.partner_id IN %s AND U.active = 't'
+            """,
+            ("%s seconds" % DISCONNECTION_TIMER,
+             "%s seconds" % AWAY_TIMER, tuple(self.ids))
```
lasley: Why type casting `self.ids` to a tuple? It already returns a list, which implements all the spec of a tuple AFAIK

bus_presence_override/models/res_partner.py
```diff
@@ -0,0 +1,83 @@
...
+        )
+        all_status = self.env.cr.dictfetchall()
```
lasley: It'd be best to just directly iterate the results from the cursor instead of processing just to reprocess.

```
return {row[0]: row[1] for row in self.env.cr}
```

You may need to `self.env.cr.fetchone()`, but I'm pretty sure you can just directly iterate the cursor per Python DBAPI spec

bus_presence_override/models/res_partner.py
```diff
@@ -0,0 +1,83 @@
...
+        all_status_dict = dict(
+            ((status['id'], status['status']) for status in all_status)
+        )
+        return all_status_dict
+
+    @api.multi
+    def _compute_im_status(self):
```
lasley: You're setting fields in this method that are not defined as computed. That won't work.

Also what should trigger a recompute of this? Shouldn't there be some sort of `api.depends`?

bus_presence_override/static/src/js/systray.js
```diff
@@ -0,0 +1,59 @@
+/* Copyright 2017 LasLabs Inc.
+   License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+
+odoo.define('bus_presence_override.systray', function (require) {
```
lasley: IMO you should use something more descriptive than `systray`. This is only for the editing of an IM status, and is not a system tray. `bus_presence_selection` or something

bus_presence_override/static/src/js/systray.js
```diff
@@ -0,0 +1,59 @@
...
+    "use strict";
+
+    var DataModel = require('web.DataModel');
+    var session = require('web.session');
+    var SystrayMenu = require('web.SystrayMenu');
+    var Widget = require('web.Widget');
+
+    var Systray = Widget.extend({
+        template:'systray',
```
lasley: space after `:`

bus_presence_override/static/src/js/systray.js
```diff
@@ -0,0 +1,59 @@
...
+                'online': 'fa fa-circle o_user_online',
+                'away': 'fa fa-circle o_user_idle',
+                'offline': 'fa fa-circle-o',
+            }
+        },
+        start: function () {
+            this._update_im_status_custom(status='online');
+            return this._super()
+        },
+        on_click_user_presence_status: function (event) {
```
lasley: use lowerCamelCase for anything you would typically snake_case in Python. Variable names, methods, etc.

bus_presence_override/static/src/js/systray.js
```diff
@@ -0,0 +1,59 @@
...
+
+    var Systray = Widget.extend({
+        template:'systray',
+        events: {
+            'click .o_user_presence_status': 'on_click_user_presence_status',
+        },
+        init: function() {
+            this._super.apply(this, arguments);
+            this.Partners = new DataModel('res.partner');
+            this.status_icons = {
```
lasley: statusIcons

bus_presence_override/static/src/js/systray.js
```diff
@@ -0,0 +1,59 @@
...
+    var Widget = require('web.Widget');
+
+    var Systray = Widget.extend({
+        template:'systray',
+        events: {
+            'click .o_user_presence_status': 'on_click_user_presence_status',
+        },
+        init: function() {
+            this._super.apply(this, arguments);
+            this.Partners = new DataModel('res.partner');
```
lasley: This naming should be more obvious. As it stands, this alludes to there being partners in this variable, not it being a data model. `this.partnerModel` or something

bus_presence_override/static/src/js/systray.js
```diff
@@ -0,0 +1,59 @@
...
+            this.status_icons = {
+                'online': 'fa fa-circle o_user_online',
+                'away': 'fa fa-circle o_user_idle',
+                'offline': 'fa fa-circle-o',
+            }
+        },
+        start: function () {
+            this._update_im_status_custom(status='online');
+            return this._super()
```
lasley: missing semicolon

bus_presence_override/static/src/js/systray.js
```diff
@@ -0,0 +1,59 @@
...
+        },
+        on_click_user_presence_status: function (event) {
+            var status = $(event.target).attr('name');
+            this._update_im_status_custom(status);
+        },
+        _get_im_status: function () {
```
lasley: `_` prefix is not a JS convention. Remove. `getImStatus`

bus_presence_override/static/src/js/systray.js
```diff
@@ -0,0 +1,59 @@
...
+            var self = this;
```
lasley: Dirty hack

bus_presence_override/static/src/js/systray.js
```diff
@@ -0,0 +1,59 @@
...
+            this.Partners.query(['im_status'])
+                         .filter([['id', '=', session.partner_id]])
+                         .first()
+                         .then(function (result) {
```
lasley: `.then(this.updateSystrayStatusIcon)`

And your `updateSystrayStatusIcon` method should directly accept the whole record and not just the IM status

bus_presence_override/static/src/js/systray.js
```diff
@@ -0,0 +1,59 @@
...
+        },
+        _update_systray_status_icon: function (status) {
+            $('#userStatus i').removeClass()
+                              .addClass('o_mail_user_status ' + this.status_icons[status]);
+        },
+        _update_im_status_custom: function (status) {
+            var self = this;
+            this.Partners.call('write', [[session.partner_id], {'im_status_custom': status}])
+                         .then(function () {
+                             self._update_systray_status_icon(status);
```
lasley: In this case, do not perform anything in `.then`. You instead should just directly run the updater after the write:

```javascript
var data = {'im_status_custom': status};
this.Partners.call('write', [[session.partner_id], data]);
this.updateSystrayStatusIcon(data);
```

bus_presence_override/static/src/js/systray.js
```diff
@@ -0,0 +1,59 @@
...
+            });
+        },
+
+    });
+
+    SystrayMenu.Items.push(Systray);
```
lasley: Further proof that `Systray` is bad naming. Why would you push a system tray into a menu, which already is inside of the system tray?

bus_presence_override/static/src/less/systray.less
```diff
@@ -0,0 +1,47 @@
+/* Copyright 2017 LasLabs Inc.
+   License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+
+#userStatus {
+    &:hover i {
+        color: #d3d3d3 !important;
+    }
+
+    i {
```
lasley: You should always put static definitions above conditional (:hover)

bus_presence_override/static/src/less/systray.less
```diff
@@ -0,0 +1,47 @@
+/* Copyright 2017 LasLabs Inc.
+   License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+
+#userStatus {
+    &:hover i {
+        color: #d3d3d3 !important;
```
lasley: `!important` is a dirty hack that should never be used, except without a damned good explanation. I assume there isn't one here though.

bus_presence_override/static/src/less/systray.less
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    i {
+        padding-top: 2px;
+    }
+}
+
+.o_user_presence_dropdown {
```
lasley: CSS classes should be hyphen separated instead of snake/camel case. `.o-user-presence-dropdown`

bus_presence_override/static/src/less/systray.less
```diff
@@ -0,0 +1,47 @@
...
+
+    .o_user_presence_status {
+        margin-top: 2px;
+        padding: 4px 8px;
+        width: 100%;
+        display: block;
+        font-size: 13px;
+    }
+
+    &:hover {
```
lasley: This looks like you're trying to imitate a link element?

bus_presence_override/static/src/less/systray.less
```diff
@@ -0,0 +1,47 @@
...
+        cursor: pointer;
+    }
+
+}
+
+@media (max-width: @screen-xs-max) {
```
lasley: Fix the tabs

bus_presence_override/static/src/less/systray.less
```diff
@@ -0,0 +1,47 @@
...
+
+        .o_user_presence_status {
+            color: #9d9d9d;
```
lasley: There's a lot of hardcoded colors here, which is bad. Any reason why you didn't use the pre-existing color variables? https://github.com/odoo/odoo/blob/10.0/addons/web/static/lib/bootstrap/less/variables.less#L6

lasley (changes requested):
Thanks @bmw95

bus_presence_override/static/src/js/systray.js
lasley: Oh you can use `$.proxy` as I previously advised, which would allow you to directly pass the status

woodbrettm:
Woot all passing now. @lasley thanks for the review. Updated. The build was failing from another module so i had to rebase unfortunately (i.e. pr history broken)

woodbrettm:
Odd. It looks like the status icon isn't showing in runbot. Will have to look into that. Code still ready for review however.

lasley:
@bmw95 What do you mean? It's right here:

image

woodbrettm:
@lasley no sorry I meant the user status icon in the systray in runbot. Should be an icon there between the admin avatar and the message icon. Works locally, so might be a module conflict or something in runbot.
cools

lasley:
I bet longpolling is screwed. Check your dev tools console for output

lasley:
(If that's the case, it's a Runbot issue)

woodbrettm:
Yeah I have a feeling it's a runbot issue. I tried it on the 10.0 branch, and direct messages in discuss won't even load when you post them, until you reload the page.

lasley (approved):
Changes look good, and module works in local. Runbot issue is Runbot. Thanks @bmw95 

Regarding the JS tests - don't bother. I think it's fine to hit those in the v11 migration, because otherwise they're going to just need to be rewritten.

woodbrettm:
Woot!

gurneyalex (approved):

PR #215, opened 2017-11-24
[imp] mail_digest: fallback to default website

simahawk:
Depending on the request  might fail website retrieval might fail
nevertheless the website should be there to be found.
Fallback to search and use it if possible for digest' site name

### DIFF
--- a/mail_digest/__openerp__.py
+++ b/mail_digest/__openerp__.py
@@ -5,7 +5,7 @@
 {
     'name': 'Mail digest',
     'summary': """Basic digest mail handling.""",
-    'version': '9.0.1.0.0',
+    'version': '9.0.1.0.1',
     'license': 'AGPL-3',
     'author': 'Camptocamp,Odoo Community Association (OCA)',
     'depends': [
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Do you have steps to reproduce the error?

simahawk:
@pedrobaeza there's no error indeed as I'm already catching the runtime one prior to this PR ;)
What you would see w/out that try/except is:
```bash
File "/usr/local/lib/python2.7/dist-packages/werkzeug/local.py", line 132, in _lookup
raise RuntimeError('object unbound')
RuntimeError: object unbound
```
This is just an improvement on top of that and retrieve the website if any.

Probably the most important thing here is the TODO comment: I'm not sure how this can be improved at config level.

The issue is: even if you have website you might get  internal notifications enabled on digest mode.
In this case you gonna get those notifications w/ a digest subject that will contain the site name instead of the company name.

Not a big deal but something we could address in the future possibly.

pedrobaeza:
OK, thanks for the explanation. You know now my last request: change module version number :stuck_out_tongue_closed_eyes:

simahawk:
hehe, was waiting for that. Done ;)

pedrobaeza (approved):

yajo (approved):

PR #217, opened 2017-12-04
[10.0][IMP] Hide the checkbox for the mass mailing in mail_optional_followe…

RobinetDenisAcsone:
…r_notification and mail_optional_autofollow

On mass mailing composition mode, those option doesn't do anything.

### DIFF
--- a/mail_optional_autofollow/wizard/mail_compose_message_view.xml
+++ b/mail_optional_autofollow/wizard/mail_compose_message_view.xml
@@ -6,7 +6,7 @@
         <field name="inherit_id" ref="mail.email_compose_message_wizard_form"/>
         <field name="arch" type="xml">
             <xpath expr="//div[field[@name='partner_ids']]" position="after">
-                <field name="autofollow_recipients" />
+                <field name="autofollow_recipients" attrs="{'invisible': [('composition_mode', '=', 'mass_mail')]}"/>
             </xpath>
         </field>
     </record>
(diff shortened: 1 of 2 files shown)


yajo:
What do you mean with that those fields do nothing? Maybe a couple of screenshots could help.

Also check Travis' ❌  please

RobinetDenisAcsone:
First, if you want to test easily the difference between mass mail and comment composition mode, you can go under setting > technical > email > template and search for the 'Event: Registration Badge' and click on the 'add context action' smart button if it's not already set.
This will allow to test the comment composition mode with the same template thought the send 'by mail button' and the mass mail thought the action > send by mail (Event: Registration Badge) on a attendee
- mail_optional_follower_notification module
on mass mailing composition mode, it's not the follower who a used for the recipients.
The mail will have the same behaviour when the checkbox is checked or not
image
If we send on mass mailing with the checkbox checked, here is the result:
image
The followers aren't notified

- mail_optional_autofollow
same scenario but with the 'Make recipients followers' checked:
image
The recipient (delta PC) isn't add to the followers list

For the Travis error, I don't know right now the problem, it's on the mass_mailing_custom_unsubscribe module, which I didn't modify, I'll come back as soon as I find the problem.

yajo (approved):

RobinetDenisAcsone:
For the Travis error, it come from the mass_mailing_custom_unsubscribe module, which didn't depends from the 2 modules changed in this pr.
It's cause by the test of the tour https://github.com/OCA/social/blob/10.0/mass_mailing_custom_unsubscribe/static/src/js/contact.tour.js#L18.
I don't know what it the cause of this error, if anyone have a idea, I'll be happy to ear it.

yajo:
It must have been a temporary timeout, I restarted the job and it worked.

RobinetDenisAcsone:
ok, thanks for the restart.

ecino (approved):
LGTM

PR #218, opened 2017-12-08
FIX mail_footer_notified_partner

eLBati:
we need to save the complete list of partners because
 _message_notification_recipients builds recipients
 grouped by users groups. Thus get_additional_footer would get a
 partial list of recipients

### DIFF
--- a/mail_footer_notified_partner/models/mail_followers.py
+++ b/mail_footer_notified_partner/models/mail_followers.py
@@ -9,9 +9,28 @@ from odoo.tools.translate import _
 class MailNotification(models.Model):
     _inherit = 'res.partner'
 
+    @api.multi
+    def _notify_by_email(
+        self, message, force_send=False, send_after_commit=True,
+        user_signature=True
+    ):
+        # we need to save the complete list of partners because
+        # _message_notification_recipients builds recipients
+        # grouped by users groups. Thus get_additional_footer would get a
+        # partial list of recipients
+        return super(
+            MailNotification, self.with_context(notified_partners=self)
+        )._notify_by_email(
+            message, force_send=force_send,
+            send_after_commit=send_after_commit,
+            user_signature=user_signature
+        )
+
     @api.model
     def _notify_send(self, body, subject, recipients, **mail_values):
-        body += self.get_additional_footer(recipients)
+        footer_recipients = self.env.context.get(
+            'notified_partners', recipients) or recipients
+        body += self.get_additional_footer(footer_recipients)
         return super(MailNotification, self).\
             _notify_send(body, subject, recipients, **mail_values)
 
--- a/mail_footer_notified_partner/__manifest__.py
+++ b/mail_footer_notified_partner/__manifest__.py
@@ -11,7 +11,7 @@
     "installable": True,
     "website": "http://acsone.eu",
     "category": "Mail",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "license": "AGPL-3",
     "depends": [
         "mail",


SimoRubi (approved):
Looks good

hurrinico:
LGTM :+1:

eLBati:
@OCA/social-maintainers please merge

pedrobaeza:
I would like to have ACSone approval. cc @lmignon

mail_footer_notified_partner/models/mail_followers.py
```diff
@@ -9,9 +9,29 @@
...
+    ):
+        # we need to save the complete list of partners because
+        # _message_notification_recipients builds recipients
+        # grouped by users groups. Thus get_additional_footer would get a
+        # partial list of recipients
+        ctx = dict(self.env.context)
+        ctx.update({"notified_partners": self})
+        return super(
+            MailNotification, self.with_context(ctx))._notify_by_email(
+                message, force_send, send_after_commit, user_signature)
```
hbrunn: pass keyword arguments as keyword arguments

mail_footer_notified_partner/models/mail_followers.py
```diff
@@ -9,9 +9,29 @@
...
+        user_signature=True
+    ):
+        # we need to save the complete list of partners because
+        # _message_notification_recipients builds recipients
+        # grouped by users groups. Thus get_additional_footer would get a
+        # partial list of recipients
+        ctx = dict(self.env.context)
+        ctx.update({"notified_partners": self})
+        return super(
+            MailNotification, self.with_context(ctx))._notify_by_email(
```
hbrunn: how about 

```python
return super(
    MailNotification, self.with_context(notified_partners=self)
)._notify_by_email(
    ...
)
```

mail_footer_notified_partner/models/mail_followers.py
```diff
@@ -9,9 +9,29 @@
...
+                message, force_send, send_after_commit, user_signature)
+
     @api.model
     def _notify_send(self, body, subject, recipients, **mail_values):
-        body += self.get_additional_footer(recipients)
+        notified_partners = self.env.context.get('notified_partners')
```
hbrunn: `.get('notified_partners', recipients)`

hbrunn (changes requested):
people really need to start using the github review mechanism, those thumbs up in the text slip through all filters

hbrunn:
@pedrobaeza any reason for needing this specific approval?

pedrobaeza:
Because they were the original authors of the module and known better the code, but your review is also the same valuable. My comment was put after the merge request from @eLBati, expressing my uncomfortability for merging this directly.

mail_footer_notified_partner/models/mail_followers.py
eLBati: @hbrunn this would produce a context with ```notified_partners``` only, right? In this case we need to keep the other items of original self.env.context

mail_footer_notified_partner/models/mail_followers.py
eLBati: @hbrunn actually ```notified_partners``` key could be present but empty. In that case, ```recipients``` should be used

mail_footer_notified_partner/models/mail_followers.py
hbrunn: https://github.com/OCA/OCB/blob/10.0/odoo/models.py#L4867

mail_footer_notified_partner/models/mail_followers.py
hbrunn: so + ` or recipients`

mail_footer_notified_partner/models/mail_followers.py
eLBati: @hbrunn thanks! fixed

mail_footer_notified_partner/models/mail_followers.py
eLBati: Fixed

mail_footer_notified_partner/models/mail_followers.py
eLBati: Fixed

gurneyalex (approved):

gurneyalex:
@hbrunn  update ok with you?

hbrunn (approved):

hbrunn:
in case the re-review didn't generate a notification: yes

PR #219, opened 2017-12-11
[8.0][WIP][ADD]mail_edit

Kiplangatdan:


.gitignore
```diff
@@ -55,3 +55,4 @@ docs/_build/
 
 # PyBuilder
 target/
+/.idea
```
thomaspaulb: Nicer would be to add this under a separate paragraph with `# PyCharm`. Also use `pylint-odoo` to fix the smallsmall OCA-style stuff.

mail_edit/__openerp__.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+
+{
+    "name": "Message Edit",
+    "summary": "Adds an option to edit mail to different partners",
+    "version": "8.0.1.0.0",
+    "category": "Social Network",
+    "website": "https://sunflowerweb.nl",
+    "author": "Sunflower IT"
+              "Odoo Community Association (OCA)",
```
thomaspaulb: We need a comma between our name and their name, current it will be written "Sunflower ITOdoo...".

mail_edit/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
+
+from . import compose_message, res_request_link
```
thomaspaulb: OCA style guide I think says to place alphabetically and each on their own line.

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,84 @@
+# -*- coding: utf-8 -*-
```
thomaspaulb, line 1: Copyright headers here

thomaspaulb (changes requested):
@Kiplangatdan Have tested this. When I click the "mail edit" button on a message that is on a certain partner, I get the "Compose Mail" wizard where I can select another partner, and also change the subject and the content of the mail. Then when I select another partner and press "Send", the mail is brought to the new partner but the contents have changed and the mail is also still in the old partner. Am I doing something wrong - could it be that I also have the "mail_forward" module installed and the function names are conflicting eg. the one from "forward" gets called instead of your new function?

thomaspaulb (approved):

mail_edit/README.rst
```diff
@@ -0,0 +1,58 @@
...
+
+This module extends the functionality of mail. You can edit the Partner ids of
+a mail and move it to any model.
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Go to *Message* *Chatter* in any model, click the Edit (plus) Icon.
```
hbrunn: any chance to make the plus a pencil? (that's Odoo's standard edit icon)

mail_edit/README.rst
```diff
@@ -0,0 +1,58 @@
...
+8.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+Credits
+=======
+
+Images
+------
+
+* Icon Finder `here <https://iconfinder.com>`_.
```
hbrunn: you need a link to the exact file. But from first glance this page looks like incompatible licenses, consider searching for a lookalike on http://picol.org, there we know it's creative commons (if the author didn't put an exception)

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,91 @@
...
+
+class MailEditComposeMessage(models.TransientModel):
+    _name = "mail_edit.compose_message"
+    _inherits = {"mail.compose.message": "original_wizard_id"}
+
+    @api.model
+    def default_get(self, fields):
+        result = self.original_wizard_id.default_get(fields)
+
+        if "subject" in result and "default_subject" in self.env.context:
```
hbrunn: i have the impression you mean `not in result` here, right? Or do you explicitly want to overwrite the key only if it's there? Or should `result` be `fields` and you want to check if the key is requested and there's a default? In that case, you don't need to do anything, because that's exactly how the mechanics of default_get work: https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L1310

hbrunn: oh, and if super (why do you call this so weirdly btw?) does something funny that you need to undo here, this merits a comment I think

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,91 @@
...
+            result["subject"] = self.env.context["default_subject"]
+
+        result["message_id"] = self.env.context["default_message_id"]
+        result["move_mail"] = True
```
hbrunn: the only reason you have to set those two manually is because you break inheritance by not calling super. any good reason for that?

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,91 @@
...
+    @api.onchange("destination_object_id")
+    def change_destination_object(self):
+        """Update some fields for the new message."""
+        if self.destination_object_id:
+            self.model = self.destination_object_id._name
+            self.res_id = self.destination_object_id.id
+
+            model_name = (self.env["ir.model"]
+                          .search([("model", "=", self.model)])
+                          .name)
```
hbrunn: this is not review but personal taste: I guess you're not happy with this formatting either, What I'd have done in this case is

```python
            model_name = self.env["ir.model"].search([
                ("model", "=", self.model)
            ]).name
```

to my mind, that tells me faster what this is about, but as said, no review, friendly advise

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,91 @@
...
+            record_name = self.destination_object_id.name_get()[0][1]
```
hbrunn: use `display_name`

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,91 @@
...
+        """Send mail and execute the attachment relocation if needed."""
+        # Let the original wizard do de hard work
+        result = self.original_wizard_id.send_mail()
+
+        # Relocate attachments if needed
+        if (self.move_attachments and
+                self.model and
+                self.res_id and
+                self.attachment_ids):
+            for attachment in self.attachment_ids:
```
hbrunn: don't do assignments in loops, that very inefficient. Just do a `.write`

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,91 @@
...
+        "Destination object",
+        help="Object where the forwarded message will be attached")
+
+    move_attachments = fields.Boolean(
+        "Move attachments",
+        help="Attachments will be assigned to the chosen destination "
+             "object and you will be able to pick them from its "
+             "'Attachments' button, but they will not be there for "
+             "the current object if any. In any case you can always "
+             "open it from the message itself.")
```
hbrunn: them?

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,91 @@
...
+
+    original_wizard_id = fields.Many2one(
+        "mail.compose.message",
+        "Original message compose wizard",
+        delegate=True,
```
hbrunn: doesn't the `_inherits` above already take care of this? that's an honest question, I learned just now this attribute exists, but my understanding is this will be true here by default: https://github.com/OCA/OCB/blob/8.0/openerp/fields.py#L1565

mail_edit/static/src/js/mail_edit.js
```diff
@@ -0,0 +1,73 @@
...
+            // Generate email subject as possible from record_name and subject
+            var subject = [];
+            if (this.record_name && (this.show_record_name ||
+                                     this.parent_id))
+            {
+                subject.push(this.record_name);
+            }
+            if (this.subject) {
+                subject.push(this.subject);
+            } else if (subject.length < 2) {
```
hbrunn: how can this be bigger than one at this point?

mail_edit/static/src/js/mail_edit.js
```diff
@@ -0,0 +1,73 @@
...
+                subject.push(_t("(No subject)"));
+            }
+
+            // Get only ID from the attachments
+            var attachment_ids = [];
+            for (var n in this.attachment_ids) {
```
hbrunn: you're looking for http://underscorejs.org/#map

most of the code here can be a bit more elegant by using underscore/jquery stuff, but it takes some time to get a knack for this kind of programming. It's an odoo standard library, so I'd suggest to keep most of it (this one really should change, because the `in` operator in javascript doesn't always give you what you expect) but look into it for the future

mail_edit/static/src/js/mail_edit.js
```diff
@@ -0,0 +1,73 @@
...
+                default_body: this.body,
+                default_model: this.model,
+                default_res_id: this.res_id,
+                default_subject: subject.join(": "),
+                default_message_id: this.id,
+            };
+
+            if (this.model && this.res_id) {
+                context.default_destination_object_id =
+                    [this.model, this.res_id].join();
```
hbrunn: I had to look up the ecma definition, but it's indeed a hard rule that it's a comma then: https://www.ecma-international.org/ecma-262/5.1/#sec-15.4.4.5 wanted to nag about this, but you're right. And given I looked up the definition, I wanted to share the link

hbrunn: TIL

mail_edit/static/src/js/mail_edit.js
```diff
@@ -0,0 +1,73 @@
...
+            }
+
+            // Get the action data and execute it to open the composer wizard
+            var do_action = this.do_action;
```
hbrunn: the canonical (and that's only odoo canonical, by no way js canonical) way is to assign `this` to `self` and then call stuff on `self` in handlers, but that's totally arbitrary and what you do is perfectly fine

hbrunn (changes requested):
a lot of nitpicks, nothing big. CI error is unrelated it seems

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,91 @@
...
+from openerp import api, fields, models
+
+
+class MailEditComposeMessage(models.TransientModel):
+    _name = "mail_edit.compose_message"
+    _inherits = {"mail.compose.message": "original_wizard_id"}
+
+    @api.model
+    def default_get(self, fields):
+        result = self.original_wizard_id.default_get(fields)
```
hbrunn: edit: I think you should call super here, you inherit the defaults. Given that the rest of the code more or less works around not calling super, could you try just getting rid of the function entirely?

thomaspaulb:
@hbrunn @Kiplangatdan I think many of the nitpicks were already in the `mail_forward` module (for example [here](https://github.com/OCA/social/blob/8.0/mail_forward/models/compose_message.py#L28)), because the main part of the code was just copied from there. So any improvements made here can also be proposed there.

That said, there is another issue with this code, because the current code uses the "Compose new mail" wizard to show the copied message to the user and then calls [send_mail](https://github.com/sunflowerit/social/blob/8.0-mail_edit/mail_edit/models/compose_message.py#L54) to create the copy and send it to the user. However, sending is more than just editing, and whereas for employees it is less of a problem, if we relocate the message to a "res.partner" who is a customer, who then gets an email, that can be confusing to the customer or even harmful to the relationship with the customer.

We seem to have the following options:

1. Redo the module to show an "Edit" form of the `mail.mail` object instead of the wizard
2. Keep it as it is now, but use the [mail_notify_noemail context key](https://github.com/OCA/OCB/blob/8.0/addons/mail/mail_followers.py#L268) to skip the sending of email. In this case we could also remove [this guy](https://github.com/sunflowerit/social/blob/8.0-mail_edit/mail_edit/views/compose_message.xml#L16), because notifications won't be sent anyways. Or we could have notifications as an optional checkbox that is turned off by default.

Opinions..?

Kiplangatdan:
The second option looks better to me considering the fact that we don't want to notify the customer every time we move a message from one model to another.

thomaspaulb:
@Kiplangatdan yes but that is also the case for option 1.

My personal feeling is that we will be better off just going for option 1. It may seem like more work now, but it feels more logical to me than trying to abuse a mail form to edit an object, and modifying the `send_mail` function to not send mail.

thomaspaulb:
@hbrunn We have changed to using an edit form instead of the 'Compose message' form, and are now ready for review.

mail_edit/README.rst
```diff
@@ -0,0 +1,71 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+============
+Message Edit
+============
+
+This module extends the functionality of mail. You can edit message and move it
```
hbrunn: a message/massages ... them

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,42 @@
...
+                display_name = "%s %s" % (model_name, display_name)
+
+            self.record_name = display_name
+        else:
+            self.model = self.res_id = self.record_name = False
+
+    destination_object_id = fields.Reference(
+        _get_model_selection,
+        "Destination object",
+        required=True,
```
hbrunn: don't make this required on the database level, if ever the constraint will be set up (it won't usually, because there are mails already in the system) things go bad. Set the flag on the view

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,42 @@
...
+    def change_destination_object(self):
+        """Update some fields for the new message."""
+        # pylint: disable=api-one-deprecated
+        if self.destination_object_id:
+            self.model = self.destination_object_id._name
+            self.res_id = self.destination_object_id.id
+
+            model_name = self.env["ir.model"].search([
+                ("model", "=", self.model)]).name
+            display_name = self.destination_object_id.name_get()[0][1]
```
hbrunn: no need to change, but for every model, `model.display_name` does the same, or something fancy in an override which you probably want anyways

mail_edit/views/compose_message.xml
```diff
@@ -0,0 +1,42 @@
...
+                <group>
+                    <group>
+                        <field name="subject" groups="mail_edit.group_mail_edit_editor"/>
+                        <field name="date" readonly="1"/>
+                        <field name="type" readonly="1"/>
+                        <field name="subtype_id" readonly="1"/>
+                    </group>
+                    <group>
+                        <field name="destination_object_id" groups="mail_edit.group_mail_edit_mover"/>
+                        <field name="model" invisible="1" groups="base.group_no_one"/>
```
hbrunn: if it's invisible, no need to restrict groups

mail_edit/static/src/js/mail_edit.js
```diff
@@ -0,0 +1,47 @@
...
+            // this.model has the model name or id
+            // this.res_id has the record id
+
+            // save the widget object in a var.
+            var self = this;
+
+            // Get the action data and execute it to open the composer wizard
+            var do_action = this.do_action;
+
+            this.rpc("/web/action/load", {
```
hbrunn: stick to the habit of always returning deferred object you create, then overrides can so something with it

mail_edit/static/src/js/mail_edit.js
```diff
@@ -0,0 +1,47 @@
...
+                "action_id": "mail_edit.mail_edit_action",
+            })
+            .done(function(action) {
+                action.res_id = self.id;
+                action.flags = {
+                    action_buttons: true,
+                };
+                action.context = context;
+                do_action(action, {
```
hbrunn, line 31: here too

hbrunn (changes requested):
travis error is unrelated

hbrunn (approved):
travis error is unrelated

hbrunn:
@Kiplangatdan please rebase for merging

hbrunn:
please make this one commit for the module, and one commit for the addition to .gitignore

thomaspaulb:
@daramousk Your review here?

mail_edit/static/src/js/mail_edit.js
```diff
@@ -0,0 +1,50 @@
...
+                    on_close: function () {
+                        // reload view
+                        var parent = self.getParent().getParent().getParent().getParent()
```
gfcapalbo: you can also write var parent =  self.parents.eq(3)  (zero base index)

gfcapalbo (approved):

PR #221, opened 2017-12-22
[9.0][FIX] mail_tracking_mailgun: Avoid key dict errors

chienandalu:
- Avoids error if dictionary keys doesn't exist.

cc @Tecnativa


pedrobaeza:
Superseeded by #226

PR #222, opened 2017-12-22
[10.0][FIX] mail_tracking_mailgun: Avoid key dict errors

chienandalu:
- Avoids error if dictionary keys doesn't exist.

cc @Tecnativa

### DIFF
--- a/mail_tracking_mailgun/models/mail_tracking_email.py
+++ b/mail_tracking_mailgun/models/mail_tracking_email.py
@@ -9,7 +9,7 @@ import json
 import requests
 from datetime import datetime
 from odoo import _, api, fields, models
-from odoo.exceptions import ValidationError
+from odoo.exceptions import UserError, ValidationError
 
 import logging
 _logger = logging.getLogger(__name__)
@@ -221,6 +221,8 @@ class MailTrackingEmail(models.Model):
         """
         api_key, api_url, domain, validation_key = self._mailgun_values()
         for tracking in self:
+            if not tracking.mail_message_id:
+                raise UserError(_('There is no tracked message!'))
             message_id = tracking.mail_message_id.message_id.replace(
                 "<", "").replace(">", "")
             res = requests.get(
@@ -242,7 +244,7 @@ class MailTrackingEmail(models.Model):
                 if not self.env['mail.tracking.event'].search(
                         [('mailgun_id', '=', item["id"])]):
                     mapped_event_type = self._mailgun_event_type_mapping.get(
-                        item["event"]) or False
+                        item["event"], item["event"])
                     metadata = self._mailgun_metadata(
                         mapped_event_type, item, {})
                     tracking.event_create(mapped_event_type, metadata)
--- a/mail_tracking_mailgun/models/res_partner.py
+++ b/mail_tracking_mailgun/models/res_partner.py
@@ -30,7 +30,8 @@ class ResPartner(models.Model):
             body = _('Email has been bounced: %s\n'
                      'Reason: %s\n'
                      'Event: %s') % (partner.email, reason,
-                                     event.get('Message-Id', _('unknown')))
+                                     event and event.get('Message-Id',
+                                                         _('unknown')))
             partner.message_post(body=body)
 
     @api.multi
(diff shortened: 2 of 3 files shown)


mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -240,9 +240,9 @@ def action_manual_check_mailgun(self):
                 raise ValidationError(_("Event information not longer stored"))
             for item in content["items"]:
                 if not self.env['mail.tracking.event'].search(
-                        [('mailgun_id', '=', item["id"])]):
+                        [('mailgun_id', '=', item.get("id"))]):
```
pedrobaeza: You can't enter here if there's no ID, but I don't know why this can happen

mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -240,9 +240,9 @@ def action_manual_check_mailgun(self):
...
                     mapped_event_type = self._mailgun_event_type_mapping.get(
-                        item["event"]) or False
+                        item.get("event", False))
```
pedrobaeza: Which type gets in this case?

mail_tracking_mailgun/models/res_partner.py
```diff
@@ -30,7 +30,7 @@ def _email_bounced_set(self, reason, event):
             body = _('Email has been bounced: %s\n'
                      'Reason: %s\n'
                      'Event: %s') % (partner.email, reason,
-                                     event['Message-Id'] or '')
+                                     event.get('Message-Id', ''))
```
pedrobaeza: Put `_('unknown')` as fallback

pedrobaeza (changes requested):
Change module verison

mail_tracking_mailgun/models/mail_tracking_email.py
chienandalu: Well, if there's an ok response there should allways be an id.

mail_tracking_mailgun/models/mail_tracking_email.py
chienandalu: Right, there was an error... :confused:

mail_tracking_mailgun/models/res_partner.py
chienandalu: Done

mail_tracking_mailgun/models/mail_tracking_email.py
pedrobaeza: But again, if there's no ID, you shouldn't perform any search IMO.

mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -240,9 +242,9 @@ def action_manual_check_mailgun(self):
                 raise ValidationError(_("Event information not longer stored"))
             for item in content["items"]:
                 if not self.env['mail.tracking.event'].search(
-                        [('mailgun_id', '=', item["id"])]):
+                        [('mailgun_id', '=', item.get("id"))]):
                     mapped_event_type = self._mailgun_event_type_mapping.get(
-                        item["event"]) or False
+                        item.get("event"), False)
```
pedrobaeza: If no event, should you launch `event_create` any way?

mail_tracking_mailgun/models/res_partner.py
```diff
@@ -30,7 +30,7 @@ def _email_bounced_set(self, reason, event):
             body = _('Email has been bounced: %s\n'
                      'Reason: %s\n'
                      'Event: %s') % (partner.email, reason,
-                                     event['Message-Id'] or '')
+                                     event.get('Message-Id', _('unknown')))
```
pedrobaeza: Include translation for this term

pedrobaeza (changes requested):

yajo:
This should be ready now. I took the freedom to amend @chienandalu's commit as per review.

rafaelbn (approved):
👍 

Please @Yajo @pedrobaeza review

pedrobaeza (approved):

pedrobaeza:
@Yajo please apply the same to v9

PR #223, opened 2017-12-28
ADD mail_notify_bounce

eLBati:
Notify bounce emails
====================

In case of bounced email messages retrieved by fetchmail, sender of original message will be notified. Also, it is possible to configure email addresses to be notified about bounced emails.

Typical use case:

 - User configures a partner with a wrong email address
 - User sends a message to that partner
 - User gets notified about wrong email address

In some cases (see https://github.com/odoo/odoo/issues/21119 ), these messages are silently discarded by odoo, so users don't know about possible failures

Configuration
----------------

To configure additional recipients of bounced emails, you can open fetchmail server form and fill 'Notify bounce emails to' field


tafaRU (approved):

mail_notify_bounce/README.rst
```diff
@@ -0,0 +1,67 @@
...
+
+In case of bounced email messages retrieved by fetchmail, sender of original message will be notified. Also, it is possible to configure email addresses to be notified about bounced emails.
+
+Typical use case:
+
+ - User configures a partner with a wrong email address
+ - User sends a message to that partner
+ - User gets notified about wrong email address
+
+In some cases (see https://github.com/odoo/odoo/issues/21119 ), these messages are silently discarded by odoo, so users don't know about possible failures.
```
hbrunn: make the link a link

mail_notify_bounce/models/mail_thread.py
```diff
@@ -0,0 +1,66 @@
...
+
+class MailThread(models.AbstractModel):
+    _inherit = 'mail.thread'
+
+    @api.model
+    def message_route(
+        self, message, message_dict, model=None, thread_id=None,
+        custom_values=None
+    ):
+        if self.env.context.get('fetchmail_server_id'):
```
hbrunn, line 79: move all of this to its own function, this allows other modules to intercept this if they need to

mail_notify_bounce/models/mail_thread.py
```diff
@@ -0,0 +1,66 @@
...
+            fetchmail = self.env['fetchmail.server'].browse(
+                self.env.context['fetchmail_server_id'])
+            if fetchmail.bounce_notify_partner_ids:
```
hbrunn: in the function, you can also short circuit here and save another level of indentation

mail_notify_bounce/models/mail_thread.py
```diff
@@ -0,0 +1,66 @@
...
+
+                email_from_localpart = (
+                    (
+                        tools.email_split(email_from) or ['']
+                    )[0].split('@', 1)[0].lower()
+                )
+                if (
+                    message.get_content_type() == 'multipart/report' or
+                    email_from_localpart == 'mailer-daemon'
+                ):
```
hbrunn: detecting bounces is hard. I'd suggest to make this configurable somehow, maybe config parameters with regexes

hbrunn (changes requested):
I'm not sure if you have the correct overall approach here. The canonical way to deal with bounces is to inject a return-path header into outgoing emails, preferably a unique address per mail (bounce+id would suffice). Then the other mailserver will send bounces there and you're sure it's a bounce. And given you encoded the mail id in the return address, you can then act accordingly.

Do you see any possibility to offer both? You approach would be simpler for less technical users, but it won't scale well for big organisations I think.

eLBati:
> The canonical way to deal with bounces is to inject a return-path header into outgoing emails, preferably a unique address per mail (bounce+id would suffice)

@hbrunn This is what odoo core does. That's OK.
In same cases, you need this module:
 - If the bounce mailbox is fetched by odoo, odoo will just set the message as bounced, without notifying the sender, or other people (see https://github.com/odoo/odoo/blob/0783318f9745c4e213016b825e9bbd8e1e9bd546/addons/mail/models/mail_thread.py#L1010)
 - If the bounce mailbox is fetched by odoo and the SMTP server does not allow to set the Return-Path (like Google, see https://github.com/odoo/odoo/issues/21119#issuecomment-348522350) the bounce message will just be discarded, without notifying the sender, or other people

hbrunn:
intersting, thank for the explanations

mail_notify_bounce/models/mail_thread.py
eLBati: @hbrunn this is the same criterion used by odoo: https://github.com/odoo/odoo/blob/0783318f9745c4e213016b825e9bbd8e1e9bd546/addons/mail/models/mail_thread.py#L1053
Comment added

eLBati:
@hbrunn thanks, I made the changes.
Let me know what you think

hbrunn (approved):
the refactoring I propose is nice to have I think, but whoever needs it should PR it then. Fine with me

eLBati:
@hbrunn tests added

hbrunn:
yay! If you now feed your message as it is (=without parsing it) to `message_process` instead of parsing it yourself and calling your function, your `message_route` override will be tested too, and you make the coverage target

eLBati:
@hbrunn thanks, done

hbrunn:
thanks for the efforts, I'm not going to block on the if branch (even though you could add a `# pragma: no cover` for such cases)

PR #224, opened 2018-01-16
[FIX] mass_mailing_partner: Allow unprivileged users to edit partners

yajo:
Without this patch, users without access to reading and editing mass mailing contact records are now unable to change a partner's name or email. They'd recieve an exception such as:

    AccessError: Sorry, you are not allowed to access this document. Only users with the following access level are currently allowed to do that:
    - Mass Mailing/User

    (Document model: mail.mass_mailing.contact)

Restrictive ACLs shouldn't restrict normal user operation nor DB consistency, so using sudo mode now and testing behavior.

@Tecnativa

### DIFF
--- a/mass_mailing_partner/models/res_partner.py
+++ b/mass_mailing_partner/models/res_partner.py
@@ -56,7 +56,8 @@ class ResPartner(models.Model):
                 mm_vals['name'] = vals['name']
             if vals.get('email'):
                 mm_vals['name'] = vals['email']
-            self.env["mail.mass_mailing.contact"].search([
+            # Using sudo because ACLs shouldn't produce data inconsistency
+            self.env["mail.mass_mailing.contact"].sudo().search([
                 ("partner_id", "in", self.ids),
             ]).write(mm_vals)
         return res
--- a/mass_mailing_partner/__manifest__.py
+++ b/mass_mailing_partner/__manifest__.py
@@ -7,7 +7,7 @@
 
 {
     "name": "Link partners with mass-mailing",
-    "version": "10.0.1.0.1",
+    "version": "10.0.1.0.2",
     "author": "Tecnativa, "
               "Odoo Community Association (OCA)",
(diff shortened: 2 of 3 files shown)


mass_mailing_partner/models/res_partner.py
```diff
@@ -56,7 +56,8 @@ def write(self, vals):
                 mm_vals['name'] = vals['name']
             if vals.get('email'):
                 mm_vals['name'] = vals['email']
-            self.env["mail.mass_mailing.contact"].search([
+            # Using sudo because ACLs shouldn't produce data inconsistency
+            self.env["mail.mass_mailing.contact"].sudo().search([
```
pedrobaeza, line 60: The problem with this is if you track changes, then user who made the change will be incorrect. Why this is needed?

mass_mailing_partner/models/res_partner.py
```diff
@@ -56,7 +56,8 @@ def write(self, vals):
                 mm_vals['name'] = vals['name']
             if vals.get('email'):
                 mm_vals['name'] = vals['email']
```
pedrobaeza, line 58: I have detected this as incorrect in https://github.com/OCA/social/pull/211/files#diff-7173efc033fbb9673276a0ff5ad30422R58. Maybe is this the problem?

pedrobaeza (changes requested):

mass_mailing_partner/models/res_partner.py
yajo: Nope, it's another problem, and as I see you're already fixing it, so nothing else to do here.

mass_mailing_partner/models/res_partner.py
yajo: Of course, although the change in the `res.partner` object, which is the one the user is actually updating, will be correct.

The explanation is in the description, please read it.

pedrobaeza (approved):

PR #227, opened 2018-01-30
[10.0] ADD mail_force_queue

eLBati:
Force outgoing emails to be queued instead of sent immediately.
Queued emails are sent by 'Email Queue Manager' cron.


hurrinico:
Not tested, code 👍 IMHO

SimoRubi (approved):
Code review, looks good

mail_force_queue/README.rst
```diff
@@ -0,0 +1,59 @@
...
+   :alt: License: LGPL-3
+
+==================
+Email: force queue
+==================
+
+Force outgoing emails to be queued instead of sent immediately.
+Queued emails are sent by 'Email Queue Manager' cron.
+
+This also solves possible TransactionRollbackError while deleting outgoing email (see https://github.com/odoo/odoo/issues/22148 )
```
hbrunn: make link a link

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -80,7 +80,9 @@ def test_recipient_address_compute(self):
 
     def test_message_post(self):
         # This message will generate a notification for recipient
-        message = self.env['mail.message'].create({
+        message = self.env['mail.message'].with_context({
+            'mail_notify_force_send': True
+        }).create({
```
hbrunn: if we need this here, we need to fix the original problem. The module needs its own tests anyways

hbrunn (changes requested):

mail_tracking/tests/test_mail_tracking.py
eLBati: @hbrunn what is the original problem?

hbrunn: what you write: This also solves possible TransactionRollbackError while deleting outgoing email (see https://github.com/odoo/odoo/issues/22148 )

We can't have one module fixing a problem for another module that doesn't depend on it, because tests will break in other contexts where this module is not available.

The other way around: Your module doesn't have its own tests, so it won't have coverage if used in a context where this module is not available.

eLBati: @hbrunn tests added.

I don't understand the problem with ```mail_tracking```: ```mail_force_queue``` establishes, as default behaviour, that outgoing email are queued instead of sent immediately (see tests for more clarity). Thus, the test of ```mail_tracking``` must be changed to keep the behaviour unchanged when tested along with ```mail_force_queue```.

The problem described at https://github.com/odoo/odoo/issues/22148 could also be fixed, but the functionality of ```mail_force_queue``` is useful anyway

eLBati: @hbrunn what do you think?

hbrunn: the same as before: When `mail_tracking`'s tests fail without this context key, it's a problem. We might be better off just isolating this module then

eLBati:
@hbrunn now mail_force_queue behaviour is enabled by ```force_mail_queue``` configuration field (instead of module installation). So, tests of ```mail_tracking``` do not need to be modified.
Please review.
Thanks

hbrunn (approved):

hbrunn:
yay, thanks

PR #235, opened 2018-02-15
[FIX] mail_tracking_mailgun: manual sync gets events from other recipients

chienandalu:
- The manual check can return events belonging to several recipients so they could be syncronized to a tracking not belonging to them.

cc @Tecnativa

### DIFF
--- a/mail_tracking_mailgun/models/mail_tracking_email.py
+++ b/mail_tracking_mailgun/models/mail_tracking_email.py
@@ -10,6 +10,7 @@ import requests
 from datetime import datetime
 from openerp import _, api, fields, models
 from openerp.exceptions import UserError, ValidationError
+from openerp.tools import email_split
 
 import logging
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -242,7 +243,9 @@ def action_manual_check_mailgun(self):
                 raise ValidationError(_("Event information not longer stored"))
             for item in content["items"]:
                 if not self.env['mail.tracking.event'].search(
-                        [('mailgun_id', '=', item["id"])]):
+                        [('mailgun_id', '=', item["id"])]) and (
+                            item.get("recipient", "")
+                            == email_split(tracking.recipient)[0]):
```
pedrobaeza: What happen if you have more than one email in the same field? Shouldn't you process both?

chienandalu: As far as I know there is a tracking per recipient. I use `email_split` to extract the address in cases like `"Someone" <someone@example.com>`

rafaelbn (approved):

mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -242,7 +243,9 @@ def action_manual_check_mailgun(self):
                 raise ValidationError(_("Event information not longer stored"))
             for item in content["items"]:
                 if not self.env['mail.tracking.event'].search(
-                        [('mailgun_id', '=', item["id"])]):
+                        [('mailgun_id', '=', item["id"])]) and (
+                            item.get("recipient", "") ==
+                            email_split(tracking.recipient)[0]):
```
yajo, line 251: :thinking: I'd really appreciate if you can rewrite this `if` in a more readable fashion, with a little comment that explains what is it doing... I've been staring them for awhile and I still don't understand them :nerd_face:

yajo (approved):

mail_tracking_mailgun/models/mail_tracking_email.py
hbrunn: you still approved, does this mean you don't care so much or should I wait with merging?

chienandalu:
@hbrunn My bad, I had to have commited the same moment I commented :slightly_smiling_face: Changes done

mail_tracking_mailgun/models/mail_tracking_email.py
yajo: I feel it was more a matter of taste, but I didn't want to block it for such.

Thanks

PR #236, opened 2018-02-16
[FIX] mail_tracking_mailgun: manual sync gets events from other recipients

chienandalu:
- The manual check can return events belonging to several recipients so they could be syncronized to a tracking not belonging to them.

cc @Tecnativa

### DIFF
--- a/mail_tracking_mailgun/models/mail_tracking_email.py
+++ b/mail_tracking_mailgun/models/mail_tracking_email.py
@@ -242,10 +242,10 @@ class MailTrackingEmail(models.Model):
             if "items" not in content:
                 raise ValidationError(_("Event information not longer stored"))
             for item in content["items"]:
-                if not self.env['mail.tracking.event'].search(
                 # mailgun event hasn't been synced and recipient is the same as
                 # in the evaluated tracking. We use email_split since tracking
                 # recipient could come in format: "example" <to@dest.com>
+                if not self.env['mail.tracking.event'].search(
                         [('mailgun_id', '=', item["id"])]) and (
                             item.get("recipient", "") ==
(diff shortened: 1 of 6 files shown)


chienandalu:
@rafaelbn I added a partner email autocheck feature you may want to test in runbot :wink:

rafaelbn (approved):
Thanks @chienandalu 👍 

Please review codecov/patch — 60% of diff hit (target 97.37%)

mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -242,7 +243,9 @@ def action_manual_check_mailgun(self):
                 raise ValidationError(_("Event information not longer stored"))
             for item in content["items"]:
                 if not self.env['mail.tracking.event'].search(
-                        [('mailgun_id', '=', item["id"])]):
+                        [('mailgun_id', '=', item["id"])]) and (
+                            item.get("recipient", "") ==
+                            email_split(tracking.recipient)[0]):
```
yajo, line 251: Also here I'd appreciate more readability if possible.

yajo (approved):

hbrunn (changes requested):
we need to increase coverage

pedrobaeza:
The problem is that this is all going to be mocked as Mailgun requires a valid API key, and thus tests reduce their efficiency. We can't invest more in this for now and it's not something very valuable in this case.

pedrobaeza:
@hbrunn can I merge then?

hbrunn:
if you think so, sure

pedrobaeza:
Thanks, merging.

PR #204, opened 2017-10-09
[ADD] mail_activity

hbrunn:
Backport of mail activities. Will need a UI to do something useful

PR #238, opened 2018-02-27
[PORT][9.0] mail_activity 

andreparames:
This a port to v9 of the port done in PR #204. 

I've had to drop the `mixin` approach, since the Odoo v9 models mechanism doesn't work well with them (in particular, the new-style computed fields of the models that got the mixin injected into them stopped working well).

Instead, I created a decorator that directly updates the class as it is being loaded by Python, which means that as far as the Odoo mechanism is concerned, we just used a regular inheritance and added the fields manually.


hbrunn:
you might be interested in the new commits I just pushed there (no review from my side, would be a bit pointless...)

andreparames:
@hbrunn thanks for the heads-up. I thought the plan was to keep the UI in a separate module?

hbrunn:
it still is, but for the time being, I keep it together because it simplifies development. When mergeability come close, I'd reasses if it's a good idea or not with the splitting

JordiBForgeFlow:
Superseded by https://github.com/OCA/social/pull/286

PR #240, opened 2018-03-01
[ADD] mail_autogenerated_header

hbrunn:
Autogenerated headers
=====================

This module was written to mark Odoo's emails as being autogenerated according to [RFC 3834](https://tools.ietf.org/html/rfc3834), section 5. This allows receiving mail servers to act accordingly by for example not sending a vacation autoreply.

On the receiving side, this module drops all autogenerated incoming e-mails.

The combination of both avoids possible mail loops with misconfigured or broken email servers on the opposite side.

Usage
=====

There's nothing the user has to do. Developers can set the context flag ``mail_autogenerated_header_disable`` in calls to ``send_email`` in order to suppress adding any headers at all, and override ``_message_route_drop_autoreply`` to fine tune dropping autogenerated mails per model.

Known issues / Roadmap
======================

* a configuration of this per message subtype would be nice


hbrunn:
tests are missing, will add them some time soon

pedrobaeza:
Isn't this interesting to be proposed to Odoo master?

mail_autogenerated_header/tests/test_mail_autogenerated_header.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openerp.tests.common import TransactionCase
+
+
+class TestMailAutogeneratedHeader(TransactionCase):
+    def test_mail_autogenerated_header(self):
+        pass
```
yajo: Missing

yajo (dismissed):

mail_autogenerated_header/tests/test_mail_autogenerated_header.py
hbrunn: thanks, finally with tests

hbrunn:
@Yajo I was about to close my old PRs, but this one actually seems fine to me. Could you re-review?

yajo:
Sorry, we don't use v10 anymore 🤷🏼‍♂️ 

Do as you please.

pedrobaeza:
@Yajo then remove your "Requested changes" review at least.

PR #244, opened 2018-03-19
[MIG] mail_tracking: Migration to 11.0

chienandalu:
Mail tracking
=============

This module shows email notification tracking status for any messages in
mail thread (chatter). Each notified partner will have an intuitive icon just
right to his name.

cc @Tecnativa


rafaelbn:
thanks @chienandalu ! please could you review travis? is in red

chienandalu:
@rafaelbn I'm on it. I had an old v11 instance in wich test worked. So I have to debug what change in `mail` is making the tests fail

chienandalu:
So travis won't pass due to this issue: https://github.com/odoo/odoo/issues/23808 Wich affects to `mail_digest` module here: https://github.com/OCA/social/blob/11.0/mail_digest/models/res_partner.py#L80

Without that module installed tests pass with no problem at all in my local...

pedrobaeza:
@simahawk maybe you can find a workaround for making green again the branch

simahawk:
@pedrobaeza branch 11.0 is green AFAIS. For the issue here I've no clue ATM. 

@chienandalu can you report here which domain do you get and which is the offending leaf [here](https://github.com/OCA/social/blob/11.0/mail_digest/models/res_partner.py#L80)?

Then you could improve the test cases or add a new one [here](https://github.com/OCA/social/blob/11.0/mail_digest/tests/test_partner_domains.py) so that we are sure that this case too is covered w/ tests.

BTW I'll have to work on mail_digest for some improvements in the next 2 weeks. I will have a look at this if not solved yet.

chienandalu:
Hi @simahawk You can test it like this:
1. Add a normal partner to the setup.
2. Add this partner to the message.
3. Assert the partner is found (it won't be due to the bug)
```python
def test_notify_domains_only_recipients(self):
    # we don't set recipients
    # because we call `_get_notify_by_email_domain` directly
    self.partner1.real_user_id.notification_type = 'email'
    self.partner2.real_user_id.notification_type = 'email'
    partners = self.partner1 + self.partner2
    # followers
    self.partner3.message_subscribe(self.partner2.ids)
    # partner1 is the only recipient
    message = self.message_model.create({
        'body': 'My Body',
        'res_id': self.partner3.id,
        'model': 'res.partner',
        'partner_ids': [(4, self.partner1.id), (4, self.partner_alt.id)]
    })
    domain = partners._get_notify_by_email_domain(message)
    # we find both of them since partner2 is a follower
    self._assert_found(self.partner1, domain)
    self._assert_found(self.partner2, domain)
    self._assert_found(self.partner_alt, domain)
    # no one here in digest mode
    domain = partners._get_notify_by_email_domain(message, digest=True)
    self._assert_found(self.partner1, domain, not_found=True)
    self._assert_found(self.partner2, domain, not_found=True)
```

simahawk:
@chienandalu if w/ "normal partner" you mean a partner w/ no user... then you are right. It does not pass. I'm checking.

chienandalu:
@simahawk Sure, that's what I meant :sweat_smile:

simahawk:
@chienandalu review wlcm :) https://github.com/OCA/social/pull/253

pedrobaeza:
Please rebase

chienandalu:
Rebased and a little commit squash

pedrobaeza (approved):

pedrobaeza:
@Yajo please review

mail_tracking/__manifest__.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
```
simahawk: drop coding headers

mail_tracking/controllers/main.py
```diff
@@ -0,0 +1,87 @@
...
+    reg = False
+    current = http.request.db and db == http.request.db
+    env = current and http.request.env
+    if not env:
+        with api.Environment.manage():
+            try:
+                reg = registry(db)
+            except OperationalError:
+                _logger.warning("Selected BD '%s' not found", db)
+            except:  # pragma: no cover
```
simahawk: `except Exception:` and remove the comment

mail_tracking/models/ir_mail_server.py
```diff
@@ -0,0 +1,96 @@
...
+    def _tracking_email_id_body_get(self, body):
+        body = body or ''
+        tracking_email_id = False
+        # https://regex101.com/r/lW4cB1/2
+        match = re.search(
+            r'<img[^>]*data-odoo-tracking-email=["\']([0-9]*)["\']', body)
+        if match:
+            try:
+                tracking_email_id = int(match.group(1))
+            except:  # pragma: no cover
```
simahawk: `except Exception` or `ValueError` if is related to `int` call

mail_tracking/models/ir_mail_server.py
```diff
@@ -0,0 +1,96 @@
...
+        smtp_server_used = False
+        mail_server = None
+        if mail_server_id:
+            mail_server = self.browse(mail_server_id)
+        elif not smtp_server:
+            mail_server_ids = self.search([], order='sequence', limit=1)
+            mail_server = mail_server_ids[0] if mail_server_ids else None
+        if mail_server:
+            smtp_server_used = mail_server.smtp_host
+        else:  # pragma: no cover
```
simahawk: comment needed for?

mail_tracking/models/mail_message.py
```diff
@@ -0,0 +1,78 @@
...
+            for tracking in trackings:
+                status = self._partner_tracking_status_get(tracking)
+                recipient = (
+                    tracking.partner_id.name or tracking.recipient)
+                partner_trackings.append((
+                    status, tracking.id, recipient, tracking.partner_id.id))
+                if tracking.partner_id:
+                    partners_already |= tracking.partner_id
+            # Search all recipients for this message
+            if message.partner_ids:
```
simahawk, line 52: recordsets operation can be slow compared to normal sets operations. 
You could build a set or a dict ID: name (the only values that you are going to use below)

mail_tracking/models/mail_tracking_email.py
```diff
@@ -0,0 +1,322 @@
...
+    @api.model
+    def _email_score_tracking_filter(self, domain, order='time desc',
+                                     limit=10):
+        """Default tracking search. Ready to be inherited."""
+        return self.search(domain, limit=limit, order=order)
+
+    @api.model
+    def email_is_bounced(self, email):
+        if email:
+            return len(self._email_score_tracking_filter([
```
simahawk: wouldn't be better to use search_count here, especially if - as stated above - we gonna have tons of records?

mail_tracking/models/mail_tracking_email.py
```diff
@@ -0,0 +1,322 @@
...
+                'tracking_email_id': self.id,
+            })
+
+    @api.multi
+    def _partners_email_bounced_set(self, reason, event=None):
+        recipients = []
+        if event and event.recipient_address:
+            recipients.append(event.recipient_address)
+        else:
+            recipients = list(
```
simahawk: `list()` needed for? `mapped` already returns an iterable...

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -0,0 +1,101 @@
...
+            this.action_manager = this.findAncestor(function(ancestor){
+                return ancestor instanceof ActionManager;
+            });
+        },
+        start: function () {
+            this._super();
+            this.bind_events();
+        },
+        render: function(messages, options) {
+            var MessageModel = new data.DataSetSearch(this, 'mail.message', session.context);
```
simahawk: `DataSet*` are deprecate and models too. You should use `_rpc`

simahawk (changes requested):

pedrobaeza:
@chienandalu please finish this so that we can merge it.

chienandalu:
Thanks for your comments @simahawk Take a look now

mail_tracking/models/mail_tracking_email.py
chienandalu: This oddities tend to come from `2to3`. I don't know exactly attending to wich rule. Fixed anyway.

mail_tracking/models/mail_message.py
```diff
@@ -35,8 +34,8 @@ def tracking_status(self):
         res = {}
         for message in self:
             partner_trackings = []
-            partners_already = self.env['res.partner']
-            partners = self.env['res.partner']
+            partners_already = []
```
simahawk: you don't need this list, the dict is enough. `lambda x: x.id not in partners` will work.

mail_tracking/controllers/main.py
```diff
@@ -22,7 +21,7 @@ def _env_get(db, callback, tracking_id, event_type, **kw):
                 reg = registry(db)
             except OperationalError:
                 _logger.warning("Selected BD '%s' not found", db)
-            except:  # pragma: no cover
+            except Exception:  # pragma: no cover
```
simahawk: remove pragma comment

mail_tracking/models/mail_message.py
```diff
@@ -48,18 +47,17 @@ def tracking_status(self):
...
-                    'unknown', False, partner.name, partner.id))
+            for p in message.partner_ids.filtered(
+                    lambda x: x.id not in partners_already):
+                partners[p.id] = p.name
+            for p in message.needaction_partner_ids.filtered(
+                    lambda x: x.id not in partners_already):
+                partners[p.id] = p.name
+            # If there is partners not included, then status is 'unknown'
+            partner_trackings += [
+                ('unkown', False, partners[x], x) for x in partners.keys()]
```
simahawk: `('unkown', False, pid, name) for pid, name in partners.items()`

simahawk (approved):
@chienandalu small glitches but I'm approving so that I won't block you anymore once you've addressed them :wink:

rafaelbn:
Please @Yajo  make a review here in order to merge. Thank you!

pedrobaeza:
Take into account #257

mail_tracking/models/mail_message.py
```diff
@@ -48,18 +47,17 @@ def tracking_status(self):
...
-            for partner in partners:
-                # If there is partners not included, then status is 'unknown'
-                partner_trackings.append((
-                    'unknown', False, partner.name, partner.id))
+            for p in message.partner_ids.filtered(
+                    lambda x: x.id not in partners_already):
+                partners[p.id] = p.name
+            for p in message.needaction_partner_ids.filtered(
+                    lambda x: x.id not in partners_already):
+                partners[p.id] = p.name
```
yajo: If I'm not wrong, with the change in this method you're doing the same thing as before, but avoiding the use of recordsets. However recordsets are not harmful; they help making code more understandable.

Would you mind leaving it as it was? Thanks :blush:

mail_tracking/models/mail_tracking_email.py
```diff
@@ -102,10 +101,10 @@ def _email_score_tracking_filter(self, domain, order='time desc',
     @api.model
     def email_is_bounced(self, email):
         if email:
-            return len(self._email_score_tracking_filter([
+            return self.search_count([
```
yajo: For what I can see here, we have 2 options:

1. Remove the `_email_score_tracking_filter` method, since it is not being used now.

1. Change the `_email_score_tracking_filter` signature to be `(*args, **kwargs)`, which get passed down to `self.search(*args, **kwargs)`.

   Then, call here to `_email_score_tracking_filter`, we add `count=True` as an argument, which acts just like `search_count`, but makes the count in the DB side, which is also more performant.

I think you should check if any submodule overwrites the method, to make the sane choice.

mail_tracking/models/mail_tracking_email.py
```diff
@@ -197,7 +196,7 @@ def _partners_email_bounced_set(self, reason, event=None):
         if event and event.recipient_address:
             recipients.append(event.recipient_address)
         else:
-            recipients = list(filter(None, self.mapped('recipient_address')))
+            recipients = [x for x in self.mapped('recipient_address') if x]
```
yajo: I see no benefit on this change.

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -1,102 +1,105 @@
...
+                view_type: 'form',
+                view_mode: 'form',
+                res_model: 'mail.tracking.email',
+                views: [[false, 'form']],
+                target: 'new',
+                res_id: tracking_email_id,
+            };
+            this.do_action(action);
+        },
+        bind_events: function () {
```
yajo: It's better to remove this method and add the events as a class property using [the extension method you can see here](https://github.com/odoo/odoo/blob/97fe1f05074a82715560cc825359663d65a39e41/addons/website/static/src/js/content/website_root.js#L38-L43).

This way we leverage Odoo framework to do the binding...

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -1,102 +1,105 @@
...
+                return ancestor instanceof ActionManager;
             });
-            render_super.apply(self, [messages, options]);
-        });
-    },
-
+        },
+        start: function () {
+            this._super();
+            this.bind_events();
```
yajo: ... and we avoid the need to overwrite this method too.

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -1,102 +1,105 @@
...
+                msgs = {},
+                msg_ids = [];
+            // Update trackings (async) each time we re-render thread
+            _.each(messages, function (message) {
+                msgs[message.id] = message;
+                msg_ids.push(message.id);
+            });
+            self._rpc({
+                model: 'mail.message',
+                method: 'tracking_status',
```
yajo: This thing you're doing here is not exactly wrong, but think about it: every time the chat thread is rendered (which can happens i.e. when removing a message), you're making an additional RPC call and waiting for it to resolve until you actually render the thread.

Since reading mail threads is so common in Odoo, they designed it in a way that uses the web bus and caches messages internally, so it reduces RPC calls and improves responsiveness. Here, we're introducing a bottleneck that honestly I'd prefer to avoid.

By digging into the made calls, you can see that [the web client ultimately calls the model's `message_fetch` method](https://github.com/odoo/odoo/blob/71087021f7a7881c023ac46ea9d0391be966a569/addons/mail/static/src/js/chat_manager.js#L867). Then, **it would be much better to simply overwrite that method in the backend** and avoid these RPC calls. The JS code would be much less too now.

This would be a really big performance improvement. Would you do it?

yajo (changes requested):
The migration looks good overall, but there are some points where things have changed with no good reason for it, and there is another where a change would be really appreciated. More details below.

Thanks!

rafaelbn:
@chienandalu please review @Yajo comments please. Thanks!

rafaelbn:
@chienandalu please remember to add https://github.com/OCA/social/pull/257 !! Thanks!

pedrobaeza:
@chienandalu please finish this

mail_tracking/static/src/js/mail_tracking.js
chienandalu: Well, I'd say that what you propose is alredy happening here: https://github.com/OCA/social/pull/244/files#diff-4e1cf268ea32e1266c1e39d18db0d429R66 Wich is a common point in the already exisiting rpc calls. Indeed, this code could be erased with no harm. Don't you think this method was made to have tighter check of the thread messages status?

mail_tracking/models/mail_tracking_email.py
chienandalu: Seems that nobody is using this. I chose to erase the method

mail_tracking/models/mail_tracking_email.py
chienandalu: Just for the sake of clarity

pedrobaeza:
@Yajo give your final review

mail_tracking/static/src/js/mail_tracking.js
yajo: I understand. For what I can see, this could be erased. Could you check please?

chienandalu: Seems to work as before. I guess it will lower down the refresh rate but, as you said, it will release a good bunch of unnecesary rpc calls.

yajo (approved):

PR #245, opened 2018-03-20
[IMP] mass_mailing_list_dynamic: make user aware of list change

chienandalu:
- Adds is_synced field to track whether a dynamic list has unsynced
changes or not so the user is aware that the definitive number of
contacts is yet to be determined.
- It fixes an issue that made impossible deleting a res.partner filter
when a list had use it to filter contacts.
- It also shows only the filters available for the user (shared and
belonging to self).

cc @Tecnativa

### DIFF
--- a/mass_mailing_list_dynamic/views/mail_mass_mailing_list_view.xml
+++ b/mass_mailing_list_dynamic/views/mail_mass_mailing_list_view.xml
@@ -15,6 +15,7 @@
                         <field
                             name="dynamic"
                         />
+                        <field name="is_synced" invisible="1"/>
                     </group>
                     <group attrs="{'invisible': [('dynamic', '=', False)]}">
                         <label for="sync_method"/>
@@ -26,7 +27,16 @@
                                 name="action_sync"
                                 type="object"
                                 string="Sync now"
+                                icon="fa-exclamation-triangle"
+                                class="btn-danger"
+                                attrs="{'invisible': [('is_synced', '=', True)]}"
+                            />
+                            <button
+                                name="action_sync"
+                                type="object"
+                                string="Sync now"
                                 icon="fa-refresh"
+                                attrs="{'invisible': [('is_synced', '=', False)]}"
                             />
                         </div>
                     </group>
@@ -61,6 +71,16 @@
                     />
                 </group>
             </xpath>
+            <field name="contact_nbr" position="attributes">
+                <attribute name="attrs">
+                    {'invisible': [('is_synced', '=', False)]}
+                </attribute>
+            </field>
+            <field name="contact_nbr" position="after">
+                <span attrs="{'invisible': [('is_synced', '=', True)]}">
+                    <strong>???</strong>
+                </span>
+            </field>
         </field>
     </record>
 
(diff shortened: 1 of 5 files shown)


mass_mailing_list_dynamic/wizards/mail_mass_mailing_load_filter.py
```diff
@@ -12,7 +12,9 @@ class MassMailingLoadFilter(models.TransientModel):
         comodel_name='ir.filters',
         string="Filter to load",
         required=True,
-        domain=[('model_id', '=', 'res.partner')],
+        domain="[('model_id', '=', 'res.partner'), '|', "
+               "('user_id', '=', uid), ('user_id','=',False)]",
+        ondelete='set null',
```
yajo: It is required, so better use `cascade`.

mass_mailing_list_dynamic/wizards/mail_mass_mailing_load_filter.py
```diff
@@ -12,7 +12,9 @@ class MassMailingLoadFilter(models.TransientModel):
         comodel_name='ir.filters',
         string="Filter to load",
         required=True,
-        domain=[('model_id', '=', 'res.partner')],
+        domain="[('model_id', '=', 'res.partner'), '|', "
+               "('user_id', '=', uid), ('user_id','=',False)]",
```
Yajo: This patch is good, although I'd prefer it if you comment something in the commit message about it.

pedrobaeza:
Change module version number

mass_mailing_list_dynamic/views/mail_mass_mailing_list_view.xml
```diff
@@ -61,6 +70,16 @@
                     />
                 </group>
             </xpath>
+            <field name="contact_nbr" position="attributes">
+                <attribute name="attrs">
+                    {'invisible': [('unsynced', '=', True)]}
+                </attribute>
+            </field>
+            <field name="contact_nbr" position="after">
+                <span attrs="{'invisible': [('unsynced', '=', False)]}">
```
yajo: I'd appreciate it if you can rename the field to `synced` and flip all the logic. Thinking about "unsynced" and "not unsynced" lists produces more head-scratches that thinking about "synced" and "not synced" IMHO :thinking:

yajo (changes requested):

chienandalu:
Comments attended

rafaelbn (approved):
Tested in runbot :+1: functionally looks great. Please @chienandalu review comments . Thanks!

chienandalu:
@rafaelbn I think all comments were already attended

mass_mailing_list_dynamic/views/mail_mass_mailing_list_view.xml
```diff
@@ -27,6 +28,14 @@
                                 type="object"
                                 string="Sync now"
                                 icon="fa-refresh"
```
yajo: I did a functional test and this one is a really good fix! Improves UX very much :blush: 

However I think here the icon should be `fa-exclamation-triangle`, and in the other button below, `fa-refresh`. What do you think?

yajo (approved):

mass_mailing_list_dynamic/views/mail_mass_mailing_list_view.xml
chienandalu: Sure, that's better. There it goes with tests for coverage

chienandalu: Also I refined behaviour to track `is_synced` only if the list is dynamic

yajo (approved):

pedrobaeza (approved):

PR #247, opened 2018-03-22
[ADD][8.0] New module: email_template_configurator

andreparames:
This module implements a simpler interface for the email template dynamic fields. A configurator allows the creation of predefined named replacement strings, which can then be selected in the template.

### DIFF
--- a/email_template_configurator/models/email_template.py
+++ b/email_template_configurator/models/email_template.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import api, fields, models
+
+
+class EmailTemplate(models.Model):
+
+    _inherit = 'email.template'
+
+    # Fake field for auto-completing placeholder
+    placeholder_id = fields.Many2one(
+        'email.template.placeholder', string="Placeholder")
+    placeholder_value = fields.Char()
+
+    @api.onchange('placeholder_id')
+    def _onchange_placeholder_id(self):
+        for tmpl in self:
+            if tmpl.placeholder_id:
+                tmpl.placeholder_value = tmpl.placeholder_id.placeholder
+                tmpl.placeholder_id = False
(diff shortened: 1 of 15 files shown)


email_template_configurator/models/email_template_placeholder.py
```diff
@@ -0,0 +1,17 @@
...
+
+class EmailTemplatePlaceholder(models.Model):
+
+    _name = 'email.template.placeholder'
+    _description = 'Email Template Placeholder'
+
+    name = fields.Char(required=True)
+    model_id = fields.Many2one(
+        'ir.model', string='Model', required=True)
+    placeholder = fields.Char(required=True)
```
Olivier-LAURENT: @andreparames 
Maybe a default value `${object.}` would help a little !

email_template_configurator/models/email_template.py
```diff
@@ -0,0 +1,22 @@
...
+    placeholder_id = fields.Many2one(
+        'email.template.placeholder', string="Placeholder")
+    placeholder_value = fields.Char()
+
+    @api.onchange('placeholder_id')
+    def _onchange_placeholder_id(self):
+        for tmpl in self:
+            if tmpl.placeholder_id:
+                tmpl.placeholder_value = tmpl.placeholder_id.placeholder
+                tmpl.placeholder_id = False
```
Olivier-LAURENT: @andreparames 
Is it really useful to reset the `placeholder_id` here ?
For the user it will produce a special effect: once a placeholder selected, the choice is immediately empty again

email_template_configurator/tests/test_configurator.py
```diff
@@ -0,0 +1,37 @@
...
+from openerp.tests import common
+
+
+class test_configurator(common.TransactionCase):
+
+    def test_configurator(self):
+        placeholders_obj = self.env['email.template.placeholder']
+        templates_obj = self.env['email.template']
+        invoice_model = self.env['ir.model'].search([
+            ('model', '=', 'account.invoice'),
```
Olivier-LAURENT: @andreparames 
Is this test will work properly if the "account" module is not loaded ?

Olivier-LAURENT (commented):
Tested. Just 3 details.
Thx for the job

Olivier-LAURENT:
@andreparames 
Strangly, your test seems to not be launched in travis. Is an `__init__.py file` missing ?

email_template_configurator/models/email_template.py
andreparames: Fair point. Since it was a fake field, I thought it didn't make much sense to leave the value there (to be saved, etc). But the strange behavior may create some confusion.

Olivier-LAURENT (approved):
Review and test: 👍

PR #248, opened 2018-03-22
FIX mail_notify_bounce UnicodeDecodeError

eLBati:
also using real email messages


hbrunn:
so this time i'll wait. But fasttrack when it's green

eLBati:
@hbrunn updated using ```ustr```. Thanks for the hint

hbrunn:
for data crunching, the function is a bad idea because it's slow, but with mails we talk about seconds, there a couple of extra milliseconds don't matter so much. thanks for following up so fast, otherwise i'd have to fix this myself

PR #252, opened 2018-04-04
[IMP] Apply follower domain on automatically added followers

etobella:
Sometimes, followers are added automatically, but the domain is not applied.

For example, if your domain is [('customer', '=', True)], and you take a partner with customer == False and send a message including the partner email, this partner will be added as follower, but it shouldn't.

This PR solves this issue.

### DIFF
--- a/mail_restrict_follower_selection/README.rst
+++ b/mail_restrict_follower_selection/README.rst
@@ -7,6 +7,8 @@ Restrict follower selection
 
 This module was written to allow you to restrict the selection of possible followers. For example, if you use the social ERP functions only internally, it makes sense to filter possible followers for being employees. Otherwise, you'll get a quite crowded list of partners to choose from.
 
+Moreover, the module disables the option to automatically add followers that do not meet the domain.
+
 Configuration
 =============
 
@@ -21,7 +23,7 @@ Usage
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/205/10.0
+   :target: https://runbot.odoo-community.org/runbot/205/11.0
 
 For further information, please visit:
 
@@ -43,6 +45,7 @@ Contributors
 
 * Holger Brunn <hbrunn@therp.nl>
 * Nguyen Tan Phuc <phuc.nt@komit-consulting.com>
+* Enric Tobella <etobella@creublanca.es>
 
 Maintainer
 ----------
--- a/mail_restrict_follower_selection/__manifest__.py
+++ b/mail_restrict_follower_selection/__manifest__.py
@@ -4,8 +4,8 @@
 
 {
     "name": "Restrict follower selection",
-    "version": "11.0.1.0.0",
-    "author": "Therp BV,Odoo Community Association (OCA)",
+    "version": "11.0.2.0.0",
+    "author": "Therp BV,Creu Blanca,Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Social Network",
     "summary": "Define a domain from which followers can be selected",
--- a/mail_restrict_follower_selection/models/mail_followers.py
+++ b/mail_restrict_follower_selection/models/mail_followers.py
@@ -1,0 +1,24 @@
+# Copyright (C) 2018 Creu Blanca
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+from odoo.tools.safe_eval import safe_eval
+
+
+class MailFollowers(models.Model):
+    _inherit = 'mail.followers'
+
+    @api.model
+    def _add_follower_command(self, res_model, res_ids, partner_data,
+                              channel_data, force=True):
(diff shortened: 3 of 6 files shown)


JordiBForgeFlow (approved):
Code review and tested

mail_restrict_follower_selection/models/mail_followers.py
```diff
@@ -0,0 +1,21 @@
...
+
+class MailFollowers(models.Model):
+    _inherit = 'mail.followers'
+
+    @api.model
+    def _add_follower_command(self, res_model, res_ids, partner_data,
+                              channel_data, force=True):
+        partners = self.env['res.partner'].search(
+            [('id', 'in', list(partner_data))] +
+            safe_eval(self.env['ir.config_parameter'].get_param(
```
pedrobaeza: This should use the same scheme than the other domain, allowing to put a domain per model. See:

https://github.com/OCA/social/blob/e640e21bcc3f35a9b45ae97843f0540b505bcea0/mail_restrict_follower_selection/models/mail_wizard_invite.py#L30

mail_restrict_follower_selection/__manifest__.py
```diff
@@ -4,8 +4,8 @@
 
 {
     "name": "Restrict follower selection",
-    "version": "11.0.1.0.0",
-    "author": "Therp BV,Odoo Community Association (OCA)",
+    "version": "11.0.1.1.0",
```
pedrobaeza: This new feature deserves to raise the version module to 11.0.2.0.0

pedrobaeza (changes requested):

mail_restrict_follower_selection/models/mail_followers.py
etobella: Good point! I have applied your suggestion

mail_restrict_follower_selection/README.rst
```diff
@@ -7,6 +7,8 @@ Restrict follower selection
 
 This module was written to allow you to restrict the selection of possible followers. For example, if you use the social ERP functions only internally, it makes sense to filter possible followers for being employees. Otherwise, you'll get a quite crowded list of partners to choose from.
 
+Moreover, the module disables the option to automatically add followers that do not meet the domain.
+
```
pedrobaeza, line 11: I'm missing here the same explanation in Configuration section as in the v8 version:

https://github.com/OCA/social/blob/e640e21bcc3f35a9b45ae97843f0540b505bcea0/mail_restrict_follower_selection/README.rst

etobella: I am using the same parameters, so the same configuration applies. I think that it should be keep the same logic, isn't it? Do you think that a second set of parameters is necessary?

pedrobaeza: Sorry, I didn't see the text. Yes, it's OK.

etobella:
@hbrunn

mail_restrict_follower_selection/models/mail_followers.py
```diff
@@ -0,0 +1,30 @@
...
+    @api.model
+    def _add_follower_command(self, res_model, res_ids, partner_data,
+                              channel_data, force=True):
+        partners = self.env['res.partner'].search(
+            [('id', 'in', list(partner_data))] +
+            safe_eval(self._mail_restrict_follower_selection_get_domain())
+        )
+        return super()._add_follower_command(
+            res_model, res_ids,
+            {p.id: None for p in partners}, channel_data, force)
```
hbrunn: you lose the subtypes here. You need to filter the dict to contain only the keys coming back from previous search. And pass `force` as keyword argument to super

mail_restrict_follower_selection/models/mail_followers.py
```diff
@@ -0,0 +1,30 @@
...
+
+from odoo import api, models
+from odoo.tools.safe_eval import safe_eval
+
+
+class MailFollowers(models.Model):
+    _inherit = 'mail.followers'
+
+    @api.model
+    def _mail_restrict_follower_selection_get_domain(self):
```
hbrunn: any subtle difference I'm missing, or did you just copy the function from https://github.com/etobella/social/blob/0b4d43ae4de03c2003aad1ceac107858c8ca095c/mail_restrict_follower_selection/models/mail_wizard_invite.py#L13 - if so, just call it instead

hbrunn (changes requested):
I was pondering if there are situations where we do want the resitriction on the UI, but want a post to actually add the person. Probably there are, but in this case it's simple to make this configurable, so I won't ask this here.

mail_restrict_follower_selection/models/mail_followers.py
etobella: @hbrunn all changes have been applied. Also, I detected an error on https://github.com/etobella/social/blob/0b4d43ae4de03c2003aad1ceac107858c8ca095c/mail_restrict_follower_selection/models/mail_wizard_invite.py#L13 , as it was missing sudo when reading ir.config_parameter. It is fixed now.

hbrunn (approved):

pedrobaeza (approved):

PR #170, opened 2017-06-07
[10.0] Improving the mail tracking value

TDu:
Improve the tracking of changed values on database fields
("mail.tracking.value" functionality) by adding many2many and one2many support.
As well as a more user friendly view to consult changes recorded.

PR #254, opened 2018-04-05
[MIG] mail_improved_tracking_value: migration to Odoo 11

mpanarin:
migration of module https://github.com/OCA/social/pull/170

### DIFF
--- a/mail_improved_tracking_value/models/mail_tracking_value.py
+++ b/mail_improved_tracking_value/models/mail_tracking_value.py
@@ -1,0 +1,85 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import json
+
+from odoo import api, fields, models
+
+
+class MailTrackingValue(models.Model):
+
+    _inherit = "mail.tracking.value"
+
+    new_value_formatted = fields.Char(
+        compute='_compute_formatted_value',
+        string='New value',
+    )
+    old_value_formatted = fields.Char(
+        compute='_compute_formatted_value',
+        string='Old value',
+    )
+    record_name = fields.Char(
+        related='mail_message_id.record_name',
+    )
+    model = fields.Char(
+        related='mail_message_id.model',
+        store='True',
+        string='Model',
+    )
+
+    @api.depends('new_value_char', 'new_value_integer', 'new_value_float',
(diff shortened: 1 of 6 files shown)


TDu (approved):
LGTM

simahawk (changes requested):
as we already have many reviews on v10 I think you have to include that commit and add a new one for v11 mig

guewen (approved):

rafaelbn:
Thank you! Agree with @simahawk  in https://github.com/OCA/social/pull/254#pullrequestreview-110357733 

Also is possible to add some test. it should be very nice codecov/patch — 23.8% of diff hit (target 87.44%)

rafaelbn (approved):

rafaelbn:
Hi @mpanarin , please could you attend @simahawk comments ? Thank you!

ThomasBinsfeld (approved):

pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

tdombos:
Gives an error message if trying to delete a one2may line tracked:
> Odoo Server Error - Missing Record 
> Record does not exist or has been deleted.

mail_improved_tracking_value/models/mail_tracking_value.py
```diff
@@ -0,0 +1,85 @@
...
+                record.old_value_formatted = record.old_value_text
+
+    @api.model
+    def create_tracking_values(self, old_value, new_value, col_name, col_info):
+        """ Add tacking capabilities for many2many and one2many fields """
+        if col_info['type'] in ('many2many', 'one2many'):
+
+            def get_values(source, prefix):
+                if source:
+                    names = ', '.join(source.mapped('display_name'))
```
MiquelRForgeFlow, line 63:
```suggestion
                    names = ', '.join(
                        [s.display_name for s in source.exists()])
```

If you do this, you avoid the `Missing Record` error for one2many, but then the message doesn't show the deleted record 😕

MiquelRForgeFlow:
In fact, the best way I found to track a one2many is creating a computed char and tracking it, like I did in https://github.com/Eficent/ao-odoo/pull/121/files#diff-4a0c09c64bab72fb4713db84f0c2358dR25.

MiquelRForgeFlow:
Superseded by https://github.com/OCA/social/pull/678.

PR #255, opened 2018-04-06
mail_digest: fix behavior w/ `force_send` enabled

simahawk:
`force_send` is an option in `mail` core module
that allows to send an email immediately.

Handling the message w/ digest is wrong and in any case
we gonna have a digest w/out the message inside it
since is going to be deleted right after send.

A typical use case is the notification sent to new followers.

TODO: after this PR is merged fwd port it to v11 and backport it to v9.

### DIFF
--- a/mail_digest/models/res_partner.py
+++ b/mail_digest/models/res_partner.py
@@ -89,7 +89,8 @@ class ResPartner(models.Model):
                 force_send=False, send_after_commit=True, user_signature=True):
         """Override to delegate domain generation."""
         # notify_by_email
-        email_domain = self._get_notify_by_email_domain(message)
+        email_domain = self._get_notify_by_email_domain(
+            message, force_send=force_send)
         # `sudo` from original odoo method
         # the reason should be that anybody can write messages to a partner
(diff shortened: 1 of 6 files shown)


mail_digest/models/res_partner.py
```diff
@@ -127,10 +129,12 @@ def _notify_by_digest(self, message):
         self.env['mail.digest'].sudo().create_or_update(self, message)
 
     @api.model
-    def _get_notify_by_email_domain(self, message, digest=False):
+    def _get_notify_by_email_domain(self, message,
+                                    force_send=False, digest=False):
         """Return domain to collect partners to be notified by email.
 
         :param message: instance of mail.message
+        :param force_send: wether the message should be sent immediately
```
gurneyalex: "whether"

gurneyalex (changes requested):
the force_send behaviour should be clarified in the readme

simahawk:
@gurneyalex updated

gurneyalex (approved):

rafaelbn (approved):

PR #256, opened 2018-04-12
[10.0][MIG] migrate module mail_forward

ecino:
This takes the module mail_forward that was available in version 8.0 and updates it in order to work on version 10.0


mail_forward/README.rst
```diff
@@ -0,0 +1,101 @@
...
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Emanuel Cino <ecino@compassion.ch>
+* Jairo Llopis <j.llopis@grupoesoc.es>
```
pedrobaeza: Please change the author to the new company: `jairo.llopis@tecnativa.com`

mail_forward/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+# © 2014-2015 Grupo ESOC <www.grupoesoc.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Message Forward",
+    "summary": "Add option to forward messages",
+    "version": "10.0.1.0.0",
+    "category": "Social Network",
+    "website": "https://grupoesoc.es",
+    "author": "Grupo ESOC Ingeniería de Servicios, "
```
pedrobaeza: Change also the company to the new one: `Tecnativa`

rafaelbn (commented):
Hi @ecino , thank!

I've tested and it work fine just one thing you should fix. After fordward the message, the chatter is not refreshed and user have to press F5... I think chatter should be refreshed in the same way that sending a new message happends

Let me know please

Thanks

ecino:
@rafaelbn I added the functionnality in cac0a83.
However I'm not sure it's the most efficient way to refresh, as I'm not familiar with what is available in the javascript functions of the mail module. This requires as well #257 to work, because it uses the `chat_manager.make_message` function.
But in the end, it does what you suggested.

ecino:
@rafaelbn The module is now working with #257 being merged.

rafaelbn (approved):
Tested in runbot functionally :+1:

mail_forward/__manifest__.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# © 2014-2015 Grupo ESOC <www.grupoesoc.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Message Forward",
+    "summary": "Add option to forward messages",
+    "version": "10.0.1.0.0",
+    "category": "Social Network",
+    "website": "https://grupoesoc.es",
```
rafaelbn: @ecino please change this for the rightone as described in https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L9

mail_forward/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+    "author": "Odoo Community Association (OCA)"
```
rafaelbn: Don't change order please https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L10

mail_forward/README.rst
```diff
@@ -0,0 +1,101 @@
...
+help us smashing it by providing a detailed and welcomed feedback.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Emanuel Cino <ecino@compassion.ch>
```
rafaelbn: @ecino you should add you name after older contributors and not before

rafaelbn (commented):
Please @Yajo as original author review this migration. Thanks!

rafaelbn (changes requested):
Please @ecino check my comments. Thank you!

ecino:
@rafaelbn Thanks for reviewing it. Corrections are done.

mail_forward/models/compose_message.py
```diff
@@ -56,10 +55,11 @@ def change_destination_object(self):
                 record_name = "%s %s" % (model_name, record_name)
 
             self.record_name = record_name
-        else:
-            self.model = self.res_id = self.record_name = False
 
-    @api.one
+    @api.multi
+    def send_mail_action(self):
+        return self.send_mail()
```
yajo: Remove this method.

mail_forward/static/src/js/mail_forward.js
```diff
@@ -2,76 +2,122 @@
...
-                                     this.parent_id))
-            {
-                subject.push(this.record_name);
-            }
-            if (this.subject) {
-                subject.push(this.subject);
-            } else if (subject.length < 2) {
-                subject.push(_t("(No subject)"));
-            }
+            MessageModel.call('read', [message_id, read_fields], {context: session.user_context}).then(function (result) {
```
yajo: Please put this into a separate, proxied, class-level method.

Just for inheritance/maintainability. :wink:

mail_forward/static/src/js/mail_forward.js
```diff
@@ -2,76 +2,122 @@
...
+        },
+        init: function (parent, options) {
+            this._super.apply(this, arguments);
+            // Add click reaction in the events of the thread object
+            this.events['click .o_forward'] = function (event) {
+                var message_id = $(event.currentTarget).data('message-id');
+                this.trigger("message_forward", message_id);
+            };
+        },
+        thread_reload: function(){
```
yajo, line 92: I'm not sure I like this new feature.

I did a test from the Odoo inbox forwarding to a different object... If you are forwarding to a different object, why to reload? It should never get displayed...

It moves me above in the message list if I was browsing somewhere below, and clears the messages. Well... it's a confusing behavior.

yajo (changes requested):

yajo:
I tested on runbot.

I go to events, and try to forward one message to another event. Events model don't appear in the dropdown list. Is this intentional?

AaronHForgeFlow (commented):
Tested on runbot and worked fine. Installed Sales and can forwarded messages between different SO successfully. However I don't see Even't models. Maybe it's been filtered somewhere. Can you check?

ecino:
Hi @aheficent, sorry for the delay of my answer. You should add the Event model in the referenceable models (see the README of the module). You find the menu in Admin Settings -> Technical -> Database Structure -> Referenceable Models

mail_forward/static/src/js/mail_forward.js
ecino: Not sure what you mean by proxied method. I'm no javascript expert, I just ported the module to v10.

mail_forward/static/src/js/mail_forward.js
ecino: @rafaelbn originally requested this feature. As I said, I'm no javascript expert so maybe I did not implement it the best way. Anyway I would need help to improve javascript.

yajo:
Yikes! that was in the README! Bad me for not reading myself :laughing: 

OK, then all you need to do is fix that :x: in Travis. The JS improvements are needed, yes, but that can be done later.

Thanks!

AaronHForgeFlow:
Thank you for your reply @ecino 

I'm getting an error when forwarding messages between events. The trace:

>   File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 896, in call_button
    action = self._call_kw(model, method, args, {})
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 884, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 685, in call_kw
    method = getattr(type(model), name)
AttributeError: type object 'mail_forward.compose_message' has no attribute 'send_mail_action'

ecino:
Oops, that was after I removed the method as suggested by @Yajo. It looks like the method is needed, so I will add it again. I will tell you when it's done.

ecino:
The pull request is ready to go. Travis errors should be fixed with #288

yajo:
Then please rebase to get those changes; right now bots are :x:

ecino:
@Yajo , successfully rebased, everything looks good. :heavy_check_mark:

yajo (changes requested):
I tested on runbot:

1. In messages view, clicking on the :star: or :heavy_check_mark: icons, has no effect
1. When forwarding message to a partner, an error appears:

   ```
   Odoo Server Error

     File "/.repo_requirements/odoo/odoo/http.py", line 642, in _handle_exception
     File "/.repo_requirements/odoo/odoo/http.py", line 684, in dispatch
     File "/.repo_requirements/odoo/odoo/http.py", line 334, in _call_function
     File "/.repo_requirements/odoo/odoo/service/model.py", line 101, in wrapper
     File "/.repo_requirements/odoo/odoo/http.py", line 327, in checked_call
     File "/.repo_requirements/odoo/odoo/http.py", line 942, in __call__
     File "/.repo_requirements/odoo/odoo/http.py", line 507, in response_wrap
       response = f(*args, **kw)
     File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 896, in call_button
       action = self._call_kw(model, method, args, {})
     File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 884, in _call_kw
       return call_kw(request.env[model], method, args, kwargs)
     File "/.repo_requirements/odoo/odoo/api.py", line 685, in call_kw
       method = getattr(type(model), name)
   AttributeError: type object 'mail_forward.compose_message' has no attribute 'send_mail_action'
   ```

ecino:
Maybe I dropped something during the rebase, sorry about that. I'll check it and let you know when it's fixed. Thanks for double checking it.

ecino:
@Yajo How do you access the runbot? I cannot find how to launch it, I just see the build logs.

yajo:
There it is:

peek 2018-07-05 13-03

ecino:
@Yajo I fixed the forward bug (method send_action was not to be removed). However your other issues with the star and check icons seem not related with this module. I checked on the current runbot of the branch 10.0 and the same issue happens.

rafaelbn (approved):
Tested again functionally :+1:

antoniocanovas:
Tested functionality

pedrobaeza:
@Yajo please update your review

mail_forward/models/compose_message.py
```diff
@@ -0,0 +1,97 @@
...
+                          .name)
+            record_name = self.destination_object_id.name_get()[0][1]
+            if model_name:
+                record_name = "%s %s" % (model_name, record_name)
+
+            self.record_name = record_name
+
+    @api.multi
+    def send_mail_action(self):
+        return self.send_mail()
```
yajo: Why to declare 2 methods to do the same? Am I missing something? :thinking: Otherwise, remove one of them please. :wink:

yajo (changes requested):

mail_forward/models/compose_message.py
ecino: The way odoo mail_compose_message is done works the same, if we don't declare them it will fail. I tried to remove this one but then it fails.

yajo: OK, I see now. However, [you are repeating here the same code as upstream](https://github.com/odoo/odoo/blob/33b698254a00862991d8596b72c3e3b45ebbfbf3/addons/mail/wizard/mail_compose_message.py#L186-L194), so it still makes no sense that it fails... (You should only override a method if anything changes)

Besides, it breaks inheritance too. In case this were needed, it should be:

```python
return super(MailComposeMessage, self).send_mail()
```

Could you double-check this please?

ecino: I know it's weird but when I removed that `send_mail_action` method, Travis was failing so I don't know why but this method is needed. If I do as you suggest, we will bypass our `send_mail` method which is not good. The other solution I can do is to call super of `send_mail_action`. I know the code seems weird to only call super method but if you want me I can remove it again and you will see by yourself that tests are failing without it.

yajo: Oh snap, excuse me! It was right as it was. I didn't realize this was a `_inherits` model, not an `_inherit` one.

This kind of inheritance inherits fields, but not methods. Possibly calling `super()` will not work, and the correct code is the one you had indeed.

Just please add a comment similar to upstream's, so reviewers can understand the reasoning behind that code.

Thanks, and sorry!

ecino: No worries, I didn't know as well that `_inherits` don't inherit the methods. Now it's clear why it's needed. I added the comment and restore the previous version. Thanks for your review.

yajo (approved):
Awesome, thanks!

PR #257, opened 2018-04-13
[10.0][FIX] mail_tracking js could break other modules using chat_manager

ecino:
This fix prevents some other modules that would use the JS class chat_manager to break. If other modules call function `chat_manager.make_message`, line 22 would call `this._make_message_super` but `_make_message_super` does not exist in other modules `this` object.

### DIFF
--- a/mail_tracking/static/src/js/mail_tracking.js
+++ b/mail_tracking/static/src/js/mail_tracking.js
@@ -8,22 +8,18 @@ var core = require('web.core');
 var session = require('web.session');
 var Model = require('web.Model');
 var ActionManager = require('web.ActionManager');
-var chat_manager = require('mail.chat_manager');
 var ChatThread = require('mail.ChatThread');
-var Chatter = require('mail.Chatter');
 
 var _t = core._t;
 var MessageModel = new Model('mail.message', session.context);
 
-// chat_manager is a simple dictionary, not an OdooClass
-chat_manager._make_message_super = chat_manager.make_message;
-chat_manager.make_message = function(data) {
-    var msg = this._make_message_super(data);
-    msg.partner_trackings = data.partner_trackings || [];
-    return msg;
-};
 
 ChatThread.include({
+    _preprocess_message: function (message) {
+        var msg = this._super.apply(this, arguments);
+        msg.partner_trackings = msg.partner_trackings || [];
+        return msg;
+    },
     on_tracking_partner_click: function (event) {
         var partner_id = this.$el.find(event.currentTarget).data('partner');
         var state = {
--- a/mail_tracking/__manifest__.py
+++ b/mail_tracking/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Email tracking",
     "summary": "Email tracking system for all mails sent",
-    "version": "10.0.1.1.0",
+    "version": "10.0.1.1.1",
     "category": "Social Network",
     "website": "http://www.tecnativa.com",
     "author": "Tecnativa, "


pedrobaeza:
Please change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

rafaelbn:
Please @Yajo @chienandalu take a look here, thanks!

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -18,7 +18,7 @@ var MessageModel = new Model('mail.message', session.context);
 // chat_manager is a simple dictionary, not an OdooClass
 chat_manager._make_message_super = chat_manager.make_message;
 chat_manager.make_message = function(data) {
-    var msg = this._make_message_super(data);
+    var msg = chat_manager._make_message_super(data);
```
yajo: AFAICS this wouldn't change anything in reality.

In JS, `this` always belongs to the object being called. To understand better:

```js
a = {c: "pikachu"}
a.b = function () {
  console.log(this.c)  // Logs "pikachu" because `this` is `a`
}
```

So I think the PR is not really needed. Did you really encounter the problem you were describing? Or was it just that it could hypothetically happen? If you encountered, could you test it or provide code to check the bug?

Thanks

yajo (changes requested):

mail_tracking/static/src/js/mail_tracking.js
ecino: Please see https://github.com/OCA/social/commit/cac0a83cb42d1ffc6e56ae9883b7d8895e10dc70#diff-c9046c334bb9af935c424584bde259a7R97
where I call the function in order to refresh the thread. This fails without the fix with as shown here:
image

chienandalu: I agree with @Yajo

chienandalu (changes requested):

mail_tracking/static/src/js/mail_tracking.js
ecino: @chienandalu @Yajo You can see the example I posted. Maybe this is due to Odoo inheritance, when we use `require('mail.chat_manager')`. It seems that other instances don't share the attributes defined here in mail_tracking.

ecino:
@chienandalu @Yajo You can see the example I posted above. Maybe this is due to Odoo inheritance, when we use `require('mail.chat_manager')`. It seems that other instances don't share the attributes defined here in mail_tracking.

mail_tracking/static/src/js/mail_tracking.js
yajo: I'm not sure you are actually calling the proper re-render method.

In any case, I just discovered what I think is [the really good place for this override to exist: `_preprocess_message`](https://github.com/odoo/odoo/blob/1ae1cac9aebab6721fcd5b441076f827a97c31aa/addons/mail/static/src/js/thread.js#L227).

Please change this PR and move the code to patch the message inside that method.

Do you agree @chienandalu ?

yajo (changes requested):

ecino:
@Yajo I made the changes you suggested.

yajo (approved):

pedrobaeza (approved):

pedrobaeza:
@chienandalu please include it in your v11 migration.

chienandalu:
@pedrobaeza done

PR #259, opened 2018-04-16
[FIX] mass_mailing_event: Integration tests

yajo:
When running tests and `mass_mailing_list_dynamic` was installed, they failed.

@Tecnativa

### DIFF
--- a/mass_mailing_event/tests/test_mass_mailing_event.py
+++ b/mass_mailing_event/tests/test_mass_mailing_event.py
@@ -3,10 +3,12 @@
 # Copyright 2017 David Vidal <david.vidal@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo.tests import common
+from odoo.tests.common import at_install, post_install, SavepointCase
 
 
-class TestMassMailingEvent(common.SavepointCase):
+@at_install(True)
+@post_install(True)
+class TestMassMailingEvent(SavepointCase):
 
     @classmethod
     def setUpClass(cls):
--- a/mass_mailing_event/__manifest__.py
+++ b/mass_mailing_event/__manifest__.py
@@ -6,9 +6,9 @@
 {
     'name': 'Mass mailing event',
     'summary': 'Link mass mailing with event for excluding recipients',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'Marketing',
-    'website': 'https://www.tecnativa.com',
+    'website': 'https://github.com/OCA/social',
     'author': 'Tecnativa, '
               'Odoo Community Association (OCA)',
     'license': 'AGPL-3',


mass_mailing_event/tests/test_mass_mailing_event.py
```diff
@@ -3,10 +3,12 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo.tests import common
+from odoo.tests.common import at_install, post_install, SavepointCase
 
 
-class TestMassMailingEvent(common.SavepointCase):
+@at_install(True)
+@post_install(True)
+class TestMassMailingEvent(SavepointCase):
```
pedrobaeza, line 11: Use HttpCase for avoiding problems with PDF rendering in tests.

yajo: Does that apply for this module?

pedrobaeza: The problem is when you integrate with other tests that perform that operation. Result: this one hangs.

yajo: Sorry but I don't understand how could that happen

pedrobaeza: @Yajo you have already changed other tests for the same. Please don't make elaborate me again the reason.

yajo: If you don't want to write it again, you can paste the link, but if you don't elaborate I won't understand.

pedrobaeza: I don't have the link, but you have made this several times. As said, if other module tests to render a PDF through wkhtmltopdf, this test will be hang as the other leave the cursor blocked. Using HttpCase creates another cursor and there's no problem then.

yajo: yes, that's true... but this test case has no reports nor http connections, that's why I don't understand your request :roll_eyes:

pedrobaeza: It's when other modules have. Try to combine this one with base_report_to_printer...

yajo: The point is that the base class you use for your tests must be `HttpCase` if you need to perform any RPC or normal HTTP calls (including reports) in any of your tests.

A module could even have `SavepointCase/TransactionCase` tests along with `HttpCase` tests in different files. As long as there's no HTTP connection to be established, you don't need `HttpCase` at all.

That's the case here, just check the tests: all are inner ORM calls: no reports, no web pages to render...

If you don't need this, then it's better to stick with other classes (such as here) because they have smaller boilerplate.

The real fix here is moving tests to `post_install` mode, and the conflict happens with `mass_mailing_list_dynamic` because it adds a required field with a default value that cannot be loaded in `at_install` mode; this makes integration environments to fail this test. If the problem were because the base class should be `HttpCase`, then all unit tests would fail.

pedrobaeza (approved):
Fast-tracking for finishing this with the possible rough edge.

mass_mailing_event/tests/test_mass_mailing_event.py
```diff
@@ -3,10 +3,12 @@
 # Copyright 2017 David Vidal <david.vidal@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo.tests import common
+from odoo.tests.common import at_install, post_install, SavepointCase
 
 
-class TestMassMailingEvent(common.SavepointCase):
+@at_install(True)
```
yajo, line 9: Yikes, typo here.

PR #266, opened 2018-04-26
[IMP][10.0] Mail digest improvements

simahawk:
TODO: fwd port to v11 once merged


gurneyalex (approved):

mail_digest/models/mail_digest.py
```diff
@@ -54,6 +53,14 @@ class MailDigest(models.Model):
         domain=[('type', '=', 'qweb')],
         oldname='template_id',
     )
+    sanitize_msg_body = fields.Boolean(
+        string='Sanitize message body',
+        help='Collected messages can have different styles applied '
+             'on each element. If this is flag is on (default) '
```
grindtildeath: If this ~is~ flag is on

grindtildeath (approved):
Out of a small typo :+1:

rafaelbn:
Hi @simahawk , I can merge but could you take a look please in : (OCA/social is very well maintained :smile: ) 

https://coveralls.io/github/OCA/social?branch=10.0



Tell me what do you think about this please.  I don't want to block any improvement :+1: 

Thanks!

simahawk:
@rafaelbn I could add a test for body sanitize but the preview file should NOT be checked for coverage since is just to generate demo data. I cannot find any doc that states how you can ignore a whole file on demand http://coverage.readthedocs.io/en/coverage-4.3.4/excluding.html without modifying configuration file :disappointed:
@moylop260 any pointer? Can we make the tool read a module-specific config file?

pedrobaeza:
@simahawk add that controller on tests folder, so that only on test it's run.

simahawk:
@pedrobaeza this will make pytest and any other auto-discover-test fail. Not an option :cry:

pedrobaeza:
Which tools are you talking about? Are they out of Odoo?

simahawk:
@pedrobaeza pytest? to name one. And in any case putting the file into tests is just an hack. That controller must be available not only for tests or demo.

pedrobaeza:
Then you should test it for coverage. No other solution.

simahawk:
rebased and added some dumb test for layout preview controller

rafaelbn:
@simahawk could you squash commits before merging?

rafaelbn (approved):

simahawk:
@rafaelbn I'd like to keep them actually ;)

PR #269, opened 2018-05-09
[FIX] mail_tracking_mailgun: json.load() won't swallow bytes

chienandalu:
- json.load() in python under 3.6 doesn't support binary input.
- https://docs.python.org/3/whatsnew/3.6.html#json
- This way, we let requests to decode the response itself.

cc @Tecnativa

### DIFF
--- a/mail_tracking_mailgun/models/mail_tracking_email.py
+++ b/mail_tracking_mailgun/models/mail_tracking_email.py
@@ -5,7 +5,6 @@
 
 import hashlib
 import hmac
-import json
 import requests
 from datetime import datetime
 from odoo import _, api, fields, models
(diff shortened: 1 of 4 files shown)


mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -238,7 +237,8 @@ def action_manual_check_mailgun(self):
             if not res or res.status_code != 200:
                 raise ValidationError(_(
                     "Couldn't retrieve Mailgun information"))
-            content = json.loads(res.content)
+            # content = json.loads(res.content)
```
pedrobaeza: Don't comment, but remove it

chienandalu:
done

pedrobaeza (approved):
Fast-tracking to solve the problem

PR #271, opened 2018-05-11
[FIX] mass_mailing_partner: compute fields performance

chienandalu:
- In DB which use large amounts of records and intesive use of
mass_mailings, not optimized compute records lead to a drastical
decrease of performance

cc @Tecnativa

### DIFF
--- a/mass_mailing_partner/models/res_partner.py
+++ b/mass_mailing_partner/models/res_partner.py
@@ -39,18 +39,25 @@ class ResPartner(models.Model):
     @api.depends('mass_mailing_contact_ids',
                  'mass_mailing_contact_ids.opt_out')
     def _compute_mass_mailing_contacts_count(self):
+        contact_data = self.env['mail.mass_mailing.contact'].read_group(
+            [('partner_id', 'in', self.ids)], ['partner_id'], ['partner_id'])
+        mapped_data = dict(
+            [(contact['partner_id'][0], contact['partner_id_count'])
+             for contact in contact_data])
         for partner in self:
-            partner.mass_mailing_contacts_count = len(
-                self.env['mail.mass_mailing.contact'].search_read(
-                    [('partner_id', '=', partner.id)], ['id']))
+            partner.mass_mailing_contacts_count = mapped_data.get(partner.id,
+                                                                  0)
 
     @api.multi
     @api.depends('mass_mailing_stats')
     def _compute_mass_mailing_stats_count(self):
+        contact_data = self.env['mail.mail.statistics'].read_group(
+            [('partner_id', 'in', self.ids)], ['partner_id'], ['partner_id'])
+        mapped_data = dict(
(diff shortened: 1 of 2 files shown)


mass_mailing_partner/models/res_partner.py
```diff
@@ -35,16 +35,22 @@ def _check_email_mass_mailing_contacts(self):
...
 
-    @api.one
+    @api.multi
     @api.depends('mass_mailing_contact_ids',
                  'mass_mailing_contact_ids.opt_out')
     def _compute_mass_mailing_contacts_count(self):
-        self.mass_mailing_contacts_count = len(self.mass_mailing_contact_ids)
+        for partner in self:
+            partner.mass_mailing_contacts_count = len(
+                self.env['mail.mass_mailing.contact'].search_read(
```
yajo: Better use [`search_count()`](https://github.com/odoo/odoo/blob/1dae7ea1d660dca32e7261c5c0ed710e5541c641/openerp/models.py#L1638-L1647), which performs a DB-side count, quite faster too.

mass_mailing_partner/models/res_partner.py
```diff
@@ -35,16 +35,22 @@ def _check_email_mass_mailing_contacts(self):
...
+                    [('partner_id', '=', partner.id)], ['id']))
 
-    @api.one
+    @api.multi
     @api.depends('mass_mailing_stats')
     def _compute_mass_mailing_stats_count(self):
-        self.mass_mailing_stats_count = len(self.mass_mailing_stats)
+        for partner in self:
+            partner.mass_mailing_stats_count = len(
+                self.env['mail.mail.statistics'].search_read(
```
yajo: Use `search_count` too.

yajo (changes requested):

mass_mailing_partner/models/res_partner.py
pedrobaeza: read_group it's even better

yajo: Yes, indeed, I didn't think on it.

chienandalu:
@yajo @pedrobaeza done

yajo (approved):

yajo:
Please do not merge yet, we have to check this really fixes the performance issue.

yajo:
Congratulations @chienandalu, you have discovered one of the performance bottlenecks! Here's the proof:

Before this patch, sending a 2K email:

captura de pantalla de 2018-05-14 12-37-51

After the patch: 

captura de pantalla de 2018-05-14 12-38-03

Look at the size of the purple square!

Can we fast-track this?

pedrobaeza (approved):

pedrobaeza:
Cherry-picked for v10 in 2a94ceb8dd0c6b4c5b7c86da7dbca2b835e32d25

PR #274, opened 2018-05-16
[FIX] mass_mailing_event: Avoid double test

yajo:
Fix a typo introduced in #259 that didn't disable tests at install

@Tecnativa


pedrobaeza (approved):
Fast-tracking as well

PR #276, opened 2018-05-19
11.0 mig mass_mailing_unique

ernestotejeda:
cc @Tecnativa


rafaelbn:
cc @Tecnativa

chienandalu (commented):
Here there's a problem with the new Odoo mailing lists implementation as well. Here's the thing (ping @Yajo ):

- Two contacts with the same email address will receive just one email, so one of them is just ignored.
- Maybe we should just add a `_sql_constraints` by email.

mass_mailing_unique/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
+# Copyright 2016 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Unique records for mass mailing",
+    "summary": "Avoids duplicate mailing lists and contacts",
+    "version": "11.0.1.0.0",
+    "category": "Marketing",
+    "website": "https://tecnativa.com",
```
yajo: https://github.com/OCA/social

mass_mailing_unique/models/mass_mailing.py
```diff
@@ -0,0 +1,30 @@
...
+
+from odoo import models, api, _
+from odoo.exceptions import ValidationError
+
+
+class MailMassMailingContact(models.Model):
+    _inherit = "mail.mass_mailing.contact"
+
+    @api.constrains('partner_id', 'list_ids')
+    def _check_partner_id_lists_ids(self):
```
yajo: Just add a unique constraint per `email` and remove this method.

mass_mailing_unique/models/mass_mailing.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+from odoo import models, api, _
+from odoo.exceptions import ValidationError
+
+
+class MailMassMailingContact(models.Model):
+    _inherit = "mail.mass_mailing.contact"
+
+    @api.constrains('partner_id', 'list_ids')
```
yajo: If I'm not wrong, `partner_id` field doesn't exist.

mass_mailing_unique/hooks.py
```diff
@@ -0,0 +1,46 @@
...
+    is useless without those constraints, we check here if all will work before
+    installing, and provide a user-friendly message in case of failure.
+    """
+    errors = list()
+
+    # Search for duplicates in emails
+    cr.execute("""SELECT c.email, l.name, COUNT(c.id)
+                  FROM
+                    mail_mass_mailing_contact AS c
+                    INNER JOIN mail_mass_mailing_contact_list_rel AS cl
```
yajo: All you need to check, if you follow my suggestion below, is to check there are no duplicate emails in `mail_mass_mailing_contact`. No joins needed.

yajo (changes requested):

rafaelbn:
This PR depends on @chienandalu approval.

yajo:
> Maybe we should just add a _sql_constraints by email.

This can't be possible, because it would break `mass_mailing_custom_unsubscribe`, which is currently much more needed.

> Two contacts with the same email address will receive just one email, so one of them is just ignored.

We just checked with @rafaelbn that it sends duplicated emails, just the 2nd (and next ones) are set in exception state. Not sure why that happens, but it can be confusing.

At the end, seeing current v11 implementation, all we need here is a python constraint that avoids 2 contacts to exist for 1 list.

ernestotejeda:
the situation here is similar to 11.0 mig mass_mailing_partner https://github.com/OCA/social/pull/277 , i had comment about that

mass_mailing_unique/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,6 @@
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Jairo Llopis <jairo.llopis@tecnativa.com>
```
pedrobaeza: Remove emails

pedrobaeza (changes requested):

mass_mailing_unique/models/mass_mailing.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class MailMassMailingContact(models.Model):
+    _inherit = "mail.mass_mailing.contact"
+
+    @api.constrains('email', 'list_ids')
+    def _check_email_list_ids(self):
+        for contact in self:
+            other_contact = self.search([
+                ('email', '=', contact.email),
```
chienandalu: To avoid duplucates due to case variations:
```python
('email', '=', contact.email.lower()),
```

mass_mailing_unique/models/mass_mailing.py
```diff
@@ -0,0 +1,30 @@
...
+            ])
+            if contact.list_ids & other_contact.mapped('list_ids'):
+                raise ValidationError(_("Cannot have the same email more "
+                                        "than once in the same list"))
+
+
+class MailMassMailingList(models.Model):
+    _inherit = "mail.mass_mailing.list"
+    _sql_constraints = [
+        ("unique_name", "UNIQUE(name)",
```
chienandalu, line 28: Same problem is going to happen here with case variations. For instance:

```
- My list of contacts
- my list of contacts
- My List Of Contacts
- MY LIST OF CONTACTS
...
```
They all are going to pass over the constrain.

mass_mailing_unique/hooks.py
```diff
@@ -0,0 +1,46 @@
...
+                      ON cl.contact_id = c.id
+                    INNER JOIN mail_mass_mailing_list AS l ON cl.list_id = l.id
+                  GROUP BY l.name, c.email, c.id
+                  HAVING COUNT(c.id) > 1""")
+    for result in cr.fetchall():
+        errors.append(
+            _("{0} appears {2} times in list {1}.").format(*result))
+
+    # Search for duplicates in list's name
+    cr.execute("""SELECT name, COUNT(id)
```
chienandalu, line 34: If we want to avoid case duplicates:
```sql
SELECT LOWER(name) AS n, COUNT(id)
[...]
GROUP BY n
```

mass_mailing_unique/hooks.py
```diff
@@ -0,0 +1,46 @@
...
+
+    If you define a unique key in Odoo that cannot be applied, Odoo will log a
+    warning and install the module without that constraint. Since this module
+    is useless without those constraints, we check here if all will work before
+    installing, and provide a user-friendly message in case of failure.
+    """
+    errors = list()
+
+    # Search for duplicates in emails
+    cr.execute("""SELECT c.email, l.name, COUNT(c.id)
```
chienandalu: We should check avoid email case variations. Something like:
```sql
SELECT LOWER(c.email) AS e, l.name, COUNT(c.id)
[...]
GROUP BY l.name, e, c.id
```

chienandalu (commented):
I think case variations should be consider to avoid duplicates due to user errors. For instance, these two email:

- test@example.com
- Test@example.com

Are going to be sent to the same recipient, but is going to pass over the duplicates filter this module tries to avoid.

pedrobaeza:
@ernestotejeda please rebase and attend latest comments

mass_mailing_unique/models/mass_mailing.py
ernestotejeda: With postgre, i think you can not make an insensitive unique constraint to not distinguish between uppercase and lowercase. To solve that, we can make an UNIQUE INDEX, for example:
CREATE UNIQUE INDEX "name_unique_idx" ON "mail_mass_mailing_list" (LOWER ('name'))
But then we must catch an IntegrityError exception in the create and write method or in another place. Actually, I do not know if it's worth doing all of this for this specific case

pedrobaeza: Fortunately it's possible to add a constraint case insensitive. Just put `UNIQUE (LOWER(email))`. See for example: http://shuber.io/case-insensitive-unique-constraints-in-postgres/

ernestotejeda: In this page it says that Unfortunately, Postgres does now allow us to define unique constraint using LOWER and according to this page one of the solutions is to make a 'UNIQUE INDEX'. That's what I said above and then it can be more work

pedrobaeza: Right, I read it bad :(

OK, let's keep it that way. @chienandalu please give your final review

pedrobaeza:
@ernestotejeda next time please drop a comment when you make the rebase, because no notification gets to our inboxes if not.

pedrobaeza (approved):
@Yajo please finish your review with latest changes

yajo (approved):

PR #277, opened 2018-05-19
11.0 mig mass_mailing_partner

ernestotejeda:
cc @Tecnativa


rafaelbn:
cc @Tecnativa

chienandalu (changes requested):
Thanks for the work, @ernestotejeda !

One important thing you have to count with in v11 mass_mailing migrations is that the relation between contacts and lists has changed: so now there should be a contact for every mail address and that contact is subscribed to *n* lists.

For what this module is concerned:
- The limitation of creating a partner or not by model list may not fit quite well already. What do you think here, @Yajo ?
- You should check if a contact with a partner already exists, and add the lists to the contact ones.
- Adding a contact to a list should not create a new contact if a contact associated with that partner already exists.

mass_mailing_partner/README.rst
```diff
@@ -0,0 +1,74 @@
...
+Contributors
+------------
+
+* Pedro M. Baeza <pedro.baeza@tecnativa.com>
+* Rafael Blasco <rafael.blasco@tecnativa.com>
+* Antonio Espinosa <antonio.espinosa@tecnativa.com>
+* Javier Iniesta <javieria@antiun.com>
+* Jairo Llopis <jairo.llopis@tecnativa.com>
+* David Vidal <david.vidal@tecnativa.com>
+
```
yajo: Add yourself @ernestotejeda

mass_mailing_partner/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+# Copyright 2015-2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+# Copyright 2015 Javier Iniesta <javieria@antiun.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Link partners with mass-mailing",
+    "version": "11.0.1.0.0",
+    "author": "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    "website": "https://www.tecnativa.com",
```
yajo: https://github.com/OCA/social

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class MailMassMailingContact(models.Model):
+    _inherit = 'mail.mass_mailing.contact'
+
+    partner_id = fields.Many2one(comodel_name='res.partner', string="Partner",
+                                 domain=[('email', '!=', False)])
+
+    @api.constrains('partner_id', 'list_ids')
+    def _check_partner_id_lists_ids(self):
```
yajo: "_check_partner_id_list_ids"

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,138 @@
...
+        for contact in self:
+            if contact.partner_id:
+                other_contact = self.search([
+                    ('partner_id', '=', contact.partner_id.id),
+                    ('id', '!=', contact.id)
+                ])
+                if contact.list_ids & other_contact.mapped('list_ids'):
+                    raise ValidationError(_("Partner already exists in one of "
+                                            "these mailing lists"))
```
yajo, line 27: Isn't it easier to just put an unique constraint on the `partner_id` field? Since now a contact has several lists, there's no need to have more than 1.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,138 @@
...
+
+    @api.onchange('partner_id')
+    def _onchange_partner(self):
+        if self.partner_id:
+            self.name = self.partner_id.name
+            self.email = self.partner_id.email
+            if self.partner_id.title:
+                self.title_id = self.partner_id.title.id
```
yajo: When doing assignment with this kind of notation, it's safer to do instead:

```py
self.title_id = self.partner_id.title
```

Because the ORM will check both records belong to the same model, and the rest will work equally.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,138 @@
...
+            if self.partner_id.company_id:
+                self.company_name = self.partner_id.company_id.name
+            if self.partner_id.country_id:
+                self.country_id = self.partner_id.country_id.id
```
yajo: Also drop last `.id` here.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,138 @@
...
+            if self.partner_id.category_id:
+                self.tag_ids = [(6, 0, self.partner_id.category_id.ids)]
```
yajo: Avoid this notation if possible, for the sake of readability. Just do:

```py
self.tag_ids = self.partner_id.category_id
```

The ORM will handle the conversion.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,138 @@
...
+            vals = self._set_partner(vals)
+        vals = self._set_name_email(vals)
+        return super(MailMassMailingContact, self).create(vals)
+
+    def write(self, vals):
+        for contact in self:
+            if vals.get('partner_id', None) is False:
+                # If removing partner, search again by email
+                vals = contact._set_partner(vals)
+            vals = contact._set_name_email(vals)
```
yajo: This is wrong. You can't do this in a loop. At the end, `vals` will contain values for the last loop iteration and apply them to all records.

Can't you just call `_onchange_partner()` after writing or creating a record and drop all methods below?

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,138 @@
...
+        return super(MailMassMailingContact, self).write(vals)
+
+    def _prepare_partner(self, vals, list_ids):
+        partner_vals = {
+            'name': vals.get('name', self.name) or vals.get('email',
+                                                            self.name),
+            'email': vals.get('email', self.email),
+            'country_id': vals.get('country_id',
+                                   self.country_id and self.country_id.id),
```
yajo: `self.country_id.id` will be `False` if `len(self.country_id) == 0`, so no need to to this kind of trick.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,138 @@
...
+            'title': vals.get('title_id',
+                              self.title_id and self.title_id.id),
```
yajo: Same here.

mass_mailing_partner/models/res_partner.py
```diff
@@ -0,0 +1,87 @@
...
+    mass_mailing_contact_ids = fields.One2many(
+        string="Mailing contacts",
+        oldname="mass_mailing_contacts",
+        domain=[('opt_out', '=', False)],
+        comodel_name='mail.mass_mailing.contact', inverse_name='partner_id')
+    mass_mailing_contacts_count = fields.Integer(
+        string='Mailing contacts number',
+        compute='_compute_mass_mailing_contacts_count', store=True,
+        compute_sudo=True)
+    mass_mailing_stats = fields.One2many(
```
yajo: `mass_mailing_stat_ids`

mass_mailing_partner/models/res_partner.py
```diff
@@ -0,0 +1,87 @@
...
+        string="Mass mailing stats",
+        comodel_name='mail.mail.statistics', inverse_name='partner_id')
+    mass_mailing_stats_count = fields.Integer(
+        string='Mass mailing stats number',
+        compute='_compute_mass_mailing_stats_count', store=True)
+
+    @api.constrains('email')
+    def _check_email_mass_mailing_contacts(self):
+        for partner in self:
```
yajo, line 32: Try just:

```py
if self.sudo().filtered("mass_mailing_contact_ids.list_ids"):
    raise ...
```

mass_mailing_partner/models/res_partner.py
```diff
@@ -0,0 +1,87 @@
...
+            if not partner.email:
+                mailing_contacts = partner.sudo().mass_mailing_contact_ids
+                for mailing_contact in mailing_contacts:
+                    if len(mailing_contact.list_ids):
+                        raise ValidationError(_(
+                            "This partner '%s' is subscribed to one or more "
+                            "mailing lists. Email must be assigned."
+                        ) % partner.name)
+
+    @api.multi
```
yajo: It's implicit, remove it.

mass_mailing_partner/models/res_partner.py
```diff
@@ -0,0 +1,87 @@
...
+        contact_data = self.env['mail.mass_mailing.contact'].read_group(
+            [('partner_id', 'in', self.ids)], ['partner_id'], ['partner_id'])
+        mapped_data = dict(
+            [(contact['partner_id'][0], contact['partner_id_count'])
+             for contact in contact_data])
+        for partner in self:
+            partner.mass_mailing_contacts_count = mapped_data.get(partner.id,
+                                                                  0)
+
+    @api.multi
```
yajo: :fire:

mass_mailing_partner/models/res_partner.py
```diff
@@ -0,0 +1,87 @@
...
+            mm_vals['company_name'] = company.name
+        if vals.get('country_id'):
+            mm_vals['country_id'] = vals['country_id']
+        if vals.get('category_id'):
+            mm_vals['tag_ids'] = vals['category_id']
+        if mm_vals:
+            # Using sudo because ACLs shouldn't produce data inconsistency
+            self.env["mail.mass_mailing.contact"].sudo().search([
+                ("partner_id", "in", self.ids),
+            ]).write(mm_vals)
```
yajo, line 81: Can't you replace all of this by calling `self.mass_mailing_contact_ids._onchange_partner()` after writing?

mass_mailing_partner/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo import fields, models
+
+
+class MailMassMailingList(models.Model):
+    _inherit = 'mail.mass_mailing.list'
+
+    partner_mandatory = fields.Boolean(string="Mandatory Partner",
+                                       default=False)
+    partner_category = fields.Many2one(comodel_name='res.partner.category',
+                                       string="Partner Tag")
```
yajo, line 15: Provide help on this field. Its purpose is unclear.

yajo (changes requested):
Some complex old code is making it harder to migrate and review, so I added mostly simplification suggestions.

rafaelbn (approved):
Tested in runbot functionally :+1: 
Please @ernestotejeda attend @Yajo and @chienandalu comments. Thanks!

ernestotejeda:
It's true, now a contact has several lists, but, if we put an unique constraint on the partner_id field, then when a contact is unsubscribed, he/she will not receive any mass-mail sent to any list asociated with him/her. But with the python constraint, we can have an unsubscribed contact 'x' with a partner asociated to a sub-set of lists 'SS1', and another contact 'y' with the same partner asociated with other sub-set of lists 'SS2' but not unsubscribed. So, the email asociated with this contacts 'x' and 'y' can receive mass mailing that will be send to 'SS2' and not receive mass mailing that will be send to 'SS1'.

mass_mailing_partner/models/mail_mass_mailing_contact.py
pedrobaeza: With that, you are limiting a lot the possibilities, so I prefer to not do it.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -14,68 +14,82 @@ class MailMassMailingContact(models.Model):
...
+                other_contact = self.search([
+                    ('partner_id', '=', contact.partner_id.id),
+                    ('id', '!=', contact.id)
+                ])
+                if contact.list_ids & other_contact.mapped('list_ids'):
+                    raise ValidationError(_("Partner already exists in one of "
+                                            "these mailing lists"))
 
     @api.onchange('partner_id')
     def _onchange_partner(self):
```
pedrobaeza: Rename this to `_onchange_partner_id_mass_mailing_partner` for having an unique method name and avoid to choke with other modules.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -14,68 +14,82 @@ class MailMassMailingContact(models.Model):
...
-    @api.returns('self', lambda x: x.id)
     def create(self, vals):
-        if not vals.get('partner_id'):
-            vals = self._set_partner(vals)
-        vals = self._set_name_email(vals)
-        return super(MailMassMailingContact, self).create(vals)
+        if not vals.get('email') and vals.get('partner_id'):
+            vals['email'] = vals['partner_id']
+        result = super(MailMassMailingContact, self).create(vals)
+        if not result.partner_id:
```
pedrobaeza: Please don't perform a create and a write. You should do it adding to vals the corresponding result. What you can do is:
```python
def create(self, vals):
    if not vals.get('partner_id'):
        record = self.new({
            'email': vals.get('email'),
            'name': vals.get('name'),
        })
        record._set_partner(vals)
        record._onchange_partner()
        new_vals = record._convert_to_write(record._cache)
        vals.update(new_vals)
```

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -14,68 +14,82 @@ class MailMassMailingContact(models.Model):
...
 
     def write(self, vals):
+        result = super(MailMassMailingContact, self).write(vals)
         for contact in self:
-            if vals.get('partner_id', None) is False:
-                # If removing partner, search again by email
-                vals = contact._set_partner(vals)
-            vals = contact._set_name_email(vals)
-        return super(MailMassMailingContact, self).write(vals)
+            if not contact.partner_id:
```
pedrobaeza: Perform something similar to create here.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -14,68 +14,82 @@ class MailMassMailingContact(models.Model):
...
-
-    def _set_name_email(self, vals):
-        partner_id = vals.get('partner_id', self.partner_id.id)
-        if not partner_id:
-            return vals
-        partner = self.env['res.partner'].browse(partner_id)
-        vals['email'] = partner.email
-        vals['name'] = partner.name
-        return vals
+            self.partner_id = m_partner.sudo().create({
```
pedrobaeza: Don't remove the `_prepare_...`  method for having a hook on things.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -14,68 +14,82 @@ class MailMassMailingContact(models.Model):
...
                                  domain=[('email', '!=', False)])
 
-    _sql_constraints = [
-        ('partner_list_uniq', 'unique(partner_id, list_id)',
-         _('Partner already exists in this mailing list.'))
-    ]
+    @api.constrains('partner_id', 'list_ids')
+    def _check_partner_id_list_ids(self):
+        for contact in self:
+            if contact.partner_id:
```
pedrobaeza: I'm thinking that it's best to not limiting this, as it will lead to constant problems

mass_mailing_partner/models/res_partner.py
pedrobaeza: I'm thinking that it's best to not limiting this, as it will lead to constant problems

pedrobaeza (changes requested):
I have squashed all the v11 history for better reviewing, but there are some comments that still apply.

mass_mailing_partner/models/res_partner.py
ernestotejeda: email field of contact is required and it depend on the email of the partner associated. So, i think, we should not permit that someone delete the email of the partner that has some contact associated.

mass_mailing_partner/models/mail_mass_mailing_contact.py
ernestotejeda: I think the same as @pedrobaeza , besides, in this way it is guaranteed that a partner is not repeated in a list and at the same time I do not limit things so much

ernestotejeda:
It doesn't pass the tests because the module mass_mailing_custom_unsubscribe throws an error, I have done a PR https://github.com/OCA/social/pull/287 that solves the problem. Can you please check it?

pedrobaeza:
I have rebased and squashed several commits. @Yajo please review again to see if we unblock this.

mass_mailing_partner/models/res_partner.py
```diff
@@ -31,12 +30,12 @@ class ResPartner(models.Model):
...
     def _check_email_mass_mailing_contacts(self):
         for partner in self:
-            if partner.sudo().mass_mailing_contact_ids and not partner.email:
-                raise ValidationError(
-                    _("This partner '%s' is subscribed to one or more "
-                      "mailing lists. Email must be assigned.") % partner.name)
+            if not partner.email and self.sudo().mass_mailing_contact_ids:
+                raise ValidationError(_(
+                    "This partner '%s' is linked to one or more mass "
+                    "mailing contact. Email must be assigned."
```
yajo: I'm thinking that it should discriminate when mass mailings are in queue. Otherwise, there would be no problem in removing the email, right?

mass_mailing_partner/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
+This module links mass-mailing contacts with partners.
+
+Features
+--------
```
yajo: Isn't this gonna crash?

yajo (approved):
It looks good. Some doubts, but since they are theories, I think it's not worth it blocking this for them.

mass_mailing_partner/readme/DESCRIPTION.rst
pedrobaeza: Yeah, good catch. The only allowed subtitle is `~~~~~~`, and I'm not sure we should keep it.

mass_mailing_partner/models/res_partner.py
chienandalu: But the field email in the mass_mailing contact model is still required. Setting the linked partner email to blank would raise an exception.

pedrobaeza (approved):
As there are too much changes difficult to see if there's going to be problems, I'm merging this and we'll try on several DBs

PR #267, opened 2018-04-27
[IMP] mass_mailing_custom_unsubscribe: GDPR compliance

yajo:
- Record resubscriptions too.
- Record action metadata.

@Tecnativa

PR #278, opened 2018-05-24
[IMP] mass_mailing_custom_unsubscribe: GDPR compliance

chienandalu:
Implement #267 in `9.0`

* [IMP] mass_mailing_custom_unsubscribe: GDPR compliance

- Record resubscriptions too.
- Record action metadata.
- Make ESLint happy.
- Quick color-based action distinction in tree view.
- Add useful quick groupings.
- Display (un)subscription metadata.
- Pivot & graph views.

cc @Tecnativa

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,10 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
-  - TESTS="1" ODOO_REPO="OCA/OCB"
+  - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="mass_mailing_custom_unsubscribe"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="mass_mailing_custom_unsubscribe"
+  - TESTS="1" ODOO_REPO="odoo/odoo"INCLUDE="mass_mailing_custom_unsubscribe"
+  - TESTS="1" ODOO_REPO="OCA/OCB" INCLUDE="mass_mailing_custom_unsubscribe"
 
 virtualenv:
   system_site_packages: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,8 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="mass_mailing_custom_unsubscribe"
-  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="mass_mailing_custom_unsubscribe"
-  - TESTS="1" ODOO_REPO="odoo/odoo"INCLUDE="mass_mailing_custom_unsubscribe"
-  - TESTS="1" ODOO_REPO="OCA/OCB" INCLUDE="mass_mailing_custom_unsubscribe"
+  - TESTS="1" ODOO_REPO="odoo/odoo"
+  - TESTS="1" ODOO_REPO="OCA/OCB""
 
 virtualenv:
   system_site_packages: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ env:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
   - TESTS="1" ODOO_REPO="odoo/odoo"
-  - TESTS="1" ODOO_REPO="OCA/OCB""
+  - TESTS="1" ODOO_REPO="OCA/OCB"
 
 virtualenv:
   system_site_packages: true
--- a/mass_mailing_custom_unsubscribe/controllers/main.py
+++ b/mass_mailing_custom_unsubscribe/controllers/main.py
@@ -108,6 +108,7 @@ class CustomUnsubscribe(MassMailController):
         if details:
             extra_context["default_details"] = details
         request.context = dict(request.context, **extra_context)
+        del request.env
         # FIXME Remove token check in version where this is merged:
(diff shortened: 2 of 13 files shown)


mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
```diff
@@ -46,6 +66,15 @@ def _selection_unsubscriber_id(self):
...
         return self.env["mail.mass_mailing"]._get_mailing_model()
 
+    @api.multi
+    @api.constrains("action", "reason_id")
+    def _check_reason_needed(self):
+        """Ensure reason is given for unsubscriptions."""
+        for one in self:
+            if one.action == "unsubscription" and not one.reason_id:
+                raise exceptions.ReasonRequiredError(
+                    _("Please indicate why are you unsubscribing."))
```
yajo, line 76: The tour fails because of an [error 500](https://travis-ci.org/OCA/social/jobs/383253541#L1719). [This seems to be the cause](https://travis-ci.org/OCA/social/jobs/383253541#L1568).

Most likely, you need to alter the tour, so it indicates the unsubscription reason.

Or maybe the error pops up in a different form, since this shouldn't be printing an error 500 in the 1st place.

yajo (changes requested):

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
chienandalu: Grrr, I can't reproduce in my local even with the whole repo CI :confused:

pedrobaeza:
@Yajo please help in this.

yajo:
You still get an error raised to the tour, resulting in a 500 error page :confused: 
https://travis-ci.org/OCA/social/jobs/385122118#L1101-L1171

chienandalu:
@Yajo Seems that the context with the reason_id is lost from here: https://github.com/odoo/odoo/blob/9.0/addons/mass_mailing/controllers/main.py#L24 to here: https://github.com/Tecnativa/social/blob/091e7147f700e194078efce8f64983847bc61f64/mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py#L56

.travis.yml
```diff
@@ -20,8 +20,10 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
-  - TESTS="1" ODOO_REPO="OCA/OCB"
+  - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="mass_mailing_custom_unsubscribe"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="mass_mailing_custom_unsubscribe"
+  - TESTS="1" ODOO_REPO="odoo/odoo"INCLUDE="mass_mailing_custom_unsubscribe"
+  - TESTS="1" ODOO_REPO="OCA/OCB" INCLUDE="mass_mailing_custom_unsubscribe"
```
yajo: Remove this change. It wasn't the problem.

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -75,11 +77,11 @@ def mailing(self, mailing_id, email=None, res_id=None, token="", **post):
...
-            request.context.update({
-                "default_reason_id": reason_id,
-                "default_details": post.get("details") or False,
-            })
-            del request.env
+            request.context = dict(
+                request.context,
+                default_reason_id=reason_id,
+                default_details=post.get("details") or False,
+            )
```
yajo: After changing `request.context`, do a `del request.env`.

yajo (changes requested):
You nailed it!

Indeed, I was forgetting that [v9 lacks the env autoreset when context changes](https://github.com/odoo/odoo/blob/f349d2f3abac89f86b0ac6c32addbfec8f689564/odoo/http.py#L244-L247).

chienandalu:
@Yajo Two problems I'm getting into:

- If the phantomjs doesn't run fast enough, it fails (I'd say that's the why travis fails)
- Debuggin my local tests, I'm having problems with this assertion: https://github.com/OCA/social/pull/278/files#diff-2418228367b122233471c3a1ff67d331R97 The phantomjs gets to write it but when it's finished the value isn't stored.

chienandalu:
To note also, that functional tests are working as far as I could try

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -90,9 +93,21 @@ def unsubscribe(self, mailing_id, opt_in_ids, opt_out_ids, email, res_id,
...
+                )
+            ),
+        }
         if reason_id:
-            request.context["default_reason_id"] = int(reason_id)
-            request.context["default_details"] = details or False
+            extra_context["default_reason_id"] = int(reason_id)
+        if details:
+            extra_context["default_details"] = details
+        request.context = dict(request.context, **extra_context)
```
yajo: After changing `request.context`, do a `del request.env`.

yajo (changes requested):
I detected one failure, let's see if that's the :x: cause.

Also, [you can add timeout to tests](https://github.com/odoo/odoo/blob/9.0/openerp/tests/common.py#L411) if you need it.

rafaelbn (approved):
Hi @chienandalu @Yajo , Travis is still RED but functionally tested :+1: in my side after getting travis green we can merge safety

chienandalu:
Yes there's still a problem with the tour I can't reproduce manually :confused:

rafaelbn:
@Yajo what can we do with travis :red_circle: here? could you check it? Functionally this works! and could be merge!

yajo (approved):

PR #283, opened 2018-06-12
[11.0][mail_activity_board] Add new module that insert activities board in boards.

dajuayen:
This module installs the Document module and generates a board with the form, tree, kanban, graph, pivot and calendar views.

It allows to manage all the activities from the same place even though their origin is of different types of objects.

The form view shows a few data of the activity, but it allows to navigate to the object with which it is related.

In the form view of the related object, in the chatter a button has been added that links to the activity board in which the activities of that object are shown.


mail_activity_board/README.rst
```diff
@@ -0,0 +1,72 @@
...
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+
+
+================
+Activities Board
+================
+
+This module adds an activity board with form, tree, kanban, calendar, pivot, graph and search views.
```
yajo: Move this paragraph to `readme/DESCRIPTION.rst`

mail_activity_board/README.rst
```diff
@@ -0,0 +1,72 @@
...
+
+
+Configuration
+=============
+
+
+Install it from APPs.
```
yajo: This is implicit, remove it :fire:

mail_activity_board/README.rst
```diff
@@ -0,0 +1,72 @@
...
+Usage
+=====
+
+Access to the views from menu Boards.
+
+A smartButton of activities is added in the opportunity form view.
+From this smartButton is linked to the activity board, to the view tree,
+which shows the activities related to the opportunity.
+
+From the form view of the activity you can navigate to the origin of the activity.
```
yajo: Move to `readm/USAGE.rst`

mail_activity_board/README.rst
```diff
@@ -0,0 +1,72 @@
...
+=======
+
+
+Contributors
+------------
+
+* SDI <http://www.sdi.es>
+
+  * David Juaneda Ayensa <djuaneda@sdi.es>
+  * Javier Garcia <jgarcia@sdi.es>
```
yajo: Move to `readme/CONTRIBUTORS.rst`

mail_activity_board/__init__.py
```diff
@@ -0,0 +1,5 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# SDI
+# author: djuaneda@sdi.es
```
yajo: I'd remove authors from such a tiny file.

mail_activity_board/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2018 SDi - David Juaneda
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+{
+    'name': 'Activities board',
+    'version': '11.0.1.0.1',
```
yajo: `11.0.1.0.0`

mail_activity_board/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+        'crm',
+        'mail',
+        'board',
+    ],
+    'data': [
+        'views/mail_activity_view.xml',
+        'views/crm_lead_opportunity_view.xml',
+    ],
+    'qweb': [],
+    'demo': [],
```
yajo: :fire:

mail_activity_board/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+{
+    'name': 'Activities board',
+    'version': '11.0.1.0.1',
+    'category': 'Social Network',
+    'author': 'David Juaneda, '
+              'Javier Garcia, '
+              'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/social',
+    'summary': 'Add Activity Boards',
+    'license': 'GPL-3',
```
yajo: AGPL-3

mail_activity_board/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2018 SDi - David Juaneda
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+{
+    'name': 'Activities board',
+    'version': '11.0.1.0.1',
+    'category': 'Social Network',
```
yajo: :thinking: Doesn't seem much of social, isn't it? Is there another better category for this?

mail_activity_board/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# SDI
+# author: djuaneda@sdi.es
```
yajo: :fire:

mail_activity_board/models/crm_lead.py
```diff
@@ -0,0 +1,36 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# SDI
+# author: djuaneda@sdi.es
```
yajo: :fire:

mail_activity_board/models/crm_lead.py
```diff
@@ -0,0 +1,36 @@
...
+from odoo import api, fields, models, _
+
+
+class Lead(models.Model):
+    _inherit = 'crm.lead'
```
yajo: Shouldn't you extend `mail.activity.mixin` directly instead? So all related models inherit this... Or is it only for leads? :thinking:

mail_activity_board/models/crm_lead.py
```diff
@@ -0,0 +1,36 @@
...
+
+    opportunity_activities_count = fields.Integer(
+        "Activities", compute='_compute_activities_count')
+
+    @api.multi
+    def _compute_activities_count(self):
+        """
+        Calculates the number of activities related to the opportunity
+        """
```
yajo: `"""Calculates the number of activities related to the opportunity"""`, in one line please :blush:

mail_activity_board/models/crm_lead.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class Lead(models.Model):
+    _inherit = 'crm.lead'
+
+    opportunity_activities_count = fields.Integer(
+        "Activities", compute='_compute_activities_count')
+
+    @api.multi
+    def _compute_activities_count(self):
```
yajo: Add the corresponding `@api.depends`

mail_activity_board/models/crm_lead.py
```diff
@@ -0,0 +1,36 @@
...
+from odoo import api, fields, models, _
+
+
+class Lead(models.Model):
+    _inherit = 'crm.lead'
+
+    opportunity_activities_count = fields.Integer(
+        "Activities", compute='_compute_activities_count')
+
+    @api.multi
```
yajo: :fire: , implicit

mail_activity_board/models/crm_lead.py
```diff
@@ -0,0 +1,36 @@
...
+                opp.opportunity_activities_count += self.env['mail.activity']\
+                    .search_count([('res_model', '=', 'crm.lead'),
+                                   ('res_id', '=', opp.id),
+                                   ('active', '=', False)])
+
+    def redirect_to_lead_actvivities(self):
+        """
+            Method that redirects to the list of opportunity activities.
+            :return: action.
+        """
```
yajo: Check https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings on how to format proper docstrings :blush:

mail_activity_board/models/mail_activity.py
```diff
@@ -0,0 +1,49 @@
...
+    _inherit = "mail.activity"
+
+    res_model_id_name = fields.Char(
+        default="", related='res_model_id.name',
+        string="Origin", store=False, readonly=True)
+    duration = fields.Float(
+        related='calendar_event_id.duration',
+        string="Duration", store=False, readonly=True)
+    calendar_event_id_start = fields.Datetime(
+        default=False, related='calendar_event_id.start',
```
yajo: Add a string for humans

mail_activity_board/models/mail_activity.py
```diff
@@ -0,0 +1,49 @@
...
+        store=False, readonly=True)
+    calendar_event_id_partner_ids = fields.Many2many(
+        related='calendar_event_id.partner_ids', string='Attendees',
+        store=False, default=False, readonly=True)
+
+    @api.multi
+    def open_origin(self):
+        """
+        Utility method used to add an "Open Company" button in partner views.
+        """
```
yajo: https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

mail_activity_board/models/mail_activity.py
```diff
@@ -0,0 +1,49 @@
...
+            if res_type == 'opportunity':
+                view_id = self.env.ref("crm.crm_case_form_view_oppor").id
+                response['view_id'] = view_id
+        return response
+
+    @api.model
+    def action_your_activities(self):
+        action = self.env.ref(
+            'mail_activity_board.open_boards_activities_form_tree').read()[0]
+        action_context = safe_eval(action['context'], {'uid': self.env.uid})
```
yajo: Hmm, this is weird. Why do you need the `uid`? That's something always available from `self.env.user.id`... :thinking:

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+<!--
+    VIEWS
+-->
+
+
+    <!-- FORM VIEW -->
+    <record id="mail_activity_view_form_board" model="ir.ui.view">
+        <field name="name">mail.activity.form.boards.sdi</field>
```
yajo: sdi? :smirk:

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+                                nolabel="1"/>
+                        </div>
+                    </group>
+                    <group>
+                        <group>
+                            <field name="activity_type_id" required="1" options="{'no_create': True, 'no_open': True}"/>
+                            <field name="res_model_id_name"/>
+
+                            <button name="open_origin" type="object" class="centre oe_link" nolabel="1">
+                                <field name="res_name"><field name="res_model_id_name"/></field>
```
yajo: field inside field? :thinking:

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+
+                            <field name="calendar_event_id" invisible="1"/>
+                            <field name="create_date" invisible="1"/>
+                            <field name="summary" placeholder="e.g. Discuss proposal"/>
+                        </group>
+                        <group>
+                            <field name="date_deadline"
+                                   attrs="{'invisible': [('activity_category', '=', 'meeting')]}"/>
+                            <field name="calendar_event_id_start" string="Start meeting"
+                                   attrs="{'invisible': [('calendar_event_id','=', False)], 'readonly': 1}"/>
```
yajo: `1` is not a domain. Instead, add directly the XML attribute `readonly="1"`

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+                            <field name="duration" widget="float_time"
+                                   attrs="{'invisible': ['|',('duration', '=', False),
+                                   ('calendar_event_id','=', False)], 'readonly': 1}"/>
```
yajo: Same here.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+                            <field name="user_id"/>
+                            <field name="calendar_event_id_partner_ids" widget="many2many_tags"
+                                   attrs="{'invisible': [('calendar_event_id','=', False)], 'readonly': 1}"/>
```
yajo: Same here.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+    <!-- SEARCH VIEW -->
+    <record id="mail_activity_view_search" model="ir.ui.view">
+        <field name="name">search.activity.view.calendar.inherit.activities_board</field>
+        <field name="model">mail.activity</field>
+        <field name="inherit_id" ref="mail.mail_activity_view_search"/>
+        <field name="mode">primary</field>
+        <field name="arch" type="xml">
+            <xpath expr='//field[@name="res_model_id"]' position='before'>
+                <field name="user_id"/>
+                <field name="res_name" string="Origin (opportunity/customer)"/>
```
yajo: I don't think that this is a good place to hardcode origin possibilities. You can add a `help` attribute on the field instead, if you need to provide examples on the field meaning.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+            </xpath>
+
+            <xpath expr='//filter[@name="activities_my"]' position='after'>
+                <filter string="Act. next month" name="activities_month"
+                        domain="[('date_deadline', '&lt;', (context_today()+datetime.timedelta(days=30)).strftime('%Y-%m-%d'))]"
+                        help="Show activities which date deadline is before than 1 month."/>
```
yajo: Show activities scheduled for next month.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+                <filter string="Act. next 6 months" name="activities_6_month"
+                        domain="[('date_deadline', '&lt;', (context_today()+datetime.timedelta(days=180)).strftime('%Y-%m-%d'))]"
+                        help="Show activities which date deadline is before than 6 months."/>
```
yajo: Show activities scheduled for next 6 months.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+                <separator/>
+            </xpath>
+
+            <xpath expr='//search/group' position='inside'>
```
yajo: It would be geat to have a groupby for past, today and future activities (I'm not sure if it exists already)

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+        </field>
+    </record>
+
+
+<!--
+    ACTION
+-->
+
+    <record model="ir.actions.act_window" id="open_boards_activities_form_tree">
+        <field name="name">ACTIVITIES</field>
```
yajo: DO NOT CRY :laughing: 

"Activities"

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+        <field name="res_model">mail.activity</field>
+        <field name="view_type">form</field>
+        <field name="view_mode">tree,form</field>
```
yajo: I think it's better to default to the kanban view.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+                <field name="res_name"/>
+                <field name="res_model"/>
+                <field name="summary"/>
+                <field name="date_deadline"/>
+                <field name="state"/>
+                <field name="icon"/>
+                <field name="activity_type_id"/>
+                <field name="activity_category"/>
+                <templates>
+                    <t t-name="kanban-box">
```
yajo, line 91: This whole template needs a little bit of :heart:. Check out the kanban views in crm or project modules, to inspire yourself.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,216 @@
...
+
+<!--
+Action
+-->
+
+    <record id="action_your_activities" model="ir.actions.server">
+        <field name="name">Boards: My Activities</field>
+        <field name="model_id" ref="model_mail_activity"/>
+        <field name="state">code</field>
+        <field name="code">action = model.action_your_activities()</field>
```
yajo: I think it's generally better the other way around: declare the activity in XML and then load it from code if ever needed (you'll find most times you don't need it).

yajo (changes requested):

mail_activity_board/__manifest__.py
dajuayen: I have corrected this and for the moment I have set Uncategorized.
I accept suggestions.

mail_activity_board/views/mail_activity_view.xml
dajuayen: There are filters for late, today and future activities but if you think it might be interesting to add the groups, it's not hard for me to do it.

mail_activity_board/views/mail_activity_view.xml
dajuayen: I will try to overcome it

mail_activity_board/views/mail_activity_view.xml
dajuayen: There are fields that I only want to show depending on the type of activity.
If not, how do I do it?

If only I put readonly, they will always show and there are fields like Duration that are only of type reunion.

This resource is not mine, I saw it in the code and reused it. Example:
<field name = "paperformat_id"
                                        attrs = "{'invisible': [('report_type', 'not in', ['qweb-pdf'])]}" />

mail_activity_board/__manifest__.py
dajuayen: I have modified the version to 'version': '11 .0.1.0.0 ', as you said to me.

mail_activity_board/__init__.py
dajuayen: I deleted all comments in the \__init__.py files.

mail_activity_board/models/__init__.py
dajuayen: I deleted all comments in the \__init__.py files.

mail_activity_board/README.rst
dajuayen: I have added the readme folder with the file CONTRIBUTORS.rst

mail_activity_board/README.rst
dajuayen: I have added the readme folder with the file USAGE.rst

mail_activity_board/README.rst
dajuayen: Removed

mail_activity_board/README.rst
dajuayen: I have added the readme folder with the file DESCRIPTION.rst

mail_activity_board/__manifest__.py
dajuayen: Modified. Lapsus linguae

mail_activity_board/__manifest__.py
dajuayen: Removed.

mail_activity_board/models/crm_lead.py
dajuayen: File deleted:
I have moved the functionality to another class.

mail_activity_board/models/crm_lead.py
dajuayen: The functionality add to mail.activity.mixin class.

mail_activity_board/models/crm_lead.py
dajuayen: Modified.

mail_activity_board/models/crm_lead.py
dajuayen: Thanks for the link.

mail_activity_board/models/mail_activity.py
dajuayen: Added.

mail_activity_board/models/mail_activity.py
dajuayen: Thanks again for the link.

mail_activity_board/views/mail_activity_view.xml
dajuayen: Modified.

mail_activity_board/views/mail_activity_view.xml
dajuayen: I can assure you that I have put a lot of love on this kanban card.

mail_activity_board/models/crm_lead.py
dajuayen: Added

mail_activity_board/models/mail_activity.py
dajuayen: I had added the user's id to the context because in other views I saw that they did it.
I understood that maybe it was necessary to filter or group by the user or for the domain to be only the instances related to the user.

But right now I have eliminated it.

mail_activity_board/views/mail_activity_view.xml
dajuayen: True ! I already removed it.

mail_activity_board/views/mail_activity_view.xml
dajuayen: I've already corrected.

mail_activity_board/views/mail_activity_view.xml
dajuayen: I have also corrected it.

mail_activity_board/views/mail_activity_view.xml
dajuayen: I've already modified it.

mail_activity_board/views/mail_activity_view.xml
dajuayen: I've already corrected it.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -14,7 +14,7 @@
             <form string="Activity Form" create="false" edit="false" delete="false">
                 <sheet string="Activity">
                     <button name="open_origin" type="object" class="centre oe_link" nolabel="1">
-                        <field name="res_name"><field name="res_model_id_name"/></field>
+                        <h1><field name="res_name"/></h1>
```
dajuayen: Similar to the titles of other forms.

mail_activity_board/models/crm_lead.py
dajuayen: Deleted.

dajuayen:
@gurneyalex @Yajo I have managed to pass the tests of travis but I fail to run on runbot.
What is wrong? As much as I searched, I have not been able to find information.
How can I see the logs?

mail_activity_board/views/mail_activity_view.xml
yajo: I'm talking about the `'readonly': 1` part at the end of this line, are we talking about the same thing? :thinking:

mail_activity_board/__manifest__.py
```diff
@@ -1,25 +1,25 @@
...
-    'category': 'Social Network',
-    'author': 'David Juaneda, '
-              'Javier Garcia, '
-              'Odoo Community Association (OCA)',
-    'website': 'https://github.com/OCA/social',
     'summary': 'Add Activity Boards',
-    'license': 'GPL-3',
+    'version': '11.0.1.0.0',
+    'development_status': 'Beta',
+    'category': 'Uncategorized',
```
yajo: Social Network was a good category IMHO.

mail_activity_board/models/mail_activity.py
```diff
@@ -1,31 +1,27 @@
...
 
 
 class MailActivity(models.Model):
     _inherit = "mail.activity"
 
     res_model_id_name = fields.Char(
-        default="", related='res_model_id.name',
-        string="Origin", store=False, readonly=True)
+        related='res_model_id.name', string="Origin",
+        store=False, default="", readonly=True)
```
yajo: `store=False` is the default, you can remove it.

Also, `default=""` can lead to confusing behavior, because in Odoo, the standard is that empty fields return `False` always. Having `""` instead of `False` is unconsistent, so please remove that too.

mail_activity_board/models/mail_activity.py
```diff
@@ -1,31 +1,27 @@
...
     duration = fields.Float(
-        related='calendar_event_id.duration',
-        string="Duration", store=False, readonly=True)
-    calendar_event_id_start = fields.Datetime(
-        default=False, related='calendar_event_id.start',
+        related='calendar_event_id.duration', string="Duration",
         store=False, readonly=True)
```
yajo: Most likely, with `related='calendar_event_id.duration', readonly=True` it should be the same, because all other attributes are inherited from the related field itself.

mail_activity_board/models/mail_activity.py
```diff
@@ -1,31 +1,27 @@
...
+    calendar_event_id_start = fields.Datetime(
+        related='calendar_event_id.start', string="Start",
+        store=False, default=False, readonly=True)
```
yajo: Same here.

mail_activity_board/models/mail_activity.py
```diff
@@ -1,31 +1,27 @@
...
     calendar_event_id_partner_ids = fields.Many2many(
         related='calendar_event_id.partner_ids', string='Attendees',
         store=False, default=False, readonly=True)
```
yajo: Same here

mail_activity_board/models/mail_activity.py
```diff
@@ -1,31 +1,27 @@
...
 
     @api.multi
     def open_origin(self):
-        """
-        Utility method used to add an "Open Company" button in partner views.
-        """
+        """  """
```
yajo: :fire:

mail_activity_board/models/mail_activity.py
```diff
@@ -41,9 +37,7 @@ def open_origin(self):
         return response
 
     @api.model
-    def action_your_activities(self):
+    def action_activities_board(self):
         action = self.env.ref(
-            'mail_activity_board.open_boards_activities_form_tree').read()[0]
-        action_context = safe_eval(action['context'], {'uid': self.env.uid})
-        action['context'] = action_context
+            'mail_activity_board.open_boards_activities').read()[0]
```
yajo: just do `self.env.ref(...).id`

mail_activity_board/models/mail_activity_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import models
+
+
+class MailActivityMixin(models.AbstractModel):
+    _inherit = 'mail.activity.mixin'
+
+    def redirect_to_activities(self, **kwargs):
+        """
+            Redirects to the list of activities of the object shown.
```
yajo: Put this line along with the `"""`

mail_activity_board/models/mail_activity_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+            Redirects to the activity board and configures the domain so that
+            only those activities that are related to the object shown are
+            displayed.
+
+            Add to the title of the view the name the class of the object from
+            which the activities will be displayed.
+
+            :param kwargs: Dictionary that contains the id of the object
+                            and the model it is about.
+            :return: action.
```
yajo: Reduce 4 spaces of indentation in all this docstring.

FTR https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

mail_activity_board/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module adds an activity board with form, tree, kanban, calendar, pivot, graph and search views.
+
```
yajo, line 2: :fire:

mail_activity_board/readme/HISTORY.rst
```diff
@@ -0,0 +1,11 @@
+11.0.0.0.0 (2018-06-12)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* [BREAKING] Breaking changes come first.
+  (`#70 <https://github.com/OCA/repo/issues/70>`_)
+* [ADD] New feature.
+  (`#74 <https://github.com/OCA/repo/issues/74>`_)
+* [FIX] Correct this.
+  (`#71 <https://github.com/OCA/repo/issues/71>`_)
```
yajo: :fire:

mail_activity_board/static/src/js/override_chatter.js
```diff
@@ -0,0 +1,34 @@
...
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+
+odoo.define('mail.Chatter.activity', function(require){
+    "use strict";
+
+    var chatter = require('mail.Chatter');
+
+    chatter.include({
+
+        events: {
```
yajo: I hope you're not overriding parent events. To protect against that, you should do:

```js
events: _.extend({}, chatter.prototype.events, {
    // your events here...
}),
```

JS is funny :yum:

mail_activity_board/static/src/js/override_chatter.js
```diff
@@ -0,0 +1,34 @@
...
+            'click .o_chatter_button_new_message': '_onOpenComposerMessage',
+            'click .o_chatter_button_log_note': '_onOpenComposerNote',
+            'click .o_chatter_button_schedule_activity': '_onScheduleActivity',
+            'click .o_chatter_button_count_activity': '_onCountActivity',
+        },
+
+        _onCountActivity: function (event) {
+            event.preventDefault();
+            var self = this;
```
yajo: You can remove this...

mail_activity_board/static/src/js/override_chatter.js
```diff
@@ -0,0 +1,34 @@
...
+            this._rpc({
+                    model: self.record.model,
+                    method: 'redirect_to_activities',
+                    args: [[]],
+                    kwargs: {'id':self.record.res_id,
+                             'model':self.record.model},
+                    context: this.record.getContext(),
+            }).then(function(action) {
```
yajo: ... if you do this here:

```js
}).then($.proxy(this, "do_action"));
```

mail_activity_board/static/src/xml/inherit_chatter.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+
+    <t t-extend="mail.Chatter.Buttons">
+         <t t-jquery="button.o_chatter_button_schedule_activity" t-operation="after">
+            <button t-if="schedule_activity_btn" class="btn btn-sm btn-link o_chatter_button_count_activity" title=" activity">
+                <i class="fa fa-list"/> <t t-if="isMobile">Activities</t><t t-else="">Activities</t>
```
yajo: Why the if?

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -108,27 +89,51 @@
...
+                                        <field name="summary"/>
                                     </strong>
                                 </div>
+                                <div>
+                                    <strong class="o_kanban_record_title"><field name="res_name"/></strong>
+                                </div>
+
+                                <div class="o_kanban_record_bottom">
+                                    <div class="oe_kanban_bottom_left">
+                                        <t t-if="record.date_deadline.raw_value and record.date_deadline.raw_value lt (new Date())">
```
yajo: `lt`?

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -108,27 +89,51 @@
...
+                                            <span t-attf-class="text-danger"><i class="fa fa-clock-o"/></span>
+                                            <t t-if="record.activity_category.raw_value!='meeting'">
+                                                <span t-attf-class="text-danger">
+                                                    <field name="date_deadline"/>
+                                                </span>
+                                            </t>
+                                            <t t-if="record.activity_category.raw_value=='meeting'">
```
yajo: Can't you use `t-else`?

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -108,27 +89,51 @@
...
+                                            </t>
+                                        </t>
+                                        <t t-else="">
+                                            <span><i class="fa fa-clock-o"/></span>
+                                            <t t-if="record.activity_category.raw_value!='meeting'">
+                                                <span>
+                                                    <field name="date_deadline"/>
+                                                </span>
+                                            </t>
+                                            <t t-if="record.activity_category.raw_value=='meeting'">
```
yajo: `t-else`?

mail_activity_board/views/mail_activity_view.xml
yajo: It should be enough then, thanks :wink:

yajo (changes requested):
See the logs here:
peek 2018-07-12 08-29

Some code comments below. Waiting for runbot to test :blush:

elicoidal:
Runbot seems not available?

mail_activity_board/__manifest__.py
dajuayen: Changed it.

mail_activity_board/models/mail_activity.py
dajuayen: Yes!!!

mail_activity_board/models/mail_activity_mixin.py
dajuayen: Corrected it.

mail_activity_board/models/mail_activity_mixin.py
dajuayen: Thanks for the link.

mail_activity_board/readme/HISTORY.rst
dajuayen: I delete this file.

mail_activity_board/static/src/js/override_chatter.js
dajuayen: When you know what little you've learned by hitting your head against the wall, ... it's very funny, it's not just.

mail_activity_board/static/src/js/override_chatter.js
dajuayen: I take note, although I do not know very well what it does

mail_activity_board/static/src/xml/inherit_chatter.xml
dajuayen: copy & paste

dajuayen: I already removed it.

mail_activity_board/views/mail_activity_view.xml
dajuayen: If record.date_deadline 
and 
its value is previous previous to today
add the class "text-danger".

Is it badly used?

mail_activity_board/views/mail_activity_view.xml
dajuayen: No problem.

mail_activity_board/views/mail_activity_view.xml
dajuayen: Oh, yeah!

gurneyalex:
If you check runbot status, it is "orange" which means that there are some warnings in the logs. 

Checking the logs show:
 
```
2018-07-13 12:43:26,081 165 INFO openerp_test odoo.modules.loading: loading mail_activity_board/views/mail_activity_view.xml
2018-07-13 12:43:26,110 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: domain
2018-07-13 12:43:26,135 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: context
```

gurneyalex:
If you check runbot status, it is "orange" which means that there are some warnings in the logs. 

Checking the logs show:
 
```
2018-07-13 12:43:26,081 165 INFO openerp_test odoo.modules.loading: loading mail_activity_board/views/mail_activity_view.xml
2018-07-13 12:43:26,110 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: domain
2018-07-13 12:43:26,135 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: context
```

gurneyalex:
If you check runbot status, it is "orange" which means that there are some warnings in the logs. 

Checking the logs show:
 
```
2018-07-13 12:43:26,081 165 INFO openerp_test odoo.modules.loading: loading mail_activity_board/views/mail_activity_view.xml
2018-07-13 12:43:26,110 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: domain
2018-07-13 12:43:26,135 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: context
```

gurneyalex:
If you check runbot status, it is "orange" which means that there are some warnings in the logs. 

Checking the logs show:
 
```
2018-07-13 12:43:26,081 165 INFO openerp_test odoo.modules.loading: loading mail_activity_board/views/mail_activity_view.xml
2018-07-13 12:43:26,110 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: domain
2018-07-13 12:43:26,135 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: context
```

gurneyalex:
If you check runbot status, it is "orange" which means that there are some warnings in the logs. 

Checking the logs show:
 
```
2018-07-13 12:43:26,081 165 INFO openerp_test odoo.modules.loading: loading mail_activity_board/views/mail_activity_view.xml
2018-07-13 12:43:26,110 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: domain
2018-07-13 12:43:26,135 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: context
```

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,209 @@
...
+        </field>
+    </record>
+
+
+    <!-- TREE VIEW -->
+    <record id="mail_activity_view_tree" model="ir.ui.view">
+        <field name="name">mail.activity.boards.view.tree</field>
+        <field name="model">mail.activity</field>
+        <field name="inherit_id" ref="mail.mail_activity_view_tree"/>
+        <field name="domain">[]</field>
```
gurneyalex: this line is causing the build failure on runbot

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,209 @@
...
+        </field>
+    </record>
+
+
+    <!-- KANBAN VIEW -->
+     <record id="mail_activity_view_kanban" model="ir.ui.view">
+         <field name="name">mail.activity.boards.view.kanban</field>
+         <field name="model">mail.activity</field>
+         <field name="priority" eval="10"/>
+         <field name="context">{}</field>
```
gurneyalex: this line is causing a red bulid on runbot

yajo:
Fix runbot:

```
2018-07-13 12:43:26,110 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: domain
2018-07-13 12:43:26,135 165 WARNING openerp_test odoo.models: ir.ui.view.create() includes unknown fields: context
```

mail_activity_board/models/mail_activity.py
dajuayen: Deleted store=False.
Deleted default="".

mail_activity_board/models/mail_activity.py
dajuayen: I have tried to leave the field with only "related" and "readonly" attributes and it has behaved as you have indicated, so I have changed it.

mail_activity_board/models/mail_activity.py
dajuayen: Me too.

mail_activity_board/models/mail_activity.py
dajuayen: Me too.

mail_activity_board/models/mail_activity.py
dajuayen: They do not return the same.
self.env.ref ('...'). read () [0] returns a dictionary with all the data of the action.
self.env.ref ('...'). id only returns the id (int) of the action.

I can not change it

mail_activity_board/views/mail_activity_view.xml
dajuayen: Thank you for answering the question.

mail_activity_board/views/mail_activity_view.xml
dajuayen: Thanks again.

mail_activity_board/models/mail_activity.py
dajuayen: As you mention in another change, the field with 'related' and 'readonly' inherits the attributes and therefore also the 'string', so the non-necessary ones I have finally deleted.

mail_activity_board/views/mail_activity_view.xml
dajuayen: The night is dark and full of terrors.

mail_activity_board/views/mail_activity_view.xml
dajuayen: I think that the "readonly" attribute is not necessary because I have already defined the form as non-editable, so I have deleted it from the fields.

mail_activity_board/views/mail_activity_view.xml
dajuayen: I think that the "readonly" attribute is not necessary because I have already defined the form as non-editable, so I have deleted it from the fields.

mail_activity_board/views/mail_activity_view.xml
dajuayen: I think that the "readonly" attribute is not necessary because I have already defined the form as non-editable, so I have deleted it from the fields.

mail_activity_board/static/src/xml/inherit_chatter.xml
dajuayen: Finally I have left it:
\<t t-if = "isMobile"> List </ t> <t t-else = ""> Activities list </ t>

because it shows a different legend if 'isMobile' or not.

dajuayen:
Hi @Yajo.
Runbot test pass, but now something called 'codecov' appeared. 
What do I have to do now to finish the push request?

mail_activity_board/models/mail_activity.py
```diff
@@ -10,14 +10,12 @@ class MailActivity(models.Model):
...
+        related='calendar_event_id.duration', readonly=True)
     calendar_event_id_start = fields.Datetime(
-        related='calendar_event_id.start', string="Start",
-        default=False, readonly=True)
+        related='calendar_event_id.start', readonly=True)
     calendar_event_id_partner_ids = fields.Many2many(
-        related='calendar_event_id.partner_ids', string='Attendees',
-        default=False, readonly=True)
+        related='calendar_event_id.partner_ids',
+        Stringreadonly=True)
```
yajo: Oops, wrong parameter.

mail_activity_board/static/src/js/override_chatter.js
```diff
@@ -5,24 +5,71 @@ odoo.define('mail.Chatter.activity', function(require){
...
             }).then($.proxy(this, "do_action"));
         },
+        _formatActivitiesList: function (count){
+            var str = '';
+            if (count <= 0) {
+                str = _t('No activities');
+            } else if (count === 1){
+                str = _t('One activity');
+            } else {
+                str = ''+count+' '+_t('activities');
```
yajo: This method makes sense in your language and some others, but did you know that Russian has different plurals, depending on if the number ends by 2-4 or by other number? And of course, they have singular too. And [it's not a corner case](https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html). You likely don't want this kind of methods that return different strings depending on the number. Gettext itself should handle that, and if it doesn't, it's upstream's problem (and also I don't think it will affect you, since your language follows the same rules as English).

It's much simpler to be dumber and just do:

```
return _.str.sprintf(_t("%d activities"));
```

... which wouldn't be so extremely bad if rendered as *1 activities*. You can even translate that into Spanish as *%d actividad/es*, which is perfectly OK.

This also should make you realize that using a method just for this is not much needed. You can just do that call inline and it should be OK.

In any case, this is being used just to add a `title` attribute to the button, but *we can simply hardcode a more useful title and forget about all of this*; so, just remove this whole `_formatActivitiesList` method please.

I'm just trying to give you a bigger picture about how translations work (or should work), so you can remove all this complexities from your mind (and code!). :wink:

mail_activity_board/static/src/xml/inherit_chatter.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+
+    <t t-extend="mail.Chatter.Buttons">
+         <t t-jquery="button.o_chatter_button_schedule_activity" t-operation="after">
+            <button t-if="schedule_activity_btn" class="btn btn-sm btn-link o_chatter_button_list_activity">
+                 <span class="o_list_activity label label-primary"/> <t t-if="isMobile">List</t><t t-else="">Activities list</t>
```
yajo: Some points here:

1. If you need to display something depending on screen size, it's better to do it this way by pure bootstrap CSS:

   ```xml
   <span class="hidden-xs">Activities list</span>
   <span class="visible-xs-inline">Activities</span>
   ```

   The less JS and the more CSS you use, the faster everything is. However...

1. ... KISS; just use "Activities". The "list" word here adds nothing useful really.

1. Additions should be done here, and not above in the `_render()` method; however, it's true that will only be possible after https://github.com/odoo/odoo/pull/25801 is merged.

So, it should end up being similar to this, as of today:

```xml
<button t-if="schedule_activity_btn" class="btn btn-sm btn-link o_chatter_button_list_activity" title="See activities list" type="button">
  Activities
</button>
```

Once https://github.com/odoo/odoo/pull/25801 is merged, if ever, it could become this:

```diff
 <button t-if="schedule_activity_btn" class="btn btn-sm btn-link o_chatter_button_list_activity" title="See activities list" type="button">
+  <span class="o_list_activity label label-primary">
+    <t t-esc="widget.fields.activity.activities.length"/>
+  </span>
   Activities
 </button>
```

mail_activity_board/static/src/js/override_chatter.js
```diff
@@ -0,0 +1,75 @@
...
+
+            var $spinner = $(QWeb.render('Spinner'));
+            concurrency.rejectAfter(concurrency.delay(500), def).then(function () {
+                $spinner.appendTo(self.$el);
+            });
+            return def.then(function () {
+                if (self.fields.activity) {
+                    self.$('.o_list_activity')
+                                .html(self.fields.activity.activities.length)
+                                .parent().attr("title", self._formatActivitiesList(self.fields.activity.activities.length));
```
yajo: I see that [you're duplicating almost all of upstream method](https://github.com/odoo/odoo/blob/3656999f5e51658e0611af100c1507abc7647d94/addons/mail/static/src/js/chatter.js#L175-L200) just to add these 2 lines, breaking the call chain. Sorry, but that shouldn't be done. :no_entry_sign: 

You have 3 options:

1. Override the `start()` method, call its `_super()`, and, once done, make the changes into `this.$topbar`, which is there rendered synchronously. All of this deferred rendering of fields has no relation with what you're modifying, which is actually the buttons bar, not the fields themselves.
1. Vote for https://github.com/odoo/odoo/pull/25801 to be merged, and just do all overrides directly in Qweb, as explained below.
1. Simply show static buttons. That can be done through Qweb only, and wouldn't need https://github.com/odoo/odoo/pull/25801 to be merged. I recommend this approach, at least for now; see below.

Remove this `_render()` method in any case.

mail_activity_board/static/src/js/override_chatter.js
```diff
@@ -0,0 +1,75 @@
...
+    var QWeb = core.qweb;
+
+    chatter.include({
+
+        events: _.extend({}, chatter.prototype.events, {
+            'click .o_chatter_button_list_activity': '_onListActivity',
+        }),
+
+        _onListActivity: function (event) {
+            event.preventDefault();
```
yajo: Not needed if you add `type="button"` to the button. Remove and see comments below.

yajo (changes requested):
You need to simplify that JS, friend :wink:

mail_activity_board/models/mail_activity.py
dajuayen: OMG :anguished:

mail_activity_board/static/src/js/override_chatter.js
dajuayen: Thanks for advice. 
When I don't know how to make something I use to find in ODOO folder a piece of code similar to I want to do and copy it. 
I found in followers.js something similar:

``` javascript
   _formatFollowers: function (count){
        var str = '';
        if (count <= 0) {
            str = _t('No follower');
        } else if (count === 1){
            str = _t('One follower');
        } else {
            str = ''+count+' '+_t('followers');
        }
        return str;
    }
```


In next times I'll try to be more efficient and follow your wise advice.

I  :fire: it  already

mail_activity_board/static/src/xml/inherit_chatter.xml
dajuayen: As I have already mentioned to you, when I have to do something, I try to look at how it is done in another part of the code and imitate it.
 
I thought that instead of using an icon, I could use the number of activities for it, in the style of followers and the number of activities when the list of activities is minimized.

What a pity
:pensive:

mail_activity_board/static/src/js/override_chatter.js
dajuayen: I removed _render().

dajuayen: But I wrote down option 1. I'm stubborn.

mail_activity_board/static/src/js/override_chatter.js
dajuayen: You mean only the line:

```
event.preventDefault();
```

mail_activity_board/views/mail_activity_view.xml
yajo: There's still this. `lt` is nothing. Did you mean to use `&lt;`? Also, I think you're comparing a `String` with a `Date`, which will very likely produce false positives (not sure, just guessing...)

yajo (changes requested):
Much better, getting closer :wink:

dajuayen:
Hi @Yajo 

Regarding your last comment, I knew that I had taken it from some part of the code, so I found it in TASK in the OCA runbot.
odoo/addons/project/views/project_views.xml
project_task_view_kanban

That's why I'm surprised that this is failing.
I can not wait to finish the PR. If you think that removing it, the PR would pass the codecov, I'll take it down.

But abusing your kindness, I do not understand what fails in that test. Reading the codecov documentation, this informs about the amount of code that is covered by the tests. Is the flaw that it is not completely covered? How do I get all the code covered by the tests?

What do I have to do?

We are in contact and  thanks,
regards

mail_activity_board/views/mail_activity_view.xml
yajo: Answering to https://github.com/OCA/social/pull/283#issuecomment-406784467 here to avoid losing context.

The fact that you copy-pasted code doesn't mean you copied good code. :stuck_out_tongue_winking_eye: 

Normally, codecov won't detect this, because the qweb view you write in a kanban is actually a client-side (JS) qweb one. This means that it's not executed when installing, but when you browse it.

See this screenshot from Firefox JS console, where you can see that you're producing a syntax error, and a false positive:

captura de pantalla de 2018-07-23 08-18-42

yajo (commented):
About codecov, you can see how to get to the report here:

peek 2018-07-23 08-25

The red-marked lines in that diff are the ones that are not covered.

"Not covered" means: that code has not been executed while running installation and unit tests of the code.

So, to make it go from :x: to :heavy_check_mark:, you need to [develop some unit tests](https://www.odoo.com/documentation/11.0/reference/testing.html) that make odoo execute that part of code and make sure the output from those methods is what's expected to be.

mail_activity_board/views/mail_activity_view.xml
dajuayen: @Yajo I think it is ready.
The field 'date_deadline' in the class 'mail.activity' is required, so I have removed the check if it is empty.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -101,28 +101,29 @@
 
                                 <div class="o_kanban_record_bottom">
                                     <div class="oe_kanban_bottom_left">
-                                        <t t-if="record.date_deadline.raw_value and record.date_deadline.raw_value lt (new Date())">
+                                        <t t-set="act_date" t-value="new Date(record.date_deadline.raw_value)"/>
+                                        <t t-if="act_date lt (new Date())">
```
yajo: What does `lt` mean? Shouldn't that be `&lt;`?

dajuayen: Yes, it is probably the most convenient.

Anyway, what luck I had when looking for an example to guide me, it is likely that he could give false positives and he stopped escaping this.

  I change it and now I upload it.

dajuayen:
@Yajo I have added a change to the form view on the activity board for those that are meeting type. Check them out to see what you think. 
The other problems are solved.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -36,10 +36,56 @@
...
                                    ('calendar_event_id','=', False)]}"/>
                             <field name="user_id" options="{'no_open': True}"/>
-                            <field name="calendar_event_id_partner_ids" widget="many2many_tags"
-                                   attrs="{'invisible': [('calendar_event_id','=', False)]}"/>
+
                         </group>
                     </group>
+                    <group attrs="{'invisible': ['|',('calendar_event_id','=', False),('calendar_event_id_partner_ids','=', False)]}">
+                        <field name="calendar_event_id_partner_ids" mode="kanban">
+                                <kanban quick_create="false" create="true" delete="true">
```
yajo: Please indent consistently.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,254 @@
...
+                                    <field name="id"/>
+                                    <field name="name"/>
+                                    <field name="image_small"/>
+                                    <field name="parent_id"/>
+                                    <field name="function"/>
+                                    <field name="type"/>
+                                    <field name="is_company"/>
+                                    <templates>
+                                        <t t-name="kanban-box">
+                                            <div class="oe_kanban_global_click" style="max-width: 250px">
```
yajo: Avoid inline styles if possible by styling through a css file.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,254 @@
...
+                                                <a t-if="! read_only_mode" type="delete" class="text-danger fa fa-times"/>
+                                                <div class="o_kanban_image">
+                                                    <t t-if="record.image_small.raw_value">
+                                                        <img t-att-src="kanban_image('res.partner', 'image_small', record.id.raw_value)" height="40" width="40" class="oe_avatar oe_kanban_avatar_smallbox mb0"/>
```
yajo: Remove height and width. That's how HTML got styling before CSS was invented!

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,254 @@
...
+                                                    </t>
+                                                    <t t-if="!record.image_small.raw_value">
+                                                        <t t-if="record.type.raw_value === 'delivery'">
+                                                            <img t-att-src='_s + "/base/static/src/img/truck.png"' class="o_kanban_image oe_kanban_avatar_smallbox"/>
```
yajo: Where is `_s` defined? :thinking:

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -0,0 +1,254 @@
...
+                            <field name="duration" widget="float_time"
+                                   attrs="{'invisible': ['|',('duration', '=', False),
+                                   ('calendar_event_id','=', False)]}"/>
+                            <field name="user_id" options="{'no_open': True}"/>
+
+                        </group>
+                    </group>
+                    <group attrs="{'invisible': ['|',('calendar_event_id','=', False),('calendar_event_id_partner_ids','=', False)]}">
+                        <field name="calendar_event_id_partner_ids" mode="kanban">
+                                <kanban quick_create="false" create="true" delete="true">
```
yajo: One suggestion: since this is just a kanban display of event attendees... Can't you just use here the default kanban view for partners? That would reduce ~50 LOC from here. :thinking: 

---

Note: ignore comments below if you choose to do that :point_up:

yajo (changes requested):

mail_activity_board/views/mail_activity_view.xml
dajuayen: Ok. 
captura de pantalla de 2018-09-17 12-28-28
I took the idea of the CRM, in the tab of Sales Channels in configuration.
The view seems cleaner to me. But if you prefer it to be standard, I'll put it like that:
captura de pantalla de 2018-09-17 12-33-08

yajo: As you prefer. I tend to prefer the solution that means less code to maintain :yum:

dajuayen: I buy it to you.

yajo (approved):

LoisRForgeFlow (approved):
Tested on runbot and functionally looks great! :+1: 

Thanks!

dajuayen:
@lreficent Thaks you!
We have developed a module that keeps mail.activiy done and when we had this fact we were going to upload it but I have seen that you have advanced. This PR was my first, between that and that I have not had time to dedicate, it has been very long.
The other day I downloaded your code and I was testing it join to the mail_activity_board and so you can see together they work perfectly.
Anyway I made a comment in the [PR](https://github.com/OCA/social/pull/296) to see what you think.
What do you think what I said? How do you carry the PR?

dajuayen:
@Yajo  Why not just upload the PR to Social? Is something missing to be done so that it can be uploaded to the repository?

yajo:
Sorry I just forgot this one. Merging.

PR #284, opened 2018-06-14
11.0 mig mass_mailing_list_dynamic

damcar:
Need mass_mailing_partner (dependency) module to works.

- [x] I tested with #277 and it's ok.


yajo:
Thanks, setting as WIP until dependency is merged.

mass_mailing_list_dynamic/models/mail_mass_mailing_contact.py
```diff
@@ -9,13 +8,16 @@
...
-        if any((one.list_id.dynamic and
-                one.list_id.sync_method == "full") for one in self):
-            raise ValidationError(
-                _("Cannot edit manually contacts in a fully "
-                    "synchronized list. Change its sync method or execute "
-                    "a manual sync instead."))
+        for one in self:
+            if any((list.dynamic and
+                    list.sync_method == "full") for list in one.list_ids):
+        # if any((one.list_id.dynamic and
```
pedrobaeza: Remove this commented text

mass_mailing_list_dynamic/models/mail_mass_mailing_contact.py
```diff
@@ -9,13 +8,16 @@
...
+        #         one.list_id.sync_method == "full") for one in self):
+                raise ValidationError(
+                    _("Cannot edit manually contacts in a fully "
+                        "synchronized list. Change its sync method or execute "
```
pedrobaeza, line 22: Align correctly these lines

mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
```diff
@@ -47,18 +46,36 @@ def action_sync(self):
             desired_partners = Partner.search(sync_domain)
             # Remove undesired contacts when synchronization is full
             if one.sync_method == "full":
-                Contact.search([
-                    ("list_id", "=", one.id),
-                    ("partner_id", "not in", desired_partners.ids),
-                ]).unlink()
-            current_contacts = Contact.search([("list_id", "=", one.id)])
-            current_partners = current_contacts.mapped("partner_id")
+                contacts = Contact.search([("list_ids", "in", [one.id]),
```
pedrobaeza: Why changing the style in this search? This need one extra line for the same arguments.

pedrobaeza (changes requested):
About the change you have made on the behavior, I think it should be at least configurable, having 2 options:

* Always create new mass mailing contacts
* Try to reuse existing mass mailing contacts

The first one will use the behavior on previous version, and the second will use your new method.

Why this? Because there's something important in the mass mailing contacts management: the unsubscription. If you always reuse existing contacts, then if the contact has opt_out checked, it won't receive any mailing, so the dynamic list won't serve at all.

That's why I would even say that the default behavior should be the first one.

What do you think?

pedrobaeza:
@damcar have you seen my comments?

damcar:
@pedrobaeza sorry for the response time!
Regarding your comments of 28/06, it's a good idea, i will take this change.
For your other comments in the code, it's ok for me.

mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
damcar: you're right, it's not necessary!

mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
```diff
@@ -47,18 +46,36 @@ def action_sync(self):
...
+                                           ("partner_id", "not in",
+                                            desired_partners.ids)])
+                for contact in contacts:
+                    # No delete contacts with other lists
+                    if len(contact.list_ids) > 1:
+                        contact.write({'list_ids': [(3, one.id, False)]})
```
yajo, line 55: You might prefer to do `contact.list_ids -= one`

mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
```diff
@@ -47,18 +46,36 @@ def action_sync(self):
...
+                    current_partners_not_in_list:
                 Contact.create({
-                    "list_id": one.id,
+                    "list_ids": [(4, one.id, False)],
                     "partner_id": partner.id,
                 })
+            # Add list in existing contacts
+            for contact in current_contacts_not_in_list:
+                contact.write({
+                    "list_ids": [(4, one.id, False)],
```
yajo: Here you might also prefer `contact.list_ids |= one`

rafaelbn:
Hello! Thanks fo this contribution! :smile: 

Travis is :red_circle:  runbot is :red_circle:  this both should be fixed

In this kinf of migration the most important is thinking that there are customer using this in v10 in production and when we migrate them to v11 the modules should make the same, could have improvements but never change behavior.

PR #285, opened 2018-06-16
[FIX] do not use section titles in readme fragments

sbidoul:
This creates inconsistency issues when assembling them in the README.

cc/ @chienandalu

### DIFF
--- a/mass_mailing_custom_unsubscribe/readme/CONFIGURE.rst
+++ b/mass_mailing_custom_unsubscribe/readme/CONFIGURE.rst
@@ -1,6 +1,3 @@
-Unsubscription Reasons
-----------------------
-
 You can customize what reasons will be displayed to your unsubscriptors when
 they are going to unsubscribe. To do it:
 


pedrobaeza:
But why we can use subtitles?

sbidoul:
Because rst infers document structure based on the style used for title underlines. Since we include these fragments in content that have their own structure, there is a risk of conflict. The only exception we made is in HISTORY.rst (one level using tilde `~` underline). Read also [here](https://github.com/OCA/maintainer-tools/blob/master/template/module/readme/USAGE.rst) and [here](https://github.com/OCA/maintainer-tools/blob/master/template/module/readme/HISTORY.rst).

pedrobaeza:
But there can be the need of several sections inside. Can we put in any way this subsections (maybe with `~`)?

Check the original README with 2 sections (now one of them is covered by core, and that's why there's only one subsection, but there can be another case where it's needed): https://github.com/OCA/social/blob/8.0/mass_mailing_custom_unsubscribe/README.rst

sbidoul:
One level with tilde is supported. If we want more we need to document very clearly what characters to use for underlines.

pedrobaeza:
OK, for now, we can merge this one (Travis fail is a glitch).

yajo:
Are there any docs on how to properly use the new readme generator?

sbidoul:
@Yajo the module template is the documentation.

https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst
https://github.com/OCA/maintainer-tools/blob/master/template/module/readme/USAGE.rst
...

PR #299, opened 2018-09-05
[IMP] mail_tracking: email score performance

chienandalu:
In instances with tons of trackings this method can resent performance. Hopefully this improves it.

cc @Tecnativa

### DIFF
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -107,8 +107,7 @@ class MailTrackingEmail(models.Model):
             return 0.
         data = self.read_group([('recipient_address', '=', email.lower())],
                                ['recipient_address', 'state'], ['state'])
-        mapped_data = dict([(state['state'], state['state_count'])
-                            for state in data])
+        mapped_data = {state['state']: state['state_count'] for state in data}
         return self.with_context(mt_states=mapped_data).email_score()
 
     @api.model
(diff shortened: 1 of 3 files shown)


mail_tracking/models/mail_tracking_email.py
```diff
@@ -103,10 +103,13 @@ def email_is_bounced(self, email):
...
-        if email:
-            return self.search([
-                ('recipient_address', '=', email.lower())]).email_score()
-        return 0.
+        if not email:
+            return 0.
+        data = self.read_group([('recipient_address', '=', email.lower())],
+                               ['recipient_address', 'state'], ['state'])
+        mapped_data = dict([(state['state'], state['state_count'])
+                            for state in data])
```
yajo: Remove the surrounding `[...]` brackets.

With brackets:

1. Produce a lazy generator.
2. Exhaust the generator converting it into a list.
3. Convert the list in a dict.

Without brackets:

1. Produce a lazy generator.
2. Exhaust the generator converting it into a dict.

You can save tons of loops! :blush: 

---

Note that another idiomatic way (equivalent to doing it without brackets) to do so is: 

```py
mapped_data = {state['state']: state['state_count'] for state in data}
```

yajo (approved):

mail_tracking/models/mail_tracking_email.py
chienandalu: Noted and fixed :wink:

Full coverage as well.

pedrobaeza (approved):

PR #305, opened 2018-09-26
10.0 Make sync method action more user friendly

thinkwelltwd:
The SQL constraint gives the user no idea how
to rectify the problem.

Instead, sync all the Contacts that can be, and
collect the ones that can't into a list to
display to the user in a UserError.

### DIFF
--- a/mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
+++ b/mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 # Copyright 2017 Tecnativa - Jairo Llopis
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from psycopg2 import IntegrityError
 
 from odoo import api, fields, models
-from odoo.tools import safe_eval
+from odoo.exceptions import ValidationError
+from ast import literal_eval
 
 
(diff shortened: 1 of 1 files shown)


thinkwelltwd:
This is going to fail tests, as it won't raise a ValidationError, but rather than fix the test at this point, I thought I'd post the commit and ask if there's any reason this is not an acceptable fix.

I'll be happy to change the test to check for UserError if the overall approach is satisfactory.

mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
```diff
@@ -55,15 +58,27 @@ def action_sync(self):
...
-                Contact.create({
-                    "list_id": one.id,
-                    "partner_id": partner.id,
-                })
+                try:
+                    Contact.create({
+                        "list_id": one.id,
+                        "partner_id": partner.id,
+                    })
+                    self.env.cr.commit()
```
pedrobaeza: Never commit inside a transaction. Instead, you can use a savepoint:
```python
with self.env.cr.savepoint():
  # do stuff
```

thinkwelltwd:
Sorry for this taking so long; switched to using `self.env.cr.savepoint()` as requested. Does this pass muster?

thinkwelltwd:
Does this pass muster now?

PR #309, opened 2018-10-12
[8.0] Add mail_from_force module

### LINKED ISSUE #308: [RFC] mail_from_force: force the "From" address to a fixed email
In some scenarios, the outgoing mail server enforces the "From" email address to be the one for User connected to the mail server.

This means that if I connect to the outgoing mail server using "admin", all sent emails must be from "admin@example.com", and an email From "daniel.reis@example.com" would be rejected by the mail server.

AFAICS this behavior is not configurable, so I'm preparing an extension to solve this:
The From would be changed to keep the sender's name, but use the Outgoing Mail user's address.

Following the example above, Daniel's email From would be something like "Daniel Reis <admin@example.com>" or "On behalf of Daniel Reis <admin@example.com>"

Technically, we just need to override the `_get_default_from` method of `mail.message`.

dreispt:
Closes #308

### DIFF
--- a/mail_from_force/__init__.py
+++ b/mail_from_force/__init__.py
@@ -1,0 +1,2 @@
+# -*- coding: utf-8 -*-
+from . import models
--- a/mail_from_force/__openerp__.py
+++ b/mail_from_force/__openerp__.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+{
+    'name': "Force From sender to be the Admin user",
+    'author': "Daniel Reis,Odoo Community Association (OCA)",
+    'summary': 'Force the header Sender and envelope From '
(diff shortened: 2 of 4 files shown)


astirpe:
@dreispt you may want to have a look at this old PR: https://github.com/OCA/server-tools/pull/869. The From is not fixed but is read from the `X-Original-From` present in the header.

dreispt:
Thank you for pointing that PR Andreas.
I feel that these are different features: my problem here was that the outgoing mail server does not allow Odoo use other people in the "From" field, impersonating them.
Am I seeing this right?

astirpe:
@dreispt right, the other module was meant for a different purpose.

dreispt:
It looks like there is prior art on this, see #388

PR #312, opened 2018-10-24
[10.0][FIX] mail_sendgrid*. Use info loglevel for import error.

NL66278:
Using _logger.warning for ImportError exception results in runbot/travis failures.

It is also more in line with how this is handled in other modules.

See also:
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#importerror


pedrobaeza (approved):
I agree. Fast-tracking this.

PR #315, opened 2018-11-02
[FIX] mass_mailing_partner: Fix and optimize batch writes

yajo:
Before this patch, all customizations done per contact in a batch write, except the last one, were ignored. It also was enforcing to write on some fields that were already fine, triggering their constraints unnecesarily, and causing unwanted side effects such as when combined with `mail_tracking_mass_mailing` and `mass_mailing_list_dynamic` in a fully synced list.

After this patch, customized writes are written one by one, and they only get the really needed customizations.

@Tecnativa

### DIFF
--- a/mass_mailing_partner/models/mail_mass_mailing_contact.py
+++ b/mass_mailing_partner/models/mail_mass_mailing_contact.py
@@ -64,7 +64,7 @@ class MailMassMailingContact(models.Model):
         m_mailing = self.env['mail.mass_mailing.list']
         m_partner = self.env['res.partner']
         list_id = vals.get('list_id', self.list_id.id)
-        mailing_list = m_mailing.browse(list_id, self._prefetch)
+        mailing_list = m_mailing.browse(list_id, prefetch=self._prefetch)
         # Look for a partner with that email
         email = email.strip()
         partner = m_partner.search([('email', '=ilike', email)], limit=1)
--- a/mass_mailing_partner/models/mail_mass_mailing_contact.py
+++ b/mass_mailing_partner/models/mail_mass_mailing_contact.py
@@ -34,12 +34,19 @@ class MailMassMailingContact(models.Model):
         return super(MailMassMailingContact, self).create(vals)
 
     def write(self, vals):
+        remaining = self
         for contact in self:
-            if vals.get('partner_id', None) is False:
-                # If removing partner, search again by email
-                vals = contact._set_partner(vals)
-            vals = contact._set_name_email(vals)
-        return super(MailMassMailingContact, self).write(vals)
+            _vals = vals.copy()
+            # If removing partner, search again by email
+            if _vals.get('partner_id') is False:
+                _vals = contact._set_partner(_vals)
+            _vals = contact._set_name_email(_vals)
+            # If changed, write it separately
+            if vals != _vals:
+                remaining -= contact
+                super(MailMassMailingContact, contact).write(_vals)
+        # Normal batch write for contacts that need no customization
+        return super(MailMassMailingContact, remaining).write(vals)
 
     def _prepare_partner(self, vals, mailing_list):
         vals = {
@@ -57,13 +64,13 @@ class MailMassMailingContact(models.Model):
         m_mailing = self.env['mail.mass_mailing.list']
         m_partner = self.env['res.partner']
         list_id = vals.get('list_id', self.list_id.id)
-        mailing_list = m_mailing.browse(list_id)
+        mailing_list = m_mailing.browse(list_id, self._prefetch)
         # Look for a partner with that email
         email = email.strip()
-        partners = m_partner.search([('email', '=ilike', email)], limit=1)
-        if partners:
-            # Partner found
-            vals['partner_id'] = partners[0].id
+        partner = m_partner.search([('email', '=ilike', email)], limit=1)
+        # Different partner found
+        if partner and partner != self.partner_id:
+            vals['partner_id'] = partner.id
         elif mailing_list.partner_mandatory:
             # Create partner
             partner = m_partner.sudo().create(
@@ -75,7 +82,9 @@ class MailMassMailingContact(models.Model):
         partner_id = vals.get('partner_id', self.partner_id.id)
         if not partner_id:
             return vals
-        partner = self.env['res.partner'].browse(partner_id)
-        vals['email'] = partner.email
-        vals['name'] = partner.name
+        partner = self.env['res.partner'].browse(partner_id, self._prefetch)
+        if vals.get("email", self.email) != partner.email:
+            vals['email'] = partner.email
+        if vals.get("name", self.name) != partner.name:
+            vals['name'] = partner.name
         return vals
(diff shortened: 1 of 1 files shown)


mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -57,13 +64,13 @@ def _set_partner(self, vals):
         m_mailing = self.env['mail.mass_mailing.list']
         m_partner = self.env['res.partner']
         list_id = vals.get('list_id', self.list_id.id)
-        mailing_list = m_mailing.browse(list_id)
+        mailing_list = m_mailing.browse(list_id, self._prefetch)
```
pedrobaeza: Please use keyword argument assignment for prefetch argument.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -75,7 +82,9 @@ def _set_name_email(self, vals):
         partner_id = vals.get('partner_id', self.partner_id.id)
         if not partner_id:
             return vals
-        partner = self.env['res.partner'].browse(partner_id)
-        vals['email'] = partner.email
-        vals['name'] = partner.name
+        partner = self.env['res.partner'].browse(partner_id, self._prefetch)
+        if vals.get("email", self.email) != partner.email:
```
pedrobaeza, line 87: s/self.email/partner.email?

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -75,7 +82,9 @@ def _set_name_email(self, vals):
...
+            vals['email'] = partner.email
+        if vals.get("name", self.name) != partner.name:
```
pedrobaeza, line 89: s/self.name/partner.name?

mass_mailing_partner/models/mail_mass_mailing_contact.py
yajo: The point here is to avoid updating the record if the email found on it is equal to the partner's, so AFAICS it's OK as it is: It gets the email passed through `vals`, and if none, it gets it from current record, to be able to compare it with the partner's.

mass_mailing_partner/models/mail_mass_mailing_contact.py
yajo: Same here.

mass_mailing_partner/models/mail_mass_mailing_contact.py
pedrobaeza: I'm talking for avoiding singleton error. Replace `self` (multi-record) by `partner` (single-record)

yajo: This method [was not meant to be called on multi records before](https://github.com/OCA/social/pull/315/files#diff-6e2abeeb2e2cc273679a065d55fcd860R82), so that's not a problem here.

pedrobaeza: OK, got it, although the variable names are confusing (specially `partner` meaning the other part).

pedrobaeza (approved):

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -75,7 +82,9 @@ def _set_name_email(self, vals):
         partner_id = vals.get('partner_id', self.partner_id.id)
         if not partner_id:
             return vals
-        partner = self.env['res.partner'].browse(partner_id)
-        vals['email'] = partner.email
-        vals['name'] = partner.name
+        partner = self.env['res.partner'].browse(partner_id, self._prefetch)
```
chienandalu: Why do you make the call positional here and named here https://github.com/OCA/social/pull/315/files#diff-6e2abeeb2e2cc273679a065d55fcd860R67 ?

yajo: I used positional at first, but they asked me to change it above, so I changed it. Should I change it here also?

chienandalu: I don't know :man_shrugging:

pedrobaeza: Yes, please, I didn't see this one, but always keyword if they are keyword in the original signature.

yajo: Keyword-only arguments are a python3-only feature that is not used in Odoo AFAIK.

The signature for keyword-only arguments is:

```python
def method(self, method, whatever=True, *, keyword_only=False):
   pass
```

By "keyword" did you mean that they have a default value?

pedrobaeza: Yes, arguments with a default value are called "keyword arguments". `keyword_only` arguments are for explicitly forcing them to be used with keyword, but a good practice is to always use keyword arguments with the argument, so that if someone use another order (or arguments are passed as dictionaries with `**kwargs`), they are kept in the proper dictionary or argument.

yajo: Usually I use keywords when the argument is not explicit enough. In this case, `self._prefetch` is quite obviously passed to the prefetch argument.

Nevermind, it's changed.

pedrobaeza: But do you understand the reasoning about keeping them keywords? Imagine you have an overriding like that:
```python
def browse(args, kwargs):
    if 'prefetch' in kwargs:
        # do stuff
    return super().browse(*args, **kwargs)
```
it won't do what you want, and you have to hardcode possible argument position in `args`.

yajo: I don't understand it, TBH.

I'd write that like this, in case I just want to use the `prefetch` arg and I don't mind about other possible args:

```python
def browse(self, ids, prefetch=None, *args, **kwargs):
    if prefetch:
        ...
    return super().browse(ids, prefetch, *args, **kwargs)
```

That kind of override you said is actually forcing an argument to be called as a kwarg, which alters the original signature and is possibly gonna lead to errors, so that should be considered a bug if ever landed. :roll_eyes:

pedrobaeza: It's not a bug, but a generic way to catch any argument. What you write implies that you have to add all the positional arguments previous to the keyword argument you pass as positional and also that you add keyword arguments in the same order, which is not true before P3.6 when you pass a dictionary to super.

yajo: Well, you're right, it's not a bug. However, your code would only work if the `prefetch` argument is passed as a keyword one. If that's what you really want, then that's the way to do it. If what you really want is to get the parameter, no matter how the method is called, then you need to write the paramteres in order as I did. That's normal python (in fact, even python 2), and has nothing to do with py3.6 dict order AFAIK.

The fact that an argument has a `=Whatever` in its signature means that the argument has a default value, _not_ that you must pass it as a keyword argument. Otherwise, Python wouldn't have a way to pass default values to ordered arguments.

In any case, it's changed as you wanted, so I guess this whole conversation is unnecessary.

hbrunn (approved):

PR #318, opened 2018-11-05
12.0 mig mail_tracking

ernestotejeda:
* Adapt tests
* Adapt javascript
* Adapt template
Cc @Tecnativa


mail_tracking/__manifest__.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2016 Antonio Espinosa - <antonio.espinosa@tecnativa.com>
+# Copyright 2018 David Vidal - <david.vidal@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
chienandalu, line 4: Put yourself here :wink:

chienandalu (changes requested):
Thanks @ernestotejeda !

I've been testing in runbot with a sandbox smtp and found some quirks.

- Email score and bounce boolean are not aligned in the contact form:


- There's this strange behaviour with tracking events log. (Maybe Odoo stopping at the first sending exception?)

peek 07-11-2018 09-55

- I've got two authorized email address in my smtp sandbox. When I send a mail only to those, everything goes fine. But when I added a failing address, it only delivered once (surely related to the previous point):

peek 07-11-2018 10-16

mail_tracking/models/res_partner.py
```diff
@@ -0,0 +1,49 @@
...
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    # tracking_emails_count and email_score are non-store fields in order
+    # to improve performance
+    # email_bounced is store=True and index=True field in order to filter
+    # in tree view for processing bounces easier
+    tracking_emails_count = fields.Integer(
+        compute='_compute_tracking_emails_count', readonly=True)
+    email_bounced = fields.Boolean(index=True)
```
chienandalu: (Thinking aloud) Shouldn't this be merged into the new `mail.blacklist.mixin`  https://github.com/odoo/odoo/blob/12.0/addons/mail/models/mail_blacklist.py#L106
https://github.com/odoo/odoo/blob/12.0/addons/mail/models/res_partner.py#L19

ernestotejeda: The name of the misxin mail.blacklist.mixin and the description of this class suggests that it's related to blacklist, not with 'mail tracking' and I don't know if this could confuse someone when using it, maybe another mixing could be created, maybe with the name mail.traking.mixin to give classes that inherit from it aspects related to mail tracking, but do you think other classes need these fields? I mean, do you think other classes need to inherit this new mixing in case we do it?
What do you think?

chienandalu: Yes, you're right :slightly_smiling_face: thats more for opt-outs.

Yes, there could be a mixin like `mail.bounced.mixin`. What do you think about it @pedrobaeza and @Yajo ? Does it deserve the effort?

pedrobaeza: I don't think it's needed, as `res.partner` is the only source of recipients.

chienandalu: Well, there's replicated functionality for the mailing contacts in mail_tracking_mass_mailing and it could be even used in other models like leads, for instance.

chienandalu: I don't have a strong opinion on it anyway :smile:

pedrobaeza: > Well, there's replicated functionality for the mailing contacts

Are you sure? Bounced check is only on partners, not on mass mailing contacts AFAIK. We don't replicate things for mass mailing, but link mail.tracking records to mails sent by mass mailing. If you think that something can be joined under a mixin, then go ahead, as I don't know exact details.

chienandalu: res.partner: https://github.com/OCA/social/blob/11.0/mail_tracking/models/res_partner.py#L16
mail.mass_mailing.contact: https://github.com/OCA/social/blob/11.0/mail_tracking_mass_mailing/models/mail_mass_mailing_contact.py#L11

pedrobaeza: Oh, I didn't know! Then it makes sense of course.

yajo: Yep, good idea!

hbrunn (approved):

pedrobaeza:
@ernestotejeda can you please do the mixin you talked about? There's other patch that my colleague is going to do for previous versions. Keep an eye in this repository for including it here.

chienandalu:
@ernestotejeda Can you review and have this fix in mind?: https://github.com/OCA/social/pull/336

ernestotejeda:
Yes, I will keep this fix in mind

ernestotejeda:
@chienandalu It's true, when you try to send an email but an exception is captured, then an exception is thrown and no more emails are sent. That is implemented in this module from the beginning and I don't know why. It's in the redefinition of the send_email method of 'ir.mail_server'. I have deleted the line where this exception is launched so odoo continues to send emails even if one of them can not be sent.

rafaelbn:
Please @ernestotejeda teake in care all the changes of https://github.com/OCA/social/pull/336 ( including my contribution :smile: ) Thanks

alan196 (approved):
👍 Code review and tested in a production server

pedrobaeza:
Please check #336 now that is merged for being included here.

ernestotejeda:
> Please check #336 now that is merged for being included here.

It's already checked and included here

yajo (approved):

PR #321, opened 2018-11-07
12.0 mig mail debrand

gdgellatly:
Complete rewrite as mail.templates are now QWeb templates so simply overwrite using inheritance.

The main change is all default mail notifications are debranded, not just the main one.


mail_debrand/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+[ This file must be max 2-3 paragraphs, and is required. ]
```
pedrobaeza: Remove this part

mail_debrand/readme/HISTORY.rst
```diff
@@ -0,0 +1,14 @@
+[ The change log. The goal of this file is to help readers
```
pedrobaeza: Remove this part

mail_debrand/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
+[ Enumerate known caveats and future potential improvements.
```
pedrobaeza: Remove this file if empty

mail_debrand/readme/USAGE.rst
```diff
@@ -0,0 +1,13 @@
+[ This file must be present and contains the usage instructions
```
pedrobaeza: Remove this part

mail_debrand/views/mail_notification_view.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
```
pedrobaeza, line 2: This is not going to work being noupdate="1".

pedrobaeza (changes requested):
The change of strategy is not going to work as noupdate="1" records are not updated on secondary modules, so you should stick to python hack. It would be also good that you generate the README directly in this PR for saving one extra commit on merge and the lack of verification of the full composition.

mail_debrand/views/mail_notification_view.xml
gdgellatly: @pedrobaeza What do you mean it won't update a secondary module? Its a completely new record. Just inheriting from an old one, just like any view inheritance, it worked absolutely fine in test.

The old python hack was required because mail.templates weren't written in QWeb and didn't support inheritance. Given QWeb views support inheritance and modification, it would seem to me the correct way to do this.

Edit - you can see below - it installs just fine.

image

gdgellatly: Oh btw, I'm happy to generate the readme, but I tried a few times before making the PR. It isn't really documented and didn't seem to work.  Wants repos and branches, I'm just developing on laptop. -- OK now it works.

pedrobaeza: Sorry, you're right about that. Indeed, previously this template wasn't Qweb and I have mixed both. Any way, please do several things:

* Are the original records noupdate="1" also? If not, you should do these ones noupdate="'0" for being consistent in the inheritance chain.
* Please put a low priority in the templates, as you are doing `replace` operations.

pedrobaeza:
About regenerating the README, you only need to install maintainer-tools in your laptop and run the command inside module folder:
```
oca-gen-addon-readme  --repo-name social --branch 12.0 --addon-dir . --no-commit
```
but if not, at least don't remove entire README and let the previous one for reducing the diffs (one removing the text and another adding it the other way).

mail_debrand/readme/HISTORY.rst
gdgellatly: is it not useful to future maintainers, just comments, I don't care.  Just seems strange to document the usage, then remove the documentation.

EDIT: Oh I see gen-addon-readme doesn't ignore the comments when copying to the main readme.  I expected them to be skipped

pedrobaeza: Well, I care because it's just junk that doesn't serve. It's put on the template for knowing what to do on each file, but not for being preserved in all the new READMEs across whole OCA.

mail_debrand/views/mail_notification_view.xml
gdgellatly: @pedrobaeza 

yes the originals are no update="1". 
I'm not sure how priority works in QWeb, I made it 99, hopefully that is low priority.

Everything else done now as well.

pedrobaeza (approved):

LoisRForgeFlow (approved):
LGTM

levkar:
Odoo changed templates inside modules and this module is not working for many templates.

Example:

https://github.com/odoo/odoo/blob/11.0/addons/auth_signup/data/auth_signup_data.xml#L131
becomes
https://github.com/odoo/odoo/blob/12.0/addons/auth_signup/data/auth_signup_data.xml#L279

Python hack will not work either unless better regex patterns are written.

@gdgellatly @pedrobaeza @lreficent

gdgellatly:
Just make a PR for those templates to add. They aren't inherited so no real
problem.

gdgellatly:
Actually that won't work. Will need a separate module for each module. Can
you give a definitive number of modules affected?

pedrobaeza:
Well, this module shouldn't depend on those modules, so better to improve regexp for them.

gdgellatly:
@pedrobaeza I took a quick look, its not too bad. The first xml replacement needs to stay. While the rest could all be covered by 3 or 4 fairly straightforward regex, I think a better solution is an etree xpath based element replacement in code. Just traverse to the node and remove and continue to each ancestor if it is empty. Some of these are in td's, some in tr's some in tables, some in div's. Just easier to xpath to the node containing the text, remove, check parent, if now empty recursively remove.

There must be some point in the rendering that the full document is available for parsing.

pedrobaeza:
Yes, but once rendered, not by XML.

rvalyi:
Are you sure this module works? I'm getting mad trying with sending a quotation by email end the "Powered by Odoo" is not removed. the xpath view extension of the mail.mail_notification_paynow looks correct however it looks like the inherit is not applied. For instance if I change something in the mail.mail_notification_paynow template, I see it change in the email, but that last ```<tr>``` where the branding stands is never removed...

gdgellatly:
They did work just fine, but I must admit I didn't retest after being made
to change view priority from review. But anyway for a different reason we
need to go back to a python based approach because these brandings are not
just in mail, but about 7 other modules incl sale.

pedrobaeza:
I told about the change on the strategy about Python code, and now my fears come true...

gdgellatly:
@pedrobaeza Well looking at it, actually its working absolutely identically to v11.0 (in fact better as it at least takes care of all mail cases) so it is nothing to do with strategy change or anything like that. It just so happens that the v11.0 module also didn't remove the powered by messages from templates either inside or outside of mail module. 

So picking on this PR and "strategy change" without checking is just a red herring, so you can be less fearful now.  The functionality never existed to be ported.

Now if we agree we do want to remove these Powered by messages, I have written the xml parsing to do so, the question is do we want the change?

EDIT: I should not, there is a significant performance tradeoff here. Every email will require parsing and at least 1 xpath, whereas in the past only 1 single template was ever checked.

pedrobaeza:
@gdgellatly when I migrated this module in v10, I took the approach of replacing text in the mail rendering because of these things, and that's why I talked about from the very first moment. I still don't understand why this can't be done the same way:

https://github.com/OCA/social/blob/10.0/mail_debrand/models/mail_template.py

gdgellatly:
@pedrobaeza But you didn't. Except for one template. And that was before they were changed to qweb. And definitely the powered by messages were never removed.

Now it is 8 templates all slightly differently formatted in 2 differing formats. Parsing xml is the right way for the html based templates, replacement is right for the qweb ones. I did a POC, works nicely.

POC is at #344

gdgellatly:
@pedrobaeza I just went back and tested based off your v11.  It doesn't work that way anymore for anything qweb based. It seems it now renders the template without the overarching qweb, and then adds them in during later based on user preferences.

pedrobaeza:
But mine is not v11, but v10. v11 version was done by Eficent.

gr22:
I am experiencing the same thing. I find no way to override those views.

> 
> 
> Are you sure this module works? I'm getting mad trying with sending a quotation by email end the "Powered by Odoo" is not removed. the xpath view extension of the mail.mail_notification_paynow looks correct however it looks like the inherit is not applied. For instance if I change something in the mail.mail_notification_paynow template, I see it change in the email, but that last `<tr>` where the branding stands is never removed...

PR #323, opened 2018-11-09
[12.0][MIG] mass_mailing_partner: Migration to v12.0

sergio-teruel:
cc @Tecnativa


chienandalu (commented):
Tests are failing on external ids of partner which were completely changed in v12

pedrobaeza:
That's why I don't want to rely on demo data for tests. On v13 demo data are going to be contextual to the sector, so it's going to be even worse. cc @Yajo

mass_mailing_partner/views/mail_mass_mailing_contact_view.xml
```diff
@@ -0,0 +1,65 @@
...
+            </field>
+    </record>
+
+    <record model="ir.ui.view" id="view_mail_mass_mailing_contact_form">
+        <field name="name">mail.mass_mailing.contact.form.partner</field>
+        <field name="model">mail.mass_mailing.contact</field>
+        <field name="inherit_id" ref="mass_mailing.view_mail_mass_mailing_contact_form"/>
+        <field name="arch" type="xml">
+            <field name="email" position="after">
+                <field name="partner_id"/>
```
chienandalu, line 26: The field is missplaced in the view:
image

mass_mailing_partner/wizard/partner_mail_list_wizard.py
```diff
@@ -0,0 +1,40 @@
...
+        default=lambda self: self.env.context.get("active_ids"))
+
+    @api.multi
+    def add_to_mail_list(self):
+        contact_obj = self.env['mail.mass_mailing.contact']
+        for partner in self.partner_ids:
+            if not partner.email:
+                raise UserError(_("Partner '%s' has no email.") % partner.name)
+            contact = contact_obj.search([('partner_id', '=', partner.id)])
+            if self.mail_list_id not in contact.mapped('list_ids'):
```
chienandalu: If the contact exists the lists should be added to its list_ids otherwise create it.

chienandalu (changes requested):

yajo:
@pedrobaeza refer to https://github.com/OCA/maintainer-tools/pull/213#issuecomment-268194045 if you want. Setting as off topic.

yajo (approved):
OK when bots get :heavy_check_mark:

mass_mailing_partner/models/res_partner.py
```diff
@@ -14,7 +14,6 @@ class ResPartner(models.Model):
     mass_mailing_contact_ids = fields.One2many(
         string="Mailing contacts",
         oldname="mass_mailing_contacts",
```
chienandalu: This should go away

mass_mailing_partner/models/res_partner.py
```diff
@@ -14,7 +14,6 @@ class ResPartner(models.Model):
...
-        domain=[('opt_out', '=', False)],
```
chienandalu: Why this is needed to be removed?

chienandalu (commented):
Thanks @ernestotejeda ! A final comment. As I stated before: wouldn't it make more sense to add mailing lists to the partner main (and unique?) contact instead of creating a new one everytime we run the wizard?

ernestotejeda:
As I said in PR #324 , I think it is feasible to add a contact _sql_constraints in mailing contact (mail.mass_mailing.contact) to add a unique (email).
If you agree with this, then all the constraints in this module would no longer be necessary to prevent more than one partner (res.partner) from being in the same mailing list more than once, just by making this module dependent of mass_mailing_unique that would be solved, because avoiding that an email is repeated in a list, implies that a contact in a list will not be repeated either. This implies that each partner is associated only with a mailing contact (Many2one instead of Many2many).
What do you think about this?

pedrobaeza:
You both should take into account that we sometimes want to have same the same partner in several mass mailing contacts. For example, with `mass_mailing_list_dynamic`, unless we put another implementation for that one. Check the v11 PR for that where I put several comments about this topic.

chienandalu:
To have the PR @pedrobaeza comments in consideration: https://github.com/OCA/social/pull/284 (abandoned?)

Indeed in v11 there's a problem in the way the unsubscription was managed due to an incomplete frontend support. I'd say that's now solved in v12 (it supports it both ways):
image

I'd go for allowing both ways of having contacts. Although I still think the logic way for the partner is to have a unique relation to a single contact.

ernestotejeda:
As i said in PR #324 , i think is better to leave that migration as it is now, so i think we should leave this PR as it is now too.

mass_mailing_partner/models/res_partner.py
rafaelbn: > This should go away

@chienandalu why? We should put reasons please :busts_in_silhouette: :1234: :smile:

rafaelbn:
Rebuilding runbot for testing

About https://github.com/OCA/social/pull/323#issuecomment-438303490 from @ernestotejeda 

- The functional purpose of this module is relate res.partner with mail.mass_mailing.contact (the README is clear
- The NEW thing / change in v12 (@sergio-teruel I think you are aware) is that now ONE single mail.mass_mailing.contact can be related to N mail.mass_mailing.list. En v11 and before you have to create ONE mail.mass_mailing.contact PER mail.mass_mailing.list . THIS means for example that now you cannot group by mailing list int the list of mass mailing contact

- ***The POINT*** This module must not prevent that 2 or more different mail.mass_mailing.contact has the same email for this purpose we have mass_mailing_unique (could be use toguether o not). We should be aware about:
  - ONE res.partner can be related to ONE mail.mass_mailing.contact.
  - If you create a NEW res.partner and add it to a mass mailing list then create a NEW mail.mass_mailing.contact in that list
  - If you take and EXISTIGN res.partner which already is related to a mail.mass_mailing.contact and you want to add it to another mass mailing list then DON'T create a NEW mail.mass_mailing.contact take the existing mail.mass_mailing.contact related to the res.partner and ADD it to another mass mailing list
  - Can you create N res.partners with the same email so you will have N mass.mailing.contract with the same email? YES (I guess this will be never forbidden)
  - Can you ADD to the same Mass Mailing Lins to N res.partner with the same email? YES if you don't want the use mass_mailing_unique
  - In the CASE you have a mass mailing list and you create a new mail.mass_mailing.contact it will check if there is a partner with that email and related before creating a new one if there is not such a partner with that email then if this option is enabled https://github.com/OCA/social/pull/323/files#diff-1d6f3cba6dabcedb707195bcf524ad25R35 will create a new partner from the mail.mass_mailing.contact
- Because you named in this PR... mass_mailing_list_dynamic should change name to mass_mailing_partner_list_dynamic in v12
- About mass_mailing_partner_list_dynamic and https://github.com/OCA/social/pull/323#issuecomment-438303490 we will take care migrating it to v12. Because now in v12 the game has change and the migration of date will be hard (we will have to merge al mass.mailing.contact of a single res.partner in ONE single mass mailing contact

VERY IMPORTANT ISSUE WE HAD and @chienandalu SOLVED is when the user (customer) MERGE res.partner . I will test this case to see what happends.

mass_mailing_partner/README.rst
```diff
@@ -0,0 +1,104 @@
...
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module links mass-mailing contacts with partners.
+
+Features
+~~~~~~~~
+
+* When creating or saving a mass-mailing contact, partners are matched through
+  email, linking matched partner, or creating a new one if no match and the
+  maling list partner mandatory field is checked.
```
rafaelbn, line 35: @ernestotejeda I have found a BUG. Module creates a NEW res.partner always and it shouldn't.
- Modules should check always if there is a res.partner with the same email to match but
- Only if the mail.mass_mailing.contact belongs to a mail.mass_mailing.list which has set to TRUE the field "Mandatory Partner" must create a NEW res.partner

mass_mailing_partner/models/res_partner.py
rafaelbn: Why is this domain removed? The reason could be important! :smile:

mass_mailing_partner/README.rst
```diff
@@ -0,0 +1,104 @@
...
+* Mailing contacts smart button in partner form.
```
rafaelbn, line 36: Yes, but I've find a new smart-button "Mailing contact" in res.partner. Why are we allowing this? In my side doesn't have any sense as I said and we should FIX this. In the migration from v10 or v11 to v12 then OpenUpgrade should merge all mass_mailing.contact related to a single res.partner to a single mass_mailing.contact

rafaelbn (changes requested):
Please review that:

- All fields from res.partner are filled in the mass_mailng.contact when it gets created
- All filelds from mass_maling.contact are set in the res.partner when created
- Which are the possibilities to have this fields updated over the time?

mass_mailing_partner/README.rst
chienandalu: @rafaelbn Yay!

ernestotejeda:
Well, I don't want to extend the debate any longer, but I think the summary that @rafaelbn has made is very good, and I take the opportunity to respond to these aspects because I think that before making any change to the code, we must have everything clearly

@rafaelbn, I agree with almost everything he has put in, but I do not understand what would be the problem with allowing a partner to be associated with more than one contact.
Since this does not contradict the functional purpose of this module (as long as all the contacts associated with this partner belong to different lists).
Indeed We could restrict that a partner is associated with a single contact (as you said), and that contact associate all the lists that we want to relate with the partner, but this, I think, instead of making everything simpler, it would complicate things, because:
- An idea would be in the create method, call write method in case the contact we are creating match with an existing partner that already has an associated contact, so that the lists are added to this, and return that record instead of calling super. About this we were debating in the PR #324 but referring to the email field instead partner_id field. But the problem is that we should also overwrite the write method when a contact is modified by changing the partner for one that already has an associated contact, we can add the lists to that contact, but then we would have to call the unlink method to eliminate the contact that was being modified, and here in this last case it does not behave in the expected way, I already implemented it to make sure, and for that reason I discard that option.

- Another idea would be to put a constraint to avoid always having two contacts associated with the same partner. But then we should make modifications in other places so there is no conflict with this. I have already found these:

1- the wizard that exists to associate several partners to a list:
* For each partner, it should be checked that if this already has a contact associated, that contact would be added to the selected list, otherwise a new one would be created. (This has already been mentioned before)

2- In res.partner:
* The Many2many with mail.mass_mailing.contact would not make sense, therefore, it would be then a many2one.
* The smart-button "Mailing contact" would not make sense either, but it could be changed by one called "Mailing list" in order to see the mailing lists associated with the partner's contact.

3- In the mass_mailing_list_dynamic module (in case you were to migrate to v12)
* During the synchronization, after finding the partners that matches with the domain, you would only have to create a new contact if this partner no longer has a contact associated with it.
This would be a change that would be made when the module is migrated.

4- As I said in the PR #324 . The module website_mass_mailing adds the snippet that gives the user the possibility to register in a mass mailing list from the website and each time a user uses this to subscribe to this list, this module creates a new contact mailing although there is another contact with that e-mail. So, if there is an AA contact with an email aa@example.com, each time that email subscribes to a list through the website, a contact will be created with that email and therefore associated with the AA partner.
What I can think of to solve this is to create an auto_install = True module that depends on website_mass_mailing and mass_mailing_partner, this module would redefine the 'subscribe' method in the snippet controller to create a contact only if one does not already exist with the same partner.

Anyway, I think that if the fact that a partner can be related to more than one contact is not wrong as a concept, it would be better to leave it as is, so as not to have to intervene in the aforementioned aspects.

mass_mailing_partner/README.rst
ernestotejeda: @rafaelbn  I tested on the runbot and I could not reproduce that error

ernestotejeda:
the domain domain = [('opt_out', '=', False)] in the field mass_mailing_contact_ids was removed because the mail.mass_mailing.contact doesn't have that field now in v12, mail.mass_mailing class. list_contact_rel is the class that has this field now.

chienandalu:
Good analysis @ernestotejeda  I'm on keeping both ways of behaviour so we avoid those issues. So:

- I'd go for the wizard approach for adding a list to the main partner contact. The code is almost done :smile: : https://github.com/OCA/social/pull/323/files#diff-e7229f99dcf57bbe7c6c20faef2c10cfR27
- (maybe as a post-PR improvement) I'd provide a tool to merge partner contacts into a single one. Something like:
```python
list_ids = partner.contact_ids.mapped('list_ids')
partner.contact_ids[1:].unlink()
partner.contact_ids[:1].list_ids |= list_ids  # Or maybe on the relation model to keep the subscription dates and states
```
- The last one could be even triggered in the `res.partner` `write()`

ernestotejeda:
I agree with you @chienandalu, thanks

ernestotejeda:
@rafaelbn when a contact is associated with a partner the following contact fields:
name, email, title_id, company_name, country_id and tag_ids are read-only and each time one of these fields is modified in the related partner, the corresponding field is updated in the contact. That is the way in which the module keeps the fields updated at the moment. The only problem is that the company is not updating well, I have to fix that.

sergio-teruel:
@ernestotejeda @chienandalu Whats happend with this? :smile:

ernestotejeda:
Well, @sergio-teruel , I think all the suggestions have been resolved.

rafaelbn (dismissed):

rafaelbn (commented):
This migration is critical I need to analyze deeper. It could be a disaster when migrating a customer from v10 to v12 for example. I cannot approve with out testing it deeper.

mass_mailing_partner/models/res_partner.py
victor-ditay: @rafaelbn oldaname is used by Odoo to rename fields automatic DB tables when upgrade modules or migration happens then if the migration occurs in order, it should be like from 9 to 10 then from 10 to 11, etc.. 
This means here makes no sense because this oldname could only exist before version 10 (is the origin of this module migration)

mass_mailing_partner/models/res_partner.py
victor-ditay: This not sure @chienandalu could you reply?

victor-ditay (commented):
I would try finish this PR as I have next module depends on

mass_mailing_partner/views/mail_mass_mailing_contact_view.xml
victor-ditay: mass_mailing_contact

Seem solved @chienandalu

chienandalu:
> I would try finish this PR as I have next module depends on

Thanks, @victormmtorres I'd say this was almost ready but I lost track of it. I you'd review if all the comments were honored and this is ready to review again let me know and I do my final one :)

victor-ditay:
@chienandalu I kind of lost on that amount of comments on many people on this PR, I fix one I found is clear. But other maybe already done by @ernestotejeda on his last commit ca741868566cb2773e9d36b41dc102e8916fd89f

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -45,6 +45,10 @@ def create(self, vals):
             record._set_partner()
         record._onchange_partner_mass_mailing_partner()
         new_vals = record._convert_to_write(record._cache)
+        new_vals.update(
+            subscription_list_ids=vals.get('subscription_list_ids', False),
```
victor-ditay, line 49: This is breaking test of previous modules on the repository as `mail_tracking_mass_mailing` one of the test asume the the contact is not subscribed before. So whether fix this module would require reduce the improvements or we should refactor first the other module. I guess the PSC of this repository should decide 

cc @chienandalu @ernestotejeda @pedrobaeza @rafaelbn

pedrobaeza: The problem was with the code in module `mail_tracking`, that was calling method `write` inside other `write` call. Fixed in 007d8ada83dec2e1e082064b7a256907baba4817.

PR #324, opened 2018-11-12
12.0 mig mass_mailing_unique

ernestotejeda:
* Adapt existing tests
* Add and tests
* Add constraints

Cc @Tecnativa


mass_mailing_unique/README.rst
```diff
@@ -0,0 +1,97 @@
...
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/205/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module extends the functionality of mass mailing lists to disable
+duplicate entries in list names and contact emails per list.
+
+This way you will avoid sending the same message more than once to the same
```
chienandalu, line 31: Indeed Odoo drops duplicate addresses in every mass mailing

chienandalu (changes requested):
There's an issue that could clash with the `mass_mailing_partner` module. Whenever you create a new contact it relates the list another contact with the same address has related with so no new contacts can be created with the same address:

peek 13-11-2018 08-36

ernestotejeda:
In v12, unlike v11, a mailing contact has a One2many with subscriptions to lists (mail.mass_mailing.list_contact_rel), so the same mailing contact (mail.mass_mailing.contact) can be related to mailing lists (mail. mass_mailing.list) from which he / she wants to receive emails and at the same time can be related to other mailing lists from which he / she doesn't want to receive mails, therefore, now in v12, to restrict that an email is no longer once in the same list, it's enough to add a contact _sql_constraints that adds a unique (email).
What do you think about this?

chienandalu:
Well, the relation in v11 was similar. Indeed, we overlooked this feature a little bit since it was not completely usable (for instance in unsubscribes)

What I was noting is that when you create a new contact with the same address another contact has, now we're getting a link automatically to the lists that contact has. This is not happening in standard. And I think both options should be available.

You dropped the lower case check as well so we can have duplicates just varying the email case again.

Besides, this is going to clash against your migration of mass_mailing_partner, that precisely creates several contacts :smile: (anyway, as I stated there, I think subscriptions should be related to a single contact as well).

pedrobaeza:
Well, I haven't dug too much on this. What I want is a happy world and that everything fits the best way :smile: , so go on with the best design you think of having into account all the possible modules.

ernestotejeda:
I have done a bit of research about this topic, and I think we should better leave it as it is now in this PR cause it continues to meet the pursued objective of this module. If we put some restriction so  there is no more than one contact with the same email, as I had said before, it would conflict with the module website_mass_mailing, because this module adds the snippet that gives the user the possibility to register in a mass mailing list from the website and each time a user uses this to subscribe to a list, this module creates a new contact mailing although there is already another contact with that email. I don't know either if exists another module out there (maybe in the enterprise edition) that does the same. We could redefine the create method, and if amount the parameters we get an email address that already exists in another contact, we should modify that contact instead of creating a new one, but I do not think it's a good idea to invoke the write method from the create method.

yajo:
> We could redefine the create method, and if amount the parameters we get an email address that already exists in another contact, we should modify that contact instead of creating a new one, but I do not think it's a good idea to invoke the write method from the create method.

Indeed it seems a clunky solution. Maybe we have to stay with current implementation and IMHO, now that you have done such research, it's a good time to open a little PR to Odoo master, so the create/write election is done via the snippet controller, and we remove this situation for v14+. Do you feel like opening that? Ping me there if you do.

mass_mailing_unique/hooks.py
```diff
@@ -0,0 +1,46 @@
...
+                  GROUP BY name
+                  HAVING COUNT(id) > 1""")
+    for result in cr.fetchall():
+        errors.append(
+            _("There are {1} lists with name {0}.").format(*result))
+
+    # Abort if duplicates are found
+    if errors:
+        raise ValidationError(
+            _("Fix this before installing:") +
```
yajo: No need to translate, there's no context language.

mass_mailing_unique/hooks.py
```diff
@@ -0,0 +1,46 @@
...
+                  FROM
+                    mail_mass_mailing_contact AS c
+                    INNER JOIN mail_mass_mailing_contact_list_rel AS cl
+                      ON cl.contact_id = c.id
+                    INNER JOIN mail_mass_mailing_list AS l ON cl.list_id = l.id
+                  GROUP BY l.name, e
+                  HAVING COUNT(c.id) > 1""")
+    for result in cr.fetchall():
+        errors.append(
+            _("{0} appears {2} times in list {1}.").format(*result))
```
yajo: No need to translate, there's no context language.

mass_mailing_unique/hooks.py
```diff
@@ -0,0 +1,46 @@
...
+
+    # Search for duplicates in list's name
+    cr.execute("""SELECT name, COUNT(id)
+                  FROM mail_mass_mailing_list
+                  GROUP BY name
+                  HAVING COUNT(id) > 1""")
+    for result in cr.fetchall():
+        errors.append(
+            _("There are {1} lists with name {0}.").format(*result))
```
yajo: No need to translate, there's no context language.

pedrobaeza:
Well, it's not so bad solution at all. We are already doing that on certain parts and it's working without problems. It's not a problem to call a write on other records on the create, and return that record instead of calling super in that case.

pedrobaeza:
But I'm thinking that this would avoid the creation of explicit isolated contacts, so better to leave it as is indeed.

mass_mailing_unique/models/mail_mass_mailing_list.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class MailMassMailingList(models.Model):
+    _inherit = 'mail.mass_mailing.list'
+
+    @api.constrains('contact_ids')
+    def _check_contact_ids_email(self):
+        contact_obj = self.env['mail.mass_mailing.contact']
```
yajo: These 3 constraints do the same thing. Can't we fall back to just one?

For instance, I think the most readable one is the one found in `mail.mass_mailing.contact`. Then, here you could just do:

```python
self.mapped("contact_ids")._check_email_list_ids()
```

mass_mailing_unique/models/mail_mass_mailing_list_contact_rel.py
```diff
@@ -0,0 +1,17 @@
...
+from odoo import api, models, _
+from odoo.exceptions import ValidationError
+
+
+class MailMassMailingContactListRel(models.Model):
+    _inherit = 'mail.mass_mailing.list_contact_rel'
+
+    @api.constrains('contact_id', 'list_id')
+    def _check_contact_id_partner_id_list_id(self):
+        for rel in self:
```
yajo: Again, here you can do:

```python
self.mapped("contact_id")._check_email_list_ids()
```

mass_mailing_unique/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,17 @@
...
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class MailMassMailingContact(models.Model):
+    _inherit = 'mail.mass_mailing.contact'
+
+    @api.constrains('email', 'list_ids')
+    def _check_email_list_ids(self):
+        for contact in self:
```
yajo, line 13: I'm afraid this for could lead to bottlenecks, mostly when using suggestions from below. How about using `set()`s, which allow to check for duplicates? One idea:

```python
others = self.mapped("list_ids.contact_ids") - self
if set(self.mapped("email")) & set(others.mapped("email")):
    raise ...
```

yajo (changes requested):

chienandalu:
Ok, for one side the unique contatc constraint is already implemented in Odoo standard: https://github.com/odoo/odoo/blob/12.0/addons/mass_mailing/models/mass_mailing.py#L61
image

So why don't we keep the original behaviour of the module so we cover all the cases?

mass_mailing_unique/models/mail_mass_mailing_contact.py
ernestotejeda: The problem of what you propose is that when you try to create a list with two contacts with the same email address, the exception will not be launched

ernestotejeda:
@chienandalu, the objective that this module pursues is that two contacts with the same email address can not be associated to the same mailing list.

chienandalu:
Sure, we're on the same page there! :sweat_smile: I tested it on runbot and it seems to work fin now. Anyway, the case issue still exists. I dont't know if there could be a way to solve it.

rafaelbn:
I've tested this PR @chienandalu and I cannot find any issue. I don't understand the problem here https://github.com/OCA/social/pull/324#pullrequestreview-174222810 :cry: 

Could you please clarify a little bit? go here , make that you will find this problem.... :pray: 

Isn't this PR a trivial migration from v11 to v12? Let's go!

rafaelbn (approved):
Tested functionally :+1: Thanks

mass_mailing_unique/models/mail_mass_mailing_contact.py
yajo: Are you sure? AFAICS that's exactly what it does... It just avoids many loops to make sure the method scales fine.

chienandalu (approved):
Sure @rafaelbn it's solved by now! :smile: Thanks @ernestotejeda !

mass_mailing_unique/models/mail_mass_mailing_contact.py
ernestotejeda: Suppose that the write method is called to modify two contacts (contact_1 and contact_2) so that both have the same email (contact@example.com) and belong to the same lists (L_1 and L_2), which had not yet been associated to no contact. In that case, the error will not be launched, because the intesection between the sets set (contact@example.com) and set () will be empty.

yajo: Well, you're absolutely right! Thanks :wink:

yajo (approved):

yajo:
Please could you squash migration-related commits to merge?

PR #326, opened 2018-11-16
[11.0][MIG] Migrate module mail_inline_css

p-tombez:


pedrobaeza:
Thanks for this, Patrick. Isn't Odoo already doing this?

About the module name, please avoid plurals on it.

yajo:
This is done by Odoo on the JS side, using this file: https://github.com/odoo/odoo/blob/11.0/addons/web_editor/static/src/js/editor/transcoder.js

chienandalu:
Besides, Odoo wipes the `<style>` tag from the email bodys: https://github.com/odoo/odoo/blob/HEAD/odoo/tools/mail.py#L28

tschanzt (approved):
Looks good.

simahawk:
@Yajo am I wrong or this works only if you edit it via editor?

It happens only here https://github.com/odoo/odoo/blob/1868713dbd07e0b518f91dffe73e62d85e6ab9a6/addons/web_editor/static/src/js/inline.js#L40

and here 

https://github.com/odoo/odoo/blob/1868713dbd07e0b518f91dffe73e62d85e6ab9a6/addons/web_editor/static/src/js/backend/fields.js#L185

so if you do not use the editor - which is very likely since if you have a nice corporate branding template, you are going to screw it - then is not enough...

mail_inline_styles/__manifest__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Mail Inline Styles",
+    "version": "11.0.1.0.0",
+    "category": "Social Network",
```
simahawk: is this categ correct?

mail_inline_styles/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Akim Juillerat <akim.juillerat@camptocamp.com>
```
simahawk: you can add me n you here ;)

tschanzt:
@chienandalu that is the purpose of it. The premailer should convert it to style attributes but you can write normal css in the <style> tag which makes it nicer for you.

mail_inline_styles/tests/test_mail_inline_styles.py
```diff
@@ -0,0 +1,62 @@
...
+            l = style.split(':')
+            res[l[0].strip()] = l[1].strip()
+        return res
+
+    def find_by_id(self, node, html_id):
+        return node.xpath('//*[@id="{}"]'.format(html_id))
+
+    def assertNodeStyle(self, node, expected):
+        self.assertIn('style', node.attrib)
+        self.assertEqual(self.parse_node_style(node), expected)
```
simahawk: isn't `expected` a dict? Then you should use `assertDictEqual`

yajo:
Oh I understand now. OK then this is awesome! :rocket: 

But, will it conflict with Odoo's system that I linked above? :thinking:

chienandalu (changes requested):
Ok, I understand now what you achieve with this. Could you explain it better in the readme, please :wink: Also fix runbot warnings.

Anyway, it'd be great to come to a solution in the way Odoo did in v12 for mass mailings, allowing to use header styles instead: https://github.com/odoo/odoo/commit/967d228a44080381d185c252ff29c993b2b88429

chienandalu:
Not really, @Yajo I tried it on runbot and the styles are already rendered inline when you open the template in the editor :smile:

simahawk (changes requested):
LG, some minor things.
plus... my bad, remove the 's' at the end :wink:

simahawk:
> But, will it conflict with Odoo's system that I linked above? thinking

not sure... we can double check on our customer project tho.

mail_inline_styles/__manifest__.py
p-tombez: Don't know, all the other modules in this repo have this one

mail_inline_styles/tests/test_mail_inline_styles.py
p-tombez: ```assertEqual``` is already calling ```assertDictEqual``` internally, as per the doc https://docs.python.org/3/library/unittest.html#type-specific-methods

simahawk: I'm sure, it just make clear that you are testing a dict. Plus, if any one changes data structure you'll get a warning. 
Not mandatory here... is just an approach that I tend to take and suggest... feel free to ignore it ;)

yajo:
I think the only critical part where you should check for conflicts is mass mailing, where the JS transpiler is actually used. For normal mail templates, it's not used AFAIK.

simahawk:
@p-tombez I've just discovered this :sweat_smile: https://github.com/OCA/social/tree/10.0/mail_inline_css

It's based on premailer as well.
We should migrate it and add our changes there I think

simahawk (approved):
LG, you have to squash the last commit.
Not a blocker but next time you could generate the final readme https://github.com/OCA/maintainer-tools#readme-generator

Another thing you can do is to set some `noqa` here https://codecov.io/gh/OCA/social/src/11.0_add_mail_styles_inline/mail_inline_css/models/mail_template.py to fix test coverage

p-tombez:
@simahawk I generated the README (I was told it is generated during the build, so I didn't add it on the first place).
```noqa``` is for flake8 checks, it has nothing to do with code coverage. I didn't find anything (despite ignoring an entire file) to solve the issue

simahawk:
`# pragma: no cover`

simahawk:
@chienandalu fine for you? :)

simahawk:
can anyone push the green button? :)

chienandalu (approved):
:+1:

PR #331, opened 2018-11-27
[11.0] Add module mail_browser_view

p-tombez:
This module enables you to add a link in your mail templates, so the users can view the resulting email in the browser in case they are badly displayed in their clients.


pedrobaeza:
Does this work for mass mailing as well?

mail_browser_view/controllers/brower_view.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2018 Camptocamp SA
```
simahawk: filename mispelled

mail_browser_view/models/mail_mail.py
```diff
@@ -0,0 +1,105 @@
...
+    _inherit = 'mail.mail'
+
+    @api.model
+    def create(self, vals):
+        rec = super(Mail, self).create(vals)
+        rec._replace_view_url()
+        return rec
+
+    @api.model
+    def get_record_for_token(self, token):
```
simahawk, line 39: docstring

mail_browser_view/models/mail_mail.py
```diff
@@ -0,0 +1,105 @@
...
+            token = urlsafe_b64decode(token).decode()
+            access_token, rec_id = token[:32], token[32:]
+            rec = self.sudo().browse(int(rec_id))
+            return (rec.access_token == access_token and
+                    rec.is_token_alive and rec or self)
+        except (ValueError, MissingError, binascii.Error):
+            return self
+
+    @api.multi
+    def _replace_view_url(self):
```
simahawk: docstring

p-tombez:
@pedrobaeza It will need to be tested, but AFAIK it should work unless mass mailing bypasses completely the mail creation process

pedrobaeza:
Thanks for telling.

mail_browser_view/models/mail_mail.py
```diff
@@ -0,0 +1,105 @@
...
+                mail_date = fields.Datetime.from_string(
+                    rec.mail_message_id.date
+                )
+                rec.is_token_alive = (
+                    (mail_date + max_delta).date() >= date.today()
+                )
+        else:
+            self.is_token_alive = True
+
+    access_token = fields.Char(
```
simahawk: fields definition should stay above

mail_browser_view/models/mail_mail.py
```diff
@@ -0,0 +1,105 @@
...
+        'Security Token',
+        compute="_compute_access_token", store=True
```
simahawk: do you really need this to be a field and be stored? You are passing the id of the record in the url...
Either you rely on the ID and you use the token just to make some noise in the URL (no need to store it)
or you use the token to retrieve the message (no need to use the ID in the URL).

mail_browser_view/models/mail_mail.py
```diff
@@ -0,0 +1,106 @@
...
+    @api.model
+    def get_record_for_token(self, token):
+        try:
+            token = urlsafe_b64decode(token).decode()
+            access_token, rec_id = token[:32], token[32:]
+            rec = self.sudo().browse(int(rec_id))
+            return (rec.access_token == access_token and
+                    rec.is_token_alive and rec or self)
+        except (ValueError, MissingError, binascii.Error):
+            return self
```
simahawk: I'd return None or self.browse() as none value since you might call this from a recordset and you'll get the same record back.

simahawk (changes requested):
@p-tombez 2 lines in the PR description won't hurt ;)

mail_browser_view/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "version": "11.0.1.0.0",
+    "category": "Social Network",
+    "website": "https://github.com/OCA/social",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "mail",
+        "email_template_qweb",
```
simahawk: `email_template_qweb` I don't think this is a dependency, you don't care about how the body_html is rendered.
But you do care about styles and to make sure they are there I think we should depend on mail_inline_style.
If not, I think we should strongly encourage ppl to use it in the readme.

rafaelbn:
Wow! Great! :smile:

mail_browser_view/models/mail_mail.py
p-tombez: IMO, I think it adds a bit of "strength", like if you had a login and a password. But if you think it's too much, I'll remove the ID from the URL and rely only on the access token

simahawk: fine, you can keep them to strength the uniqueness but then you use both to search the record ;)

p-tombez: Currently, I do a ```browse``` with the ID and check if the access_token is matching if a record is found. Should I do a ```search``` with both values instead ? Or use ```read``` ? What's best ?

simahawk: search using both

mail_browser_view/models/mail_mail.py
```diff
@@ -0,0 +1,106 @@
...
+        else:
+            self.is_token_alive = True
+
+    access_token = fields.Char(
+        'Security Token',
+        compute="_compute_access_token", store=True
+    )
+    view_in_browser_url = fields.Char(
+        'View URL',
+        compute="_compute_browser_url", store=True
```
simahawk: this should not be stored

mail_browser_view/models/mail_mail.py
```diff
@@ -0,0 +1,118 @@
...
+            max_delta = relativedelta(hours=expiration_time)
+            for rec in self:
+                mail_date = fields.Datetime.from_string(
+                    rec.mail_message_id.date
+                )
+                rec.is_token_alive = (
+                    (mail_date + max_delta).date() >= date.today()
+                )
+        else:
+            self.is_token_alive = True
```
yajo: Shouldn't this be `self.update({'is_token_alive': True})`? For multiple records...

yajo (approved):
Cool!

mail_browser_view/models/mail_mail.py
p-tombez: Absolutely

mail_browser_view/models/mail_mail.py
```diff
@@ -0,0 +1,136 @@
...
+        ).replace(
+            scheme=base.scheme, netloc=base.netloc
+        ).to_url()
+
+    @api.multi
+    def _replace_view_url(self):
+        """Replace placeholders with record URL.
+
+        Replace the 'href' attribute of all `<a></a>` tags
+        having the 'id' attribute equal to 'view_in_browser_url'
```
simahawk: must be a class for multiple links ;)

mail_browser_view/models/mail_mail.py
```diff
@@ -0,0 +1,136 @@
...
+            max_delta = relativedelta(hours=expiration_time)
+            for rec in self:
+                mail_date = fields.Datetime.from_string(
+                    rec.mail_message_id.date
+                )
+                rec.is_token_alive = (
+                    (mail_date + max_delta).date() >= date.today()
+                )
+        else:
+            self.write({'is_token_alive': True})
```
simahawk: use `self.update`

mail_browser_view/readme/USAGE.rst
```diff
@@ -0,0 +1,14 @@
+Upon module installation, a secure token will be generated for each mail,
+allowing it to be reached *via* a constructed URL.
+You can then put the following placeholder::
+
+    <a href="#" id="view_in_browser_url">View this mail in browser</a>
```
simahawk: you know... use `class` ;)

yajo:
You can also put a href that it should search and replace.

simahawk:
@Yajo right, but having a fixed class makes sure that we can style it always in the same way across templates :wink:

requirements.txt
```diff
@@ -0,0 +1 @@
+cssselect
```
simahawk: you don't need this, you can use xpath

p-tombez: I found it cleaner to use this lib (which was initially included in ```lxml```) instead of writing some long obscure xpath expression like: ```a[@class and contains(concat(' ', normalize-space(@class), ' '), ' view_in_browser_url ')]```
But I understand if it's better not to use it. Your opinion ?

yajo: If you want to select by class, odoo contains one helper: `//*[hasclass("view_in_browser_url")]`. It should be enough.

p-tombez: Oh, didn't know it can be used outside qweb views

simahawk:
@p-tombez LG. One last thing to do: move test cov next to 100%.
Can you add a test on the controller? No need for HttpTestCase.
Just instantiate a controller instance and call `email_view`:
1. test not found is raised
2. test email content is rendered

Also, this condition is not tested https://codecov.io/gh/OCA/social/compare/e9c655c1c60f915262ed25283c6f54304bc1d5a3...3b3c1047de63dae7c0718295ea87eda2a2d131ad/diff#D13-136

simahawk (approved):
LG, thanks!

PR #333, opened 2018-11-28
[10.0] ADD mail_check_mailbox_size

eLBati:
Send an email with a summary of remote IMAP mailbox folders, with their total space used.

TODO
check v12 fixes at https://github.com/OCA/social/pull/342


pedrobaeza:
Do you think this is interesting at Odoo level? I see this more like a sysadmin task to be monitored with Sentry or similar...

hbrunn (approved):
I think you should consider moving your html generation to qweb an feed that with some generator for the imap folders, this gives you customization for free, but I consider this a future extension.

@pedrobaeza I agree that in a more professional setup people should be using specialized tools, but I see the use case for small organizations that struggle with their IT stuff anyways

pedrobaeza:
OK, but this is very limited also to IMAP (when a POP with no message removal can have the same problem...)

eLBati:
@pedrobaeza I think it's interesting at odoo level because the functionality becomes accessible to users without sysadmin skills.

eLBati:
> when a POP with no message removal can have the same problem

True, but I think we cannot do much in that case

eLBati:
@OCA/social-maintainers what do you think?

mail_check_mailbox_size/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "maintainers": ["eLBati"],
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "fetchmail",
+    ],
+    "data": [
+        "views/fetchmail_view.xml",
+        "data/cron_data.xml"
```
lepistone:
```suggestion
        "data/cron_data.xml",
```

mail_check_mailbox_size/models/fetchmail.py
```diff
@@ -0,0 +1,136 @@
...
+        if self.env.context.get('fetchmail_cron_running'):
+            return res
+        try:
+            cron = self.env.ref(
+                'mail_check_mailbox_size.ir_cron_check_mailbox_size')
+            cron.toggle(model=self._name, domain=[
+                ('state', '=', 'done'), ('type', '=', 'imap'),
+                ('check_mailbox_size', '=', True),
+            ])
+        except ValueError:
```
lepistone, line 134: Please comment when this can happen.

mail_check_mailbox_size/README.rst
```diff
@@ -0,0 +1,90 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/social-10-0/social-10-0-mail_check_mailbox_size
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/205/10.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+Send an email with a summary of remote IMAP mailbox folders, with their total space used.
```
lepistone, line 28: It might be interesting to mention that in some context (for example the Italian electronic invoices sent by email) the business can benefit to know that emails can go through.

lepistone (approved):
Thanks! Please check a couple minor comments.

eLBati:
@lepistone thanks, done

rafaelbn:
Hi @eLBati so great! just missing some test? :smiley: 

> codecov/patch — 25.39% of diff hit (target 95.15%)

eLBati:
@rafaelbn tests would be useful of course. Any experience with IMAP testing? Using `localmail`?

hbrunn:
better use mock

rafaelbn:
I don't know how to make this test, maybe @chienandalu or @Yajo .... as Holder say... Mock

https://stackoverflow.com/questions/3180192/what-is-the-easiest-way-to-mock-an-imap-or-pop-server-for-unit-tests

In any case I'm not blocking! This could be hard to test but I ask trying to maintain the coverage

yajo:
@rafaelbn that link is for java!

Mock is documented here, indeed possibly the best approach for tests here: https://docs.python.org/3/library/unittest.mock.html

hbrunn:
please rebase

eLBati:
Done

PR #336, opened 2018-11-29
[FIX] mail_tracking: wrong bounced partner

chienandalu:
- Partner get bounced when his associated email is in a hard bounced
tracking. It didn't matter if the hard bounce event was solved in that
recipient later and leaded to an increasing number of false positives.
- We also use the email_bounced_set() method to get the whole bounce
info in case positive.

cc @Tecnativa

### DIFF
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -94,12 +94,16 @@ class MailTrackingEmail(models.Model):
 
     @api.model
     def email_is_bounced(self, email):
-        if email:
-            return self.search_count([
-                ('recipient_address', '=', email.lower()),
-                ('state', 'in', ('error', 'rejected', 'spam', 'bounced')),
-            ]) > 0
-        return False
+        if not email:
+            return False
+        res = self._email_last_tracking_state(email)
+        return res and res[0].get('state', '') in ['rejected', 'error',
+                                                   'spam', 'bounced']
+
+    @api.model
+    def _email_last_tracking_state(self, email):
+        return self.search_read([('recipient_address', '=', email.lower())],
+                                ['state'], limit=1, order='time DESC')
 
     @api.model
     def email_score_from_email(self, email):
--- a/mail_tracking/models/res_partner.py
+++ b/mail_tracking/models/res_partner.py
@@ -40,10 +40,15 @@ class ResPartner(models.Model):
         return partners.write({'email_bounced': True})
 
     def write(self, vals):
-        email = vals.get('email')
-        if email is not None:
-            vals['email'] = email.lower() if email else False
-            vals['email_bounced'] = (
-                bool(email) and
-                self.env['mail.tracking.email'].email_is_bounced(email))
+        if 'email' not in vals:
+            return super(ResPartner, self).write(vals)
+        email = vals['email'].lower() if vals['email'] else False
+        mte_obj = self.env['mail.tracking.email']
+        if not mte_obj.email_is_bounced(email):
+            vals['email_bounced'] = False
+            return super(ResPartner, self).write(vals)
+        res = mte_obj._email_last_tracking_state(email)
+        tracking = mte_obj.browse(res[0].get('id'))
+        event = tracking.tracking_event_ids[:1]
+        self.email_bounced_set(tracking, event.error_details, event)
         return super(ResPartner, self).write(vals)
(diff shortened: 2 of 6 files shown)


mail_tracking/models/mail_tracking_email.py
```diff
@@ -94,12 +94,16 @@ class MailTrackingEmail(models.Model):
...
-        return False
+        if not email:
+            return False
+        res = self._email_last_tracking_state(email)
+        return res and res[0].get('state', '') in ['rejected',
+                                                   'spam', 'bounced']
+
+    @api.model
+    def _email_last_tracking_state(self, email):
+        return self.search_read([('recipient_address', '=', email.lower())],
```
pedrobaeza, line 105: Now there's no real benefit in using `read` instead of `browse`. Even worse, it doesn't use cache.

mail_tracking/models/res_partner.py
```diff
@@ -40,10 +40,15 @@ def email_bounced_set(self, tracking_emails, reason, event=None):
...
     def write(self, vals):
-        email = vals.get('email')
-        if email is not None:
-            vals['email'] = email.lower() if email else False
-            vals['email_bounced'] = (
-                bool(email) and
-                self.env['mail.tracking.email'].email_is_bounced(email))
+        if 'email' not in vals:
+            return super(ResPartner, self).write(vals)
+        vals['email'] = vals['email'].lower() if vals['email'] else False
```
pedrobaeza: You shouldn't lower the email here.

mail_tracking/models/res_partner.py
```diff
@@ -40,10 +40,15 @@ def email_bounced_set(self, tracking_emails, reason, event=None):
...
+        mte_obj = self.env['mail.tracking.email']
+        if not mte_obj.email_is_bounced(vals['email']):
+            vals['email_bounced'] = False
+            return super(ResPartner, self).write(vals)
+        res = mte_obj._email_last_tracking_state(vals['email'])
+        tracking = mte_obj.browse(res.get('id'))
+        event = tracking.event_ids[:1]
+        self.email_bounced_set(reason=event.reason, event=event)
```
pedrobaeza: Why logging this on write?

pedrobaeza (changes requested):
Travis is red

ernestotejeda:
With the changes that have been made in this PR, when the email of a partner is modified the value of the field *Email Bounced* will be established correctly, but what happens if a partner that already has in True the field *Email Bounced* but receives an email correctly, I think that in that case, the *Email Bounced* field must be set to false. Maybe in the event_create method of the MailTrackingEmail class that can be handled.

mail_tracking/models/res_partner.py
chienandalu: I agree, but this behaviour is expected by the original code and there's even a test for it.

mail_tracking/models/res_partner.py
chienandalu: Well, we're checking that the email we're setting to the partner isn't already in a bounce state and logging that state to the partner. For instance, you merge a partner into another, or you use an email that is in hard-bounce state. This will only be checked when the email is set.

mail_tracking/models/mail_tracking_email.py
chienandalu: Is it the same with search()?

chienandalu:
@ernestotejeda The problem there is that you'd have to check every time you send a mail if:

- the recipient has a previous hard-bounce state
- case positive: the recipient is associated to a partner with bounced set.

I'd say it's a little overloading. In any case, the spirit of the field is to avoid sending emails to erroneous recipients.

rafaelbn:
@pedrobaeza travis is :green_apple: 

Just  

> codecov/patch — 76.47% of diff hit (target 96.78%) 

I'm testing this in production and by the moment :+1:

mail_tracking/models/res_partner.py
pedrobaeza: But this is not correct. You must change it.

pedrobaeza:
There are unattended comments and something very ugly, like lowering the email. That should be done on a different field or when comparing, never on the original field.

mail_tracking/models/res_partner.py
chienandalu: Ok, I dropped this behavior

mail_tracking/models/mail_tracking_email.py
chienandalu: I can't browse here.

yajo (approved):

pedrobaeza (approved):

PR #344, opened 2018-12-27
POC Depower Mail

gdgellatly:
@rvalyi @pedrobaeza Try this and see what you think. It is significantly different to v11 module which didn't bother with the powered by messages you are seeing. Those messages are quite pervasive, and in all different structures so I tried to smartly remove any empty parents.


gdgellatly:
After doing some tests, I decided it looked nicer leaving the blank space and elements in place, so now we just remove children and set text to None.

rvalyi:
thanks @gdgellatly I will test this today. My goal is to partially get rid of our own Akretion module for this: https://github.com/akretion/odoo-usability/tree/10.0/mail_usability and join forces here.

gdgellatly:
@rvalyi @pedrobaeza Finally I realised the issue was translations, so in a sense I went back to the old way, but added in the powered by support. Because the new views are just views, and post_process is automatically called for each email it was just easier to validate their arch for tests.

ernestotejeda (commented):
For me the code is fine. I have done some functional tests and they have worked well.

Maybe this PR can be used to put the website correctly in the manifest: `"website": "https://github.com/OCA/social",`
In the migration commit to v12, the authors 'Tecnativa' and 'Eficent' were remove in the manifest. I don't know if that is correct, @pedrobaeza  what you think about it?. If not, then this PR could also be used for that.

pedrobaeza:
Is there some code kept from our version? If so, authorship should be preserved. If not, let's reset it.

gdgellatly:
@pedrobaeza Not in the original port, there was zero lines of actual code. In this PR, I'm not so sure. Its been a while.

mail_debrand/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Mail Debrand",
     "summary": "Remove Odoo branding in sent emails",
-    "version": "12.0.1.0.0",
+    "version": "12.0.2.0.0",
     "category": "Social Network",
     "website": "https://odoo-community.org/",
```
chienandalu: https://github.com/OCA/social

chienandalu (changes requested):
It doesn't seem to work. There seem to be some tag left after the clean up that is spoiling the render action

mail_debrand_error

gdgellatly:
@devishree-git can you look at for me please, no time atm.

devishree-git:
> @devishree-git can you look at for me please, no time atm.
Hello Graeme,
Please check following for fix I applied.
https://github.com/odoonz/social/compare/12.0-depower-mail...devishree-git:patch-1

Thanks You,
Devishree

chienandalu:
here: https://github.com/odoonz/social/compare/12.0-depower-mail...devishree-git:patch-1#diff-3ab12b61215d29e15c8a26cde5d92b9bR15

@devishree-git I'd try to get something more deterministic, don't you think? If you just search for "Odoo" string you can end up getting rid of things you didn't want to

gdgellatly:
@devishree-git That won't work. The existing code was mostly good. Just need to work out the tag mismatches. Check the error from @chienandalu , the current code is easy to test with, you can just pass in any view arch in shell to test and check the output.

ernestotejeda:
About the error from @chienandalu, the problem is that the etree.fromstring method is receiving by parameter an incorrect xml 
(https://github.com/OCA/social/pull/344/files#diff-3ab12b61215d29e15c8a26cde5d92b9bR19).
In the specific case of the template `Auth Signup: Odoo Account Created` (template related to the error)
(https://github.com/odoo/odoo/blob/2e7cbca0e957578518e206f5a9a773fbf2d58be4/addons/auth_signup/data/auth_signup_data.xml#L197), 
in one part it is incorporated the user signature (https://github.com/odoo/odoo/blob/2e7cbca0e957578518e206f5a9a773fbf2d58be4/addons/auth_signup/data/auth_signup_data.xml#L242). 
This signature is an editable html field that does not always keep a correct xml sintax. In the case of the administrator's signature for example, it is not a correct xml, and even if it is correct, the user can edit his signature, keeping that as an html that works but not as a correct xml.
The solution I see here is to change 
`root = etree.fromstring(html)` to `root = lxml.html.fromstring(html)` 
and `return str(etree.tostring(root))` to `return htmltree.tostring(root).decode("utf-8")`

mail_debrand/models/mail_template.py
```diff
@@ -0,0 +1,39 @@
+from lxml import etree
```
ernestotejeda:
```suggestion
from lxml import html as htmltree
```

mail_debrand/models/mail_template.py
```diff
@@ -0,0 +1,39 @@
...
+    def _debrand_body(self, html):
+        using_word = _('using')
+        odoo_word = _('Odoo')
+        html = re.sub(
+            using_word + "(.*)[\r\n]*(.*)>" + odoo_word + r"</a>", "", html,
+        )
+        powered_by = _("Powered by")
+        if powered_by not in html:
+            return html
+        root = etree.fromstring(html)
```
ernestotejeda:
```suggestion
        root = htmltree.fromstring(html)
```

mail_debrand/models/mail_template.py
```diff
@@ -0,0 +1,39 @@
...
+            if any(
+                [
+                    "www.odoo.com" in child.get("href", "")
+                    for child in elem.getchildren()
+                ]
+            ):
+                for child in elem.getchildren():
+                    elem.remove(child)
+                elem.text = None
+        return str(etree.tostring(root))
```
ernestotejeda:
```suggestion
        return htmltree.tostring(root).decode("utf-8")
```

devishree-git:
> @devishree-git That won't work. The existing code was mostly good. Just need to work out the tag mismatches. Check the error from @chienandalu , the current code is easy to test with, you can just pass in any view arch in shell to test and check the output.

[https://drive.google.com/file/d/1_YOOrJoWNHx8DWQ_7SYSX0Kow9LqidcV/view](url)

Above link is suggest changes with improvment of just 2 lines with import lxml.html

chienandalu:
@ernestotejeda What do you think about @devishree-git proposal?

chienandalu:
@gdgellatly Would you mind if we superseed this PR in order to speed it up?

gdgellatly:
@chienandalu So 6 months nothing matters, now it needs to happen in hours? Just do the PR against this branch and I'll merge it in.

mail_debrand/__manifest__.py
gdgellatly:
```suggestion
     "website": "https://github.com/OCA/social/",
```

mail_debrand/__manifest__.py
```diff
@@ -3,16 +3,13 @@
 {
     "name": "Mail Debrand",
     "summary": "Remove Odoo branding in sent emails",
-    "version": "12.0.1.0.0",
+    "version": "12.0.2.0.0",
     "category": "Social Network",
-    "website": "https://odoo-community.org/",
+     "website": "https://github.com/OCA/social/",
```
chienandalu:
```suggestion
    "website": "https://github.com/OCA/social/",
```

chienandalu (commented):
Works great now. Just a tiny lint error and we can go :)

mail_debrand/models/mail_template.py
```diff
@@ -0,0 +1,39 @@
+from lxml import html as htmltree
```
chienandalu, line 4:
```suggestion
# Copyright 2019 O4SB - Graeme Gellatly
# Copyright 2019 Tecnativa - Ernesto Tejeda
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
from lxml import html as htmltree
```

ernestotejeda:
For me it works great and code is OK. 
Only the sugested changes by @chienandalu are still pending.

gdgellatly:
Sorry, storms and power cuts. I committed those changes, feel free to approve. I'll rebase and squash one of those commits tomorrow.

chienandalu (approved):

gdgellatly:
Authorship was lost on the code review changes on rebase. Anyone care, I can re rebase if needed and add them in.

chienandalu:
No, let's move on :)

ernestotejeda (approved):
No, it's ok

pedrobaeza:
Thanks to all!

PR #357, opened 2019-01-24
[ADD] Adds mail_thread_track_x2many module.

daramousk:
Mail thread does not do a good job tracking changes on x2many fields, this
module fixes that.


daramousk:
@hbrunn On `one2many` fields that are changed from outside the form, their changes are not shown on the chatter (since the write is not invoked) yet they are shown when the `read` returns. I fixed this using `message_post` on the specific places that I want the update to be shown. Perhaps there is a better solution? Or that is not a bug?

daramousk:
@hbrunn The tests have been added quite some time back, things break here but it does not seem it is the fault of this module.

hbrunn:
https://travis-ci.org/OCA/social/jobs/509924640#L1832 looks very much like a problem in the tests

daramousk:
`mail_thread` should exist at that point, I specifically declare a dependency on `mail`.

hbrunn:
mail_thread is an abstract model that doesn't have a table

PR #358, opened 2019-02-03
[ADD][8.0] mail_queue_send_limit

hbrunn:
This module allows you to rate limit the mail queue job for situation when you have to use an SMTP server which has a rate limit on connections or mails to send.

The error message you'll get in such a case is

```
4.7.0 Too many login attempts, please try again later.
```


hbrunn:
I took the liberty to fix some RST errors in another commit, but the timeout on odoo/odoo seems to point to some incompatibility of the affected module (not this one) with upstream

PR #360, opened 2019-02-11
[11.0][message_auto_subscribe_notify_own]

JordiBForgeFlow:
With this module you will receive a notification when you assign yourself
to an object that has a responsible user, like the salesperson in
leads/opportunities.

In general Odoo will notify other users in this situation, but not the own
user who has performed the action.

But in certain situations it is better that you receive this notification, as
a reminder sent to your mailbox that you are now responsible for that object.

Usage
=====

Assign yourself as salesperson of a lead/opportunity, for example, and you
will see that you'll receive a notification.


pedrobaeza:
Why not creating an activity instead?

JordiBForgeFlow:
Because it's just a remainder. People may assign leads to various
salesperson, including themselves, as a first step.

Then they often take the time later to review what they need to work on
through the notifications.

It is an optional usability feature. Perhaps other companies just do not
need that

JordiBForgeFlow:
@ageficent @aheficent @bjeficent  Can you review?

message_auto_subscribe_notify_own/models/mail_thread.py
```diff
@@ -0,0 +1,25 @@
...
+class MailThread(models.AbstractModel):
+    _inherit = 'mail.thread'
+
+    @api.multi
+    def message_auto_subscribe(self, updated_fields, values=None):
+        super(MailThread, self).message_auto_subscribe(updated_fields,
+                                                       values=values)
+        user_field_lst = self._message_get_auto_subscribe_fields(
+            updated_fields)
+        to_add_users = self.env['res.users'].sudo().browse(
```
AaronHForgeFlow, line 17: Tested on local. I created a lead and assign administrator as a sales person. Also tried to modified an existng lead and assign admin as sales person. I could not make it work. `user_field_lst` is always empty. Not sure what is wrong.

AdriaGForgeFlow (approved):
Tested on local. First I have created a new user, then (once logged in as the new user) I have auto assigned an opportunity and the email has been received correctly to the specified address. 

LGTM :+1:

JordiBForgeFlow:
@aheficent can you update your review, based on comments by @ageficent ?

AaronHForgeFlow (approved):
tested again. :+1:  It's working now. I was missing to configure the user to handle notifications by email.

PR #362, opened 2019-02-20
[FIX] mass_mailing_partner: Remove dead and wrong code

yajo:
These lines are useless, and, besides, wrong. Removing them.

@Tecnativa


pedrobaeza (approved):
Fast-tracking as straight-forward

PR #363, opened 2019-02-20
[11.0][IMP] message_auto_subscribe_notify_own: Allow to select models…

HviorForgeFlow:
… to be notified

CC @jbeficent


chienandalu:
@hveficent Is it WIP?

HviorForgeFlow:
@chienandalu, more or less it is. I can't figure out how to implement unit testing. This approach is failing on Test shutdown.

Any idea to move it forward

message_auto_subscribe_notify_own/tests/test_fake_model.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2019 Eficent Business and IT Consulting Services, S.L.
+# Copyright 2019 Aleph Objects, Inc.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class FakeModelToTest(models.Model):
+    """A Fake model to Test."""
+
+    _name = 'fake.model'
```
MiquelRForgeFlow: `message_auto_subscribe_notify_own.test` should be a better name. This way we can identify the table from which module comes from.

MiquelRForgeFlow (approved):

message_auto_subscribe_notify_own/tests/test_fake_model.py
HviorForgeFlow: Ok! it makes sense, changed

JordiBForgeFlow (approved):

PR #374, opened 2019-04-09
[11.0][ADD] mass_mailing_newsletter_welcome_mail

cristinamartinrod:
This module extends the functionality of the website popup newsletter subscription form sending to new subscribers an automatic welcome email and allow you to customize it.

cc @Tecnativa


yajo:
Is this ready to review?

mass_mailing_welcome_mail/data/base_automation_data.xml
```diff
@@ -0,0 +1,38 @@
...
+            <field name="model_id" ref="mass_mailing.model_mail_mass_mailing_contact"/>
+            <field name="use_default_to" eval="True"/>
+            <field name="email_from">${object.company_id.email}</field>
+            <field name="lang">${object.env.context.get("lang")}</field>
+            <field name="subject">Welcome to the mailing list</field>
+            <field name="body_html" type="html">
+                <div>Dear <b>${object.name}</b>,
+                    <br/>
+                    <br/>
+                    <h3>Thanks for signing up four our newsletter!</h3>
```
pedrobaeza: s/four/for

mass_mailing_welcome_mail/data/base_automation_data.xml
```diff
@@ -0,0 +1,38 @@
...
+                    <p>
+                        Enjoy a discount code for 20% off your first order
```
pedrobaeza: This shouldn't be included here in this mail, but customize the template on the customer.

mass_mailing_welcome_mail/data/base_automation_data.xml
```diff
@@ -0,0 +1,38 @@
...
+
+    <data>
+        <record id="automated_action" model="base.automation">
+            <field name="name">Send Welcome Mail to new subscribers</field>
+            <field name="model_id" ref="mass_mailing.model_mail_mass_mailing_contact"/>
+            <field name="state">email</field>
+            <field name="trigger">on_create</field>
+            <field name="active" eval="True"/>
+            <field name="template_id" ref="email_template"/>
+            <field name="filter_domain">[["list_ids","=","Newsletter"]]</field>
```
pedrobaeza: Why this filter? By default, it should be in any mass mailing list, and be customized on customer.

pedrobaeza (dismissed):

mass_mailing_welcome_mail/data/base_automation_data.xml
cristinamartinrod: Done

mass_mailing_welcome_mail/data/base_automation_data.xml
cristinamartinrod: This is part of the default template of the newsletter in Odoo, something very common, that's why I included it here. I delete it.

mass_mailing_welcome_mail/data/base_automation_data.xml
cristinamartinrod: The same that the previous default text, the subscription is usually to the newsletter and to set a filter for default, easier to modify for the user. I delete it

cristinamartinrod:
Now is ready for you review @Yajo :)

mass_mailing_welcome_mail/data/base_automation_data.xml
pedrobaeza: But this doesn't count with translations for example. If there's already a list in data with an XML-ID, use `eval` and `ref` for setting this field with the ID of this mass mailing list by default. That's something reasonable, but not to make the comparison by name.

cristinamartinrod: Yes, just was a little help/guide for users, but it's ok, I removed it to avoid doubts.

pedrobaeza: But my question remains unanswered: is there a mailing list defined in data with that name? If so, definitively it's advisable to set it as said, due to Odoo intention is to have that one as newsletter. If not, then put all of them.

cristinamartinrod: Yes, it's a default list set on data in mass_mailing

pedrobaeza: OK, then it can be interesting to be added, but I let you to decide. If we add it, then:

- Change module title to reflect this: s/mailing list/newsletter.
- Should the module be renamed to `mass_mailing_newsletter_welcome_mail`?
- Put `<field name="filter_domain" eval="[['list_ids', '=', ref('xml_id_of_newsletter')]]"/>`

If not added:

- Change module description to say that is for all mailing list, as right now references the popup only.

cristinamartinrod: I think it's useful if it's configured by default a specific list, so I changed name, descriptions and settings for it.

mass_mailing_newsletter_welcome_mail/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    'name': "Welcome mail to new subscribers",
+    "summary": "Send an automated welcome mail to new newsletter subscribers",
+    'license': 'AGPL-3',
+    'author': 'Tecnativa, '
+              'Odoo Community Association (OCA)',
+    'category': 'Social',
+    'version': '11.0.1.0.0',
+    'website': 'https://github.com/OCA/website',
```
yajo:
```suggestion
    'website': 'https://github.com/OCA/social',
```

mass_mailing_newsletter_welcome_mail/data/base_automation_data.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+    <data>
+        <record id="email_template" model="mail.template">
+            <field name="name">Welcome mail to newsletter subscribers</field>
+            <field name="model_id" ref="mass_mailing.model_mail_mass_mailing_contact"/>
+            <field name="use_default_to" eval="True"/>
+            <field name="lang">${object.env.context.get("lang")}</field>
+            <field name="subject">Welcome to the mailing list</field>
```
yajo:
```suggestion
            <field name="subject">Welcome to the mailing list ${object.list_id.name}</field>
```

mass_mailing_newsletter_welcome_mail/data/base_automation_data.xml
```diff
@@ -0,0 +1,32 @@
...
+            <field name="body_html" type="html">
+                <div>Dear <b>${object.name}</b>,
+                    <br/>
+                    <h3>Thanks for signing up for our newsletter!</h3>
```
yajo:
```suggestion
                    <h3>Thanks for signing up for ${object.list_id.name}!</h3>
```

mass_mailing_newsletter_welcome_mail/data/base_automation_data.xml
```diff
@@ -0,0 +1,32 @@
...
+            </field>
+        </record>
+    </data>
+
+    <data>
+        <record id="automated_action" model="base.automation">
+            <field name="name">Send Welcome Mail to new subscribers on Newsletter</field>
+            <field name="model_id" ref="mass_mailing.model_mail_mass_mailing_contact"/>
+            <field name="state">email</field>
+            <field name="trigger">on_create</field>
```
yajo: This won't take into account if the user is added by a backend odoo user or if he subscribes himself. Is that what we want? :thinking:

mass_mailing_newsletter_welcome_mail/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module was written to extend the functionality of the website popup
+newwsletter subscription form to support sending to new subscribers an
```
yajo:
```suggestion
newsletter subscription form to support sending to new subscribers an
```

mass_mailing_newsletter_welcome_mail/i18n/mass_mailing_welcome_mail.pot
```diff
@@ -0,0 +1,35 @@
...
+"Plural-Forms: \n"
+
+#. module: mass_mailing_welcome_mail
+#: model:mail.template,body_html:mass_mailing_welcome_mail.email_template
+msgid "<div>Dear <b>${object.name}</b>,
+                    <br/>
+                    <br/>
+                    <h3>Thanks for signing up four our newsletter!</h3>
+                    <p>
+                        Enjoy a discount code for 20% off your first order
```
yajo: You have to update the translations.

mass_mailing_newsletter_welcome_mail/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    'depends': [
+        'mass_mailing',
+        'website',
+        'website_crm',
```
yajo: Why `website` and `website_crm`? :thinking:

mass_mailing_newsletter_welcome_mail/data/base_automation_data.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+    <data>
```
yajo: You don't need `<odoo>` and `<data>`. Just use one.

yajo (changes requested):

mass_mailing_newsletter_welcome_mail/data/base_automation_data.xml
pedrobaeza: I don't think we are able to differentiate both.

mass_mailing_newsletter_welcome_mail/data/base_automation_data.xml
cristinamartinrod: We changed this yesterday, customizing only the Newsletter list. That's the reason we re-named the module. I understand that this is a good contribution for any list, even if it's only for Newsletter, it's the same.

mass_mailing_newsletter_welcome_mail/data/base_automation_data.xml
cristinamartinrod: Now I can not remember any email marketing software that does not send an automated welcome email for new subscribers to a list regardless of the source (import, create, add, forms, ...). It's basic functionality, I suppose that if an Odoo user wants that action to exclude x subscribers they'll have to modify the filter of the action or create a customized one.

mass_mailing_newsletter_welcome_mail/__manifest__.py
cristinamartinrod: Done

mass_mailing_newsletter_welcome_mail/data/base_automation_data.xml
cristinamartinrod: The same

mass_mailing_newsletter_welcome_mail/readme/DESCRIPTION.rst
cristinamartinrod: Done

mass_mailing_newsletter_welcome_mail/i18n/mass_mailing_welcome_mail.pot
cristinamartinrod: Yes, with last change of this template doesn't update the .pot. It's already changed

mass_mailing_newsletter_welcome_mail/__manifest__.py
cristinamartinrod: Testing website forms, but it's not necessary that dependency. I remove it

mass_mailing_newsletter_welcome_mail/data/base_automation_data.xml
yajo: We can differentiate if the mail sending is triggered via the subscription controller instead of from an automated action.

I think current approach can lead to surprises...

cubells (approved):
:+1: 

LGTM

Code review and test, but do the changes taht @Yajo recommend.

yajo (approved):
I fixed all myself:

- Allow user to customize template.
- Only react when someone subscribes from the website. Not when creating a contact from the backend.
- Proper usage of context values.
- Full instructions.

I cannot write tests because of https://github.com/odoo/odoo/pull/30768, so I added it to roadmap.

Tardo:
website_mass_mailing can be used to increase database, has no limits in email length... malicious people can create an script to generate records on demand.

I think that 'mass_mailing_newsletter_welcome_mail' can be used to spam an email account... because never checks if the contact is new. But i don't test this.

pedrobaeza:
@Tardo interesting thoughts about this. I also think it's redundant, but it was something specifically asked by a customer against our advice. Any way, possible spammers won't control the mail template that is sent, so they wouldn't get their goal of spamming with something specific. Only "noise". And this can't be done several times (only the first one), as Odoo control if an email already belongs to the mailing list.

Tardo:
@pedrobaeza yeah.. I had to say "flood" instead of "spam": Perhaps fill the email quota... or (i don't know how really this work) get a ban for the domain from email providers :/

pedrobaeza:
OK, then let's continue adding that warning in the README, OK?

yajo:
This is a problem in the base addon. I reported it here: https://github.com/odoo/odoo/issues/33234, you can put the link in the README If you want.

pedrobaeza:
Can you please put it in the roadmap?

Issue #375, opened 2019-04-24
[mail_attach_existing_attachment][12.0] not usable for invoices


tbaden:
Invoice wizard form is new in version 12.0 

Model:
account.invoice.send
Invoice Mail Wizard Form:
account.account_invoice_send_wizard_form

pedrobaeza:
Closing as there's a PR already.

PR #376, opened 2019-04-24
[IMP][12.0] mail_attach_existing_attachment

tbaden:
since 12.0 account invoice has it's own wizard and own model.
account.invoice.send inherits mail.compose.message so no changes needed here.
the view account.account_invoice_send_wizard_form just needs the same customization as email_compose_message_wizard_inherit_form

issue https://github.com/OCA/social/issues/375

### DIFF
--- a/mail_attach_existing_attachment/__manifest__.py
+++ b/mail_attach_existing_attachment/__manifest__.py
@@ -32,10 +32,12 @@
     'version': '12.0.1.0.0',
     'license': 'AGPL-3',
     'depends': [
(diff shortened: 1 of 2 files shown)


pedrobaeza:
You must create a new module `mail_attach_existing_attachment_account` that adds this, as changing dependencies in a module already merged is not usually allowed, and it's not logical to force to install account for having the general feature.

tbaden:
ok I'll check if there is a cleaner way to do this. thanks.

PR #384, opened 2019-05-21
[10.0][FIX][mass_mailing_list_dynamic] Fixed sync domain problem with escape characters. Added context…

wpichler:
… variable auto_created=True when creating contacts. So a possible send double opt in mail function can check for this context variable

### DIFF
--- a/mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
+++ b/mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
@@ -5,7 +5,6 @@
 from odoo import api, fields, models
 from odoo.tools import safe_eval
 
-
 class MassMailingList(models.Model):
     _inherit = "mail.mass_mailing.list"
 
@@ -43,7 +42,7 @@ class MassMailingList(models.Model):
         # Skip non-dynamic lists
         dynamic = self.filtered("dynamic")
         for one in dynamic:
-            sync_domain = safe_eval(one.sync_domain) + [("email", "!=", False)]
+            sync_domain = safe_eval(one.sync_domain.replace('u\'', '\'')) + [("email", "!=", False)]
             desired_partners = Partner.search(sync_domain)
             # Remove undesired contacts when synchronization is full
             if one.sync_method == "full":
@@ -55,7 +54,7 @@ class MassMailingList(models.Model):
             current_partners = current_contacts.mapped("partner_id")
             # Add new contacts
             for partner in desired_partners - current_partners:
-                Contact.create({
+                Contact.with_context(auto_created=True).create({
                     "list_id": one.id,
                     "partner_id": partner.id,
                 })


mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
```diff
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
-from odoo.tools import safe_eval
+from odoo.tools.safe_eval import safe_eval
```
pedrobaeza: Why changing this? It was working before.

mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
```diff
@@ -55,7 +55,7 @@ def action_sync(self):
             current_partners = current_contacts.mapped("partner_id")
             # Add new contacts
             for partner in desired_partners - current_partners:
-                Contact.create({
+                Contact.with_context(auto_created=True).create({
```
pedrobaeza, line 57: Can you provide here a comment pointing to why it's done this way. I have read your main comment and I still don't understand it...

hbrunn:
please rebase

rafaelbn:
H @wpichler , thanks for this contribution. Could you please rebase and answer questions from the review please? Best!

mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
wpichler: correct - was working before - i thought it would be better to directly only import the function... - i will undo this

mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
wpichler: in an other module i extend the create function of mail.mass_mailing.contact to automatically send a double opt in mail to the new created contact. To be able to not send it, in the case the contact got created in some other way - i've added this context var.

mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
```diff
@@ -43,7 +42,7 @@ def action_sync(self):
         # Skip non-dynamic lists
         dynamic = self.filtered("dynamic")
         for one in dynamic:
-            sync_domain = safe_eval(one.sync_domain) + [("email", "!=", False)]
+            sync_domain = safe_eval(one.sync_domain.replace('u\'', '\'')) + [("email", "!=", False)]
```
yajo, line 45: Nothing forbids you to use unicode strings to search AFAIK, so this change doesn't seem to be actually fixing something.

OTOH, it's introducing a new real bug, where a domain like `[('name', '=', 'Maru')]` would be converted to `[('name', '=', 'Mar')]` and produce unexpected results.

I bet you're trying to fix a bug in the wrong place.

mass_mailing_list_dynamic/models/mail_mass_mailing_list.py
yajo: Which module?

yajo (changes requested):

PR #385, opened 2019-05-21
[MIG] website_mass_mailing_name: Migration to 12.0

victor-ditay:


victor-ditay:
Once test fixed I'll set to review

victor-ditay:
cc @Tecnativa

website_mass_mailing_name/static/src/css/website_mass_mailing_name.scss
```diff
@@ -0,0 +1,8 @@
+/*! Copyright 2016-2017 Jairo Llopis <jairo.llopis@tecnativa.com>
+ *  License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+.js_subscribe .form-control {
+  width: 50%;
```
yajo: Follow the OCA standard of 4 spaces indentation please.

website_mass_mailing_name/static/src/css/website_mass_mailing_name.scss
```diff
@@ -0,0 +1,8 @@
...
+}
+.js_subscribe[data-subscribe=on] .js_subscribe_name:not([disabled]) {
+  display: none;
```
yajo: 4 spaces please.

website_mass_mailing_name/static/src/js/editor_and_public_tour.js
```diff
@@ -45,11 +45,12 @@ odoo.define("website_mass_mailing_name.editor_and_public_tour", function (requir
             {
                 content: "Open user menu",
                 extra_trigger: ".js_subscribe .alert-success",
-                trigger: "#top_menu span:contains('Administrator')",
+//                extra_trigger: ".js_subscribe",
```
yajo: :fire:

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
```diff
@@ -4,7 +4,8 @@
 odoo.define("website_mass_mailing_name.subscribe", function (require) {
     "use strict";
     require("mass_mailing.website_integration");
-    var animation = require("web_editor.snippets.animation");
+    // var animation = require("web_editor.snippets.animation");
```
yajo: :fire:

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
```diff
@@ -27,7 +28,9 @@ odoo.define("website_mass_mailing_name.subscribe", function (require) {
                 };
             // Stop on error
             if (email_error || name_error) {
-                this.$target.addClass("has-error")
+                this.$target.addClass("o_has-error");
```
yajo: The old `has-error` class came from Bootstrap 3 and you shouldn't be using it anymore. It's probably still working, but only [due to a backwards-compatibility layer that Odoo 12.0 has](https://github.com/odoo/odoo/blob/c52b9dfc3761c166f5f1b475e7a14ab2ec9fb66f/addons/website/static/src/scss/compatibility/bs3_for_12_0.scss#L81-L83).

The class `o_has-error` you are using... I cannot find it in the source code... Where does it come from?

In any case, please check [this section about form validation in BS4](https://getbootstrap.com/docs/4.3/components/forms/#how-it-works). I think you can use the constraint validation API, or the classes `is-valid` and `is-invalid` (although this is more a fallback for server-side validation; not your case).

yajo (changes requested):

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
victor-ditay: I found it by inspect elements but true is no way it comes from Odoo. So I will change it as you suggest and test again

victor-ditay:
Changes done @Yajo

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
```diff
@@ -0,0 +1,61 @@
...
+        _onClick: function() {
+            var email_error = !this.$email.val().match(/.+@.+/),
+                name_error = this.$name.length && !this.$name.val(),
+                values = {
+                    "list_id": this.$target.data('list-id'),
+                    "email": this.$email.val(),
+                };
+            // Stop on error
+            if (email_error || name_error) {
+                this.$target.addClass("is-invalid");
```
yajo:
```suggestion
```

I tested in runbot and this one actually is not necessary. Are the ones below that do the actual work, so you can remove it.

yajo (changes requested):

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
victor-ditay: I was unsure, thanks for your review :)

victor-ditay:
@Yajo Changes done

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
```diff
@@ -0,0 +1,60 @@
...
+ * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+
+odoo.define("website_mass_mailing_name.subscribe", function (require) {
+    "use strict";
+    require("mass_mailing.website_integration");
+    var animation = require("website.content.snippets.animation");
+
+    animation.registry.subscribe.include({
+        start: function(editable_mode) {
+            this.$email = this.$target.find(".js_subscribe_email");
```
Tardo: Can use this.$() instead of this.$target.find()...

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
```diff
@@ -0,0 +1,60 @@
...
+            // Thanks upstream for your @$&#?!! inheritance-ready code.
+            // Injecting ajax events to modify behavior of snippet.
+            if (this.$name) {
+                $(document).ajaxSend($.proxy(this.on_ajax_send, this));
+            }
+            return this._super(editable_mode);
+        },
+
+        _onClick: function() {
+            var email_error = !this.$email.val().match(/.+@.+/),
```
Tardo: Here uses regex only to "check" that string have an @... perhaps can be better use "search" instead of "match".

Tardo (approved):
LGTM, tested on runbot

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
victor-ditay: Not against but see your review is LGTM but still asking for changes, would be better move forward instead spend more time waiting for Travis?

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
victor-ditay: Same as the comment above. Would be better move forward instead spend more time?

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
Tardo: Yeah, without these changes all works fine... it's just a small suggestion that can be completely ignored ;)
** Same for the other comment.

pedrobaeza:
When someone approves a PR, but put comments, they are only informative ones, that can be changed if you want, but not blocking. I understand those 2 as this type, for knowing more JS tricks, accelerators, and so.

I'm seeing a warning in runbot, but logs are not accessible. I have launched a rebuild. Please check when finished, and if you want, make the proposed improvements, that are not blocking.

victor-ditay:
Changes done @pedrobaeza @Tardo please review

Tardo:
@victormmtorres thx, good work!

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
```diff
@@ -0,0 +1,69 @@
...
+            // Injecting ajax events to modify behavior of snippet.
+            if (this.$name) {
+                $(document).ajaxSend($.proxy(this.on_ajax_send, this));
+            }
+            return this._super(editable_mode);
+        },
+
+        _onClick: function() {
+            function validateEmail(sEmail) {
+                var filter = /^([\w-\.]+)@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.)|(([\w-]+\.)+))([a-zA-Z]{2,4}|[0-9]{1,3})(\]?)$/;
```
yajo: This regexp is wrong, sorry.

No matter what you could possibly imagine, you'll almost always fail email regexp validation. :laughing: Go http://emailregex.com and see for yourself.

However, the good news is that you won't need a regexp. See below. You'll be able to drop the `validateEmail()` function.

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
```diff
@@ -0,0 +1,69 @@
...
+                if (filter.test(sEmail)) {
+                    return true;
+                }
+                else {
+                    return false;
+                }
+            }
+            var email_error = validateEmail(!this.$email.val()),
```
yajo:
```suggestion
            var email_error = !this.$email[0].checkValidity(),
```

This would still be a false positive if the email is empty. See below on how to fix that.

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
```diff
@@ -0,0 +1,69 @@
...
+                name_error = this.$name.length && !this.$name.val(),
```
yajo:
```suggestion
                name_error = this.$name[0].checkValidity(),
```

This will also be a false positive if the name is empty. See below for the fix.

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
```diff
@@ -0,0 +1,69 @@
...
+                values = {
+                    "list_id": this.$target.data('list-id'),
+                    "email": this.$email.val(),
+                };
+            // Stop on error
+            if (email_error || name_error) {
+                this.$(".js_subscribe_name").addClass("is-invalid");
+                this.$(".js_subscribe_email").addClass("is-invalid");
```
yajo: Since we're going to use native HTML5 validation, you'll possibly can remove these 2 lines.

website_mass_mailing_name/templates/snippets.xml
```diff
@@ -0,0 +1,18 @@
...
+
+<odoo>
+
+<template id="s_newsletter_subscribe_form"
+          inherit_id="website_mass_mailing.s_newsletter_subscribe_form">
+    <xpath expr="//input[hasclass('js_subscribe_email','form-control')]"
+           position="before">
+        <input
+            name="name"
+            class="js_subscribe_name form-control"
```
yajo, line 13: Now, the magic to make it required:

```suggestion
            class="js_subscribe_name form-control"
            required="required"
```

You should do the same for the email input, by using another xpath with `position="attributes"` and adding the `required=""` attribute.

---

Note: Boolean attributes might work unexpectedly in HTML. [Read the docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Boolean_Attributes).

yajo (changes requested):

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
victor-ditay: My bad, ok I will change it asap

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
```diff
@@ -18,16 +18,26 @@ odoo.define("website_mass_mailing_name.subscribe", function (require) {
             return this._super(editable_mode);
         },
 
-        on_click: function() {
-            var email_error = !this.$email.val().match(/.+@.+/),
+        _onClick: function() {
+            function validateEmail(sEmail) {
```
chienandalu: Sorry @victormmtorres I don't fully understand what you want to achieve here :slightly_smiling_face: You should not need to validate email here don't you? You already have that in the main method: https://github.com/odoo/odoo/blob/12.0/addons/website_mass_mailing/static/src/js/website_mass_mailing.js#L41-L64

chienandalu (changes requested):

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
victor-ditay: It looks like need to involve `form` element. I tested on a and works fine if input are embed on form elements.

https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_required

website_mass_mailing_name/templates/snippets.xml
victor-ditay: Please see my comment above

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
victor-ditay: I will change this part

victor-ditay:
@Yajo please help me out as none of the suggested changes are working and I spend some time investigating and I believe at least the validation for email should work

victor-ditay:
Changes done @chienandalu @Yajo

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
```diff
@@ -18,16 +18,10 @@ odoo.define("website_mass_mailing_name.subscribe", function (require) {
...
-                name_error = this.$name.length && !this.$name.val(),
-                values = {
-                    "list_id": this.$target.data('list-id'),
-                    "email": this.$email.val(),
-                };
-            // Stop on error
-            if (email_error || name_error) {
-                this.$target.addClass("has-error")
+        _onClick: function() {
+            var email_valid = this.$email[0].reportValidity(),
```
chienandalu, line 24: I still think we're checking this twice. Anyway, not blocking.

chienandalu (approved):
Thanks @victormmtorres !

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
yajo: Yes, but there's a reason for that.

If we do it like it is now, UX is way better. The user gets a clear bubble indicating where the problem is. Maybe it's a wrong email, or maybe it's an empty field.

If we rely on upstream validation, it would just set all red (in fact, both fields) and say nothing. It's OK if there's only one field, but now that there are 2, if they are validated in different ways, it gets inconsistent.

The way odoo does it supports very old browsers, but has the problems explained in https://github.com/odoo/odoo/issues/14942.

However it's true that you can add a comment explaining why we're duplicating this validation here, @victormmtorres.

yajo (approved):

website_mass_mailing_name/static/src/js/website_mass_mailing_name.js
pedrobaeza: Add the comment and I'll merge

pedrobaeza: Add the comment and I'll merge

victor-ditay: comment added

PR #389, opened 2019-06-03
[11.0][IMP] mail_activity_team

jarroyomorales:
This PR extends the activities widget on the systray. It creates 2 separate windows to separate personal activities and team activities. 
Captura de pantalla de 2019-06-03 13-01-40

@jbeficent @aheficent @lreficent


AaronHForgeFlow:
This is awesome! :heart: 

Is it correct to set the user_id field not required? I remember leaving user_id blank caused error when saving the activity, it was related followers.

jarroyomorales:
Yes, this is a problem. You have to set a user anyways because it needs it no make it follower, I have been thinking and couldn't find an easy solution, but it would be great to be able to do so, since at the end, this is what this module is for.

JordiBForgeFlow (approved):
Great addition! Please check lint errors in travis.

mail_activity_team/static/src/js/systray.js
```diff
@@ -0,0 +1,72 @@
...
+                this.do_action({
+                    type: 'ir.actions.act_window',
+                    name: data.model_name,
+                    res_model:  data.res_model,
+                    views: [[false, 'kanban'], [false, 'form']],
+                    search_view_id: [false],
+                    domain: [
+                        '|',
+                        ['activity_user_id', '=', session.uid],
+                        ['activity_team_user_ids', 'in', session.uid]
```
LoisRForgeFlow, line 66: From the team activities action you are providing both user and team activities

mail_activity_team/models/res_users.py
```diff
@@ -11,3 +11,53 @@ class ResUsers(models.Model):
...
+                                    WHEN %(today)s::date -
+                                    act.date_deadline::date = 0 Then 'today'
+                                    WHEN %(today)s::date -
+                                    act.date_deadline::date > 0 Then 'overdue'
+                                    WHEN %(today)s::date -
+                                    act.date_deadline::date < 0 Then 'planned'
+                                END AS states
+                            FROM mail_activity AS act
+                            JOIN ir_model AS m ON act.res_model_id = m.id
+                            WHERE team_id in (
```
LoisRForgeFlow, line 30: But you are displaying only team's activities.

LoisRForgeFlow (changes requested):
A comment in line.

Also for discussion:
Should the number in the clock icon of the systray display the sum of team and indiviual activities?
image

mail_activity_team/static/src/js/systray.js
jarroyomorales: True, I just changed the domain, it didnt make any sense since with the query you couldnt get any activity that was yours but not of your team.

jarroyomorales:
The ideal regarding to the notifications count would be getting the sum but being careful that team activities where you are the responsible are not counted double. I'm still working on it.

jarroyomorales:
@lreficent @jbeficent I just modified the counter. I was testing it and I think it displays properly what we talked. The sum of your activities + team activities but without counting double the team activities where you are `user_id` .

LoisRForgeFlow (approved):
Works perfect now! :+1: 

(sorry for the delay on the review)

AaronHForgeFlow (approved):
:+1: functional tests

PR #397, opened 2019-06-13
[FIX] green runbot 2019-06-13

hbrunn:


pedrobaeza:
Add a warning also on installation section of the module.

hbrunn:
@pedrobaeza done, can we fasttrack?

pedrobaeza (approved):

PR #398, opened 2019-06-20
[11.0][ADD] message_security_author

MiquelRForgeFlow:
Message Security Author
=======================

Some modules like `mail_tracking` give the users the possibility to access to
the messages by indirect ways. Odoo have not yet regulated what the users can
do to messages because by default in Odoo no users can access to them.

Thus, this `message_security_author` module enters in scene.

This module restricts who can edit/delete messages. Specifically, a message
may be edited or deleted if and only if:

- The user is who created the message previously.
- The user have special permission (is in the group `Mail Message Manager`)


LoisRForgeFlow:
travis should be fixed with https://github.com/OCA/social/pull/400

MiquelRForgeFlow:
thanks

message_security_author/models/mail_message.py
```diff
@@ -0,0 +1,47 @@
...
+    @api.multi
+    def write(self, vals):
+        user = self.env["res.users"].sudo().browse(self._uid)
+        message_manager_group = self.env.ref(
+            'message_security_author.group_mail_message_manager')
+        if self._uid == 1:
+            rec = super(Message, self).write(vals=vals)
+        elif self._name != 'mail.message':
+            rec = super(Message, self).write(vals=vals)
+        elif message_manager_group in user.groups_id:
```
LoisRForgeFlow: why not `user.has_group('message_security_author.group_mail_message_manager')`?

message_security_author/models/mail_message.py
```diff
@@ -0,0 +1,47 @@
...
+from odoo import api, models, _
+from odoo.exceptions import AccessError
+
+
+class Message(models.Model):
+    _inherit = 'mail.message'
+
+    @api.multi
+    def write(self, vals):
+        user = self.env["res.users"].sudo().browse(self._uid)
```
LoisRForgeFlow: why not `self.env.user`?

LoisRForgeFlow (commented):
Some comments inline

LoisRForgeFlow (approved):
:+1:

HviorForgeFlow (approved):

message_security_author/models/mail_message.py
```diff
@@ -0,0 +1,50 @@
...
+
+from odoo import api, models, _
+from odoo.exceptions import AccessError
+
+
+class Message(models.Model):
+    _inherit = 'mail.message'
+
+    @api.multi
+    def write(self, vals):
```
yajo, line 12: You should overload methods `check_access_rights` (by model) and/or `check_access_rule` (by record; according to your module description you should use this one possibly).

message_security_author/models/mail_message.py
```diff
@@ -0,0 +1,50 @@
...
+        user = self.env.user
+        if self._uid == 1:
+            rec = super(Message, self).write(vals=vals)
```
yajo, line 15: Calling to super 6 times is a bit ugly. Can't you think in another way?

message_security_author/models/mail_message.py
```diff
@@ -0,0 +1,50 @@
...
+        elif self._name != 'mail.message':
+            rec = super(Message, self).write(vals=vals)
+        elif user.has_group(
+                'message_security_author.group_mail_message_manager'):
```
yajo, line 19: This can be done through normal permission system, no need to hardcode it.

message_security_author/models/mail_message.py
```diff
@@ -0,0 +1,50 @@
...
+            rec = super(Message, self).write(vals=vals)
+        elif not self.filtered(lambda m: user not in m.author_id.user_ids):
```
yajo, line 21: This also.

message_security_author/models/mail_message.py
```diff
@@ -0,0 +1,50 @@
...
+            rec = super(Message, self).write(vals=vals)
+        elif self.model == 'crm.lead' and vals.get('subject') and \
+                vals.get('res_id'):
```
yajo, line 24: This is weird... So if I cannot write a crm.lead, to be able to write it I just have to modify these 2 fields? :roll_eyes:

message_security_author/models/mail_message.py
```diff
@@ -0,0 +1,50 @@
...
+    def unlink(self):
+        user = self.env.user
+        if self._uid == 1:
+            rec = super(Message, self).unlink()
+        elif self._name != 'mail.message':
+            rec = super(Message, self).unlink()
+        elif self._context.get('deleting_mail_thread'):
+            rec = super(Message, self).unlink()
+        elif user.has_group(
+                'message_security_author.group_mail_message_manager'):
```
yajo, line 43: This can be done with normal permissions.

message_security_author/readme/INSTALL.rst
```diff
@@ -0,0 +1 @@
+To install this module, simply follow the standard install process.
```
yajo, line 1: You can remove this file. It's more targeted to non-obvious installation cases.

yajo (changes requested):
> Some modules like `mail_tracking` give the users the possibility to access to
> the messages by indirect ways. Odoo have not yet regulated what the users can
> do to messages because by default in Odoo no users can access to them.

If this is true, it should be considered a security problem in `mail_tracking` (and fixed there too)?

What exactly do you mean by "accessing messages by indirect ways"? :thinking:

PR #400, opened 2019-06-21
[FIX] travis. set postgres version to 9.6 as 9.3 is not available in xenial.

LoisRForgeFlow:
Same as in all repos.


pedrobaeza (approved):
Fast-tracking this

PR #402, opened 2019-06-25
12.0 mig mass_mailing_custom_unsubscribe

ernestotejeda:
From Odoo 11 to Odoo 12 have changed several things regarding the subscription/unsubscription of a mass mailing contact, so I had to make several adaptations to the module. Some adaptations are described here:
- Now in v12 this module does not need to depend on website_mass_mailing or website, just mass_mailing and uses its own layout for the UI
- Now the contacts don't have the field opt_out, the class that has this field is a new one that is responsible for managing subscribtions/unsubscribtions from a contact to a list. So now each contact, is related to one or more mass mailing lists and also he/she can be unsubscribed to some of them and at the same time be subscribed to others of them.
For the reason explained above for this version of the module I have eliminated what had been done in v11 of the module to force users to choose only one mailing list per contact.
- I have also deleted the `not_cross_unsubscriptable` field from the mailing lists because already in v12 there is a field called` is_public` for the same purpose.
- In v11, when a user access the unsubscribe link in the email, odoo first shows you the status of your subscriptions and then the user chooses which list he/she wants to subscribe/unsubscribe to, now in v12 ordoo first unsubscribe the user from the list and then it shows the status of subscriptions/unsubscriptions. Because of that the code of the controller was adapted so that a new page is displayed where reasons for unsubscription are requested before that normal unsubscription is executed.
- Now in v12, if the email that is going to be de-subscribed is one that does not correspond to a mass mailing contact, for example a partner, it goes to the blacklist. Because of that the module was adapted so that a new page is displayed where reasons for blacklisting are requested  before going to the blacklist and the blacklistings are registered as well as the unsubscriptions are registered.

The unit tests using the phantom_js are not included here because I have not been able to make them work, so I hope that someone can help me with this.

Cc @Tecnativa


mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
```diff
@@ -0,0 +1,120 @@
...
+                raise exceptions.ReasonRequiredError(
+                    _("Please indicate why are you unsubscribing."))
+
+    @api.multi
+    @api.constrains("details", "reason_id")
+    def _check_details_needed(self):
+        """Ensure details are given if required."""
+        for one in self:
+            if not one.details and one.details_required:
+                raise exceptions.DetailsRequiredError(
```
Tardo, line 97: Why need a custom exception?

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -44,48 +42,54 @@ def mailing(self, mailing_id, email=None, res_id=None, token="", **post):
...
+        reasons = request.env["mail.unsubscription.reason"].search([])
         try:
             # Check if we already have a reason for unsubscription
             reason_id = int(post["reason_id"])
         except (KeyError, ValueError):
             # No reasons? Ask for them
-            return self.reason_form(mailing, email, res_id, token)
+            return self.reason_form(mailing_id, email, res_id, reasons, token)
         else:
-            # Unsubscribe, saving reason and details by context
```
victor-ditay: think you could keep a comment not sure by the content if is still valid

victor-ditay (changes requested):
I would like have the main changes on the history.rst file because this much more than a migration and will help reviewers

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
ernestotejeda: Maybe this makes the code more readable. Do you think I should remove this? I don't think it's wrong

mass_mailing_custom_unsubscribe/controllers/main.py
ernestotejeda: Yes, I added the comment again

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
Tardo: Yeah isn't wrong... but actually this exception is never being catched. Perhaps @Yajo can tell us if is really necessary.

chienandalu (changes requested):
Great work here, @ernestotejeda 

I've been testing the functionality and I'm finding major changes from v11. This is how it worked in v11:

Peek 04-07-2019 11-14

But this is how it works now in v12 (wich is less usable I think):

Peek 04-07-2019 11-16

If only the main screen when the unsubscribe screen was the subscription lists update screen, it would fit the previous behavior.

ernestotejeda:
Thanks for your comment, @chienandalu . I understand your point, but since this is the new unsubscription philosophy now in v12 and I don't think it's less usable, I don't know if it's better to go back to the previous version. Look at the following and tell me what you think ..
In general, the link that appears at the end of the newsletters is 'unsubscribe', but it is not 'Update subscriptions/unsubscription from mailing lists', therefore, I don't see so bad that first the user is unsubscribed after click on that link and then the status of the subscriptions is displayed.
At the beginning, when I started to make the migration, I thought about doing what you propose (keep the previous behavior), but I was getting complicated and also the main objectives of the module are:
- Know why and when a contact has subscribed or canceled his subscription to a bulk email.
- Provide evidence of why you are sending bulk emails to a specific contact, as required by the GDPR in Europe.

and that's not related to the way you subscribe or unsubscribe

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+        except (KeyError, ValueError):
+            # No reasons? Ask for them
+            return self.reason_form(mailing_id, email, res_id, reasons, token)
+        else:
+            # Unsubscribe, saving reason and details by context
+            details = post.get("details", False)
+            self._add_extra_context(mailing_id, res_id, reason_id, details)
+            # You could get a DetailsRequiredError here, but only if HTML5
+            # validation fails, which should not happen in modern browsers
+            result = super(CustomUnsubscribe, self).mailing(
```
yajo:
```suggestion
            result = super().mailing(
```

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+    def unsubscribe(self, mailing_id, opt_in_ids, opt_out_ids, email, res_id,
+                    token, reason_id=None, details=None):
+        """Store unsubscription reasons when unsubscribing from RPC."""
+        # Update request context
+        self._add_extra_context(mailing_id, res_id, reason_id, details)
+        _logger.debug(
+            "Called `unsubscribe()` with: %r",
+            (mailing_id, opt_in_ids, opt_out_ids, email, res_id, token,
+             reason_id, details))
+        return super(CustomUnsubscribe, self).unsubscribe(
```
yajo:
```suggestion
        return super().unsubscribe(
```

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+            mailing_id, opt_in_ids, opt_out_ids, email, res_id, token)
+
+    @route()
+    def blacklist_add(self, mailing_id, res_id, email, token, reason_id=None,
+                      details=None):
+        self._add_extra_context(mailing_id, res_id, reason_id, details)
+        return super(CustomUnsubscribe, self).blacklist_add(
```
yajo:
```suggestion
        return super().blacklist_add(
```

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+            mailing_id, res_id, email, token)
+
+    @route()
+    def blacklist_remove(self, mailing_id, res_id, email, token,
+                         reason_id=None, details=None):
+        self._add_extra_context(mailing_id, res_id, reason_id, details)
+        return super(CustomUnsubscribe, self).blacklist_remove(
```
yajo:
```suggestion
        return super().blacklist_remove(
```

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+            mailing_id, res_id, email, token)
+
+    def _add_extra_context(self, mailing_id, res_id, reason_id, details):
+        environ = request.httprequest.headers.environ
+        # safe mailing_id and res_id to register a blacklisting or a
```
yajo: I don't understand this comment. Could you reword it please?

mass_mailing_custom_unsubscribe/models/mail_blacklist.py
```diff
@@ -0,0 +1,36 @@
...
+from odoo import models
+
+
+class MailBlackList(models.Model):
+    _inherit = 'mail.blacklist'
+
+    def _add(self, email):
+        mailing_id = self.env.context.get('mailing_id')
+        if mailing_id:
+            mailing = self.env['mail.mass_mailing'].browse(mailing_id)
```
yajo:
```suggestion
            mailing = self.env['mail.mass_mailing'].browse(mailing_id, self._prefetch)
```

mass_mailing_custom_unsubscribe/models/mail_blacklist.py
```diff
@@ -0,0 +1,36 @@
...
+            model_name = mailing.mailing_model_real
+            res_id = self.env.context.get('res_id')
+            self.env["mail.unsubscription"].create({
+                "email": email,
+                "mass_mailing_id": mailing_id,
+                "unsubscriber_id": "%s,%d" % (model_name, res_id),
+                "action": "de_blacklisting",
```
yajo: Apply here the same change as suggested below.

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+        extra_context = {
+            "default_metadata": "\n".join(
+                "%s: %s" % (val, environ.get(val)) for val in (
+                    "REMOTE_ADDR",
+                    "HTTP_USER_AGENT",
+                    "HTTP_ACCEPT_LANGUAGE",
+                )
+            ),
+            "mailing_id": mailing_id,
+            "res_id": int(res_id),
```
yajo: This key is too generic. To avoid possible conflicts:

```suggestion
            "unsubscription_res_id": int(res_id),
```

mass_mailing_custom_unsubscribe/models/mail_blacklist.py
```diff
@@ -0,0 +1,36 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class MailBlackList(models.Model):
+    _inherit = 'mail.blacklist'
+
+    def _add(self, email):
+        mailing_id = self.env.context.get('mailing_id')
```
yajo, line 11:
```suggestion
        mailing_id = self.env.context.get('mailing_id')
        res_id = self.env.context.get('unsubscription_res_id')
```

The `res_id` is required for the whole `if` block to work fine.

mass_mailing_custom_unsubscribe/models/mail_blacklist.py
```diff
@@ -0,0 +1,36 @@
...
+        if mailing_id:
```
yajo:
```suggestion
        if mailing_id and res_id:
```

mass_mailing_custom_unsubscribe/models/mail_blacklist.py
```diff
@@ -0,0 +1,36 @@
...
+            mailing = self.env['mail.mass_mailing'].browse(mailing_id)
+            model_name = mailing.mailing_model_real
+            res_id = self.env.context.get('res_id')
```
yajo:
```suggestion
```

mass_mailing_custom_unsubscribe/models/mail_blacklist.py
```diff
@@ -0,0 +1,36 @@
...
+            self.env["mail.unsubscription"].create({
+                "email": email,
+                "mass_mailing_id": mailing_id,
+                "unsubscriber_id": "%s,%d" % (model_name, res_id),
+                "action": "blacklisting",
+            })
+        return super(MailBlackList, self)._add(email)
+
+    def _remove(self, email):
+        mailing_id = self.env.context.get('mailing_id')
```
yajo, line 26:
```suggestion
        mailing_id = self.env.context.get('mailing_id')
        res_id = self.env.context.get('unsubscription_res_id')
```

The `res_id` is required for the whole `if` block to work fine.

mass_mailing_custom_unsubscribe/models/mail_blacklist.py
```diff
@@ -0,0 +1,36 @@
...
+        if mailing_id:
```
yajo:
```suggestion
        if mailing_id and res_id:
```

mass_mailing_custom_unsubscribe/models/mail_blacklist.py
```diff
@@ -0,0 +1,36 @@
...
+            mailing = self.env['mail.mass_mailing'].browse(mailing_id)
```
yajo:
```suggestion
            mailing = self.env['mail.mass_mailing'].browse(mailing_id, self._prefetch)
```

mass_mailing_custom_unsubscribe/models/mail_blacklist.py
```diff
@@ -0,0 +1,36 @@
...
+            model_name = mailing.mailing_model_real
+            res_id = self.env.context.get('res_id')
```
yajo:
```suggestion
```

mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,34 @@
...
+            mailing_list_ids = [r.list_id.id for r in subscriptions]
+            # reason_id and details are expected from the context
+            self.env["mail.unsubscription"].create({
+                "email": email,
+                "mass_mailing_id": self.id,
+                "unsubscriber_id": "%s,%d" % (model_name, contact.id),
+                'mailing_list_ids': [(6, False, mailing_list_ids)],
+                "action": action,
+            })
+        return super(MailMassMailing, self).update_opt_out(
```
yajo:
```suggestion
        return super().update_opt_out(
```

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
```diff
@@ -0,0 +1,120 @@
...
+        default=lambda self: self._default_date(),
+        required=True)
+    email = fields.Char(
+        required=True)
+    action = fields.Selection(
+        selection=[
+            ("subscription", "Subscription"),
+            ("unsubscription", "Unsubscription"),
+            ("blacklisting", "Blacklisting"),
+            ("de_blacklisting", "De-blacklisting"),
```
yajo: `de_blacklisting` sounds to me like blacklisting for German people. :sweat_smile: 

How about changing these 2 options to `blacklist_add` and `blacklist_rm`?

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
```diff
@@ -0,0 +1,120 @@
...
+        model = self.env['ir.model'].search(
+            [('model', 'in', MASS_MAILING_BUSINESS_MODELS)]).mapped('model')
+        return self.map_mailing_list_models(model)
+
+    @api.multi
+    @api.constrains("action", "reason_id")
+    def _check_reason_needed(self):
+        """Ensure reason is given for unsubscriptions."""
+        for one in self:
+            if one.action == "unsubscription" and not one.reason_id:
```
yajo: Should we change it to this now?:

```suggestion
            if one.action in {"unsubscription", "blacklist_add"} and not one.reason_id:
```

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
```diff
@@ -0,0 +1,120 @@
...
+        """Ensure details are given if required."""
+        for one in self:
+            if not one.details and one.details_required:
+                raise exceptions.DetailsRequiredError(
+                    _("Please provide details on why you are unsubscribing."))
+
+    @api.model
+    def create(self, vals):
+        # No reasons for subscriptions
+        if vals.get("action") == "subscription":
```
yajo: Also here should it be changed?

```suggestion
        if vals.get("action") in {"subscription", "blacklist_rm"}:
```

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,279 @@
...
+/* This JS module replaces core AJAX submission because it is impossible
+ * to extend it as it is currently designed. */
+odoo.define('mass_mailing_custom_unsubscribe.unsubscribe', function (require) {
+    'use strict';
+
+    var ajax = require('web.ajax');
+    var core = require('web.core');
+    require('web.dom_ready');
+
+    var _t = core._t;
```
yajo, line 15: You need to add [these lines](https://github.com/odoo/odoo/blob/57320593f76bc7f6e8c477b6da06b9e3eb49d813/addons/mass_mailing/static/src/js/unsubscribe.js#L15-L17) around here to avoid executing the rest of the code where not needed.

About the code in the rest of the file, I can see that you have copy-pasted the one from Odoo v12 themselves. I imagine there must be some legal issue, since you removed their copyright here and relicensed it here AGPL (Odoo is LGPL). Also, the code will not validate with our linter.

In v11, this code was object-oriented, more properly organized. Wouldn't it be possible to do it like that? If you see this is not feasible (which I would understand), then please at least disable the linter in this file.

Yikes! This is such an ugly part of Odoo... :man_facepalming:  We should some day publish to master properly organized code...

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,279 @@
...
+                $info_state.removeClass('alert-success').removeClass('alert-info').removeClass('alert-warning').addClass('alert-error');
+            });
+    }
+    else {
+        $('#div_blacklist').hide();
+    }
+
+    var unsubscribed_list = $("input[name='unsubscribed_list']").val();
+    if (unsubscribed_list) {
+        $('#subscription_info').html(_t('You have been <strong>successfully unsubscribed from ' + unsubscribed_list + "</strong>."));
```
yajo: Include https://github.com/odoo/odoo/pull/34643 please.

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,279 @@
+    /* Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
```
yajo:
```suggestion
/* Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
```

mass_mailing_custom_unsubscribe/templates/general_reason_form.xml
```diff
@@ -0,0 +1,47 @@
...
+            </div>
+            <textarea class="form-control" name="details" cols="60" rows="3"></textarea>
+        </div>
+    </template>
+
+    <template id="reason_form" name="Unsubscription Reason Form">
+        <t t-call="mass_mailing.layout">
+            <div class="container">
+                <div class="row">
+                    <form id="reason_form" t-attf-action="/mail/mailing/#{mailing_id}/unsubscribe?token=#{token}&amp;debug=assets" method="post" class="col-lg-6 offset-lg-3 mt-4">
```
yajo:
```suggestion
                    <form id="reason_form" t-attf-action="/mail/mailing/#{mailing_id}/unsubscribe?token=#{token}" method="post" class="col-lg-6 offset-lg-3 mt-4">
```

mass_mailing_custom_unsubscribe/templates/mass_mailing_contact_reason.xml
```diff
@@ -0,0 +1,46 @@
...
+<!-- Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <template id="unsubscribe" inherit_id="mass_mailing.unsubscribe" name="Add Reasons to Mailing List Management Form">
+        <!-- Disable core AJAX submission of form, because it is impossible to
+             extend it as it is designed right now. -->
+        <xpath expr="//div[hasclass('container', 'o_unsubscribe_form')]" position="attributes">
+            <attribute name="class" value="container o_unsubscribe_form_custom"/>
```
yajo: IMHO it makes more sense to just disable upstream JS file in the assets definition. Remove this part and see below.

mass_mailing_custom_unsubscribe/templates/mass_mailing_contact_reason.xml
```diff
@@ -0,0 +1,46 @@
...
+        </xpath>
+    </template>
+
+    <template id="unsubscribed"
+              inherit_id="mass_mailing.unsubscribed"
+              name="Add Reasons to Blacklist Management Form">
+        <!-- Disable core AJAX submission of form, because it is impossible to
+             extend it as it is designed right now. -->
+        <xpath expr="//div[hasclass('container', 'o_unsubscribe_form')]" position="attributes">
+            <attribute name="class" value="container o_unsubscribe_form_custom"/>
```
yajo: Same here.

mass_mailing_custom_unsubscribe/views/assets.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+<template id="assets_backend"
+          inherit_id="web.assets_backend">
```
yajo:
```suggestion
          inherit_id="mass_mailing.assets_backend">
```

mass_mailing_custom_unsubscribe/views/assets.xml
```diff
@@ -0,0 +1,15 @@
...
+    <xpath expr=".">
```
yajo:
```suggestion
    <xpath expr="//script[@src='/mass_mailing/static/src/js/unsubscribe.js']" position="replace">
```

This way you disable upstream JS file and enable yours.

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
yajo: Tests are using these exceptions to make sure the constraints work fine.

yajo (changes requested):
There are a few stylistic suggestions, some questions and a few requirements to do.

In general terms, the next time you cannot get a tour to pass, just publish the PR with the failing tour, so we reviewers can help you make it pass again. Publishing the PR removing the tour feels like "tricking" somehow. So, please, restore the tours so we can see the failure and fix it. :wink: 

Thanks!

mass_mailing_custom_unsubscribe/models/mail_unsubscription.py
ernestotejeda: haha, this is where I first saw that term https://github.com/odoo/odoo/blob/98f5ab8489cec60b82d73b5c65d51791bddcbca0/addons/mass_mailing/controllers/main.py#L139 and then I search a bit about that and i think that term is used in that scope. Anyway, I have changed it to avoid confusion

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
ernestotejeda: I haven't found the way to do this code as a object-oriented code as it was done in the previous version because now you can't include `website.content.snippets.animation` to extend it so that now in v12 this module doesn't depend on the website module which includes the Animation class. What I did now was disable the linters as you suggested.

There was no copyright in the original js odoo file https://github.com/odoo/odoo/blob/743c790fa15e9137bd9b8e4212b58d73447d5b88/addons/mass_mailing/static/src/js/unsubscribe.js#L1
Anyway, should I relicense it here AGPL? Is it just replace `License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).` by `License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html) .`?

ernestotejeda:
@yajo I have made all the changes except some to which I have responded with a comment

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
yajo: > this module doesn't depend on the website module which includes the Animation class

Oh interesting. I knew there had to be a reason!

> There was no copyright in the original js odoo file

That's because they copyright the whole repo, not file by file; but you should add it here since they have it in this particular file.



> Anyway, should I relicense it here AGPL?

You must license this file as LGPL. That's its original license.

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,286 @@
+/* Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
```
yajo:
```suggestion
 * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
```

mass_mailing_custom_unsubscribe/views/assets.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+<template id="assets_backend"
+          inherit_id="mass_mailing.assets_backend">
+    <xpath expr="//script[@src='/mass_mailing/static/src/js/unsubscribe.js']" position="replace">
```
yajo, line 10: Please add this little comment to explain this weird replacement we're doing.

```suggestion
    <!-- Need to replace the asset because upstream is impossible to extend -->
    <xpath expr="//script[@src='/mass_mailing/static/src/js/unsubscribe.js']" position="replace">
```

yajo (changes requested):
Just a couple of details to finish.

ernestotejeda:
Changes done

yajo (approved):

mass_mailing_custom_unsubscribe/models/mail_mass_mailing_list.py
```diff
@@ -1,14 +0,0 @@
-# Copyright 2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
-from odoo import fields, models
-
-
-class MailMassMailing(models.Model):
-    _inherit = "mail.mass_mailing.list"
-
-    not_cross_unsubscriptable = fields.Boolean(
```
pedrobaeza: Why this feature has been removed? It's still needed in our customer.

pedrobaeza (changes requested):
Why the module has no tours now? Why there are removed features? If the scope of the module has changed as well, README should reflect these changes.

@Yajo please support @ernestotejeda for a perfect migration, or tell me directly in a meeting the design decisions

mass_mailing_custom_unsubscribe/models/mail_mass_mailing_list.py
yajo: Check the PR description, it's explained there. :wink:

yajo (changes requested):
Good point, I forgot the tours thing.

You have to restore them. Please don't remove tours just because they fail; on migrations it's a common thing, and it's not expected that we hide that behind the carpet.

If you cannot make them pass, just push the code and we'll help you :wink:

chienandalu:
> In general, the link that appears at the end of the newsletters is 'unsubscribe', but it is not 'Update subscriptions/unsubscription from mailing lists', therefore, I don't see so bad that first the user is unsubscribed after click on that link and then the status of the subscriptions is displayed.

Ok, @ernestotejeda Adapt tours to the new behavior and document it if necessary

mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,33 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+from itertools import groupby
+
+
+class MailMassMailing(models.Model):
+    _inherit = "mail.mass_mailing"
+
+    def update_opt_out(self, email, list_ids, value):
```
Tardo, line 11: @api.multi

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,286 @@
...
+            $radios.prop("required", true);
+        }
+    };
+
+    $mailing_lists.change(function (e) {
+        toggle_reasons();
+        $('#info_state').addClass('invisible');
+    });
+
+    if (email != '' && email != undefined) {
```
Tardo, line 68:
```suggestion
    if (!_.isEmpty(email)) {
```

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,286 @@
...
+        ajax.jsonRpc('/mailing/blacklist/check', 'call', {
+            'email': email,
+            'mailing_id': mailing_id,
+            'res_id': res_id,
+            'token': token
+        })
+            .then(function (result) {
+                if (result == 'unauthorized') {
```
Tardo, line 76:
```suggestion
                if (result === 'unauthorized') {
```

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,286 @@
...
+                    $('#button_add_blacklist').hide();
+                    $('#button_remove_blacklist').hide();
+                }
+                else if (result == true) {
```
Tardo, line 80:
```suggestion
                else if (result === true) {
```

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,286 @@
...
+                    $('#button_remove_blacklist').show();
+                    toggle_opt_out_section(false);
+                }
+                else if (result == false) {
```
Tardo, line 84:
```suggestion
                else if (result === false) {
```

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,286 @@
...
+    else{
+        $('#subscription_info').html(_t('You have been <strong>successfully unsubscribed</strong>.'));
+    }
+
+    $('#unsubscribe_form').on('submit', function (e) {
+        e.preventDefault();
+
+        var checked_ids = [];
+        $("input[type='checkbox']:checked").each(function (i) {
+            checked_ids[i] = parseInt($(this).val(), 10);
```
Tardo, line 118: Here perhaps is best use 'unary plus' instead of parseInt()

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,286 @@
...
+        });
+
+        var unchecked_ids = [];
+        $("input[type='checkbox']:not(:checked)").each(function (i) {
+            unchecked_ids[i] = parseInt($(this).val(), 10);
```
Tardo, line 123: Here perhaps is best use 'unary plus' instead of parseInt()

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,286 @@
...
+        if (!$("#unsubscribe_form")[0].reportValidity()) {
+            return;
+        }
+
+        ajax.jsonRpc('/mailing/blacklist/add', 'call', {
+            'email': email,
+            'mailing_id': mailing_id,
+            'res_id': res_id,
+            'token': token,
+            'reason_id': parseInt($reasons.find("[name='reason_id']:checked").val(), 10),
```
Tardo, line 193: Here perhaps is best use 'unary plus' instead of parseInt()

Tardo (changes requested):

mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
pedrobaeza: In fact, it's not needed and we should tend to the contrary: all methods are `@api.multi` by default, and it's going to be removed in v13.

yajo:
@Tardo, I think @ernestotejeda just copy-pasted [upstream's file](https://github.com/odoo/odoo/blob/12.0/addons/mass_mailing/static/src/js/unsubscribe.js) and tried to stay as close as possible to upstream source code.

In this case, upstream code isn't inheritable and we have to replace it completely (@ged-odoo could you help with that? :innocent:), so it's one of the cases where it's more maintainable to disable the linter than to obey it (so, when we migrate, we copy the new source code from odoo v13 and just apply our little diff).

victor-ditay (approved):
Code LGTM but I would like someone to show me funcional use

cc @chienandalu @rafaelbn

ernestotejeda:
I have made two important changes:
1. I have added again the field not_cross_unsubscriptable to the mail.mass_mailing.list class and I have made the required adaptations so that this field has the same effect as in the previous version of the module. I had removed it because I thought it was no longer necessary with the new is_public field, but no, I was wrong.
2. I was able to make Odoo see the tours and I add the UI tests and I  adapted its to the changes made in the module in this migration. to make Odoo see the tours I had to add `<t t-call =" web.assets_frontend "/>` to the <head> of the custom template that odoo uses for the unsubscription.

ernestotejeda:
@victormmtorres , Here you have a GIF with a use case so you can see how it works. The execution is a bit fast, but I hope it helps you:

Peek 18-07-2019 15-38

mass_mailing_custom_unsubscribe/models/mail_mass_mailing_list.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class MailMassMailing(models.Model):
+    _inherit = "mail.mass_mailing.list"
+
+    not_cross_unsubscriptable = fields.Boolean(
```
yajo, line 10: I personally dislike the negative booleans, where a "no" means "yes". They lead to double negation blocks such as:

```python
if not record.not_cross_unsubscriptable:
    # It's cross_unsubscriptable! 😵
```

All you'd need is call this field `cross_unsubscriptable`, do code changes, add `default=True` and add a migration script that inverts its value wherever present.

Well, this is not a strict requirement, since this was there before as it is now, but I'd really appreciate it.

mass_mailing_custom_unsubscribe/tests/test_ui.py
```diff
@@ -0,0 +1,176 @@
...
+    def setUp(self):
+        super(UICase, self).setUp()
+        self.email = "test.contact@example.com"
+        self.mail_postprocess_patch = mock.patch(
+            "odoo.addons.mass_mailing.models.mail_mail.MailMail."
+            "_postprocess_sent_message",
+            autospec=True,
+            side_effect=self.extract_url,
+        )
+        with self.tempenv() as env:
```
yajo: Remove this `with` block and dedent inner lines, and make them use `self.env` instead of `env`.

This is a workaround for v11 and lower versions, but in v12 it's not needed nor helpful.

mass_mailing_custom_unsubscribe/tests/test_ui.py
```diff
@@ -0,0 +1,176 @@
...
+                "name": "test mailing %d" % n,
+                "mailing_model_id": self.env.ref(
+                    'mass_mailing.model_mail_mass_mailing_list').id,
+                "contact_list_ids": [
+                    (6, 0, [self.lists[0].id, self.lists[3].id])
+                ],
+                "reply_to_mode": "thread",
+            })
+            self.mailing._onchange_model_and_list()
+            # HACK https://github.com/odoo/odoo/pull/14429
```
yajo: I wonder if this is still needed. Maybe you could try to remove it.

mass_mailing_custom_unsubscribe/tests/test_ui.py
```diff
@@ -0,0 +1,176 @@
...
+                    </a>
+                </div>
+            """
+
+    def tearDown(self):
+        del self.email, self.lists, self.contact, self.mailing, self.url
+        super(UICase, self).tearDown()
+
+    @contextmanager
+    def tempenv(self):
```
yajo: :fire: remove this method.

mass_mailing_custom_unsubscribe/tests/test_ui.py
```diff
@@ -0,0 +1,176 @@
...
+                self.lists = self.lists.with_env(env)
+                self.contact = self.contact.with_env(env)
+                self.mailing = self.mailing.with_env(env)
+            except AttributeError:
+                pass  # We are in :meth:`~.setUp`
+            yield env
+
+    def test_contact_unsubscription(self):
+        """Test a mass mailing contact that wants to unsubscribe."""
+        with self.tempenv() as env:
```
yajo: Remove this block and do as explained above.

mass_mailing_custom_unsubscribe/tests/test_ui.py
```diff
@@ -0,0 +1,176 @@
...
+
+        tour = "mass_mailing_custom_unsubscribe_tour_contact"
+        self.browser_js(
+            url_path=self.url,
+            code=self._tour_run % tour,
+            ready=self._tour_ready % tour,
+            login='demo')
+
+        # Check results from running tour
+        with self.tempenv() as env:
```
yajo: Remove this block and do as explained above.

mass_mailing_custom_unsubscribe/tests/test_ui.py
```diff
@@ -0,0 +1,176 @@
...
+            # tour and they are all the existing unsubscriptions
+            self.assertEqual(
+                unsubscription_1 | unsubscription_2 | unsubscription_3,
+                env["mail.unsubscription"].search([])
+            )
+            self.assertEqual(3, len(env["mail.unsubscription"].search([])))
+
+    def test_partner_unsubscription(self):
+        """Test a partner that wants to unsubscribe."""
+        with self.tempenv() as env:
```
yajo: Remove this block and do as explained above.

mass_mailing_custom_unsubscribe/tests/test_ui.py
```diff
@@ -0,0 +1,176 @@
...
+
+        tour = "mass_mailing_custom_unsubscribe_tour_partner"
+        self.phantom_js(
+            url_path=self.url,
+            code=self._tour_run % tour,
+            ready=self._tour_ready % tour,
+            login='demo')
+
+        # Check results from running tour
+        with self.tempenv() as env:
```
yajo: Remove this block and do as explained above.

yajo (changes requested):
@Tardo I took the freedom to mark all of your JS comments as resolved, as I saw your :+1: on https://github.com/OCA/social/pull/402#issuecomment-511707513. Maybe you want to update your review?

mass_mailing_custom_unsubscribe/tests/test_ui.py
ernestotejeda: That PR was made for version odoo:9.0 but is still needed for v12: https://github.com/odoo/odoo/blob/20eca78cb9afd6288f5900d4833dc83cbe4fa811/addons/mass_mailing/models/mass_mailing.py#L628

mass_mailing_custom_unsubscribe/models/mail_mass_mailing_list.py
pedrobaeza: @Yajo, although I somehow agree on the basis, we can't spend more time in this migration, so we can put this in the ROADMAP and change it in other round

rafaelbn: I think @pedrobaeza spend time in this migration, it will save us a lot of time after the migration.

@victormmtorres I wish after our meeting | training of today can review this part meticulously

rafaelbn (changes requested):
This module has been lots of issues in the past so please @chienandalu @Yajo review in deep.

mass_mailing_custom_unsubscribe/models/mail_mass_mailing_list.py
pedrobaeza: What do you mean? This is only semantic, so the change is optional. The feature is there.

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,286 @@
+/* Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+ * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+
+/* This JS module replaces core AJAX submission because it is impossible
+ * to extend it as it is currently designed. */
```
yajo: We explained in https://github.com/OCA/social/pull/402#issuecomment-511707513 that this file is almost a copy+paste from upstream, and that's why it's so ugly.

So, to avoid making readers and linters crazy, let's add this here:

```suggestion
 * to extend it as it is currently designed. It is almost a copy+paste from
 * upstream, to allow easier version/patch updates, so linter is disabled. */
 /* eslint-disable */
```

yajo (changes requested):
Nice, now you have some linters to fix and you're done.

- JS lints: https://travis-ci.org/OCA/social/jobs/561045770#L432-L644 (see also comment below)
- Python lints: https://travis-ci.org/OCA/social/jobs/561045770#L322-L323

ernestotejeda:
Changes done

yajo (approved):

PR #242, opened 2018-03-12
[ADD] mail_drop_target

hbrunn:
Drag & drop emails to Odoo
==========================

This module was written to allow users to drag&drop emails from their desktop to Odoo.

Usage
=====

To use this module, you need to:

- save your emails on the desktop / somewhere in the file system
- drag them to your browser, and drop them on the chatter of the record you want to attach your email to

Known issues / Roadmap
======================

* most mail clients won't allow you to drag mails directly from the mail client, you'll need some plugin for that
* for corporate environments, it might be feasible to support imap URLs and get the mail in question on the server side

PR #403, opened 2019-06-25
[MIG][11.0] mail_drop_target

etobella:
Migration of module mail_drop_target #242 

Depends on OCA/web#1313


etobella:
ping @jbeficent @jarroyomorales

hbrunn:
please add a commit to temporarily change `web` to your branch in the dependencies file in order to run tests

jarroyomorales (approved):

JordiBForgeFlow (approved):
:+1: tested, and works like a charm!

mail_drop_target/static/src/js/mail_drop_target.js
```diff
@@ -0,0 +1,78 @@
...
+            if (drop_file.name.endsWith('.msg')) {
+                mail_processor = 'message_process_msg';
+                data = base64js.fromByteArray(
+                    new Uint8Array(reader.result)
+                );
+            } else {
+                mail_processor = 'message_process';
+                var reader_array = new Uint8Array(reader.result);
+                data = ""
+                for (var i = 0; i < reader_array.length; i++) {
```
etobella, line 59: I detected an issue when uploading big emails, now it should be solved.

jarroyomorales:
@etobella Could you add some tests?

etobella:
@jarroyomorales I think that all possible tests have been added.

etobella:
Also, dependency to PR of web is no longer required. Now it is ready to review :green_apple: 
@hbrunn

jarroyomorales:
@etobella Perfect!!!

jarroyomorales:
@OCA/social-maintainers I think this is ready to merge

rafaelbn:
I'm going to test it now :smile:

rafaelbn:
Hi, I'm trying but it look like runbot doesn't build :cry:

jarroyomorales:
@etobella Any idea?

etobella:
It worked as expected for me on the runbot. But the build ended like 40 minutes ago. Can you retest now @rafaelbn ?

rafaelbn:
Hi @etobella , 

Are you kidding me? This module is amazing!

- I have tested and I cannot drag and drop the same email in 2 diferente objets. For example in two diferent crm.lead o in a res.partner and a crm.lead
- When you drag and drop the email and it doesn't work you dont' know why and you try and try until de picture of a monkey reflecting in the screed notice you that you must stop!
- Why this name???? it should be mail_draganddrop_target . Isn't it?

Thanks!

rafaelbn:
Please @etobella what do you think about my comments? :smile:  :arrow_up_small:

etobella:
Sorry @rafaelbn I had some issues, and I couldn't review it. It is true that odoo checks if the message has already been imported, but it comes from core:
https://github.com/odoo/odoo/blob/11.0/addons/mail/models/mail_thread.py#L1298
We could try to raise an exception. Probably it is the best approach I will check today how to do it.
The names comes from the original module name `web_drop_target`

hbrunn:
`drop target` is a technical term that describes exactly what those modules do, providing something (a target) we can drop stuff on.

The exception seems fine to me, with extra points for telling the user where the mail is attached already. So if they really want to have it on the current record, they can use `mail_edit` to change the mail's record. Would be nice to implement this in a way that (to my knowledge not existing) `mail_chatter_multi_records` or the like could intercept there and attach the mail anyways, but I don't think that's something this module should do itself.

hbrunn:
@etobella I can return the favor with https://github.com/OCA/social/pull/242/commits/95cf9c89b1e7ead1591fc3b6b7921dc42cc78737

etobella:
Now a warning is raised, but we cannot show the records with a standard warning as it is html escaping. We can try to make a custom widget for this case if necessary, but I don't know it this is the best approach.

hbrunn:
oh, my proposal was nothing fancy, just it already hangs on $model $display_name or something like that

hbrunn:
(but leave out if it poses problems, I don't think that's a show stopper)

etobella:
Well I tried to attach the html on the warning but it is not working

StefanRijnhart:
@hbrunn if I understand correctly, you agree with the other reviewers that this is fit to merge in the current state? Will you do the honours?

LoisRForgeFlow (approved):
Tested locally, LGTM :+1:

pedrobaeza:
Let's see if this one is now accepted: /ocabot merge

PR #404, opened 2019-06-26
[WIP] [ADD] mail_embed_image module

daramousk:


daramousk:
This works when sent on our mail server, but gmail seems to strip these (as tested)

mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,89 @@
...
+            else:
+                if msg.get_content_subtype() == 'html':
+                    body = msg.get_payload(decode=True)
+                    root = fromstring(body)
+                    for img in root.xpath(
+                            "//img[starts-with(@src, '/web/image')]"):
+                        base_url = self.env['ir.config_parameter'].get_param(
+                            'web.base.url')
+                        src = base_url + img.get('src')
+                        response = requests.get(src)
```
hbrunn: as you now enforce that this is an image linked from Odoo, you can just as well determine which attachment this is from the address and read this via the ORM, less overhead

daramousk: I am not sure how you want me to do this.

At this point I do know that the url for an image stored in the filestore begins with `/web/image/` after that a lot of things can exist - including the url containing dimensions of the image that when given the image is resized. All this happens in the controller, so doesn't it make sense to just have the controller take care of everything? Let alone that that request might be forwarded to a different server in different configurations.

FYI at this point all outlook, gmail and Therp's MUAs show the email correctly

hbrunn: look at the [controller for images](https://github.com/OCA/OCB/edit/10.0/addons/web/controllers/main.py#L1007) and then at the definition of `binary_content`

Great the PoC works, now you can deploy this for the customer to test with their version of outlook, add proper tests and open the PR here

hbrunn:
read travis' output, your code fails a lot of other tests.

hbrunn:
concerning gmail, this is what gmail produces itself when embedding an image:

```
MIME-Version: 1.0
Date: Fri, 16 Aug 2019 19:11:59 +0200
Message-ID: <CAMApDL39co2S1X03WB5dA5p98Qe5uUXe794=nMQSVHsC_p=7xg@mail.gmail.com>
Subject: imagetest2
From: XXX
To: XXX
Content-Type: multipart/related; boundary="0000000000008ef16605903f160e"

--0000000000008ef16605903f160e
Content-Type: multipart/alternative; boundary="0000000000008ef16505903f160d"

--0000000000008ef16505903f160d
Content-Type: text/plain; charset="UTF-8"

this should be the therp logo:
[image: website_logo.png]
and some text below to be sure the client shows that too

--0000000000008ef16505903f160d
Content-Type: text/html; charset="UTF-8"

<div dir="ltr">this should be the therp logo:<div><div><img src="cid:ii_jzeddpaa0" alt="website_logo.png" width="542" height="143"><br></div></div><div>and some text below to be sure the client shows that too</div></div>

--0000000000008ef16505903f160d--
--0000000000008ef16605903f160e
Content-Type: image/png; name="website_logo.png"
Content-Disposition: attachment; filename="website_logo.png"
Content-Transfer-Encoding: base64
X-Attachment-Id: ii_jzeddpaa0
Content-ID: <ii_jzeddpaa0>


--0000000000008ef16605903f160e--
```

in some future extension, we could do one thing or the other depending on the recipient address(es), but for now just refactor in a way that this is simple to adapt for other modules.

mail_embed_image/README.rst
```diff
@@ -0,0 +1,80 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: https://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+================
+mail_embed_image
+================
```
hbrunn, line 3: for the final version, add a split readme here and fill it

mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,132 @@
...
+                    # attach the image into the email as attachment
+                    email.attach(filepart)
+                    img.set('src', 'cid:%s' % (cid))
+                part.set_payload(b64encode(tostring(root)))
+        return email
+
+    def _get_image(self, base_url, src):
+        """ This function emulates the operations of
+        odoo.addons.web.controllers.main.Binary.content_image() because
+        we cannot call this from here
```
hbrunn: this was what I expected you to figure out. To be sure that I don't ask the impossible, here a version that runs in python_odoo:

```python
from odoo import http
from werkzeug.test import EnvironBuilder
from mock import Mock, patch

builder = EnvironBuilder(path='/web/image/1')
env = builder.get_environ()
endpoint, arguments = http.routing_map(
    session.env.registry._init_modules, 
    False, 
    converters=session.env['ir.http']._get_converters()
).bind_to_environ(
    env
).match(return_rule=False)

result_dict = {}
with patch(
    'odoo.addons.web.controllers.main.request', 
    env=session.env, 
    registry=session.env.registry,
    make_response = lambda x, y: result_dict.update(
        data=x,
        headers=y,
    ) or Mock(),
) as mock_request, patch(
    'odoo.addons.website.models.ir_http.request',
    env=session.env,
):
    result = endpoint.method(**arguments)
# without the mocking, result contains what you need
print len(result_dict['data'])
print result_dict['headers']
```

The problem with your code is twofold: The code duplication is already bad, but it's worse that you miss possible overrides of the controller that might change the image in whatever ways we can't foresee.

Adapt the above to your situation, and research why and how this works. Add comments with explanations and links to code/documentation that prove you did so. The mocking is ugly, better detect if `odoo.http.request` is bound, and if not, bind it to some fake request and unbind afterwards.

mail_embed_image/tests/test_mail_embed_image.py
```diff
@@ -0,0 +1,88 @@
...
+from mock import patch, mock
+from lxml import html
+from requests import get
+from odoo import api, registry, SUPERUSER_ID
+
+
+reg = registry(common.get_db_name())
+env = api.Environment(reg.cursor(), SUPERUSER_ID, {})
+MockBuildEmail = mock.Mock(wraps=env['ir.mail_server'].build_email)
+MockSendEmail = mock.Mock(wraps=env['ir.mail_server'].send_email)
```
hbrunn: explain the scope of this and how it interacts with the rest of the code, and then move this into your test function

mail_embed_image/tests/test_mail_embed_image.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class TestMailEmbedImage(common.TransactionCase):
+
+    post_install = True
+    at_install = False
```
hbrunn, line 15: don't, there's no reason to do this and it can fall on your feet if some other module later does stuff you don't expect

mail_embed_image/tests/test_mail_embed_image.py
```diff
@@ -0,0 +1,88 @@
...
+        with patch(
+                'odoo.addons.mail_embed_image.models.ir_mail_server.'
+                'IrMailServer.build_email',
+                new=MockBuildEmail,
+                ) as build_email:
+            with patch(
+                    'odoo.addons.base.ir.ir_mail_server.'
+                    'IrMailServer.send_email',
+                    new=MockSendEmail,
+                    ) as send_email:
```
hbrunn: read the python documentation on how to stack context managers

hbrunn (changes requested):

rafaelbn:
Hi @daramousk , travis red :red_circle:

gfcapalbo:
@daramousk  please fix small lints and rebase.

gfcapalbo:
@george , please rebase and make branch green . tests fail.

george:
@gfcapalbo You pinged the wrong @george 😄

gfcapalbo:
@daramousk please rebase and make branch green . tests fail. @george apologies

gfcapalbo:
now replaced by  https://github.com/OCA/social/pull/458
going through comments and implementing them there.
branch can be closed.

PR #407, opened 2019-07-04
[11.0][IMP] mail_tracking: Cc

Tardo:
Adds support to view 'Cc' in messages

mail_cc

Cc @Tecnativa


pedrobaeza:
Please take the occasion for splitting the README by fragments, and add in the description this new feature, both the icon and the suggestion when writing the mail. You can even add the GIF if you consider so.

mail_tracking/models/mail_message.py
```diff
@@ -58,20 +62,41 @@ def tracking_status(self):
...
             res[message.id] = partner_trackings
         return res
 
+    def _get_email_cc(self):
+        res = {}
+        ResPartnerObj = self.env['res.partner']
+        for message in self:
+            email_cc = (message.email_cc or '').split(COMMASPACE)
+            email_cc_no_partner = []
+            groups = ResPartnerObj.read_group([('email', 'in', email_cc)],
```
pedrobaeza: Isn't any way to already obtain this information from upstream code?

mail_tracking/static/src/xml/mail_tracking.xml
```diff
@@ -4,39 +4,46 @@
...
-            <i class="fa fa-check fa-stack-1x" style="margin-left:-3px"></i>
-        </span>
-    </t>
-    <t t-if="tracking[0] == 'opened'">
-        <span class="fa-stack mail_tracking_opened mail_tracking_pointer">
-            <i class="fa fa-check fa-stack-1x" style="margin-left:1px"></i>
-            <i class="fa fa-check fa-inverse fa-stack-1x" style="margin-left:-2px;"></i>
-            <i class="fa fa-check fa-stack-1x" style="margin-left:-3px"></i>
-        </span>
+    <t t-else="">
```
pedrobaeza: Instead of changing the indentation, I think you can put `t-elif` in all tags

mail_tracking/models/mail_message.py
Tardo: I don't see any stuff that can help here...

pedrobaeza: OK, maybe others that know more about the module can say.

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -2,35 +2,35 @@
...
 
     ChatThread.include({
         events: _.extend(ChatThread.prototype.events, {
-            'click .o_mail_action_tracking_partner': 'on_tracking_partner_click',
+            'click .o_mail_action_tracking_partner':
+                'on_tracking_partner_click',
             'click .o_mail_action_tracking_status': 'on_tracking_status_click',
         }),
-        _preprocess_message: function (message) {
+        _preprocess_message: function () {
```
chienandalu, line 30: Is this ok? Aren't you cropping the _super arguments?

mail_tracking/models/mail_thread.py
```diff
@@ -0,0 +1,44 @@
...
+            content_subtype=content_subtype, **kwargs)
+        email_cc = kwargs.get('cc')
+        if email_cc:
+            new_message.sudo().write({
+                'email_cc': email_cc,
+            })
+        return new_message
+
+    @api.multi
+    def message_get_suggested_recipients(self):
```
chienandalu, line 24: Please cover this with tests

chienandalu (changes requested):
Thanks Alex.

I made a functional test and I couldn't make it work:

- I made a new message in a res.partner thread with a CC mail address.
- I got the email in the CC address right.

But:

- No tracking events were registered for the CC, just for the main one.
- The CC address is not in the trackings of the chatter
image

Maybe I'm doing the tests wrong?

pedrobaeza:
@chienandalu the idea is the contrary: when you receive a message with a CC email that is not in your contact list (or being there, that for now appears with the forbidden sign).

mail_tracking/models/mail_message.py
```diff
@@ -1,12 +1,17 @@
...
-from odoo import models, api
+from odoo import models, api, fields
+from email.utils import COMMASPACE
 
 
 class MailMessage(models.Model):
     _inherit = "mail.message"
 
+    # Recipients
+    email_cc = fields.Char("Cc")
```
```suggestion
    email_cc = fields.Char("Cc", help='Additional recipients that receive a "Carbon Copy" of the e-mail')
```

mail_tracking/models/mail_message.py
```diff
@@ -58,20 +63,42 @@ def tracking_status(self):
             for partner in partners:
                 # If there is partners not included, then status is 'unknown'
                 partner_trackings.append((
-                    'unknown', False, partner.name, partner.id))
+                    'unknown', False, partner.name, partner.id, partner.email))
             res[message.id] = partner_trackings
         return res
 
+    @api.model
```
yajo: According to your code, this fits better:

```suggestion
    @api.multi
```

mail_tracking/models/mail_message.py
```diff
@@ -58,20 +63,42 @@ def tracking_status(self):
...
+    def _get_email_cc(self):
+        res = {}
+        ResPartnerObj = self.env['res.partner']
+        for message in self:
+            email_cc_list = (message.email_cc or '').split(COMMASPACE)
```
yajo: It's better that you use one of `email_split()` or `email_split_and_format()` methods from the `odoo.tools.mail` package.

This should detect properly things like `I'm someone <i+some.one@a-dns-rec.ow.ner.net>,Other addr <other@add.res.co>` which have lots of possible corner cases.

mail_tracking/models/mail_message.py
```diff
@@ -58,20 +63,42 @@ def tracking_status(self):
...
+            email_cc_list_checked = []
+            if any(email_cc_list):
+                partners = ResPartnerObj.search([
+                    ('email', 'in', email_cc_list)
+                ])
+                for cc in email_cc_list:
+                    partners_cc = partners.filtered(lambda x: x.email == cc)
```
yajo: This loop in loop in loop will perform badly, possibly.

How about this?

```python
found_cc = set(partners.mapped("email"))
email_cc_list_checked = [(cc, cc in found_cc) for cc in email_cc_list]
```

mail_tracking/models/mail_thread.py
```diff
@@ -0,0 +1,44 @@
...
+        res = super().message_get_suggested_recipients()
+        ResPartnerObj = self.env['res.partner']
+        for record in self:
+            messages = record.message_ids.filtered(lambda x: x.email_cc)
```
yajo:
```suggestion
            messages = record.message_ids.filtered("email_cc")
```

It's the same.

mail_tracking/models/mail_thread.py
```diff
@@ -0,0 +1,44 @@
...
+            for msg in messages:
+                email_cc = msg.email_cc.split(COMMASPACE)
```
yajo: Also here it's better that you use one of `email_split()` or `email_split_and_format()` methods from the `odoo.tools.mail` package.

mail_tracking/models/mail_thread.py
```diff
@@ -0,0 +1,44 @@
...
+                for cc in email_cc:
+                    partner_id = record.message_partner_info_from_emails(
+                        [cc])[0].get('partner_id')
```
yajo: Can't you adapt the code to call this method with a list of emails instead, something like this?:

```suggestion
                        email_cc)
```

Of course you'd have to take it outside the loop and adapt a little bit more the code. This will help the backend code working in batch and reducing queries.

mail_tracking/models/mail_thread.py
```diff
@@ -0,0 +1,44 @@
...
+                    if not partner_id:
+                        res[record.ids[0]].append((False, cc, "Cc"))
```
yajo: `"Cc"` is a string displayed to the user, right? If so:

```suggestion
                        res[record.ids[0]].append((False, cc, _("Cc")))
```

mail_tracking/models/mail_thread.py
```diff
@@ -0,0 +1,44 @@
...
+                    else:
+                        partner = ResPartnerObj.browse(partner_id)
```
yajo:
```suggestion
                        partner = ResPartnerObj.browse(partner_id, partner._prefetch)
```

This will save you some queries.

mail_tracking/models/mail_thread.py
```diff
@@ -0,0 +1,44 @@
...
+                        record._message_add_suggested_recipient(
+                            res, partner=partner, reason='Cc')
```
yajo: I'm also not sure about this, but possibly it's needed. Could you check it please?:

```suggestion
                            res, partner=partner, reason=_('Cc'))
```

mail_tracking/static/src/js/mail_tracking.js
yajo: No, it's OK. Just one of those JS weirdies. He's using the built-in `arguments` keyword below, which holds them all, no matter how they are defined in the method signature.

mail_tracking/static/src/xml/mail_tracking.xml
```diff
@@ -46,11 +52,15 @@
...
+                <t t-set="isCc" t-value="false" />
+                <t t-foreach="message.email_cc" t-as="cc">
+                    <t t-if="cc[0] == tracking[4]" t-set="isCc" t-value="true" />
+                </t>
                 <t t-if="!first_tracking">
                       -
                 </t>
                 <t t-if="tracking[3]">
-                    <a class="o_mail_action_tracking_partner"
+                    <a t-attf-class="o_mail_action_tracking_partner #{ isCc?'o_mail_cc':'' }"
```
yajo:
```suggestion
                    <a t-attf-class="o_mail_action_tracking_partner #{isCc ? 'o_mail_cc' : ''}"
```

mail_tracking/models/mail_thread.py
```diff
@@ -0,0 +1,44 @@
...
+from odoo import models, api
+from email.utils import COMMASPACE
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    @api.multi
+    @api.returns('self', lambda value: value.id)
+    def message_post(self, body='', subject=None, message_type='notification',
```
yajo: Since all you care about is a `kwarg`, you can use this here:

```suggestion
    def message_post(self, *args, **kwargs):
```

mail_tracking/models/mail_thread.py
```diff
@@ -0,0 +1,44 @@
...
+                     subtype=None, parent_id=False, attachments=None,
+                     content_subtype='html', **kwargs):
+        new_message = super().message_post(
```
yajo: ... and then this here:

```suggestion
        new_message = super().message_post(*args, **kwargs)
```

And it will do exactly the same.

yajo (changes requested):
Some changes suggested or requested. Thanks!

mail_tracking/models/mail_message.py
Tardo: Thanks!

mail_tracking/models/mail_thread.py
Tardo: Done

Tardo:
Changes done, thanks!

mail_tracking/models/mail_thread.py
yajo: This one is still missing.

pedrobaeza:
@Yajo can you please re-review?

pedrobaeza:
2 possible improvements/bugs:

* If the CC mail is not yet registered as partner, there's no information when you click on the icon:

Selección_043

Shouldn't it be linked to the corresponding mail? Or if not possible (not being any mail.message.tracking or similar record), don't do that icon clickable.

* Linked to this, when we have later created the contact for that CC replying in the thread, if I refresh the window, the CC disappears from the tracking of the previous message, although it appears in the new one.

mail_tracking/models/mail_message.py
```diff
@@ -58,20 +64,42 @@ def tracking_status(self):
...
+        ResPartnerObj = self.env['res.partner']
+        for message in self:
+            email_cc_list = email_split(message.email_cc)
+            email_cc_list_checked = []
+            if any(email_cc_list):
+                partners = ResPartnerObj.search([
+                    ('email', 'in', email_cc_list)
+                ])
+                found_cc = set(partners.mapped("email"))
+                # Tuple: str:email, bool:hasPartner
```
yajo: I think this method deserves a docstring, and this comment fits better in it.

Tardo:
Changes done. Thanks!

pedrobaeza:
Travis is red

yajo (approved):

pedrobaeza:
Please check Travis in order to merge this

mail_tracking/models/mail_message.py
```diff
@@ -58,20 +64,49 @@ def tracking_status(self):
...
                 # If there is partners not included, then status is 'unknown'
                 partner_trackings.append((
-                    'unknown', False, partner.name, partner.id))
+                    'unknown', False, partner.name, partner.id, partner.email))
             res[message.id] = partner_trackings
         return res
 
+    @api.multi
+    def _get_email_cc(self):
+        """This method gets all Cc mails and the associated partner if exsits.
```
victor-ditay:
```suggestion
        """This method gets all Cc mails and the associated partner if exist.
```

mail_tracking/models/mail_message.py
```diff
@@ -1,12 +1,18 @@
 # Copyright 2016 Antonio Espinosa - <antonio.espinosa@tecnativa.com>
+# Copyright 2019 Alexandre Díaz
```
victor-ditay, line 2: missing email info about you

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -112,6 +112,33 @@ def test_message_post(self):
...
+        # mail cc
+        email_cc = message_dict['email_cc'][0]
+        self.assertEqual(email_cc[0], 'unnamed@test.com')
+        self.assertFalse(email_cc[1])
+        email_cc = message_dict['email_cc'][1]
+        self.assertEqual(email_cc[0], 'sender@example.com')
+        self.assertTrue(email_cc[1])
+        # suggested recipients
+        recipients = self.recipient.message_get_suggested_recipients()
+        self.assertEqual(recipients[self.recipient.id][0][1],
```
victor-ditay, line 140: Travis is failing here:
```
FAIL: test_email_cc (odoo.addons.mail_tracking.tests.test_mail_tracking.TestMailTracking)
`   File "/home/travis/build/OCA/social/mail_tracking/tests/test_mail_tracking.py", line 132, in test_email_cc
`     self.assertEqual(email_cc[0], 'unnamed@test.com')
` AssertionError: 'sender@example.com' != 'unnamed@test.com'
` - sender@example.com
` + unnamed@test.com
```

mail_tracking/models/mail_message.py
pedrobaeza: We prefer nowadays to not put emails

Tardo:
Changes done, thanks!

PR #411, opened 2019-07-15
[11.0][IMP] mail_tracking: Filter failed messages

Tardo:
Changes made in this PR:
- Adds 'Failed' section on Discuss
- Adds 'Failed Messages' on top of the chatter
- Adds 'Failed Sent Messages' filter

Because of how chatter & discuss is implemented I had to use some hacks to get the main functionality working.
Main 'hacks':
- Can't add new 'static' channel to discuss: Uses python to send it like a regular channel.
- Can't update channel changes on discuss: Show a clickable text to indicate that need refresh data
- Can't set the channel domain: Uses python to modify domain
- Can't override chat_manager 'start': Relaunched 'mail/client_action' rpc
- To avoid retry to all followers, added new message subtype "Failed Retry"

Peek 17-07-2019 18-00

** Demo data in demo res.partner

Cc @Tecnativa


mail_tracking/models/__init__.py
```diff
@@ -1,9 +1,10 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from . import ir_mail_server
+from . import mail_thread
```
yajo: Why the reorder?

mail_tracking/models/mail_composer.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2019 Alexandre Díaz
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields, api
+
+
+class MailComposer(models.TransientModel):
+    _inherit = 'mail.compose.message'
+
+    no_followers = fields.Boolean(string="Hide follower message",
```
yajo:
```suggestion
    hide_followers = fields.Boolean(string="Hide follower message",
```

mail_tracking/models/mail_composer.py
```diff
@@ -0,0 +1,20 @@
...
+                                  default=False)
+
+    @api.multi
+    def send_mail(self, auto_commit=False):
+        res = super().send_mail(auto_commit=auto_commit)
+        message = self.env['mail.message'].browse(
+            self._context.get('message_id'))
+        if message:
```
yajo: Contrary to what it might seem, `.browse()` doesn't actually call the DB backend. It does that lazily. You need to force it to check the message exists:

```suggestion
        if message.exists():
```

mail_tracking/models/mail_message.py
```diff
@@ -12,6 +12,16 @@ class MailMessage(models.Model):
     # Recipients
     email_cc = fields.Char("Cc", help='Additional recipients that receive a '
                                       '"Carbon Copy" of the e-mail')
+    mail_tracking_ids = fields.One2many(
+        comodel_name='mail.tracking.email',
+        inverse_name='mail_message_id',
+        string="Mail Trackings Associated with this message",
```
```suggestion
        string="Mail trackings",
        help="Mail tracking records associated with this message",
```

mail_tracking/models/mail_message.py
```diff
@@ -12,6 +12,16 @@ class MailMessage(models.Model):
...
+    )
+    track_needs_action = fields.Boolean(
+        string="The message tracking will be considered"
```
```suggestion
        help="The message tracking will be considered"
```

mail_tracking/models/mail_message.py
```diff
@@ -12,6 +12,16 @@ class MailMessage(models.Model):
     # Recipients
     email_cc = fields.Char("Cc", help='Additional recipients that receive a '
                                       '"Carbon Copy" of the e-mail')
+    mail_tracking_ids = fields.One2many(
+        comodel_name='mail.tracking.email',
+        inverse_name='mail_message_id',
+        string="Mail Trackings Associated with this message",
+    )
+    track_needs_action = fields.Boolean(
```
```suggestion
    mail_tracking_needs_action = fields.Boolean(
```

mail_tracking/models/mail_message.py
```diff
@@ -95,6 +105,18 @@ def _get_email_cc(self):
...
         return res
 
+    @api.multi
+    def _get_failed_message(self):
+        res = {}
+        for message in self:
+            res.update({
+                message.id: message.track_needs_action
+                and any(message.mail_tracking_ids.filtered(
+                    lambda x: x.state in ['error', 'rejected', 'spam',
```
yajo: Appart from what's commented above, this will make it faster:

```suggestion
                    lambda x: x.state in {'error', 'rejected', 'spam',
```

mail_tracking/models/mail_message.py
```diff
@@ -95,6 +105,18 @@ def _get_email_cc(self):
...
+                                          'bounced', 'soft-bounced']))
```
```suggestion
                                          'bounced', 'soft-bounced'}))
```

mail_tracking/models/mail_message.py
```diff
@@ -95,6 +105,18 @@ def _get_email_cc(self):
             })
         return res
 
+    @api.multi
+    def _get_failed_message(self):
+        res = {}
+        for message in self:
+            res.update({
+                message.id: message.track_needs_action
+                and any(message.mail_tracking_ids.filtered(
```
yajo: I think this is not properly done.

Correct me if I'm wrong, but the `track_needs_action` by itself already indicates that the state is not good, right? In such case, the next check wouldn't be needed.

mail_tracking/models/mail_message.py
```diff
@@ -103,10 +125,66 @@ def _message_read_dict_postprocess(self, messages, message_tree):
...
+                })
         return res
+
+    @api.model
+    def _prepare_dict_failed_message(self, message):
+        failed_trackings = message.mail_tracking_ids.filtered(
+            lambda x: x.state in ['error', 'rejected', 'spam', 'bounced',
+                                  'soft-bounced'])
+        failed_partners = failed_trackings.mapped('partner_id')
+        failed_recipients = [[partner.id, partner.name]
```
yajo: This is the standard way to do it:

```suggestion
        failed_recipients = failed_partners.name_get()
```

mail_tracking/models/mail_message.py
```diff
@@ -103,10 +125,66 @@ def _message_read_dict_postprocess(self, messages, message_tree):
...
+                             for partner in failed_partners]
```
```suggestion
```

mail_tracking/models/mail_message.py
```diff
@@ -103,10 +125,66 @@ def _message_read_dict_postprocess(self, messages, message_tree):
...
+                    'partner_trackings': partner_trackings[mail_message_id],
+                    'email_cc': email_cc[mail_message_id],
+                    'failed_message': failed_message[mail_message_id],
+                })
         return res
+
+    @api.model
+    def _prepare_dict_failed_message(self, message):
+        failed_trackings = message.mail_tracking_ids.filtered(
+            lambda x: x.state in ['error', 'rejected', 'spam', 'bounced',
```
yajo: I see you using this pattern often. I think it would make sense a `self._failed_states()` method.

Also, remember to use `set`s when using `in` to boost performance:

```suggestion
            lambda x: x.state in {'error', 'rejected', 'spam', 'bounced',
```

mail_tracking/models/mail_message.py
```diff
@@ -103,10 +125,66 @@ def _message_read_dict_postprocess(self, messages, message_tree):
...
+                                  'soft-bounced'])
```
```suggestion
                                  'soft-bounced'})
```

mail_tracking/models/mail_message.py
```diff
@@ -103,10 +125,66 @@ def _message_read_dict_postprocess(self, messages, message_tree):
...
+        failed_partners = failed_trackings.mapped('partner_id')
+        failed_recipients = [[partner.id, partner.name]
+                             for partner in failed_partners]
+        return {
+            'id': message.id,
+            'date': message.date,
+            'author_id': [message.author_id.id,
```
```suggestion
            'author_id': message.author_id.name_get(),
```

mail_tracking/models/mail_message.py
```diff
@@ -103,10 +125,66 @@ def _message_read_dict_postprocess(self, messages, message_tree):
...
+                          message.author_id.display_name],
```
```suggestion
```

mail_tracking/models/mail_message.py
```diff
@@ -103,10 +125,66 @@ def _message_read_dict_postprocess(self, messages, message_tree):
...
+            'body': message.body,
+            'subtype_id': self.env.ref(
+                'mail_tracking.mail_message_subtype_failed_retry').id,
```
yajo: I guess we don't want this to raise an exception if somebody deletes the record by accident, isn't it?

```suggestion
                'mail_tracking.mail_message_subtype_failed_retry', False).id,
```

mail_tracking/models/mail_message.py
```diff
@@ -103,10 +125,66 @@ def _message_read_dict_postprocess(self, messages, message_tree):
...
+    def _get_failed_message_domain(self):
+        return [
+            ('mail_tracking_ids.state', 'in',
+                ['error', 'rejected', 'spam', 'bounced', 'soft-bounced']),
+            ('track_needs_action', '=', True)
+        ]
+
+    @api.model
+    def get_failed_count(self):
+        return self.env[self._name].search_count(
```
yajo: :sweat_smile: Isn't this the same?

```suggestion
        return self.search_count(
```

mail_tracking/models/mail_message.py
```diff
@@ -103,10 +125,66 @@ def _message_read_dict_postprocess(self, messages, message_tree):
...
+            self._get_failed_message_domain())
+
+    @api.model
+    def message_fetch(self, domain, limit=20):
+        if self.env.context.get('filter_failed_message'):
+            domain = self._get_failed_message_domain()
```
yajo, line 197: Shouldn't it be like this?

```suggestion
            domain = self._get_failed_message_domain() + domain
```

mail_tracking/models/mail_thread.py
```diff
@@ -40,3 +48,44 @@ def message_get_suggested_recipients(self):
...
+        elif view_type == 'form':
+            # Create field
+            field_failed_messages = etree.Element('field', {
+                'name': 'failed_message_ids',
+                'widget': 'mail_failed_message',
+            })
+            # Modify view to add new filter element
+            nodes = doc.xpath(
+                "//field[@name='message_ids' and @widget='mail_thread']")
+            if nodes is not None and len(nodes) > 0:
```
yajo: I guess this is the same:

```suggestion
            if nodes:
```

mail_tracking/models/mail_thread.py
```diff
@@ -40,3 +48,44 @@ def message_get_suggested_recipients(self):
...
+                    ])
+                })
+            # Modify view to add new filter element
+            nodes = doc.xpath("//search")
+            if nodes is not None and len(nodes) > 0:
+                nodes[0].append(etree.Element('separator'))
+                nodes[0].append(new_filter)
+        elif view_type == 'form':
+            # Create field
+            field_failed_messages = etree.Element('field', {
```
yajo: This sentence could go inside the `if` below to skip it if it's not gonna be appended anywhere.

mail_tracking/models/mail_thread.py
```diff
@@ -40,3 +48,44 @@ def message_get_suggested_recipients(self):
...
+                'filter', {
+                    'string': _('Failed sent messages'),
+                    'name': "failed_message_ids",
+                    'domain': str([
+                        ('failed_message_ids.track_needs_action', '=', True)
+                    ])
+                })
+            # Modify view to add new filter element
+            nodes = doc.xpath("//search")
+            if nodes is not None and len(nodes) > 0:
```
yajo: This should be the same:

```suggestion
            if nodes:
```

mail_tracking/models/mail_thread.py
```diff
@@ -40,3 +48,44 @@ def message_get_suggested_recipients(self):
...
+        """
+        res = super().fields_view_get(
+            view_id=view_id, view_type=view_type, toolbar=toolbar,
+            submenu=submenu)
+        if view_type != 'search' and view_type != 'form':
+            return res
+        doc = etree.XML(res['arch'])
+        if view_type == 'search':
+            # Create filter element
+            new_filter = etree.Element(
```
yajo: Move the sentence inside the next `if` to avoid creating it without need.

mail_tracking/models/mail_tracking_email.py
```diff
@@ -92,6 +92,15 @@ class MailTrackingEmail(models.Model):
         string="Tracking events", comodel_name='mail.tracking.event',
         inverse_name='tracking_email_id', readonly=True)
 
+    @api.multi
+    def write(self, vals):
+        if 'state' in vals and vals['state'] in \
+                ['error', 'rejected', 'spam', 'bounced', 'soft-bounced']:
```
yajo: As said above, use a method if possible. Anyways:

```suggestion
                {'error', 'rejected', 'spam', 'bounced', 'soft-bounced'}:
```

mail_tracking/models/mail_tracking_email.py
```diff
@@ -92,6 +92,15 @@ class MailTrackingEmail(models.Model):
...
+            for tracking_mail in self:
```
yajo:
```suggestion
            self.mapped("mail_message_id").write({"track_needs_action": True})
```

mail_tracking/static/src/css/failed_message.less
```diff
@@ -0,0 +1,106 @@
+.o_mail_failed_message {
+    &.o_field_widget {
+        display: block;
+    }
+
+    .o_thread_date_separator
+    {
+        margin-top: 15px;
```
yajo: Please use `rem` units wherever possible, to ease migration to v12. I'm not saying it's wrong in this line, just take a look at the file and change those you consider that fit better with that unit please.

mail_tracking/static/src/css/mail_tracking.less
```diff
@@ -0,0 +1,24 @@
+/* Copyright 2016 Antonio Espinosa - <antonio.espinosa@tecnativa.com>
```
yajo, line 1: Add yourself too.

mail_tracking/static/src/css/mail_tracking.less
```diff
@@ -0,0 +1,24 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). */
```
yajo:
```suggestion
   License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). */
```

mail_tracking/static/src/css/mail_tracking.less
```diff
@@ -0,0 +1,24 @@
...
+
+.mail_tracking {
+    span {
+        color: #909090;
```
yajo: No variables available for this?

mail_tracking/static/src/css/mail_tracking.less
```diff
@@ -0,0 +1,24 @@
...
+
+        &.mail_tracking_opened {
+            color: #a34a8b;
```
yajo: Same about using variables if possible.

mail_tracking/static/src/js/failed_message.js
```diff
@@ -0,0 +1,366 @@
...
+    var session = require('web.session');
+    var config = require('web.config');
+
+    var QWeb = core.qweb;
+    var _t = core._t;
+
+
+    /* DISCUSS */
+    var failed_counter = 0;
+    var is_channel_failed_outdated = false;
```
yajo, line 24: Based on the usage I see here, you'll possibly better use a deferred.

```suggestion
    var is_channel_failed_outdated = $.Deferred();
```

mail_tracking/static/src/js/failed_message.js
```diff
@@ -0,0 +1,366 @@
...
+    ChatAction.include({
+        init: function () {
+            this._super.apply(this, arguments);
+            // HACK: Custom event to update messsages
+            core.bus.on('force_update_message', this, function (data) {
+                is_channel_failed_outdated = true;
```
yajo, line 30:
```suggestion
                is_channel_failed_outdated.resolve();
```

mail_tracking/static/src/js/failed_message.js
```diff
@@ -0,0 +1,366 @@
...
+                this._onMessageUpdated(data);
+                this.throttledUpdateChannels();
+            });
+        },
+        _renderSidebar: function (options) {
+            options.failed_counter = chat_manager.get_failed_counter();
+            return this._super(options);
```
yajo: In general terms, always use this:

```suggestion
            return this._super.apply(this, arguments);
```

The way arguments are handled in ES is so error-prone that it's the only safe way to pass them to upstream method in ES5.

mail_tracking/static/src/js/failed_message.js
```diff
@@ -0,0 +1,366 @@
...
+        },
+        _onMessageUpdated: function (message, type) {
+            var self = this;
+            var current_channel_id = this.channel.id;
+            // HACK: broked inheritance because can't override properly
```
yajo:
```suggestion
            // HACK: break inheritance because can't override properly
```

mail_tracking/static/src/js/failed_message.js
```diff
@@ -0,0 +1,366 @@
...
+                    domain: this.domain,
+                }).then(function (messages) {
+                    var options = self._getThreadRenderingOptions(messages);
+                    self.thread.remove_message_and_render(
+                        message.id, messages, options).then(function () {
+                        self._updateButtonStatus(messages.length === 0, type);
+                    });
+                });
+            } else {
+                this._super(message, type);
```
yajo:
```suggestion
                this._super.apply(this, arugments);
```

mail_tracking/static/src/js/failed_message.js
```diff
@@ -0,0 +1,366 @@
...
+            }
+        },
+        _updateChannels: function () {
+            var self = this;
+            // HACK: broked inheritance because can't override properly
```
yajo:
```suggestion
            // HACK: break inheritance because can't override properly
```

mail_tracking/static/src/js/failed_message.js
```diff
@@ -0,0 +1,366 @@
...
+            // FIXME: Because can't refresh "channel_failed" we add a flag
+            // to indicate that the data is outdated
+            var refresh_elm = this.$(
+                ".o_mail_chat_sidebar .o_mail_failed_message_refresh");
+            refresh_elm.click(function (event) {
+                event.preventDefault();
+                event.stopPropagation();
+                location.reload();
+            });
+            if (is_channel_failed_outdated) {
```
yajo, line 91: This should change now that it's a deferred.

mail_tracking/static/src/js/failed_message.js
```diff
@@ -0,0 +1,366 @@
...
+                _.each(['dm', 'public', 'private'], function (type) {
+                    var $input = self.$(
+                        '.o_mail_add_channel[data-type=' + type + '] input');
+                    self._prepareAddChannelInput($input, type);
+                });
+            } else {
+                this._super();
+            }
+
+            // FIXME: Because can't refresh "channel_failed" we add a flag
```
yajo, line 82: This can be fixed by using the deferred.

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -12,6 +12,7 @@ odoo.define('mail_tracking.partner_tracking', function (require) {
 
     var _t = core._t;
 
+
```
yajo:
```suggestion
```

Not needed.

mail_tracking/static/src/js/failed_message.js
```diff
@@ -0,0 +1,366 @@
...
+                event.stopPropagation();
+                location.reload();
+            });
+            if (is_channel_failed_outdated) {
+                refresh_elm.removeClass('hidden');
+            }
+        },
+    });
+
+    chat_manager.get_failed_counter = function () {
```
yajo, line 97: It seems to me that all of this will have to be refactored for v12, where the chat manager is a service. Maybe you could add a note to the roadmap, or just target v12 directly.

mail_tracking/static/src/js/failed_message.js
```diff
@@ -0,0 +1,366 @@
...
+        return failed_counter;
+    };
+
+    chat_manager._onMailClientAction_failed_message_super =
+        chat_manager._onMailClientAction;
+    chat_manager._onMailClientAction = function (result) {
+        failed_counter = result.failed_counter;
+        this._onMailClientAction_failed_message_super(result);
```
yajo:
```suggestion
        return this._onMailClientAction_failed_message_super(result);
```

mail_tracking/static/src/js/failed_message.js
```diff
@@ -0,0 +1,366 @@
...
+    // to '/mail/client_action' again with overrided '_onMailClientAction'
+    session.is_bound.then(function () {
+        var context = _.extend({isMobile: config.device.isMobile},
+            session.user_context);
+        return session.rpc('/mail/client_action', {context: context});
+    }).then(chat_manager._onMailClientAction.bind(chat_manager));
+
+
+    /* FAILED MESSAGES CHATTER WIDGET */
+    function time_from_now (date) {
```
yajo, line 169:
```suggestion
    // TODO: Use timeFromNow() in v12
    function time_from_now (date) {
```

yajo (changes requested):
I haven't been able to finish the review (phew, this is loooong :sweat_smile:). However, some things already have to be changed. See comments.

Besides comments, there are these general important topics:

- Lots of cumbersome JS that is not commented. I'm having a hard time on understanding the JS code. Please add comment and explain things a little bit more.
  - Many parts of it and  many hacks in it will have to be refactored in v12. Have you considering adding the features directly in v12 and let v11 as it is? I think that at this point of history, adding hundreds of hacky LOC in v11 is not worth it (although you might have reasons for it, of course!)
- No screenshot or screencast to know what are you changing and how should it look.
- No instructions to reproduce a mail failure and actually see the new UI controls.
  - If it's not possible, then it would be nice to have demo data to test this functionally.

Thanks!

mail_tracking/models/mail_message.py
Tardo: Perhaps can need an action but not related with 'failed message'

mail_tracking/models/mail_message.py
Tardo: No, this is another "hack" related with https://github.com/OCA/social/pull/411/files#diff-d84e32af67b478b77fc10bc4225d49f9R147

mail_tracking/static/src/js/failed_message.js
Tardo: 'super' doesn't have return statement, this is really necessary?

mail_tracking/static/src/js/failed_message.js
Tardo: I can't see the benefits of use deferred... The main problem is that can't modify "channels": https://github.com/odoo/odoo/blob/11.0/addons/mail/static/src/js/chat_manager.js#L385

mail_tracking/static/src/js/failed_message.js
Tardo: This problem is currently on the module https://github.com/OCA/social/blob/11.0/mail_tracking/static/src/js/mail_tracking.js#L16
I will add the note on the roadmap

Tardo:
Some changes done, updated pr description

mail_tracking/models/mail_tracking_email.py
```diff
@@ -92,13 +92,21 @@ class MailTrackingEmail(models.Model):
...
+
     @api.model
     def email_is_bounced(self, email):
         if not email:
             return False
         res = self._email_last_tracking_state(email)
-        return res and res[0].get('state', '') in ['rejected', 'error',
-                                                   'spam', 'bounced']
+        return res and res[0].get('state', '') in \
+            self.env['mail.message']._failed_states
```
chienandalu: This isn't correct for soft-bounce errors, wich are temporary (like full mailbox, recipient mail server down, etc).
We're even questioning the 'error' event as it triggers in some cases wich are not the recipients fault like smtp server errors, etc.

mail_tracking/readme/USAGE.rst
```diff
@@ -44,3 +44,8 @@ These are all available status icons:
 |opened|  **Opened**: Opened by partner
 
 |cc|  **Cc**: It's a Carbon-Copy recipient. Can't know the status so is 'Unknown'
+
+
+When the message generates and 'error' status, it will apear on discuss 'Failed'
+channel. Any view that uses 'mail_thread' widget can show the failed messages
+too.
```
chienandalu, line 51: Some screenshots would be welcome

chienandalu (changes requested):
Great work @Tardo I've tested it and is a superfeature to have :)

Some comments:

- Add spanish translations please
- Document code as much as possible, please.
- Add tests to improve coverage if possible

Tardo:
@chienandalu changes done, thanks!

mail_tracking/static/src/js/failed_message.js
yajo: Oh it's just a good practice. I.e.: when you migrate code to newer version, if the newer version does have a return, you'd probably miss it.

However, it's not required as you said.

mail_tracking/models/mail_message.py
```diff
@@ -8,10 +8,21 @@
 
 class MailMessage(models.Model):
     _inherit = "mail.message"
+    _failed_states = {'error', 'rejected', 'spam', 'bounced', 'soft-bounced'}
```
yajo: Since `mail_tracking` is a base abstract module with several concrete implementations, I think this is better as a method, so it's easier to expand if needed.

mail_tracking/models/mail_message.py
```diff
@@ -8,10 +8,21 @@
...
     email_cc = fields.Char("Cc", help='Additional recipients that receive a '
                                       '"Carbon Copy" of the e-mail')
+    mail_tracking_ids = fields.One2many(
+        comodel_name='mail.tracking.email',
+        inverse_name='mail_message_id',
+        string="Mail Trackings",
+    )
+    mail_tracking_needs_action = fields.Boolean(
+        help="The message tracking will be considered"
+             " for filter tracking issues",
```
```suggestion
             " to filter tracking issues",
```

mail_tracking/models/mail_message.py
yajo: OK, then just a little change:

```suggestion
                and bool(message.mail_tracking_ids.filtered(
```

mail_tracking/models/mail_thread.py
```diff
@@ -1,17 +1,26 @@
...
+        'mail.message', 'res_id', string='Failed Messages',
+        domain=lambda self:
+            [('model', '=', self._name)]
+            + self.env['mail.message']._get_failed_message_domain(),
+        auto_join=True)
+
     @api.multi
     @api.returns('self', lambda value: value.id)
     def message_post(self, *args, **kwargs):
+        """ Store email_cc in mail.message """
```
yajo:
```suggestion
        """Store email_cc in mail.message."""
```

This weird spacing is a tendency I see and I don't quite understand... :thinking: 

You can review [PEP 257 on the single-line docstrings section](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings). It's not there in any example.

The same for all single-line docstrings here and there.

mail_tracking/models/mail_thread.py
```diff
@@ -40,3 +51,48 @@ def message_get_suggested_recipients(self):
                         record._message_add_suggested_recipient(
                             res, partner=partner, reason=_('Cc'))
         return res
+
+    @api.model
+    def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
+                        submenu=False):
+        """Add a filter to any model with mail.thread that will show up records
```
yajo: ... and now that we speak about docstrings, in the case of [multiline docstrings](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings):

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line.

An idea of what would be a proper multiline docsting here:

```python
"""Add filters for failed messages.

These filters will show up on any form or search views of any
model inheriting from ``mail.thread``.
"""
```

I guess you didn't know there's a PEP on docstrings and you're gonna like to have a standard on that. :blush:

mail_tracking/static/src/js/failed_message.js
yajo: Hmm OK, maybe I misunderstood something.

We can handle that in the migration if needed, I think in v12 this will be easier to write.

yajo (changes requested):
I didn't have time to finish the review, this is what I have for now...  sorry!

chienandalu:
Not the main purpose of this PR, but it would be great to have pivot for tracking events and mail trackings: https://github.com/OCA/social/blob/85690b0ae87bd7f6bd969016a3a0e5dfeebc4da4/mail_tracking/views/mail_tracking_event_view.xml#L115
https://github.com/OCA/social/blob/85690b0ae87bd7f6bd969016a3a0e5dfeebc4da4/mail_tracking/views/mail_tracking_email_view.xml#L111

pedrobaeza:
You can put that in the ROADMAP, but indeed out of scope on this PR.

yajo:
One doubt I have: when using this new section of failed messages, messages exist only there, or they exist also in the normal chatter history?

I'm thinking that the chatter position of a message is important for the user to know which of those messages failed, and to know if the conversation continued since then or not, and also know what he was answering to, etc.

Tardo:
You can see failed messages in three sites:
- New section on discuss
- New section in the thread (like activities)
- Normal message with failed status

rafaelbn:
Please @Tardo "This branch has conflicts that must be resolved"

rafaelbn:
Please @Yajo @chienandalu @Tardo review together thanks :+1:

chienandalu:
Rebuilding runbot to test functionally . Is there something left @Tardo ?

Could we achieve the later coverage?: 92.59% of diff hit (target 95.39%)

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -12,13 +12,29 @@ odoo.define('mail_tracking.partner_tracking', function (require) {
...
     // Chat_manager is a simple dictionary, not an OdooClass
     chat_manager._make_message_super = chat_manager.make_message;
     chat_manager.make_message = function (data) {
         var msg = this._make_message_super(data);
         msg.partner_trackings = data.partner_trackings || [];
+<<<<<<< HEAD
```
chienandalu: Take care with these git conflicts

mail_tracking/static/src/xml/failed_message.xml
```diff
@@ -0,0 +1,69 @@
...
+            </span>
+        </div>
+        <div id="o_chatter_failed_message" class="collapse">
+            <t t-foreach="failed_messages" t-as="message">
+                <t t-call="mail_tracking.failed_message" />
+            </t>
+        </div>
+    </t>
+
+    <!--t t-extend="mail.ChatThread.Message">
```
chienandalu: Remove if not needed

chienandalu (changes requested):
Put yourself in CONTRIBUTORS.rst as well :)

mail_tracking/models/mail_message.py
```diff
@@ -110,16 +124,82 @@ def tracking_status(self):
...
+    @api.model
+    def _prepare_dict_failed_message(self, message):
+        failed_trackings = message.mail_tracking_ids.filtered(
+            lambda x: x.state in self.get_failed_states())
+        failed_partners = failed_trackings.mapped('partner_id')
+        failed_recipients = failed_partners.name_get()
+        return {
+            'id': message.id,
+            'date': message.date,
+            'author_id': message.author_id.name_get()[0],
```
yajo, line 164:
```suggestion
            'author_id': message.author_id.display_name,
```

It's the same, but more readable. :wink:

mail_tracking/models/mail_message.py
```diff
@@ -110,16 +124,82 @@ def tracking_status(self):
...
+            'body': message.body,
+            'subtype_id': self.env.ref(
+                'mail_tracking.mail_message_subtype_failed_retry', False).id,
```
yajo: This is a bit inconsistent.

If you add `False`, it means that you expect it to maybe not exist, but if you add `.id` it means that you expect it to always exist.

Choose one.

mail_tracking/models/mail_message.py
```diff
@@ -110,16 +124,82 @@ def tracking_status(self):
...
+        for msg in self:
+            failed_messages.append(self._prepare_dict_failed_message(msg))
+        return failed_messages
+
+    @api.multi
+    def toggle_tracking_status(self):
+        """Toggle message tracking action needed to ignore them in the tracking
+           issues filter"""
+        # a user should always be able to star a message he can read
+        self.check_access_rule('read')
```
yajo: Isn't this implicit? :thinking:

yajo (approved):
Code OK, some little remarks.

mail_tracking/models/mail_message.py
pedrobaeza: But we are not showing full display name on other parts...

mail_tracking/models/mail_message.py
Tardo: Copy/paste issue... deleted. Thanks

Tardo:
Changes done

rafaelbn (approved):
:+1:

yajo (approved):

mail_tracking/README.rst
```diff
@@ -58,25 +58,25 @@ status icon will appear just right to name of notified partner.
 
 These are all available status icons:
 
-.. |sent| image:: static/src/img/sent.png
+.. |sent| image:: ../static/src/img/sent.png
```
pedrobaeza: Why relative images on main README? This should be on fragments, but not on main README, and the README generator will convert them to absolute. Doing what you have done, will get to work on certain places, but not on others.

mail_tracking/README.rst
```diff
@@ -93,6 +93,27 @@ These are all available status icons:
...
+  .. image:: https://raw.githubusercontent.com/OCA/social/11.0/mail_tracking/static/img/failed_message_discuss.png
+
+* Chatter
+
+  .. image:: https://raw.githubusercontent.com/OCA/social/11.0/mail_tracking/static/img/failed_message_widget.png
+
+Known issues / Roadmap
+======================
+
+* Handle message updates on discuss 'channel_failed' instead of show the
```
pedrobaeza: s/show/showing

mail_tracking/data/mail_message_subtype_data.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="1">
+        <record id="mail_message_subtype_failed_retry" model="mail.message.subtype">
```
pedrobaeza: Why do we need a new subtype?

mail_tracking/models/mail_composer.py
```diff
@@ -0,0 +1,22 @@
...
+
+class MailComposer(models.TransientModel):
+    _inherit = 'mail.compose.message'
+
+    hide_followers = fields.Boolean(string="Hide follower message",
+                                    default=False)
+
+    @api.multi
+    def send_mail(self, auto_commit=False):
+        """ This method mark as reviewed the message when use the 'Retry'
```
pedrobaeza: s/mark/marks
s/use/using

mail_tracking/models/mail_message.py
pedrobaeza: Please confirm this

pedrobaeza (changes requested):
Please review my comments. It seems you have an outdated README generator

mail_tracking/data/mail_message_subtype_data.xml
Tardo: To avoid retry to all followers

pedrobaeza: But can we do it other way? I don't like too much to need a new subtype. And what happen on multiple failed messages in same thread? Are they notified again although they are not failed this time?

mail_tracking/README.rst
Tardo: This regex https://github.com/OCA/maintainer-tools/blob/master/tools/gen_addon_readme.py#L148 doesn't support "image substitutions": https://sublime-and-sphinx-guide.readthedocs.io/en/latest/images.html#use-image-substitutions

Using a regex like this fix the issue:
```
r'\s*\.\. (?:\|.+\|\s)?(figure|image)::\s+(?P<path>.*?)\s*$'
```

pedrobaeza: This is working, @Tardo. What the regexp do is to replace by the absolute URL in raw.github.com, for being accessible no matter where the file is hosted. And that happens when converting fragments to README.rst.

Tardo: You can see it, the actual regex doesn't recognize the image line when use image substitution...

Actual behavior: https://regex101.com/r/Txgcx0/3
With proposed regex: https://regex101.com/r/Txgcx0/4

Parhaps i'm doing something wrong... but after change regex i can generate valid readme's:
```
.. |sent| image:: https://raw.githubusercontent.com/OCA/social/11.0/mail_tracking/static/src/img/sent.png
   :width: 10px
```

rafaelbn:
@Tardo please review comments from @pedrobaeza and @chienandalu

pedrobaeza:
@Tardo please retake this and finish it

mail_tracking/models/mail_message.py
Tardo: Yes, on other parts the module shows the same "name"...

mail_tracking/data/mail_message_subtype_data.xml
pedrobaeza: What about this?

Tardo: Removed the message subtype... Added another hack to the collection: https://github.com/OCA/social/pull/411/files?file-filters%5B%5D=.py#diff-4e1cf268ea32e1266c1e39d18db0d429R207

Tardo:
Changes applied

pedrobaeza (approved):
Let's go for it! Please forward-port it to v12 reusing what you can of new core retry mechanism

rafaelbn:
@Tardo @pedrobaeza is this already in v12? Thanks

pedrobaeza:
Not yet. It's working on it.

PR #413, opened 2019-07-16
[12.0][FIX] mail_debrand: debrand regular notifications.

LoisRForgeFlow:
without this fix, any regular notification (added as follower,
message notifications...) were not parsed at all and as a
consequence also not debranded.

### DIFF
--- a/mail_debrand/__manifest__.py
+++ b/mail_debrand/__manifest__.py
@@ -1,3 +1,7 @@
+# Copyright 2016 Tecnativa - Jairo Llopis
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# Copyright 2019 Eficent Business and IT Consulting Services S.L.
+#   - Lois Rilo <lois.rilo@eficent.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
@@ -6,7 +10,7 @@
     "version": "12.0.2.0.0",
     "category": "Social Network",
     "website": "https://github.com/OCA/social/",
-    "author": "Odoo Community Association (OCA)",
+    "author": "Tecnativa, Eficent, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "installable": True,
     "depends": [
--- a/mail_debrand/readme/CONTRIBUTORS.rst
+++ b/mail_debrand/readme/CONTRIBUTORS.rst
@@ -1,1 +1,3 @@
+* Pedro M. Baeza <pedro.baeza@tecnativa.com>
+* Lois Rilo <lois.rilo@eficent.com>
 * Graeme Gellatly <graeme@o4sb.com>
--- a/mail_debrand/README.rst
+++ b/mail_debrand/README.rst
@@ -60,9 +60,17 @@ Do not contact contributors directly about support or help with technical issues
 Credits
 =======
 
+Authors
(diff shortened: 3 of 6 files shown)


mail_debrand/readme/CONTRIBUTORS.rst
```diff
@@ -1 +1,7 @@
+* Rafael Blasco <rafabn@tecnativa.com>
+* Jairo Llopis <jairo.llopis@tecnativa.com>
+* Jordi Ballester Alomar <jordi.ballester@eficent.com>
+* Darshan Patel <darshan.patel.serpentcs@gmail.com>
```
pedrobaeza: As the module was totally redone, contributors were reset, and this one hasn't contributed in current one.

mail_debrand/readme/CONTRIBUTORS.rst
```diff
@@ -1 +1,7 @@
+* Rafael Blasco <rafabn@tecnativa.com>
```
pedrobaeza: As the module was totally redone, contributors were reset, and this one (and the following) hasn't contributed in current one. Any way, the email address is incorrect.

pedrobaeza (changes requested):
Isn't enough then with the `mail.thread` method (and remove `mail.template` one)?

LoisRForgeFlow:
@pedrobaeza I'm not sure, as it was added there in the first place, I just wanted to respect that because maybe in some other process it actually goes thought `render_post_process` method, maybe in mass_mailing?? maybe you have more deep knowledge on this... :wink:

mail_debrand/readme/CONTRIBUTORS.rst
LoisRForgeFlow: Fixed!

mail_debrand/__manifest__.py
```diff
@@ -6,7 +10,7 @@
     "version": "12.0.2.0.0",
     "category": "Social Network",
     "website": "https://github.com/OCA/social/",
-    "author": "Odoo Community Association (OCA)",
+    "author": "Tecnativa, Eficent, Odoo Community Association (OCA)",
```
pedrobaeza, line 13: @gdgellatly deserves here his authorship as well. Which company, Graeme?

pedrobaeza:
@lreficent just remove it and see if tests pass.

LoisRForgeFlow:
> @pedrobaeza I'm not sure, as it was added there in the first place, I just wanted to respect that because maybe in some other process it actually goes thought render_post_process method, maybe in mass_mailing?? maybe you have more deep knowledge on this... wink

cc @gdgellatly maybe you know better?

LoisRForgeFlow:
@pedrobaeza test will pass for sure, because test is not sending any email, is just parsing a qweb template.

EDIT: I mean, the test is only testing the debranding method explicitly, outside of any odoo functional process.

yajo (approved):

pedrobaeza:
Waiting @gdgellatly's opinion and company name for putting in authorship

yajo:
I agree, although IMHO, as long as we're not *removing* an author, there's no reason to block this PR on that base. @gdgellatly can always open a new PR and add himself if he considers it important.

pedrobaeza (approved):

pedrobaeza:
OK, let's move

PR #416, opened 2019-07-22
[11.0][FIX] mail_tracking: Repeated suggested cc emails

Tardo:
This PR fixes repeated suggested cc emails introduced in https://github.com/OCA/social/pull/407

### DIFF
--- a/mail_tracking/models/mail_thread.py
+++ b/mail_tracking/models/mail_thread.py
@@ -24,19 +24,21 @@ class MailThread(models.AbstractModel):
     def message_get_suggested_recipients(self):
         res = super().message_get_suggested_recipients()
         ResPartnerObj = self.env['res.partner']
+        email_cc_formated_list = []
         for record in self:
-            messages = record.message_ids.filtered('email_cc')
-            for msg in messages:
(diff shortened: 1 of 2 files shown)


mail_tracking/models/mail_thread.py
```diff
@@ -33,7 +33,8 @@ def message_get_suggested_recipients(self):
                     partner_id = record.message_partner_info_from_emails(
                         [email_parts[1]])[0].get('partner_id')
                     if not partner_id:
-                        res[record.id].append((False, cc, _('Cc')))
+                        record._message_add_suggested_recipient(
```
pedrobaeza: How performant is this? Isn't better to have a set?

yajo (approved):

mail_tracking/models/mail_thread.py
Tardo: Done, thanks!

pedrobaeza (approved):

pedrobaeza:
Please forward-port it to v12.

Tardo:
Done: https://github.com/OCA/social/pull/418

PR #418, opened 2019-07-22
[12.0][MIG] mail_tracking: Fix repeated cc

Tardo:
Cc @tecnativa


pedrobaeza (approved):
Fast-tracking as it's the same as v11

PR #419, opened 2019-07-23
[IMP] mail_tracking: Adapt test to set (not order preserved)

Tardo:
This PR fix the test

Cc @tecnativa

### DIFF
--- a/mail_tracking/tests/test_mail_tracking.py
+++ b/mail_tracking/tests/test_mail_tracking.py
@@ -137,8 +137,10 @@ class TestMailTracking(TransactionCase):
         self.assertFalse(email_cc[1])
         # suggested recipients
         recipients = self.recipient.message_get_suggested_recipients()
-        self.assertEqual(recipients[self.recipient.id][0][1],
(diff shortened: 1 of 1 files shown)


mail_tracking/tests/test_mail_tracking.py
```diff
@@ -137,8 +137,10 @@ def test_email_cc(self):
         self.assertFalse(email_cc[1])
         # suggested recipients
         recipients = self.recipient.message_get_suggested_recipients()
-        self.assertEqual(recipients[self.recipient.id][0][1],
-                         'unnamed@test.com')
+        suggested_mails = {
+            email[1] for email in recipients[self.recipient.id]
+        }
+        self.assertTrue('unnamed@test.com' in suggested_mails)
```
pedrobaeza: use `assertIn` instead

Tardo: Done, thanks!

pedrobaeza (approved):
Fast-tracking for fixing Travis (it gets a random error)

PR #420, opened 2019-07-23
[12.0][ADD] - new module: mail template substitute

sbejaoui:
This module allows you to create substitution rules for mail templates.
A typical use case is to replace a standard template by alternative templates
when some conditions are met. For instance, it allows to configure alternate
templates for different companies.

Todo:

- [x] Add unit tests


sbejaoui:
ping @sbidoul , @gva-acsone

gva-acsone:
Very interesting module !

Two small observations:

1. IMO, the list of available substitute mail templates should be limited to the templates that relate to the same model. It should be impossible to select an 'invoice' mail template as substitute for a 'sales order' mail template.

2. We need to find a solution to combine the mail template substitute module and the report substitute module. Indeed, there is a use case that is still problematic --> When a mail template substitute is applicable, Odoo always takes the report related to that mail template substitute and adds it as an attachment, even though this report has substitutes itself that should be applied. 

Odoo should start by looking at the mail templates substitutes and choose the right mail template to apply. Afterwards, it should look at the report related to the mail template and look at the report substitutes to choose the right report to add as an attachment.

sbejaoui:
Thank you @gva-acsone for your feedback.

> 1. IMO, the list of available substitute mail templates should be limited to the templates that relate to the same model. It should be impossible to select an 'invoice' mail template as substitute for a 'sales order' mail template.

Fixed

> 2. We need to find a solution to combine the mail template substitute module and the report substitute module. Indeed, there is a use case that is still problematic --> When a mail template substitute is applicable, Odoo always takes the report related to that mail template substitute and adds it as an attachment, even though this report has substitutes itself that should be applied.
> 
> Odoo should start by looking at the mail templates substitutes and choose the right mail template to apply. Afterwards, it should look at the report related to the mail template and look at the report substitutes to choose the right report to add as an attachment.

I made the `mail_template_substitute` depends on `report_substitute` and it resolved the use case you mentioned.

gva-acsone:
Thanks you @sbejaoui for the two modifications, all good for me.

Three other observations:

1° When we manually send a new message based on a template (chatter), the report substitution is not automatically done.

2° If the original email template doesn't have a related report template, the related report of the substitution email template will be ignored.

3° If a mail template is added as an action in the list view, the user has the ability to choose whether he wants to send the email to the selected records or the default search. If he selects the default search, the substitution only applies to the selected records.

gva-acsone:
@sbejaoui : 
The first two points are good for me (functional test), thanks for the correction !

mail_template_substitute/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+{
+    "name": "Mail Template Substitute",
+    "summary": """
+        This module allows to create substitution rules for mail templates.
+        """,
+    "version": "12.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,"
+              "Odoo Community Association (OCA)",
+    "website": "https://github.com/acsone/social",
```
rousseldenis: Change this according to https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py

rousseldenis (changes requested):
Code review. Some comment.

Don't forget to change oca_dependencies before merge

RobinetDenisAcsone (approved):
Code review, LGTM

sbejaoui:
@rousseldenis , can you update your review please

rousseldenis (approved):
Code review

sbejaoui:
@OCA/core-maintainers , This PR is ready to merge

PR #421, opened 2019-07-23
[IMP] mail_tracking: Cc name & check if mail is Cc

Tardo:
This PR improve how the Cc is showed, now only display the basic name.
Now, if a recipient (that is Cc too) have a tracking record it will display checks instead of "Cc" mark.

Cc @Tecnativa

### DIFF
--- a/mail_tracking/models/mail_message.py
+++ b/mail_tracking/models/mail_message.py
@@ -45,6 +45,11 @@ class MailMessage(models.Model):
             trackings = self.env['mail.tracking.email'].sudo().search([
                 ('mail_message_id', '=', message.id),
             ])
+            # Get Cc recipients
+            email_cc_list = email_split(message.email_cc)
+            if any(email_cc_list):
+                partners |= partners.search([('email', 'in', email_cc_list)])
+            email_cc_list = set(email_cc_list)
             # Search all trackings for this message
             for tracking in trackings:
                 status = self._partner_tracking_status_get(tracking)
@@ -52,8 +57,9 @@ class MailMessage(models.Model):
                     tracking.partner_id.name or tracking.recipient)
                 partner_trackings.append((
                     status, tracking.id, recipient, tracking.partner_id.id,
-                    tracking.partner_id.email))
+                    False))
                 if tracking.partner_id:
+                    email_cc_list.discard(tracking.partner_id.email)
                     partners_already |= tracking.partner_id
             # Search all recipients for this message
             if message.partner_ids:
@@ -64,36 +70,18 @@ class MailMessage(models.Model):
             partners -= partners_already
             for partner in partners:
                 # If there is partners not included, then status is 'unknown'
-                partner_trackings.append((
-                    'unknown', False, partner.name, partner.id, partner.email))
-            res[message.id] = partner_trackings
-        return res
-
-    @api.multi
-    def _get_email_cc(self):
-        """This method gets all Cc mails and the associated partner if exist.
-            The result is a dictionary by 'message id' with a list of tuples
-            (str:email_cc, list:[partner id, partner display_name] or False)
-        """
-        res = {}
-        ResPartnerObj = self.env['res.partner']
-        for message in self:
-            email_cc_list = email_split(message.email_cc)
-            email_cc_list_checked = []
-            if any(email_cc_list):
-                partners = ResPartnerObj.search([
-                    ('email', 'in', email_cc_list)
-                ])
-                email_cc_list = set(email_cc_list)
-                for partner in partners:
+                # Because can be an Cc recipinet
+                isCc = False
+                if partner.email in email_cc_list:
                     email_cc_list.discard(partner.email)
-                    email_cc_list_checked.append(
-                        (partner.email, [partner.id, partner.name]))
-                for email in email_cc_list:
-                    email_cc_list_checked.append((email, False))
-            res.update({
-                message.id: email_cc_list_checked
-            })
+                    isCc = True
+                partner_trackings.append((
+                    'unknown', False, partner.name, partner.id, isCc))
+            for email in email_cc_list:
+                # If there is Cc without partner
+                partner_trackings.append((
+                    'unknown', False, email, False, True))
+            res[message.id] = partner_trackings
         return res
 
     @api.model
@@ -103,11 +91,9 @@ class MailMessage(models.Model):
         mail_message_ids = {m.get('id') for m in messages if m.get('id')}
         mail_messages = self.browse(mail_message_ids)
         partner_trackings = mail_messages.tracking_status()
-        email_cc = mail_messages._get_email_cc()
         for message_dict in messages:
             mail_message_id = message_dict.get('id', False)
             if mail_message_id:
                 message_dict['partner_trackings'] = \
                     partner_trackings[mail_message_id]
-                message_dict['email_cc'] = email_cc[mail_message_id]
         return res
--- a/mail_tracking/models/mail_message.py
+++ b/mail_tracking/models/mail_message.py
@@ -51,7 +51,8 @@ class MailMessage(models.Model):
                 recipient = (
                     tracking.partner_id.name or tracking.recipient)
                 partner_trackings.append((
-                    status, tracking.id, recipient, tracking.partner_id.id))
+                    status, tracking.id, recipient, tracking.partner_id.id,
+                    tracking.partner_id.email))
                 if tracking.partner_id:
                     partners_already |= tracking.partner_id
             # Search all recipients for this message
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):
Tried on our instance

mail_tracking/static/src/xml/mail_tracking.xml
```diff
@@ -54,7 +54,7 @@
             <t t-foreach="message.partner_trackings" t-as="tracking">
                 <t t-set="isCc" t-value="false" />
                 <t t-foreach="message.email_cc" t-as="cc">
-                    <t t-if="cc[0] == tracking[4]" t-set="isCc" t-value="true" />
+                    <t t-if="!tracking[1] and cc[0] == tracking[4]" t-set="isCc" t-value="true" />
```
yajo:
```suggestion
                    <t t-if="!tracking[1] && cc[0] === tracking[4]" t-set="isCc" t-value="true" />
```

This is JS!

mail_tracking/static/src/xml/mail_tracking.xml
```diff
@@ -83,6 +83,9 @@
                     <t t-if="cc[0] == tracking[4]" t-set="needPrint" t-value="false" />
                 </t>
                 <t t-if="needPrint">
+                    <t t-if="!first_tracking">
```
yajo: Qweb already provides handy loop variables by default: https://www.odoo.com/documentation/12.0/reference/qweb.html#loops

```suggestion
                    <t t-if="!cc_first">
```

mail_tracking/static/src/xml/mail_tracking.xml
```diff
@@ -98,6 +101,7 @@
                           title="Status: unknown">
                           <t t-call="mail.tracking.status"/>
                     </span>
+                    <t t-set="first_tracking" t-value="false"/>
```
yajo: Not needed, for the same reason.

```suggestion
```

mail_tracking/models/mail_message.py
```diff
@@ -87,7 +88,7 @@ def _get_email_cc(self):
                 for partner in partners:
                     email_cc_list.discard(partner.email)
                     email_cc_list_checked.append(
-                        (partner.email, [partner.id, partner.display_name]))
+                        (partner.email, [partner.id, partner.name]))
```
yajo: Why? :thinking:

mail_tracking/models/mail_message.py
```diff
@@ -51,7 +51,8 @@ def tracking_status(self):
                 recipient = (
                     tracking.partner_id.name or tracking.recipient)
                 partner_trackings.append((
-                    status, tracking.id, recipient, tracking.partner_id.id))
+                    status, tracking.id, recipient, tracking.partner_id.id,
+                    tracking.partner_id.email))
```
yajo: I don't see where are you using this new data in the code below. Could you explain a little bit about how this patch works? :thinking: :sweat_smile:

yajo (changes requested):

mail_tracking/static/src/xml/mail_tracking.xml
Tardo: Sure! works because qweb replace it: https://github.com/odoo/odoo/blob/12.0/addons/web/static/lib/qweb/qweb2.js#L33

yajo: Wow! I was wondering why it worked... Maybe it's to avoid things such as `&amp;&amp;`... Then better leave it as it is.

mail_tracking/models/mail_message.py
Tardo: partner_trackings store email and email_cc too, if a partner_tracking without tracking record is present on email_cc it will be marked with "Cc". If an email_cc is currently present on partner_tracking it will not be printed.

Used here: https://github.com/OCA/social/blob/be273331180f017b931231049a097159ca945a6f/mail_tracking/static/src/xml/mail_tracking.xml#L57
And here: https://github.com/OCA/social/blob/be273331180f017b931231049a097159ca945a6f/mail_tracking/static/src/xml/mail_tracking.xml#L83

mail_tracking/models/mail_message.py
Tardo: Because "partner_tracking" only prints the name too: 
https://github.com/OCA/social/blob/be273331180f017b931231049a097159ca945a6f/mail_tracking/models/mail_message.py#L52
https://github.com/OCA/social/blob/be273331180f017b931231049a097159ca945a6f/mail_tracking/models/mail_message.py#L68

Tardo:
Refactored how Cc is printed. Now uses "partner_trackings" to simplify all.

mail_tracking/models/mail_message.py
```diff
@@ -63,50 +70,30 @@ def tracking_status(self):
             partners -= partners_already
             for partner in partners:
                 # If there is partners not included, then status is 'unknown'
+                # Because can be an Cc recipinet
```
```suggestion
                # Because can be an Cc recipient
```

mail_tracking/models/mail_message.py
```diff
@@ -63,50 +70,30 @@ def tracking_status(self):
...
+                if partner.email in email_cc_list:
+                    email_cc_list.discard(partner.email)
+                    isCc = True
                 partner_trackings.append((
-                    'unknown', False, partner.name, partner.id, partner.email))
+                    'unknown', False, partner.name, partner.id, isCc))
+            for email in email_cc_list:
+                # If there is Cc without partner
+                partner_trackings.append((
+                    'unknown', False, email, False, True))
```
yajo: I don't know if this is something you can actually alter, but if you can, would it make sense to use a list of `dict()`s instead? If all is positional and we have so many positions, it gets confusing, don't you think?

yajo (changes requested):
It looks much better, thanks.

Just a little typo and a question.

Tardo:
Changes done @Yajo, yeah more readable using dictionaries instead of "large" lists ;)

mail_tracking/static/src/xml/mail_tracking.xml
```diff
@@ -54,19 +54,19 @@
...
-                <t t-if="!tracking[3]">
-                    <span t-attf-class="#{tracking[4] ? 'o_mail_cc' : ''}"><t t-esc="tracking[2]"/></span>
+                <t t-else="">
+                    <span t-attf-class="#{tracking['isCc'] ? 'o_mail_cc' : ''}"><t t-esc="tracking['recipient']"/></span>
                 </t>
                 <span class="mail_tracking o_mail_action_tracking_status"
-                      t-att-data-tracking="tracking[1]"
-                      t-attf-title="Status: #{tracking[0]}">
+                      t-att-data-tracking="tracking['tracking_id']"
+                      t-attf-title="Status: #{tracking['status']}">
```
yajo: This should be translatable IMHO.

yajo (changes requested):
Thanks! Last change improved code quality **a lot**!

I now noticed one last thing.

Tardo:
@Yajo changes done.

mail_tracking/static/src/xml/mail_tracking.xml
```diff
@@ -66,7 +66,7 @@
                 </t>
                 <span class="mail_tracking o_mail_action_tracking_status"
                       t-att-data-tracking="tracking['tracking_id']"
-                      t-attf-title="Status: #{tracking['status']}">
+                      t-att-title="_.str.sprintf(_t('Status: %s'), tracking['status'])">
```
yajo: AFAIK that's not gonna work.

For a translation string to appear in the `.pot` file, a static code scan is done across all source files. In each file, something specific is searched:

- In `.py` files, it searches for calls to `_()`, 
- in `.js` files it searches for calls to `_t()` and `_lt()`, 
- and in `.xml` files it searches for certain translatable elements and attributes. For example, the `title` attribute is translatable, but any attribute that starts with `t-` is not translated.

So if I'm not wrong, if you try to produce the `.pot` file, you'll see that this string doesn't appear. That's because you're using the JS translation system in an XML file, and the scanner won't detect it.

The easiest way to fix this problem is to provide a `status_for_humans` key in the dict, produced directly from the Python side. Also think that this way of fixing the problem could produce (if it worked), i.e. in Spanish, something like: "Estado: unknown", when you want "Estado: descconocido". Translating in the server-side will let you fix this also.

Tardo:
@Yajo changes done. Thanks!

rafaelbn (approved):
Functional :+1: 

@Yajo your comments were attended

pedrobaeza:
@Yajo please give your final blessing

yajo (approved):

pedrobaeza:
/ocabot merge minor

Please forward-port it to v12

PR #422, opened 2019-07-24
[FIX] don't break when a partner has a user

hbrunn:
this fixes cases where we run into https://github.com/OCA/OCB/blob/8.0/addons/mail/mail_mail.py#L172, as the model the email applies to is a notification

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,41 +1,26 @@
 language: python
 sudo: false
-cache:
-  apt: true
-  directories:
-    - $HOME/.cache/pip
+cache: pip
 
 python:
   - "2.7"
 
 addons:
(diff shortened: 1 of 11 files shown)


rafaelbn:
Hi @hbrunn could you see travis :red_circle: please?

hbrunn:
@rafaelbn thanks for the heads up, this branch was a mess indeed. I fixed most problems, but there's some lockup in the odoo branch that I don't have time to debug right now. Not related to this change anyways.

Let's try to get this merged fast so that there's only one problem left to debug for 8.0

rafaelbn (approved):

rafaelbn:
I agree @hbrunn in merging this :+1:

PR #429, opened 2019-08-26
[ADD] mail_tracking_mailgun: manage failed state from mailgun

zaoral:
This change let Odoo process a state sent from Mailgun legacy Webhooks that seems to apply when the message is not sent because the related email has been mark as spam or have bounced before. 

For solve this we only add  two new state mapping to _mailgun_event_type_mapping_mailgun_event_type_mapping method:

* failed: Mailgun could not deliver the email to the recipient email server
* rejected: Mailgun rejected the request to send/forward the email

source in https://documentation.mailgun.com/en/latest/quickstart-events.html#events

### DIFF
--- a/mail_tracking_mailgun/models/mail_tracking_email.py
+++ b/mail_tracking_mailgun/models/mail_tracking_email.py
@@ -45,6 +45,8 @@ class MailTrackingEmail(models.Model):
             'bounced': 'hard_bounce',
             'dropped': 'reject',
             'accepted': 'sent',
+            'failed': 'error',
+            'rejected': 'error',
         }
 
     def _mailgun_event_type_verify(self, event):


mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -45,6 +45,8 @@ def _mailgun_event_type_mapping(self):
             'bounced': 'hard_bounce',
             'dropped': 'reject',
             'accepted': 'sent',
+            'failed': 'error',
+            'rejected': 'error',
```
chienandalu, line 49: Shouldn't this be `hard_bounce` or `reject`?

zaoral: Hi @chienandalu I set error in this case for both cases because in the documentation of Mailgun events they are not really specific, we know by testing that for early spam complaint it return an event as failed, but now sure if there is another case when it also return failed state but could be for a difference reason. To avoid inform Odoo with a bad assumption I preferred to inform Odoo that an error occurs instead. Same case for rejected.

IMHO, I think that failed does not match with hard_bounce. because the event really do not has been hard bounce, simply is Mailgun informing that will not sent the message for whatever reason. For the case of rejected is occurs the same, Mailgun don't bother to try to sent the message and then dropped, only does not sent the message for x reason. I think we should not assume that reason why the message was not sent

mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -45,6 +45,8 @@ def _mailgun_event_type_mapping(self):
             'bounced': 'hard_bounce',
             'dropped': 'reject',
```
yajo, line 46: I see in https://documentation.mailgun.com/en/latest/quickstart-events.html#events there's no `dropped` nor `bounced` statuses, but yes there's a `stored` state that we don't have here. Should we fix that now BTW?

chienandalu: They come from the hooks: https://documentation.mailgun.com/en/latest/api-webhooks-deprecated.html?highlight=bounced#bounces-parameters

yajo: Ah ok. And about `stored`?

chienandalu: > Mailgun has stored an incoming message

I'd say it doesn't apply here

yajo (approved):

chienandalu (approved):

PR #431, opened 2019-08-26
[FIX] mail_tracking: Fix message_post inherited method arguments

sergio-teruel:
cc @Tecnativa


pedrobaeza (approved):
Fast-tracking this as definitive solution

PR #432, opened 2019-09-01
[12.0] [ADD] mail_recipients_uncheck

nilshamerlinck:
The purpose of this module is to make sure that the recipients suggested by Odoo are not checked by default.


pedrobaeza:
This is the same as https://github.com/OCA/social/tree/10.0/mail_optional_follower_notification

nilshamerlinck:
Hi guys :-)

`mail_optional_follower_notification` "adds the possibility to choose if you want to automatically notify followers on `mail.compose.message`".

This  module has a [different purpose, described here in its rendered documentation](https://htmlpreview.github.io/?https://raw.githubusercontent.com/OCA/social/6ffdc187c2c978810e27c746010a07be9bda6cde/mail_recipients_uncheck/static/description/index.html).

pedrobaeza:
OK then, but please avoid plurals in module name as stated in guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules

nilshamerlinck:
Hi @pedrobaeza, module has been renamed to respect guidelines.

pedrobaeza:
OK, thanks. I can't say more about the module itself (I only check general guidelines across OCA), so waiting for more reviewers.

sergiocorato (approved):
LGTM but `development_status: Mature` perhaps is not possibile without tests

sergiocorato (changes requested):
This module has unintended effects when used with https://github.com/OCA/social/tree/12.0/mail_full_expand (no options available, followers are not proposed by default and mail are not sent) so unusable with a well-known module.

PR #437, opened 2019-09-23
FIX multiple body fields that were conflicting in mass mailing when using Sendgrid

ecino:
- Instead of copying body of template to mass mailing, directly use
  the template text for mass mailings when using Sendgrid.
- If needed, the template must be edited instead of the mass mailing.
- This simplifies the code complexity


StefanRijnhart (approved):
Does reduce code complexitiy indeed. And given that you are the original author you probably know best that no side effects will occur.

ecino:
It only changes how the user edits the content of the mass mailing. He cannot change it directly from the mass mailing but must do it from the template used. Can we merge it?

StefanRijnhart:
Sure!

/ocabot merge

PR #445, opened 2019-10-16
13.0 mig mail activity done

emagdalenaC2i:


emagdalenaC2i:
@ageficent @jbeficent could you review this PR?

AdriaGForgeFlow (changes requested):
When I try to Mark as done one activity Odoo throws the following error:
image

It seems that Odoo's original function action_feedback now has two arguments:
`def action_feedback(self, feedback=False, attachment_ids=None):`

emagdalenaC2i:
@ageficent fixed

AdriaGForgeFlow (approved):
Functional test and code review :+1:

mail_activity_done/models/res_users.py
```diff
@@ -0,0 +1,58 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+from odoo import api, fields, models, modules
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    @api.model
+    def systray_get_activities(self):
+        # Here we totally override the method. Not very nice, but
```
jarroyomorales, line 11: Maybe we could add a context here to use the old method?

if self.env.context.get('use_old_systray_get_activities', False):
    return super().systray_get_activities

emagdalenaC2i:
@pegonzalezspesol and @jbeficent Could you review this PR?

JordiBForgeFlow:
@ageficent @hveficent @joansis can you review?

pegonzalezspesol:
@emagdalenaC2i , as soon as I click on "Done" on an activity I get next issue

image

hveficent:
I think this module does not make sense anymore. On v13 mail activities are DELETED when they are done, and a message is posted with proper Activity kind, Note and Feedback.

HviorForgeFlow:
You must override method _action_done instead of action_feedback.

emagdalenaC2i:
@pegonzalezspesol I can't reproduce that error. The done button works fine for me.
Can you explain the steps to follow?
Have you tried it on Runbot or have you installed it?

pegonzalezspesol:
> @pegonzalezspesol I can't reproduce that error. The done button works fine for me.
> Can you explain the steps to follow?
> Have you tried it on Runbot or have you installed it?
On Runbot. Steps:
1.- Schedule an Activity
2.- Mark as Done
3.- Write a feedback
4.- Click on Done
image

HviorForgeFlow:
feedback field has been removed from mail.activity model. Now feedback is prompt on mail message, so write feedback field break module.
We might store that mail message reference on done activities to not lose that information.
js parts has been refactores (camelCase function naming) as we are overwriting tat parts I would be interest to be as similar as new version.

AdriaGForgeFlow:
Travis is red, could you check? Thanks!

HviorForgeFlow:
@emagdalenaC2i could you close it? Thanks,

PR #446, opened 2019-10-17
[11.0][FIX] mail_tracking: Reload mail fields

### LINKED ISSUE #443: [11.0][mail_tracking] Incompatibility with OCB commit
https://github.com/OCA/social/blob/eeedcbd41dc960ffdf19e8f7924ecb5b7aa9d84d/mail_tracking/static/src/js/mail_tracking.js#L11

Hello,

I have a problem with the module mail_tracking in Odoo 11.0.

Two days ago I updated the repositories, and customers who have installed the mail_tracking module, they are having trouble creating activities. When I create an activity, until I refresh the page, the activity does not appear.

I think the module has some incompatibility with the latest OCB changes:
https://github.com/OCA/OCB/commit/d7623ddcfb886194c5501bc5db36360caed15da8
There are changes in the file chatter.js

If I uninstall mail_tracking the problem disappears

Thanks a lot.

Tardo:
Fixes #443 

cc @Tecnativa

### DIFF
--- a/mail_tracking/static/src/js/failed_message.js
+++ b/mail_tracking/static/src/js/failed_message.js
@@ -348,15 +348,14 @@ odoo.define('mail_tracking.FailedMessage', function (require) {
         },
 
         _onReloadMailFields: function (event) {
-            this._super.apply(this, arguments);
-            var fieldNames = [];
             if (this.fields.failed_message && event.data.failed_message) {
-                fieldNames.push(this.fields.failed_message.name);
+                this.trigger_up('reload', {
+                    fieldNames: [this.fields.failed_message.name],
+                    keepChanges: true,
+                });
+            } else {
+                this._super.apply(this, arguments);
             }
-            this.trigger_up('reload', {
-                fieldNames: fieldNames,
-                keepChanges: true,
-            });
         },
     });
 


yajo:
Could you provide some info on how to check this works and why it was broken, etc.? Thanks!

pedrobaeza:
Check the issue it tries to fix.

pedrobaeza:
I'm afraid this is not working. Look how this still doesn't refresh activity counter until you click, while it should be automatic:

Peek 17-10-2019 12-52

This is the same test done on fresh runbot:

Peek 17-10-2019 12-53

oscars8a:
> Could you provide some info on how to check this works and why it was broken, etc.? Thanks!

When you create an activity, the normal behavior is that it appears in the activities section without having to refresh the page. But now you have to force to show the changes with f5.
If you uninstall mail_tracking and the longpolling is working, Odoo refreshes the change automatically. Is it possible that longpolling has to do? I have doubts that longpolling is enabled in runbot.

Tardo:
On my local env. it refresh as expected:
update_activities

pedrobaeza:
OK, maybe OCA runbot doesn't have longpolling (Odoo runbot does or at least works as shown). @Tardo, can you please confirm that before this patch, it wasn't working?

Tardo:
Yes, the problem is how the main method is implemented: https://github.com/odoo/odoo/blob/13.0/addons/mail/static/src/js/chatter.js#L615
Before this patch the method sends the same event two times... one with 'failed_message' and other empty... this causes the event not to propagate properly.

yajo (approved):

pedrobaeza (approved):
Please take into account for 12.0

oscars8a:
Thanks a lot! @pedrobaeza, @Tardo and @Yajo.

PR #450, opened 2019-10-22
[12.0][ADD] create module mail_header_notified_partner

Tonow-c2c:
Result: 
image

### DIFF
--- a/mail_header_notified_partner/__init__.py
+++ b/mail_header_notified_partner/__init__.py
@@ -1,0 +1,2 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from . import models
--- a/mail_header_notified_partner/__manifest__.py
+++ b/mail_header_notified_partner/__manifest__.py
@@ -1,0 +1,15 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Notified partners in mail header",
+    "summary": """
+        This module adds the list of notified partners in the header of
+        notification e-mails sent by Odoo.
+        """,
+    "author": "Camptocamp,Odoo Community Association (OCA)",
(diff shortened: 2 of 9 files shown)


pedrobaeza:
But this is what `mail_tracking` does...

rafaelbn:
I think mail_tracking module is "The hidden diamond" When C2C discover it... will use every where :thinking:

Tonow-c2c:
@pedrobaeza not exactly in this case the recipients list is in the mail.message body so we keep this trace after forwarding

Tonow-c2c:
But you right may be we should in future improve mail_tracking module to add possibility to add the recipients list in the message body

StefanRijnhart:
Ah, so this module actually tells the recipients of the notification emails who else was notified? Maybe this module could be modified to add the header to the conjoined email itself and not in the body as it is displayed in the thread so as not to overlap with `mail_tracking` which is really great by itself indeed.

len-foss:
We are also interested in having this module, which is a bit orthogonal to what mail_tracking does.
It is also essentially the same as the older [mail_footer_notified_partners](https://github.com/OCA/social/tree/8.0/mail_footer_notified_partners).
Or should it be added as a checkbox "show notified followers in email" feature to mail_tracking?

pedrobaeza:
I think it can be a system/company parameter in `mail_tracking`, and traversing not in the body, but at sending time (as it also depends on the recipient) as @StefanRijnhart commented is the right way - similar to what is done with mail.notification QWeb template in sales order/invoice.

For a simpler solution, `mail_footer_notified_partners` should be migrated instead (there are some migration PRs right there).

len-foss:
@pedrobaeza 
You're right, and it also saw some renewed interest... I'll test that version then, and will see later if I can squeeze that feature in mail_tracking. Cheers!

PR #452, opened 2019-10-24
[MIG][12.0] mass_mailing_newsletter_welcome_mail

ernestotejeda:
Cc @Tecnativa TT19367


mass_mailing_newsletter_welcome_mail/controllers/main.py
```diff
@@ -16,16 +16,19 @@ def subscribe(self, list_id, email, **post):
         if not template:
             return result
         # Welcome new subscribers
-        contact = request.env["mail.mass_mailing.contact"].sudo().search([
-            ('list_ids', 'in', list_.ids),
-            ('email', '=', email),
-            ("opt_out", "=", False),
+        contact_obj = request.env["mail.mass_mailing.contact"].with_context(
+            default_list_ids=[list_id])
```
pedrobaeza, line 20: Why this change?

mass_mailing_newsletter_welcome_mail/controllers/main.py
```diff
@@ -16,16 +16,19 @@ def subscribe(self, list_id, email, **post):
...
+            ('is_blacklisted', '=', False),
         ], limit=1)
-        template.with_context(list_name=list_.name).send_mail(
-            contact.id,
-            # Must send now to use context
-            force_send=True,
-            # If we cannot notify, the visitor shouldn't be bothered
-            raise_exception=False,
-        )
+        if contact:
```
pedrobaeza: Do this instead for avoiding too much diff:
```python
if not contact:
    return
```

pedrobaeza (changes requested):

Tardo (approved):
:+1: LGTM

mass_mailing_newsletter_welcome_mail/controllers/main.py
pedrobaeza: Can you please answer about this?

ernestotejeda: Now in v12 the contacts are not opted out by themselves, but they are opted out or not from a particular list. One way to search contacts that are opted out from a particular list is that, by searching by the opt_oup field of mass.mailing.contact (which is now a calculated field) and passing in the context the list id.
check this
https://github.com/odoo/odoo/blob/fe35870fa0024a1e8adebb386f343f25fdea2c7b/addons/mass_mailing/models/mass_mailing.py#L250
On the other hand, I assumed that so in v11 we didn't want to send mail to contacts that had been opted out from the list, now in v12 that blacklisting exists, we don't whant either to send mail to those who are blacklisted

ernestotejeda:
Changes done

mass_mailing_newsletter_welcome_mail/controllers/main.py
pedrobaeza: Well, if they are subscribing just now, how they are going to be opted-out?

If you keep it (and for the rest of cases), please always explain this not easy to deduce things as a comment in the code.

ernestotejeda: Maybe in Odoo 11.0 this is not necessary but in v12 yes because the re-subscription of a contact doesn't work well. I made a issue to Odoo for this: https://github.com/odoo/odoo/issues/39604

pedrobaeza: OK, then please add that as comment:
```python
# Needed for being re-subscribed while odoo/odoo#39604 is not fixed
```

pedrobaeza (approved):

PR #447, opened 2019-10-18
[12.0][IMP] mail_tracking: Failed messages

Tardo:
Added failed messages feature to 12.0

failed_messages

cc @tecnativa TT18533

PR #454, opened 2019-10-25
[11.0][IMP] mail_tracking: Buttons, discuss, empty mails

Tardo:
- Updated readme and demo
- Now only can see the failed messages in discuss app where you are the author or recipient
- Buttons 'Mark Reviewed' & 'Retry' are displayed in the same place as star & reply buttons
- Partners with empty mails are don't marked as bounced email and have their tracking icon:
2019-10-25-140900_1920x1080_scrot

cc @tecnativa


mail_tracking/models/mail_message.py
```diff
@@ -130,7 +135,12 @@ def _get_failed_message(self):
...
                 message.id: message.mail_tracking_needs_action
-                and bool(message.mail_tracking_ids.filtered(
+                and (
+                    message.author_id.id == self.env.user.partner_id.id
+                    or any(message.partner_ids.filtered(
+                        lambda x: x.id == self.env.user.partner_id.id))
+                )
+                and any(message.mail_tracking_ids.filtered(
                     lambda x: x.state in self.get_failed_states()))
             })
```
yajo: Please make this more readable. See https://github.com/OCA/social/pull/447#discussion_r339488173 as it is pretty similar.

mail_tracking/models/mail_message.py
```diff
@@ -161,7 +171,7 @@ def _prepare_dict_failed_message(self, message):
         return {
             'id': message.id,
             'date': message.date,
-            'author_id': message.author_id.name_get()[0],
+            'author_id': self.author_id and self.author_id.name_get()[0] or '',
```
yajo: I don't understand this change.

This is an `@api.model`, so by definition `self` shouldn't have a record attached. Also, `name_get()[0]` would return a tuple usually, but returning a tuple or an empty string doesn't seem to have much sense.

mail_tracking/models/mail_message.py
```diff
@@ -178,10 +188,13 @@ def toggle_tracking_status(self):
         return self.mail_tracking_needs_action
 
     def _get_failed_message_domain(self):
```
yajo, line 211: As you're using it in a `default`, without a record attached, I think this would clarify that:

```suggestion
    @api.model
    def _get_failed_message_domain(self):
```

mail_tracking/models/mail_tracking_email.py
```diff
@@ -162,7 +162,9 @@ def email_score(self):
     @api.depends('recipient')
     def _compute_recipient_address(self):
         for email in self:
-            if email.recipient:
+            is_empty_recipient = not email.recipient or re.search(
+                r'<False>', email.recipient)
```
yajo: Do you really need a regexp for that? `'<False>' in email.recipient` should do the same...

mail_tracking/models/mail_tracking_email.py
```diff
@@ -216,13 +218,25 @@ def _partners_email_bounced_set(self, reason, event=None):
...
-        self.sudo()._partners_email_bounced_set('error')
+        values = {}
+        IrMailServer = self.env['ir.mail_server']
+        if str(exception) == IrMailServer.NO_VALID_RECIPIENT \
+                and not self.recipient_address:
+            values.update({
+                'state': 'error',
+                'error_type': 'no_recipient',
+                'error_description':
+                    _("The partner doesn't have a defined email"),
```
yajo: I think you don't want translations in a log:

```suggestion
                    "The partner doesn't have a defined email",
```

Otherwise your chinese colleague could produce logs impossible to understand to you, or viceversa.

mail_tracking/static/src/xml/failed_message.xml
```diff
@@ -13,7 +13,16 @@
                     <strong class="o_thread_author">
                         <t t-esc="message.author_id[1]"/>
                     </strong>
-                    - <small class="o_mail_timestamp" t-att-title="message.date.format(date_format)"><t t-esc="message.hour"/></small><br/>
+                    - <small class="o_mail_timestamp" t-att-title="message.date.format(date_format)"><t t-esc="message.hour"/></small>
+                    <span t-attf-class="o_thread_icons">
+                        <a href="#" class="btn btn-link btn-success text-muted btn-sm o_failed_message_reviewed o_activity_link" t-att-data-message-id="message.id">
+                            <i class="fa fa-check"/> Mark Reviewed
```
yajo, line 19:
```suggestion
                            <i class="fa fa-check"/> Set as Reviewed
```

yajo (changes requested):

pedrobaeza:
@Yajo can we finish this review for merging it?

mail_tracking/static/src/js/failed_message.js
```diff
@@ -456,6 +453,7 @@ odoo.define('mail_tracking.FailedMessage', function (require) {
                     keepChanges: true,
                 });
             } else {
+                // Workarround to avoid trigger reload event two times.
```
yajo:
```suggestion
                // Workaround to avoid trigger reload event twice.
```

yajo:
Please apply last suggestion and squash.

Tardo:
@Yajo done

yajo (approved):

PR #455, opened 2019-10-28
[FIX] avoid bounce on 'error' event

chienandalu:
Catching this event as a bounce leads quite often to false bounces, like
in the cases the error is due to our own instance SMTP errors. Errors
like a too large message should no bounce a thread's recipients as the
problem isn't in the recipients side but ours.

We leave anyway some true positives aside (like in the case of malformed
email addresses). We should be able to detect and fix those anyway.

cc @Tecnativa TT20293

### DIFF
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -105,8 +105,7 @@ class MailTrackingEmail(models.Model):
         if not email:
             return False
         res = self._email_last_tracking_state(email)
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):

yajo (approved):

yajo:
Check travis please.

pedrobaeza:
Can you please check Travis status?

chienandalu:
Fixed

PR #458, opened 2019-11-08
WIP [ADD] mail_embed_image module

gfcapalbo:
replaces https://github.com/OCA/social/pull/404


gfcapalbo:
@all reviewers. I am having problems pinpointing the test errors. 
My local tests are passing and I have no idea how to replicate.
any ideas? I have installed all repo modules on local to try to replicate this situation.

gfcapalbo:
tests passing

mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class IrMailServer(models.Model):
+    _inherit = 'ir.mail_server'
+
+    @contextmanager
+    def fetch_image(self, path):
```
hbrunn: just out of habit, make this private

mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,130 @@
...
+        public_user = self.env.ref('base.public_user')
+        session_store = root_wsgi.session_store
+        session = session_store.new()
```
hbrunn, line 29: I don't have a good intuition how expensive this is. We might want to do that only once per mail, not once per image

mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,130 @@
...
+        # construct an odoo request with this werkzeug request.
+        request = http.HttpRequest(werkzeug_request)
+        with request:
+            request._env = self.env(user=public_user)
+            endpoint, arguments = http.routing_map(
+                self.env.registry._init_modules,
+                False,
+                self.env['ir.http']._get_converters()
+            ).bind_to_environ(
+                werkzeug_env).match(return_rule=False,)
```
hbrunn, line 48: I'm still not sure if we can keep the reference to the rule, and somehow make it just parse the new arguments, then we'd only have to do the above once. But don't change anything now, let's see if we fine a better way in the future.

mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,130 @@
...
+            if part.get_content_type() == 'text/html':
+                body = part.get_payload(decode=True)
+                if not body or body == '\n':
+                    continue
+                root = self.process_img_body(fromstring(body), email)
+                # encodestring will put a newline every 74 char
+                part.set_payload(encodestring(tostring(root)))
+        return email
+
+    def process_img_body(self, root, email):
```
hbrunn: this should be private too. It would make sense to prefix this function too with `_build_email` as the above does, indicating they belong together

mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,130 @@
...
+                    base_url or '', )):
+            # check if there is a bound request
+            image_path = img.get('src').replace(base_url, '')
+            with self.fetch_image(image_path) as (endpoint, arguments):
+                # now go ahead and call the endpoint and fetch the data
+                response = endpoint.method(**arguments)
+                if not response:
+                    logger.warning('Could not get %s', img.get('src'))
+                    continue
+                cid = uuid.uuid4().hex
```
hbrunn, line 116: this is a comparatively complex operation. The id string just needs to be unique, so you can just as well simply count

mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def process_img_body(self, root, email):
+        base_url = self.env['ir.config_parameter'].get_param(
+            'web.base.url')
+        for img in root.xpath(
+                "//img[starts-with(@src, '%s/web/image')]"
+                "|"
+                "//img[starts-with(@src, '/web/image')]" % (
+                    base_url or '', )):
+            # check if there is a bound request
```
hbrunn: we do none of that any more

mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,130 @@
...
+            image_path = img.get('src').replace(base_url, '')
+            with self.fetch_image(image_path) as (endpoint, arguments):
+                # now go ahead and call the endpoint and fetch the data
+                response = endpoint.method(**arguments)
+                if not response:
+                    logger.warning('Could not get %s', img.get('src'))
+                    continue
+                cid = uuid.uuid4().hex
+                filename_rfc2047 = encode_header_param(cid)
+                filepart = MIMEImage(response.data)
```
hbrunn, line 118: I assume this takes care of setting the correct mimetype et al?

mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,130 @@
...
+                filepart.set_param('name', filename_rfc2047)
+                filepart.add_header(
+                    'Content-Disposition',
+                    'inline',
+                    cid=cid,
+                    filename=filename_rfc2047,
+                )
+                # attach the image into the email as attachment
+                email.attach(filepart)
+                img.set('src', 'cid:%s' % (cid))
```
hbrunn: please add a comment here that we should avoid attaching the same image twice. Not necessary for version 1 I think

mail_embed_image/tests/test_mail_embed_image.py
```diff
@@ -0,0 +1,92 @@
...
+class TestMailEmbedImage(common.TransactionCase):
+
+    post_install = True
+    at_install = False
+
+    @patch(
+        'odoo.fields.html_sanitize',
+        side_effect=lambda *args, **kwargs: args[0],
+    )
+    def test_mail_embed_image(self, html_sanitize):
```
hbrunn: we discussed this elsewhere

hbrunn (changes requested):

mail_embed_image/models/ir_mail_server.py
gfcapalbo: the MIMEImage class is used to create MIME message objects of major type image. _imagedata is a string containing the raw image data. If this data can be decoded by the standard Python module imghdr, then the subtype will be automatically included in the Content-Type header. Otherwise you can explicitly specify the image subtype via the _subtype parameter

mail_embed_image/models/ir_mail_server.py
```diff
@@ -117,14 +117,20 @@ def process_img_body(self, root, email):
...
-                filepart.add_header(
-                    'Content-Disposition',
-                    'inline',
-                    cid=cid,
-                    filename=filename_rfc2047,
-                )
-                # attach the image into the email as attachment
-                email.attach(filepart)
+                # check if filepart exists (do not attach twice)
+                filepart_exists = [
```
hbrunn: if you want to do this, use a dict mapping src content to cid, but I asked you to just add a comment...

gfcapalbo:
fixing tests will remove wip when done

mail_embed_image/models/ir_mail_server.py
gfcapalbo: removed for now.

mail_embed_image/models/ir_mail_server.py
gfcapalbo: added simple incremental cid.

mail_embed_image/tests/test_mail_embed_image.py
```diff
@@ -0,0 +1,58 @@
...
+from base64 import b64encode
+from odoo.tests import common
+from lxml import html
+from requests import get
+
+
+class TestMailEmbedImage(common.TransactionCase):
+
+    post_install = True
+    at_install = False
```
hbrunn: why this?

mail_embed_image/models/ir_mail_server.py
hbrunn: this seems to be lost, I don't see it in the diff

mail_embed_image/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* George Daramouskas <gdaramouskas@therp.nl>
```
hbrunn, line 1: you need to update this file, and rerun the readme generator

mail_embed_image/models/ir_mail_server.py
gfcapalbo: removed. was causing errors. retesting on another branch.

hbrunn (approved):
please squash your commits

thomaspaulb (approved):
Great addition to OCA and including a first-class test.

Remaining two comments by @hbrunn are performance-related only, so non-critical.

hbrunn:
I agree we can address those when/if they hurt

/ocabot merge nobump

PR #460, opened 2019-11-12
[FIX][12.0][mass_mailing_partner] add api.multi to ResPartner write

Tonow-c2c:


pedrobaeza:
`@api.multi` is the default when nothing is put, so this doesn't add any value, so closing.

Tonow-c2c:
Sorry @pedrobaeza but this fix solve issue when you try to create a new contact with multi child's :disappointed_relieved:

pedrobaeza:
I can assure you the problem should be in other place. Please provide steps to reproduce the problem in runbot if not and I will test.

PR #461, opened 2019-11-13
[FIX][12.0][mass_mailing_partner] _check_email_mass_mailing_contacts

Tonow-c2c:
_check_email_mass_mailing_contacts loop not on self but partner


Tonow-c2c:
Thanks @pedrobaeza you totally right on https://github.com/OCA/social/pull/460

pedrobaeza (approved):
Thanks for detecting the problem

pedrobaeza:
Let's fast-track it as straight-forward.

/ocabot merge patch

PR #463, opened 2019-11-14
[FIX] mail_tracking: Set as reviewed notification

Tardo:
Previous this commit, failed messages widget data wasn't updated properly when you set as reviewed/retry the last message displayed.

This commit resolve the problem and improves the "set as reviewed" action to avoid multi-user inconsistencies (don't toggle the value).

Cc @tecnativa

### DIFF
--- a/mail_tracking/models/mail_message.py
+++ b/mail_tracking/models/mail_message.py
@@ -203,18 +203,17 @@ class MailMessage(models.Model):
                 for msg in self.sorted('date', reverse=True)]
 
     @api.multi
-    def toggle_tracking_status(self):
-        """Toggle message tracking action.
+    def set_need_action_done(self):
+        """Set message tracking action as done
 
-        This will mark them to be (or not) ignored in the tracking issues
-        filter.
+        This will mark them to be ignored in the tracking issues filter.
         """
         self.check_access_rule('read')
-        self.mail_tracking_needs_action = not self.mail_tracking_needs_action
+        self.write({'mail_tracking_needs_action': False})
         notification = {
             'type': 'toggle_tracking_status',
-            'message_ids': [self.id],
-            'needs_actions': self.mail_tracking_needs_action
+            'message_ids': self.ids,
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Please put steps to reproduce the problem in the commit message and text of the PR.

mail_tracking/models/mail_message.py
```diff
@@ -203,16 +203,16 @@ def get_failed_messages(self):
...
 
         This will mark them to be (or not) ignored in the tracking issues
         filter.
         """
         self.check_access_rule('read')
-        self.mail_tracking_needs_action = not self.mail_tracking_needs_action
+        self.mail_tracking_needs_action = False
         notification = {
-            'type': 'toggle_tracking_status',
+            'type': 'mail_tracking_needs_action_changed',
```
pedrobaeza: I would let the notification type the same, for being able to reuse if we want to notify also a new error in the future.

pedrobaeza: Forget it. I'm seeing that the name is more specific and cover both cases.

mail_tracking/static/src/js/failed_message/discuss.js
```diff
@@ -106,13 +106,13 @@ odoo.define('mail_tracking.FailedMessageDiscuss', function (require) {
 
         /**
          * This method updates messages in the failed mailbox when the flag
-         * 'mail_tracking_needs_action' is toggled. This can remove/add
+         * 'mail_tracking_needs_action' was change to False. This can remove/add
```
yajo:
```suggestion
         * 'mail_tracking_needs_action' was changed to False. This can remove/add
```

yajo (approved):

Tardo:
changes done

pedrobaeza:
Travis is red

Tardo:
Done :green_heart:

Tardo:
OCA Bot got stuck

PR #469, opened 2019-11-28
[12.0][FIX] mail_tracking: Add 'failed_message_ids' field description

Tardo:
Changed injection method for 'failed_message_ids' field.

Before this PR, some methods with views crashes trying to process 'failed_message_ids' field.
After this commit the 'failed_message_ids' field description is properly generated.

cc @tecnativa TT20818

### DIFF
--- a/mail_tracking/models/mail_thread.py
+++ b/mail_tracking/models/mail_thread.py
@@ -69,14 +69,14 @@ class MailThread(models.AbstractModel):
         return res
 
     @api.model
-    def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
-                        submenu=False):
+    def _fields_view_get(self, view_id=None, view_type='form', toolbar=False,
+                         submenu=False):
         """Add filters for failed messages.
 
         These filters will show up on any form or search views of any
         model inheriting from ``mail.thread``.
         """
-        res = super().fields_view_get(
+        res = super()._fields_view_get(
             view_id=view_id, view_type=view_type, toolbar=toolbar,
             submenu=submenu)
         if view_type not in {'search', 'form'}:


pedrobaeza:
Isn't possible to check if exists where processed instead of adding that everywhere? Something like:
```python
            if res['fields'].get('failed_message_ids'):
                # Create field
                field_failed_messages = etree.Element('field', {
                    'name': 'failed_message_ids',
                    ....
```

Tardo:
done!

yajo (approved):

mail_tracking/models/mail_thread.py
```diff
@@ -103,7 +103,8 @@ def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
                     })
                 nodes[0].append(etree.Element('separator'))
                 nodes[0].append(new_filter)
-        elif view_type == 'form':
+        elif view_type == 'form' \
+                and not res['fields'].get('failed_message_ids'):
```
pedrobaeza:
```suggestion
                and res['fields'].get('failed_message_ids'):
```

mail_tracking/models/mail_thread.py
```diff
@@ -114,5 +115,12 @@ def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
                     'widget': 'mail_failed_message',
                 })
                 nodes[0].addprevious(field_failed_messages)
+                res['fields'].update({
```
pedrobaeza: Remove this

pedrobaeza (changes requested):
This was not my idea... What I was saying is that if the field doesn't exist (or it's not requested to be rendered), you don't have to add that defaults. Can that work?

Tardo:
Ok, after deep debugging the problem is finally fixed using the properly method to inject the 'failed_message_ids' field.
After call it, the public method generates field descriptions.

pedrobaeza (approved):

PR #472, opened 2019-12-02
[12.0] Add mail_layout_preview WIP

simahawk:


simahawk:
moving on w/ v13 #489

PR #476, opened 2019-12-06
[13.0][MIG] mail_optional_follower_notification: Migration to 13.0

lmignon:
includes #639


ThomasBinsfeld (approved):

mail_optional_follower_notification/models/mail_thread.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    def _notify_compute_recipients(self, message, msg_vals):
+        """ Compute recipients to notify based on subtype and followers. This
+        method returns data structured as expected for ``_notify_recipients``. """
+        recipient_data = super()._notify_compute_recipients(message, msg_vals)
+        if not self.env.context.get("notify_followers", False):
```
isabellerichard:
```suggestion
        if 'notify_followers' in self.env.context and \
                not self.env.context.get("notify_followers", False):
```

isabellerichard: When I send a direct message via chat, notify_followers is not in the context, and messages are never sent.

Adding this condition fixes the issue.

isabellerichard (changes requested):
When I send a direct message via chat, notify_followers is not in the context, and messages are never sent.

Adding this condition fixes the issue.

lmignon:
This one is in production for months... Can we merge-it?

ThomasBinsfeld (approved):

BT-aestebanez (approved):

BT-aestebanez:
imho, this can be merged

BT-aestebanez:
@pedrobaeza 
Could this PR be merged to 13.0?
Thanks

pedrobaeza:
Sorry, I don't know this module. Other maintainer/PSC should merge it if he/she is sure about it.

BT-aestebanez:
@simahawk 
Could this PR be merged to 13.0? (you merged the migration to 12.0)
Thanks

mail_optional_follower_notification/models/mail_thread.py
```diff
@@ -0,0 +1,27 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    def _notify_compute_recipients(self, message, msg_vals):
+        """ Compute recipients to notify based on subtype and followers. This
```
simahawk, line 11: this overrides the original docstring - if any - you should use a comment instead.

mail_optional_follower_notification/models/mail_thread.py
```diff
@@ -0,0 +1,27 @@
...
+        method returns data structured as expected for ``_notify_recipients``. """
+        recipient_data = super()._notify_compute_recipients(message, msg_vals)
+        if "notify_followers" in self.env.context and not self.env.context.get(
```
simahawk, line 14: why not just `if self.env.context.get("notify_followers")`?

mail_optional_follower_notification/wizard/mail_compose_message.py
```diff
@@ -0,0 +1,18 @@
...
+    _inherit = "mail.compose.message"
+
+    notify_followers = fields.Boolean(default=True)
+
+    def send_mail(self, auto_commit=False):
+        ctx = self.env.context.copy()
+        for wizard in self:
+            ctx["notify_followers"] = wizard.notify_followers
+            wizard = wizard.with_context(ctx)
+            super(MailComposeMessage, wizard).send_mail(auto_commit=auto_commit)
```
simahawk: each wizard could have potentially it's own ctx, I'd do:
```suggestion
        for wizard in self:
            wizard = wizard.with_context(notify_followers=wizard.notify_followers)
            super(MailComposeMessage, wizard).send_mail(auto_commit=auto_commit)
```

mail_optional_follower_notification/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2016 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Mail optional follower notification",
+    "summary": "Choose to notify followers on mail.compose.message",
+    "author": "ACSONE SA/NV," "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/social",
+    "category": "Social Network",
```
simahawk, line 8: social network?

simahawk (changes requested):
minor glitches

mail_optional_follower_notification/models/mail_thread.py
lmignon: @simahawk otherwise you change the default behaviour if the key is not explicitly provided into the context.

mail_optional_follower_notification/wizard/mail_compose_message.py
lmignon: @simahawk I admit that the syntax is more elegant (a remaining odoo v8 syntax inheritance). But AFAIK, the context is the same for all the records into a recordset since the context is provided by the environment.

mail_optional_follower_notification/models/mail_thread.py
lmignon: That's the docstring of the overridden method.

mail_optional_follower_notification/models/mail_thread.py
simahawk: if the key is None, very likely is not there. Not a blocker of course.

mail_optional_follower_notification/models/mail_thread.py
simahawk: But then you should:

* mention that it extends the original behavior
* multi line docstring should be separated 

Anyway, there's no official OCA rule for this (yet).
Feel free to ignore my remark :wink:

simahawk (approved):

simahawk:
can you check the build?
FTR if you guys put yourselves as maintainers in the manifest you'll be able to merge PRs for the module :wink:

lmignon:
@simahawk Tests fails due to method signature change into odoo 13 :expressionless: https://github.com/odoo/odoo/commit/fbd30f4266d86b53aa4b13b72a93bccee0d7eb69

lmignon:
@simahawk Travis is fixed. related pr included into this one: #639

simahawk:
@lmignon you have to rebase right?

lmignon:
@simahawk no need to rebase but rebase is done. PLZ bump version `ocabot merge patch` at merge to ensure that the addon is pushed to pypi with an official version and not a .dev version.

PR #487, opened 2020-01-16
[11.0][ADD] mail_private

etobella:
This module allows to define messages as private and show them according to the defined groups on the `ir.model`.

Usage
======
Access the model and define the private groups. If the user, is on any of the groups, he will be able to define a private message
imagen


etobella:
Now, many groups can be defined on the same model

mail_private/security/security.xml
```diff
@@ -0,0 +1,16 @@
...
+        <field name="model_id" ref="mail.model_mail_message"/>
+        <field name="name">Message private rules</field>
+        <field name="domain_force">['|', '|', ('mail_group_id', '=', False), ('model', '=', False), ('mail_group_id', 'in', user.message_group_ids.ids)]</field>
+    </record>
+
+
+    <record id="attachment_private_rule" model="ir.rule">
+        <field name="model_id" ref="base.model_ir_attachment"/>
+        <field name="name">Attachment private rules</field>
+        <field name="domain_force">['|', '|', '|', ('related_message_ids', '=', False),('related_message_ids.mail_group_id', '=', False), ('res_model', '=', False), ('related_message_ids.mail_group_id', 'in', user.message_group_ids.ids)]</field>
```
etobella: After checking on a big database, this query is extremely inefficient. I will try to find another solution, although it does not seem easy...

etobella: Well, I think I found a solution for this issue, setting the group of the attachment. Not as clean as I would like, but a solution after all

jarroyomorales (approved):

jarroyomorales:
@OCA/social-maintainers This is more or less the same as https://github.com/OCA/social/pull/552 but in v11. Can someone check? Thank you!

albariera (approved):

Issue #480, opened 2019-12-27
error when sending mail without followers as recipients


enclustra-odoo:
module: mail_optional_follower_notification
version: 12.0

**Steps to reproduce**
- Send e-mail from purchase or sales with the mail composer
- Mail template is `mail.mail_notification_paynow`

**Current behavior**
- e-mail is not sent and the following error is displayed:
```
Error:
Odoo Server Error

Traceback (most recent call last):
  File "/mnt/c/Users/tbrauchli/odoo/odoo-ce/odoo/addons/base/models/qweb.py", line 347, in _compiled_fn
    return compiled(self, append, new, options, log)
  File "<template>", line 1, in template_330_728
  File "/mnt/c/Users/tbrauchli/odoo/odoo-ce/odoo/models.py", line 1530, in get_access_action
    return self[0].get_formview_action(access_uid=access_uid)
  File "/mnt/c/Users/tbrauchli/odoo/odoo-ce/odoo/models.py", line 5127, in __getitem__
    return self._browse((self._ids[key],), self.env)
IndexError: tuple index out of range

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/mnt/c/Users/tbrauchli/odoo/odoo-ce/odoo/http.py", line 656, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/mnt/c/Users/tbrauchli/odoo/odoo-ce/odoo/http.py", line 314, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/mnt/c/Users/tbrauchli/odoo/odoo-ce/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/mnt/c/Users/tbrauchli/odoo/odoo-ce/odoo/http.py", line 698, in dispatch
    result = self._call_function(**self.params)
  File "/mnt/c/Users/tbrauchli/odoo/odoo-ce/odoo/http.py", line 346, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/mnt/c/Users/tbrauchli/odoo/odoo-ce/odoo/service/model.py", line 97, in wrapper
    return f(dbname, *args, **kwargs)
  File "/mnt/c/Users/tbrauchli/odoo/odoo-ce/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
[...]

PR #493, opened 2020-01-31
[FIX] mail_optional_follower_notification (https://github.com/OCA/social/issues/480)

remihb:
The aim of this PR is to
- fix template problem (see [https://github.com/OCA/social/issues/480](https://github.com/OCA/social/issues/480))
- add feature on send invoice wizard (not sure if it belongs to this module)

### DIFF
--- a/mail_optional_autofollow/__init__.py
+++ b/mail_optional_autofollow/__init__.py
@@ -1,1 +1,2 @@
+from . import models
 from . import wizard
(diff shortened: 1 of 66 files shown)


len-foss (commented):
I believe that the of https://github.com/OCA/social/pull/496 apply here; however once the commits are split it should be mergeable.

PR #494, opened 2020-02-03
[11.0][ADD] mail_activity_creator

jarroyomorales:
Very simple module to show who is the creator of activities since it is missing in Odoo core.

image

@etobella


etobella (approved):
LGTM :+1:

etobella:
It can be useful when users wants to check who created the task in order to ask questions

etobella (changes requested):
It is not working properly. Odoo creates activities as sudo, so, the creator is never stored properly
https://github.com/odoo/odoo/blob/11.0/addons/mail/models/mail_activity.py#L192

etobella (approved):
Now it works as expected. Also, a test was added showing the issue with create_uid

jarroyomorales:
Thank you for the fix! @etobella

JordiBForgeFlow (approved):

jarroyomorales:
Thank you @JordiBForgeFlow !
I think in 12.0 it is already implemented, so no migration will be necessary.

PR #496, opened 2020-02-06
[IMP] mail_optional_autofollow : add feature on send invoice wizard

remihb:
Add module feature on the wizard for sending invoices.

Similar to #493 for the improvement part

Bypass this : [https://github.com/OCA/OCB/blob/12.0/addons/account/models/account_invoice.py#L704](https://github.com/OCA/OCB/blob/12.0/addons/account/models/account_invoice.py#L704) !

### DIFF
--- a/mail_optional_autofollow/__init__.py
+++ b/mail_optional_autofollow/__init__.py
@@ -1,1 +1,2 @@
+from . import models
 from . import wizard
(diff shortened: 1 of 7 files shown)


len-foss (commented):
Looks pretty good to me, however adding `account` to dependencies is not acceptable.
The solution is to split into two commits, one that adds the override context key in this module, and a bridge module that depends on this and account to add the feature.

Have you tried reporting to Odoo to use the context key at https://github.com/OCA/OCB/blob/b15cf7e5bc6c6dfd27a81943d1b554d947f21bfa/addons/account/models/account_invoice.py#L703 ? (also use this Github link syntax in your comments, so that the line you refer to make sense -- subsequent commits can change it on 12.0)

PR #501, opened 2020-02-13
[12.0][ADD] mail_activity_reminder

alexey-pelykh:
This module allows setting reminders for various Activity Types.


bennals:
Would this module work in 11.0, or if not, any thought to making a version for it?

alexey-pelykh:
@bennals I think it can be backported once it's ironed out for v12

alexey-pelykh:
@pedrobaeza this one is polished I guess and ready for review, since you were interested in it :D

mail_activity_reminder/models/mail_activity.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @api.model
+    def _get_activities_to_remind(self):
+        utc_now = self.env.context.get(
+            'mail_activity_reminder_now',
+            fields.Datetime.now()
+        ).replace(tzinfo=UTC)
+
+        activities_to_remind = self.env['mail.activity']
+        for activity in self.search([('active', '=', True)]):
```
CasVissers-360ERP: Module should depend on mail_activity_done since activity is not a base field?

CasVissers-360ERP (changes requested):
Only small functional review, will do additional reviewing later.

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: Sorry, but I don't get you. What field do you refer to?

alexey-pelykh: `active` is very much field of base model https://github.com/odoo/odoo/blob/2ad6cb4feb01dbb394ab50ce9f8c1fd6da30bf8c/addons/mail/models/mail_activity.py#L39

alexey-pelykh: And I'm not accessing any field named `activity`

CasVissers-360ERP: Weird I got an error on it while installing, let me check again.

alexey-pelykh: A stacktrace would be helpful

CasVissers-360ERP:
```
  File "/Users/brahoo/PycharmProjects/odoo/community/mail_activity_reminder/models/mail_activity.py", line 47, in _process_reminders
    activities = self._get_activities_to_remind()
  File "/Users/brahoo/PycharmProjects/odoo/community/mail_activity_reminder/models/mail_activity.py", line 22, in _get_activities_to_remind
    for activity in self.search([('active', '=', True)]):
  File "/Users/brahoo/PycharmProjects/odoo/odoo-12.0/odoo/models.py", line 1579, in search
    res = self._search(args, offset=offset, limit=limit, order=order, count=count)
  File "/Users/brahoo/PycharmProjects/odoo/odoo-12.0/odoo/models.py", line 4127, in _search
    query = self._where_calc(args)
  File "/Users/brahoo/PycharmProjects/odoo/odoo-12.0/odoo/models.py", line 3919, in _where_calc
    e = expression.expression(domain, self)
  File "/Users/brahoo/PycharmProjects/odoo/odoo-12.0/odoo/osv/expression.py", line 673, in __init__
    self.parse()
  File "/Users/brahoo/PycharmProjects/odoo/odoo-12.0/odoo/osv/expression.py", line 854, in parse
    raise ValueError("Invalid field %r in leaf %r" % (left, str(leaf)))
```

alexey-pelykh: complains about `active` field?

CasVissers-360ERP: What I can read from the traceback yes. But seems to working from time to time too. The above traceback I got in a clean DB with demo data loaded. Only installed this new module.

pedrobaeza: No need to add `('active', '=', True)`, as that's the default behavior.

@CasVissers if you are not able to reproduce it on runbot, then it can be due to outdated code or any bad interaction

CasVissers-360ERP: Consider closed.

alexey-pelykh: Updated the code to not mention `active`

alexey-pelykh:
I don't know why codecov on patch includes completely unrelated files as part of PR

mail_activity_reminder/models/mail_activity.py
```diff
@@ -0,0 +1,98 @@
...
+
+            reminder = (local_deadline - local_today).days
+            if reminder not in reminders:
+                continue
+
+            if activity.last_reminder \
+                    and activity.last_reminder >= local_today:
+                continue
+
+            activities_to_remind |= activity
```
simahawk: please use a set of ids and browse all at once at the end instead of recordset intersection

mail_activity_reminder/models/mail_activity.py
```diff
@@ -0,0 +1,98 @@
...
+            activity.last_reminder = local_today
+
+    @api.multi
+    def action_remind(self):
+        IrModel = self.env['ir.model']
+        MailThread = self.env['mail.thread']
+        message_activity_assigned = self.env.ref(
+            'mail.message_activity_assigned'
+        )
+        utc_now = self.env.context.get(
```
simahawk: Used 3 times already: worth to move this to `_reminder_utc_now` method?

mail_activity_reminder/tests/test_mail_activity_reminder.py
```diff
@@ -0,0 +1,197 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from dateutil.relativedelta import relativedelta
+from pytz import timezone, UTC
+
+from odoo import fields
+from odoo.tests import common
+
+
+class TestMailActivityReminder(common.TransactionCase):
```
simahawk: pls use a SavepointCase and move all global setup to setUpClass instead of setUp.
Also:
* disable tracking globally + reset pwd as you create users, w/ `cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True, no_reset_password=True))`
* call `self.MailActivityType.create` on setUpClass and in tests change only the field you need
* for testing dates, instead of using `mail_activity_reminder_now` freeze the time (unless you think `mail_activity_reminder_now` is something you plan to use functionally and not only for testing)

mail_activity_reminder/models/mail_activity.py
```diff
@@ -0,0 +1,98 @@
...
+            'mail_activity_reminder_now',
+            fields.Datetime.now()
+        ).replace(tzinfo=UTC)
+        for activity in self:
```
simahawk, line 119: would be nice to group such notifications by partner and send only one notification.
Could you add a TODO?

simahawk (changes requested):
nice module Alexey, thanks. LG but I have few remarks ;)

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: please elaborate a bit more, I didn't get it

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: that feels a bit too much, as it does not have any specific logic except for giving ability to override value for testing

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: Added to roadmap

mail_activity_reminder/models/mail_activity.py
simahawk:
```py
to_remind_ids = []
....
to_remind_ids.append(activity.id)
...
return self.browse(to_remind_ids)
```

mail_activity_reminder/models/mail_activity.py
simahawk: then see my comment on tests ;)

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: won't that be less readable and more work? instead of working with recordsets?

simahawk: less readable? Why? In any case it boosts performance. Keep it in mind every time you think about using recordsets ;)

alexey-pelykh: but I already have all the recordset, why making a `browse` would be faster? It would be great to get a link to code with the "why it's bad to use recordsets" part :D

simahawk: no, you don't have the recordset. You create an empty one at line 17. The "why" is: it's slower :wink: 
We knew it since a while but Guewen recently did a test, see here https://twitter.com/guewenb/status/1196775451739459584

alexey-pelykh: Got it, curious unexpected behaviour. I've updated code a bit differently

mail_activity_reminder/models/mail_activity.py
```diff
@@ -0,0 +1,84 @@
...
+class MailActivity(models.Model):
+    _inherit = 'mail.activity'
+
+    last_reminder = fields.Date()
+
+    @api.model
+    def _get_activities_to_remind(self):
+        utc_now = fields.Datetime.now().replace(tzinfo=UTC)
+
+        def _should_remind(activity):
```
simahawk: don't hide this inside the method: move it to an instance method (then don't pass `activity` and use `self)

alexey-pelykh: also true

requirements.txt
```diff
@@ -1,3 +1,4 @@
 # mail_inline_css
 premailer
 extract-msg
+freezegun==0.3.11
```
simahawk: I know @pedrobaeza will complain about this :wink:

mail_activity_reminder/models/mail_activity.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class MailActivity(models.Model):
+    _inherit = 'mail.activity'
+
+    last_reminder = fields.Date()
+
+    @api.model
+    def _get_activities_to_remind(self):
+        return self.search([]).filtered(lambda x: x._should_be_reminded())
```
simahawk: in a ctx whereas activities are used heavily, loading all activities every time could be an issue, no?
Maybe worth to add a TODO for future refactoring?

simahawk (approved):
LG

requirements.txt
alexey-pelykh: With latest freezegun it just fails with some voodoo errors :) And since it's first time I'm using freezegun, I have no idea what those were about. @pedrobaeza are you going to complain? :)

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: Sorry, didn't get the issue. This code is called only via cron

requirements.txt
pedrobaeza: Well, firstly I would complain why you need an extra library for such basic feature...

alexey-pelykh: @simahawk requested to use it for substituting the "current time" in tests, previously it was done using context

mail_activity_reminder/models/mail_activity.py
pedrobaeza: He is saying that a cron that loads all activities is too much. Can you restrict the search to mimic the `_should_be_reminded` method? Maybe it's worth to precompute in a stored computed field a field like `next_reminder_date` or similar.

simahawk: exactly. I bet this cron won't run once a day. Since it's going to load potentially thousands or records - if not more - we should find a way to limit the impact.
I don't want to block it because we can consider this an alpha stage but we should think about the improvement ASAP IMO.

requirements.txt
pedrobaeza: Yeah, and that's why I don't like to use a library only for testing and for this specific thing, just as it happened with one of his modules. Please revert to other mechanism. I remember that Odoo has added a mechanism to manipulate dates obtained by `fields.Date/Datetime.today/now`, but I would say this is for v13.

mail_activity_reminder/models/mail_activity.py
pedrobaeza: Taking into account that activities are removed when done (at least for now - in v14 there are plans to keep them), the scope is limited, but any way, crons should be as faster as possible.

requirements.txt
simahawk: @alexey-pelykh was a joke because Pedro complained about adding the same dep in another repo. I'm totally in favor of it because it's pointless to replicate the same patch all around.
And many times the patches do not work like freezegun because it does it right and everywhere, plus it has some nice features.
IMO it should be included in our tests toolset and recommended as the right way to tests w/ dates. Then we can remove all the hacks that have been made (eg: ctx key for tests only).

pedrobaeza: I don't like a library only for tests...

alexey-pelykh: But that library is only for tests, and QUnit is only for tests. I do agree that including library only for tests in every repo is a bit silly, yet injecting code in _production_ to achieve the same is as silly to me. So in this case I'd vote for actually adding this library to the test tools and to the guidelines "how to simulate datetime"

P.S. I have no idea about what's changed in v13

alexey-pelykh: Yet in any case, please make up your minds :) Changing back and forth is a bit depressing

pedrobaeza: Why don't you use `context_today` https://github.com/odoo/odoo/blob/939bca38212d5eba830bbf70d23f500ace578b1a/odoo/fields.py#L1666 and simply pass the call with `timestamp` in a context provided in the test? No need of voodoo or library.

alexey-pelykh: That's how it was done before, almost. A ctx variable defaulting to `now()` (and I need datetime not date). As I've said, there are opposite requests to change the PR: one to use freezegun and don't workaround with context, other one to workaround with context and not to use library. Once @pedrobaeza and @simahawk are in agreement how to proceed, I'd gladly adopt the OCA way :)

pedrobaeza: I have it totally clear: this is not a workaround, it's the way Odoo intends to force other timestamps than now, so we should use it.

alexey-pelykh: Technically, there's no method that would return UTC now or timestamp if provided in `fields.Datetime`, so I'll have to use:
```
self.env.context.get('mail_activity_reminder_now', fields.Datetime.now())
```
which is kind-a workaround, and is not as universal as replacing what `now` returns. Yet definitely is more "in-place" than using library. @simahawk do you agree in a non-blocking way with Pedro?

pedrobaeza: What I suggest is the following in your code (inexact code):
```python
fields.Datetime.context_timestamp(timestamp=self.env.context('override_timestamp'))
```
and in tests, pass that context:
```python
model.with_context(override_timestamp=x).call_cron()
```

alexey-pelykh: I got that, it was almost implemented like that, but context_timestamp is useless for me, so it's similar to ```self.env.context.get('mail_activity_reminder_now', fields.Datetime.now())```

pedrobaeza: OK, that is perfectly fine:

- no extra libraries
- no overhead
- almost the same code as adding the freezegun stuff
- no extra knowledge needed

simahawk: IMO the point is that is silly to have code that is needed only for tests.
And in EVERY place where you need the date to match yours, you'll have to take care of it.

@pedrobaeza how is this different from using openupgradelib for JUST running an upgrade step that can be done w/ pure odoo code? You do it because it's handy and it avoids copy/pasting things all around. I understand it. Here is the same.

"no extra knowledge needed" is wrong: you must know that to test that module you must pass that ctx key and every time someone will read the code he/she will try to figure out why you need that thing. Furthermore, if this is the pattern you establish, you'll have ppl adapting it to their use cases which will make it every time a bit different to understand.

If you use freezegun you don't care. You test always in the same way.

Moreover, tests are more clear: if the test is decorated w/ `@freeze_time(...)` you know since the beginning at glance what you are going to do.

Also, maybe not the case here, but if someone extends the module and does something related to dates he/she will have to replicate again the same trick just because you make it functional.

My $0.02 :)

@sbidoul @lmignon @sebastienbeau @guewen can you share your thoughts regarding this?
Would be nice to solve this once for all :pray:

sbidoul: I quickly skimmed through this thread. So the problem is about adding a python lib dependency that is necessary for testing only?

In an ideal world Odoo should have a manifest key for test dependencies. Or `python-tests` in `external_dependencies`. Or something like that. Did someone request that upstream yet?

Now MQT recently gained a [new way](https://github.com/OCA/maintainer-quality-tools/blob/9423a4623e10ee4e91decb38c7eb546768fb555b/travis/travis_install_nightly#L119-L134) to discover dependencies automatically, and incidentally has a mechanism to declare test dependencies at the repo level. (note this will ignore `requirements.txt` and `oca_dependencies.txt`)

So in this case you can add `freezegun` to `test-requirements.txt` and set `MQT_DEP=PIP` in `.travis.yml`.

lmignon: Personally, I find freezegun very useful to manage the dates into our unittests. I know it's an extra dependency just required for tests but it's not the first one and IMO it add a plus value. I am not in favour of having to write specific code into the business logic used only by the tests.  Even if the intention is good, this type of code could be interpreted as part of the API of the module. We are going to see people using this code in their development when we must ban the use of the context as a means of passing arguments. For all of this, I'm in favour of using freezegun.

pedrobaeza: OK, proceed that way

alexey-pelykh: So, that way being...?

pedrobaeza: The library one

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: Refactored to rely on SQL filtering

rafaelbn (approved):
:+1:

mail_activity_reminder/models/mail_activity.py
```diff
@@ -0,0 +1,132 @@
...
+                continue
+            activity.next_reminder = next_reminder_local \
+                .astimezone(UTC) \
+                .replace(tzinfo=None)
+
+    @api.multi
+    def action_notify(self):
+        super().action_notify()
+        utc_now = fields.Datetime.now().replace(tzinfo=UTC)
+        for activity in self.filtered(lambda x: not x.last_reminder_local):
```
simahawk: this way you loop twice. Check last_reminder_local in the loop and continue if not set

mail_activity_reminder/models/mail_activity.py
```diff
@@ -0,0 +1,132 @@
...
+    def _process_reminders(self):
+        activities = self._get_activities_to_remind()
+        activities.action_remind()
+        return activities
+
+    @api.multi
+    def _should_be_reminded(self):
+        """Hook for extensions"""
+        self.ensure_one()
+        if self.next_reminder > fields.Datetime.now():
```
simahawk: this IF is already taken into account in the domain now

mail_activity_reminder/models/mail_activity.py
```diff
@@ -0,0 +1,132 @@
...
+    @api.model
+    def _get_activities_to_remind_domain(self):
+        """Hook for extensions"""
+        return [('next_reminder', '<=', fields.Datetime.now())]
+
+    @api.model
+    def _get_activities_to_remind(self):
+        return self \
+            .search(self._get_activities_to_remind_domain()) \
+            .filtered(lambda activity: activity._should_be_reminded())
```
simahawk: here you will loop twice as well, because later in `action_remind` you loop already on each record. 
I'd move the call to `_should_be_reminded` inside the loop of `action_remind`.

simahawk (changes requested):
Much better Alexey, tnx! Some small remarks left.

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: Why in this case I loop twice?

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: That domain is optimization, so for clarity this method should perform a full check, not assuming "oh everything else is checked elsewhere"

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: I'd prefer not doing that, to keep checks from actions a bit separate.

mail_activity_reminder/models/mail_activity.py
simahawk: because filtered loops once then your loop does it again.
As a rule of thumb: never use filtered if you don't need the filtered recordset as-is and you don't loop on it right after.

mail_activity_reminder/models/mail_activity.py
simahawk: fine

mail_activity_reminder/models/mail_activity.py
simahawk: as you prefer, but you loop twice for nothing (see my comment above on filtered).

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: aah, @pedrobaeza usually said to move filtering to filtered if possible, if I got him correctly

pedrobaeza: What I meant is to do `filtered` instead of list comprehensions or pure Python `filter`, but as @simahawk says, here you don't need to read whole recordset at the beginning if the loop is going to be interrupted and iterate twice. Any way, a performance note: in regular cases, whole recordset values are prefetched the same first time you accessed one field through dot notation, so that's not a game changer (or you explicitly put to not prefetch for improving that bits).

alexey-pelykh: so, how would it better in this case? move filtering inside loop?

mail_activity_reminder/models/mail_activity.py
alexey-pelykh: Yep, but values are going to be reused due to being prefetched, so I don't really get the harm of it.

alexey-pelykh:
@simahawk please update your review and mark the blocking parts

mail_activity_reminder/models/mail_activity.py
simahawk:
```py
        for activity in self:
            if activity.last_reminder_local:
                continue
```
as simple as this

alexey-pelykh: > Any way, a performance note: in regular cases, whole recordset values are prefetched the same first time you accessed one field through dot notation

If when `last_reminder_local` is first accessed, entire recordset of `activity` is fetched, and that's done for every item in self, what's the difference? O(n) here and O(n) there, same number of entries are accessed and thus prefetched.

lmignon: @alexey-pelykh The difference is that the filtering is done into the same loop as the rest of the logic...

`self.filtered(lambda x: not x.last_reminder_local)` -> 1 loop
` for activity in ` -> 1 loop
=> `for activity in self.filtered(lambda x: not x.last_reminder_local)`  -> 2 loops

With the @simahawk proposal you only have 1 loop: `for activity in self: `

alexey-pelykh: Done

alexey-pelykh:
@simahawk refactored to move even more to SQL query

simahawk (approved):

alexey-pelykh:
Let's merge :D

mail_activity_reminder/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+        'Brainbean Apps, '
+        'Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'installable': True,
+    'application': False,
+    'summary': 'Reminder notifications about planned activities',
+    'depends': [
+        'mail',
+    ],
+    'external_dependencies': {
```
simahawk: This should not stay here as is not a functional dependency (I wish we have test only dependencies here :/ ).
If you want to not include this lib in your prod build you should be able to. This prevents it.
MQT=PIP + test-requirerements.txt allows it.

alexey-pelykh: > Now MQT recently gained a new way to discover dependencies automatically, and incidentally has a mechanism to declare test dependencies at the repo level. (note this will ignore requirements.txt and oca_dependencies.txt)

> So in this case you can add freezegun to test-requirements.txt and set MQT_DEP=PIP in .travis.yml.

Did I get it right that having `test-requirerements.txt` will completely ignore `requirements.txt` and `oca_dependencies.txt` and leading to quite an unpleasant things?

simahawk: if your dependencies are properly declared, you'll get the same result because you'll have functional dependencies installed via PIP and test deps installed via requirements.
In any case, even if you don't want to use MQT=PIP here, you should remove the dependency.

alexey-pelykh: What I want is to get over this module finally :D

alexey-pelykh: Added, now it's not working. What did I do wrong?

alexey-pelykh: Oh, misspelled the file name

alexey-pelykh: Done! Can we please merge it now? :)

test-requirements.txt
```diff
@@ -0,0 +1 @@
+freezegun==0.3.11
```
lmignon: Do we really need to pin a specific version? If a minimal version is required, the dependency must be delcared as `freezegun>=0.3.11`  otherwise we should remove the version part of the dependency declaration `freezegun`

lmignon (changes requested):
@alexey-pelykh One more comment about the way to declare a dependency.  (Sorry :smirk: )

test-requirements.txt
alexey-pelykh: More recent `freezegun` fails

lmignon: in such a case you can specify the dependency as follow `freezegun<=0.3.11` and add a comment to explain that tests are failing with freezegun> 3.0.11.
```
# Tests fails with freezegun > 3.0.11
freezegun <= 3.0.11
```
@simahawk Are you aware of some problems with the latest version of `freezegun`.

alexey-pelykh:
```
"/home/travis/build/brainbeanapps/social/setup/mail_activity_reminder/odoo/addons/mail_activity_reminder/tests/test_mail_activity_reminder.py", line 6, in <module>
    from freezegun import freeze_time
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/freezegun/__init__.py", line 9, in <module>
    from .api import freeze_time
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/freezegun/api.py", line 26, in <module>
    _EPOCHTZ = datetime.datetime(1970, 1, 1, tzinfo=dateutil.tz.UTC)
AttributeError: module 'dateutil.tz' has no attribute 'UTC'
```

simahawk: nope AFAIK

simahawk:
@lmignon good for you?
@alexey-pelykh thanks for the nice contrib and for the patience you've put here :wink:

alexey-pelykh:
Not related: https://travis-ci.org/github/OCA/social/jobs/683945644

alexey-pelykh:
@pedrobaeza any ideas why https://travis-ci.org/github/OCA/social/jobs/684150152 ?

sbidoul:
Travis config was quite outdated. Please retry when https://github.com/OCA/social/pull/542 is in.

alexey-pelykh:
Done, let's merge it, finally :D

PR #511, opened 2020-03-05
[12.0][MIG] mail_restrict_follower_selection

JoanSForgeFlow:
**mail_restrict_follower_selection** migration to v12.

cc ~ @ForgeFlow @LoisRForgeFlow


jarroyomorales:
Yesterday we were making the same migration but we used a different strategy
https://github.com/tegin/social/commit/4869d23e8afcfc947cebb37ad5a7ed8cf1b1e409

What do you think is better?

JoanSForgeFlow:
I think your strategy is better, you are editing the main function and I am editing a shortcut of this function, so I assume yours is going to be more stable for all cases. I am closing this PR.
Thanks for your review!

PR #498, opened 2020-02-10
[13.0][MIG] mass_mailing_partner: Migration to 13.0

manuelcalerosolis:
This is the migration to v 13.0 module mass_mailing_partner

@Tecnativa TT19777

PR #518, opened 2020-03-10
[13.0][MIG] mass_mailing_list_dynamic: Migration to 13.0

jaredkipe:
Requires #498

There is an issue in the tests for the @api.constrains (around line 82 in tests files)
I can log the `mailing.contact` before entering the constrains and during the constrains and the `list_ids` and `subscription_list_ids` are present before entering the method, and empty in the method itself.  I've flushed various things (flushing on the list is needed now as the count of subscribers has no @api.depends to invalidate the cache) and haven't been able to figure it out, any advice would be helpful.


pedrobaeza:
@ernestotejeda @manuelcalerosolis please review this TT22860

pedrobaeza:
@jaredkipe please rebase after merging the dependency

jaredkipe:
@pedrobaeza all done with rebase

Any ideas on testing the @api.constrains ?

pedrobaeza:
I haven't checked. My colleagues may say when reviewing.

ernestotejeda:
Hello @jaredkipe , it has to do with this https://github.com/OCA/social/blob/5c87e6b7e2b6e566ed09e0c688cce8a719faec22/mass_mailing_partner/models/mailing_contact.py#L67. Take a look at the redefinition of the create method too that is in that same file a few lines above. Maybe those two method should be reviewed again.

jaredkipe:
Thanks that is a really great observation!

I was concerned with the constraint in this module and didn't think to look up into dependencies for something like this... probably because I thought the constraint would trigger off of write itself, which of course in this case doesn't happen (the super/base implementation) until the new values are already calculated.

mass_mailing_list_dynamic/models/mail_mass_mailing_contact.py
```diff
@@ -9,13 +9,15 @@
...
-                    "synchronized list. Change its sync method or execute "
-                    "a manual sync instead."))
+        for one in self:
+            if any((
+                list_id.dynamic and list_id.sync_method == "full"
+            ) for list_id in one.mapped('list_ids')):
+                raise ValidationError(
+                    _("""Cannot edit manually contacts in a fully synchronized list.
+                        Change its sync method or execute
+                        a manual sync instead."""))
```
yajo: If you don't touch this string, we will not have to translate it again.

yajo (changes requested):

simahawk:
@jaredkipe any chance to move fwd?

rafaelbn:
Hello @jaredkipe ,

Please could you update this PR including https://github.com/OCA/social/pull/571

Thank you!!

cc @chienandalu

mass_mailing_list_dynamic/models/mail_mass_mailing_contact.py
jaredkipe: I'm not editing things for giggles, if this changed it would be from running `black` or other linting/clean up as prescribed by the documentation.

jaredkipe: Probably was an error on my part, however this file was renamed to  `mailing_contact.py` and does not have this problem.

jaredkipe:
@rafaelbn Rebased, cherry pick wasn't clean because of other changes, including that this method is renamed in 13

mass_mailing_list_dynamic/models/mailing.py
```diff
@@ -1,4 +1,5 @@
 # Copyright 2017 Tecnativa - Jairo Llopis
+# Copyright 2020 Hibou Corp. - Jared Kipe
```
chienandalu, line 2: Why do you author this change and commit? :thinking:

chienandalu (changes requested):
Works as expected. Anyway, you should fix the commit history respecting authorship, please

jaredkipe:
I have re-done the history to put the [FIX] #571 , with the correct author before the migration.

However, it came to my attention after I wrote this migration that I should be adding myself to the copyright on materially changed files. I understand why it looked hinky, so I've done that in the migration commit now.

Also, it worked "as expected" before yesterday.  There was no reason that I was aware of that this wouldn't have been merged before the sudden need to put a 6 hour old fix into a 4 month old migration, which would have been a smoother process for everyone.

@chienandalu would you mind checking to see if 13.0 `mass_mailing_partner` works as expected, and consider merging this 4 month old [FIX] #523

chienandalu:
> I have re-done the history to put the [FIX] #571 , with the correct author before the migration.

Thanks :)

> However, it came to my attention after I wrote this migration that I should be adding myself to the copyright on materially changed files. I understand why it looked hinky, so I've done that in the migration commit now.

Well, not really. The ettiquette would be the following:

- Standard migration tweaks (like renames of fields, models, views...) don't qualify for authorship
- Typos or translations: contributor
- Tests adaptations to the new version: depending on the entity of the change.
- Functional changes to adapt to new version: qualify you as author.
- Bug fixes: qualify you as author

Of course, for everyone of these points, you're considered a contributor and you can put your credentials in the CONTRIBUTORS.rst

> Also, it worked "as expected" before yesterday. There was no reason that I was aware of that this wouldn't have been merged before the sudden need to put a 6 hour old fix into a 4 month old migration, which would have been a smoother process for everyone.

Well, the thing is it didn't, as it's missing a fix that was detected and fixed 6 hours ago :slightly_smiling_face: 

> @chienandalu would you mind checking to see if 13.0 `mass_mailing_partner` works as expected, and consider merging this 4 month old [FIX] #523

ping me there and I check it

pedrobaeza:
Correction: bug fixes don't usually qualify for co-authorship.

mass_mailing_list_dynamic/wizards/mailing_load_filter.py
```diff
@@ -1,12 +1,13 @@
...
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import fields, models
 
 
-class MassMailingLoadFilter(models.TransientModel):
-    _name = "mail.mass_mailing.load.filter"
-    _description = "Mail Mass Mailing Load Filter"
+class MailingLoadFilter(models.TransientModel):
+    _name = "mailing.load.filter"
```
yajo, line 9: @pedrobaeza Please correct me if I'm wrong: this is OK and it doesn't need a migration script because it is a transient model and there's no problem if the previous table gets dumped and a new one is created. Right?

yajo (approved):

mass_mailing_list_dynamic/wizards/mailing_load_filter.py
pedrobaeza: Yes, transient models are not mandatory to be renamed.

pedrobaeza (approved):

PR #521, opened 2020-03-12
[ADD] mail_template_conditional_attachment

ivantodorovich:
Allows to add conditional attachments to email templates based on a domain

Usage:
1. Go to the corresponding Email Template record.
2. You should see the Conditional Attachments under Advance Settings
3. Create a new conditional attachment, and select the domain condition.

attch-config
attch-domain
attch-attachments


codeagencybe:
Very cool :) 
Also v13?

ivantodorovich:
> Very cool :)
> Also v13?

Sure :) after it's merged we'll migrate to 13.0 ^^
I would appreciate your review :smile:

mail_template_conditional_attachment/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "maintainers": [
+        "ivantodorovich"
+    ],
+    "license": "AGPL-3",
+    "depends": [
+        "mail",
+    ],
+    "data": [
+        "views/mail_template.xml",
+        "security/ir.model.access.csv",
```
mamcode, line 19: I think that is a best practice place the security files before all XML files. Can you load first "security/ir.model.access.csv" ?

mail_template_conditional_attachment/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_mail_template_conditional_attachment_user,mail.template.conditional.attachment.user,model_mail_template_conditional_attachment,base.group_user,1,1,1,1
```
mamcode, line 2: I think is better:
`access_mail_template_conditional_attachment_user,mail.template.conditional.attachment.user,model_mail_template_conditional_attachment,base.group_user,1,1,1,0`
`access_mail_template_conditional_attachment_system_user,mail.template.conditional.attachment.system.user,model_mail_template_conditional_attachment,base.group_system,1,1,1,1`

mamcode (changes requested):

mail_template_conditional_attachment/security/ir.model.access.csv
ivantodorovich: Hmmm. Not sure I agree with this one.
Users are able to edit mail.templates, why wouldn't they be able to unlink conditional attachments? It's a field in the template

mail_template_conditional_attachment/__manifest__.py
ivantodorovich: Thanks for the review! 
Done here https://github.com/OCA/social/pull/521/commits/bb97067a14ea2708862b9ce9292f7fb2bb557cb1 :+1:

mail_template_conditional_attachment/security/ir.model.access.csv
mamcode: I think that normally an admin user is who write the email templates, in fact I only would leave access rights of read for users of base.group_user and all access rights to users of base.group_system but in this suggestion of changes I'm following the concept that addon `mail` applies for access rights in model mail.template.

mamcode:
@ivantodorovich This module should only send attachments that match with the domain right? Am I getting it right ?

ivantodorovich:
> @ivantodorovich This module should only send attachments that match with the domain right? Am I getting it right ?

that's correct :)

ivantodorovich:
@simahawk would this be interesting to you?

mail_template_conditional_attachment/models/mail_template.py
```diff
@@ -0,0 +1,38 @@
...
+
+    conditional_attachment_ids = fields.One2many(
+        'mail.template.conditional.attachment',
+        'mail_template_id',
+        string='Conditional Attachments',
+    )
+
+    @api.multi
+    def generate_email(self, res_ids, **kwargs):
+        """ Overload this method to attach conditional attachments """
```
simahawk: Avoid overriding the docstring because you wipe the original. Use a comment instead.

mail_template_conditional_attachment/models/mail_template.py
```diff
@@ -0,0 +1,38 @@
...
+        # Placed before super(), so we force multi_mode=True there
+        multi_mode = True
+        if isinstance(res_ids, pycompat.integer_types):
+            res_ids = [res_ids]
+            multi_mode = False
+
+        results = super().generate_email(res_ids, **kwargs)
+
+        # Add conditional attachments
+        if self.conditional_attachment_ids:
```
simahawk, line 29: this method is multi: here it would be broken if `len(self) > 1`.
I'd do:
```py
cond_attachments = {
    x.id: conditional_attachment_ids.get_attachment_ids(x.id)
    for x in self if x.conditional_attachment_ids
}
```
then use this mapping in the loop (if valued at all).

mail_template_conditional_attachment/models/mail_template_conditional_attachment.py
```diff
@@ -0,0 +1,60 @@
...
+    _name = 'mail.template.conditional.attachment'
+    _description = 'Mail Template Conditional Attachment'
+
+    name = fields.Char('Description', required=True)
+
+    mail_template_id = fields.Many2one(
+        'mail.template',
+        ondelete='cascade',
+        required=True,
+        index=True,
```
simahawk: do  you really need an index here?

mail_template_conditional_attachment/models/mail_template_conditional_attachment.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+
+    attachment_ids = fields.Many2many(
+        'ir.attachment',
+        string='Attachments',
+    )
+
+    model_name = fields.Char(related='mail_template_id.model_id.model')
```
simahawk: readonly?

mail_template_conditional_attachment/models/mail_template_conditional_attachment.py
```diff
@@ -0,0 +1,60 @@
...
+    def _get_eval_context(self):
+        return {
+            'datetime': datetime,
+            'dateutil': dateutil,
+            'time': time,
+            'uid': self.env.uid,
+            'user': self.env.user,
+        }
+
+    def _check_condition(self, res_id):
```
simahawk, line 41: res_id is? docstring or better var naming ;)

mail_template_conditional_attachment/models/mail_template_conditional_attachment.py
```diff
@@ -0,0 +1,60 @@
...
+        self.ensure_one()
+        if self.filter_domain:
```
simahawk: not a blocker but in such cases I usually do:
```py
if not self.filter_domain:
    return True
....
```
makes clear that you don't do anything in that case and avoid indentation of the whole code.
Feel free to ignore ;)

mail_template_conditional_attachment/models/mail_template_conditional_attachment.py
```diff
@@ -0,0 +1,60 @@
...
+            domain = [('id', '=', res_id)]
+            domain += safe_eval(self.filter_domain, self._get_eval_context())
+            model = self.env[self.model_name]
+            return bool(model.search_count(domain))
+        else:
+            return True
+
+    @api.multi
+    def get_attachment_ids(self, res_id):
```
simahawk:
```suggestion
    def get_attachments(self, res_id):
```
you return records, not ids

mail_template_conditional_attachment/security/ir.model.access.csv
simahawk: I'd use the same permissions as per mail.template -> simple user cannot unlink

mail_template_conditional_attachment/views/mail_template.xml
```diff
@@ -0,0 +1,38 @@
...
+                            <field name="filter_domain"/>
+                            <field name="attachment_ids"/>
+                        </tree>
+                        <form>
+                            <sheet>
+                                <div class="oe_title">
+                                    <label for="name" class="oe_edit_only"/>
+                                    <h1><field name="name" required="1"/></h1>
+                                </div>
+                                <group>
```
simahawk: name the group pls

simahawk (changes requested):
any test on your radar?

mail_template_conditional_attachment/models/mail_template.py
ivantodorovich: Actually this method is meant to be called only on one template. That's why there's `self.ensure_one()` at the beggining.

https://github.com/odoo/odoo/blob/12.0/addons/mail/models/mail_template.py#L422

The `multi_mode` mentioned before is because the template may generate more than 1 email. But the template will always be 1. It shouldn't be confused with `api.multi`

mail_template_conditional_attachment/models/mail_template.py
ivantodorovich: Got it :+1:

mail_template_conditional_attachment/models/mail_template_conditional_attachment.py
ivantodorovich: Done :+1:

mail_template_conditional_attachment/models/mail_template_conditional_attachment.py
ivantodorovich: Not really, removed

mail_template_conditional_attachment/models/mail_template_conditional_attachment.py
ivantodorovich: Related fields are readonly by default

https://github.com/odoo/odoo/commit/52a8ed3c0c72b21baf589bc9b91a030d9da661a8

mail_template_conditional_attachment/models/mail_template_conditional_attachment.py
ivantodorovich: done

ivantodorovich (commented):
.

mail_template_conditional_attachment/models/mail_template.py
simahawk: haha my bad, I totally missed the ensure_one call :sweat_smile:

mail_template_conditional_attachment/security/ir.model.access.csv
simahawk: @ivantodorovich any feedback here?

ivantodorovich: Conditional attachments are basically a new field in mail.templates.

If simple users are able to write on mail.templates, they should be able to create/write/remove conditional attachments. Otherwise they won't be able to edit this field.

PR #523, opened 2020-03-18
[13.0][FIX] mass_mailing_partner: Unsubscribe should not unlink the subscription.

jaredkipe:
Unsubscribe should not unlink the subscription.

Before, on the backend if you check the `opt_out` checkbox and saving on a `mailing.contact`, the entire row (`mailing.contact.subscription`) would be unlinked.  Combined with `mass_mailing_list_dynamic` this means the contact would just be re-subscribed over and over.

After, check `opt_out` will fill `unsubscription_date` and the row will stay.  Tests in `mass_mailing_list_dynamic` pass. 

Needed for https://github.com/OCA/social/pull/518

### DIFF
--- a/mass_mailing_partner/README.rst
+++ b/mass_mailing_partner/README.rst
@@ -23,7 +23,7 @@ Link partners with mass-mailing
     :target: https://runbot.odoo-community.org/runbot/205/13.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| |badge5| 
+|badge1| |badge2| |badge3| |badge4| |badge5|
 
 This module links mass-mailing contacts with partners.
 
--- a/mass_mailing_partner/models/mailing_contact.py
+++ b/mass_mailing_partner/models/mailing_contact.py
@@ -3,6 +3,7 @@
 # Copyright 2015 Javier Iniesta <javieria@antiun.com>
 # Copyright 2017 David Vidal <david.vidal@tecnativa.com>
 # Copyright 2020 Tecnativa - Manuel Calero
+# Copyright 2020 Hibou Corp.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import _, api, fields, models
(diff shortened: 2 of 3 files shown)


mass_mailing_partner/README.rst
```diff
@@ -88,6 +88,8 @@ Contributors
     * Victor M.M. Torres
     * Manuel Calero
 
+* `Hibou Corp. <https://hibou.io>`_
```
pedrobaeza: You should add this to readme/CONTRIBUTORS.rst instead

pedrobaeza (commented):
I don't get how changing from `False` to `[]` fix the problem you describe.

mass_mailing_partner/README.rst
jaredkipe: I have done so.  Could clarify why README.rst's still have this content? Like.. should the README.rst link to the readme/CONTRIBUTORS.rst (replacing the current list)?

jaredkipe (commented):
@pedrobaeza

I imagine the answer is as simple as because there is a big difference in how a O2m, M2m field interprets a write with `False` (unlink) from `[]` (empty command list).

This difference is new in 13, and you can try for yourself of that by doing something like this in Odoo12 vs Odoo13

```
# get a shell, note category_id is a M2m
p = env['res.partner'].search([('category_id', '!=', False)], limit=1)
assert p.category_id
p.write({'category_id': []})
assert p.category_id
p.write({'category_id': False})
assert p.category_id
```

You should get an assertion error in 13, but not 12.  If this shouldn't be the case, then Odoo has a bigger problem (I cannot find anywhere this is explicitly mentioned in the ORM docs...)

I got paranoid (could I have a long standing ORM bug that was fixed but not in my builds?!) and went and reproduced this on runbot as well.  If you can get to it in time, https://3630038-13-0-f1b455.runbot44.odoo.com/web?debug=1#id=14&action=171&model=res.partner&view_type=form&cids=1&menu_id=121 should have server actions on `res.partner` that you can use to ensure that writing False does in fact unlink. (named "Write Tags []" and "Write Tags False")

mass_mailing_partner/README.rst
pedrobaeza: Because the bot only regenerates it on merge (with extra commit). You can regenerate it locally and include the new file downloading OCA/maintainer-tools and running `oca-gen-addon-readme`, but OK if this has been added in the other file.

pedrobaeza (approved):

pedrobaeza:
Let's fast-track this until some time for unblocking the other PR:

/ocabot merge patch

PR #524, opened 2020-03-19
[IMP] mail_tracking: Store To recipients and omit aliases

Tardo:
- Don't lose 'TO' recipients without partner
- Don't show recipients that are defined as aliases

This need be ported to 13.0

cc @tecnativa TT22167


mail_tracking/models/mail_message.py
```diff
@@ -155,12 +166,34 @@ def tracking_status(self):
...
             }
         return res
 
+
+    @api.model
+    def _drop_aliases(self, mail_list):
+        aliases_list = []
+        for email in mail_list:
+            aliases_list.append(getaddresses([email])[0][1].split('@')[0])
+        aliases = self.env['mail.alias'].search([
```
pedrobaeza: This search for each message is totally under-performant. We should keep a cached set of aliases (directly the alias name, not the records) that we invalidate when an alias is created.

mail_tracking/models/mail_message.py
```diff
@@ -155,12 +166,34 @@ def tracking_status(self):
...
+            ('alias_name', 'in', aliases_list),
+        ])
+        for alias in aliases:
+            for mail in mail_list:
+                if alias.alias_name == mail.split('@')[0]:
+                    mail_list.remove(mail)
```
pedrobaeza: With the cached set
```suggestion
        if mail.split('@')[0] in alias_set:
            mail_list.remove(mail)
```

mail_tracking/models/mail_message.py
```diff
@@ -155,12 +166,34 @@ def tracking_status(self):
                     'isCc': True,
                 })
                 partner_trackings.append(tracking_unkown_values.copy())
+            for email in email_to_list:
```
pedrobaeza: We should avoid to duplicate similar logic:
```python
        for cc, lst in [(True, email_cc_list), (False, email_to_list)]:
            for email in lst:
                # If there is CC/To without partner
                tracking_unkown_values.update({
                    'recipient': email,
                    'partner_id': False,
                    'isCc': cc,
                })
                partner_trackings.append(tracking_unkown_values.copy())
```
The variable also contains a typo

mail_tracking/models/mail_thread.py
```diff
@@ -35,38 +35,53 @@ def message_post(self, *args, **kwargs):
         """
         new_message = super().message_post(*args, **kwargs)
         email_cc = kwargs.get('cc')
+        toWrite = {}
```
pedrobaeza: Don't user uppercasing in local Python variables

mail_tracking/models/mail_thread.py
```diff
@@ -35,38 +35,53 @@ def message_post(self, *args, **kwargs):
...
         if email_cc:
-            new_message.sudo().write({
+            toWrite.update({
```
pedrobaeza: Why not using a more compact form (both variable name and expression):
```suggestion
            vals['email_cc'] = email_cc
```

mail_tracking/models/mail_thread.py
```diff
@@ -35,38 +35,53 @@ def message_post(self, *args, **kwargs):
...
                 'email_cc': email_cc,
             })
+        # This is here because Odoo only store recipients that have a partner.
+        toWrite.update({
```
pedrobaeza: What happens if no `to` key. Are we forcing a writing in the object any way? Please take the performance into account always as this is a very low level operation.

mail_tracking/models/mail_thread.py
```diff
@@ -35,38 +35,53 @@ def message_post(self, *args, **kwargs):
...
+                email_extra_formated_list.extend(email_split_and_format(email))
+        email_extra_formated_list = set(email_extra_formated_list)
+        for mail in email_extra_formated_list:
+            email_parts = getaddresses([mail])[0]
             partner_id = record.message_partner_info_from_emails(
                 [email_parts[1]])[0].get('partner_id')
             if not partner_id:
-                record._message_add_suggested_recipient(
-                    res, email=cc, reason=_('Cc'))
+                aliases = AliasObj.search([
```
pedrobaeza: Again a non performant search

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -169,7 +169,7 @@ def test_email_cc(self):
             'login': 'sender-test',
         })
         message = self.recipient.sudo(user=sender_user).message_post(
-            body='<p>This is a test message</p>',
+            body=_('<p>This is a test message</p>'),
```
pedrobaeza: Why translated? This translation is not even loaded in POT file. If you have being testing in local, you must do it always in English.

pedrobaeza (changes requested):

mail_tracking/models/mail_thread.py
Tardo: Can exists a message without a 'to' defined?

mail_tracking/tests/test_mail_tracking.py
Tardo: If not use it travis shows: https://travis-ci.org/github/OCA/social/jobs/664560869#L319

Tardo:
Changes done

mail_tracking/models/mail_thread.py
pedrobaeza: Yeah, indeed, but my problem is to perform an extra write each time. Isn't any `prepare` values function or pass this by context as intercept it in the write/create method for injecting it in the dictionary before dumping it to DB?

mail_tracking/tests/test_mail_tracking.py
pedrobaeza: That's only a warning, not an error, and in this case, it's not correct. We should fix it in MQT, but any way, please remove it and you can flag the line with `# pylint: disable=C8107` for avoiding the warning message.

manuelcalerosolis (approved):

mail_tracking/models/mail_message.py
```diff
@@ -86,6 +88,11 @@ def _get_error_description(self, tracking):
 
     def tracking_status(self):
         """Generates a complete status tracking of the messages by partner"""
+        db_aliases = set(map(
+            lambda x: x['alias_name'],
+            self.env['mail.alias'].search_read([
+                ('alias_name', '!=', False),
+            ], ['alias_name'])))
```
yajo: I'll try to help save some loops here because I see big chances of performance impact.

```suggestion
        catchall_domain = self.env['ir.config_parameter'].get_param('mail.catchall.domain')
        db_aliases = []
        if catchall_domain:
            db_aliases = [
                "%s@%s" % (rec.alias_name, catchall_domain)
                for self.env['mail.alias'].search([
                    ('alias_name', '!=', False),
                ])
            ]
```

mail_tracking/models/mail_message.py
```diff
@@ -94,11 +101,20 @@ def tracking_status(self):
             trackings = self.env['mail.tracking.email'].sudo().search([
                 ('mail_message_id', '=', message.id),
             ])
-            # Get Cc recipients
+            # String to List
             email_cc_list = email_split(message.email_cc)
+            email_to_list = email_split(message.email_to)
+            # Remove aliases recipients
+            self._drop_aliases(email_cc_list, db_aliases)
+            self._drop_aliases(email_to_list, db_aliases)
```
```suggestion
```

mail_tracking/models/mail_message.py
```diff
@@ -94,11 +101,20 @@ def tracking_status(self):
...
+            # Search related partners recipients
             if any(email_cc_list):
                 partners |= partners.search([('email', 'in', email_cc_list)])
+            if any(email_to_list):
+                partners |= partners.search([('email', 'in', email_to_list)])
```
```suggestion
            partners |= partners.search([
                ('email', 'in', email_cc_list + email_to_list),
                ('email', 'not in', db_aliases),
            ])
```

mail_tracking/models/mail_message.py
```diff
@@ -117,50 +133,67 @@ def tracking_status(self):
...
             }
         return res
 
+    @api.model
+    def _drop_aliases(self, mail_list, alias_set):
+        for email in mail_list:
+            mail_alias = getaddresses([email])[0][1].split('@')[0]
+            if mail_alias in alias_set:
+                mail_list.remove(email)
+
```
```suggestion
```

mail_tracking/models/mail_thread.py
```diff
@@ -30,43 +30,58 @@ def _get_failed_message_domain(self):
...
+        """Adds email 'extra' recipients as suggested recipients.
 
         If the recipient has a res.partner, use it.
         """
         res = super().message_get_suggested_recipients()
+        db_aliases = set(map(
+            lambda x: x['alias_name'],
+            self.env['mail.alias'].search_read([
+                ('alias_name', '!=', False),
+            ], ['alias_name'])))
```
yajo: This is the same but more readable:

```suggestion
        db_aliases = set(
            x['alias_name'] for x in
            self.env['mail.alias'].search_read([
                ('alias_name', '!=', False),
            ], ['alias_name']))
        )
```

mail_tracking/models/mail_thread.py
```diff
@@ -30,43 +30,58 @@ def _get_failed_message_domain(self):
...
-                email_cc_formated_list.extend(email_split_and_format(email))
-        email_cc_formated_list = set(email_cc_formated_list)
-        for cc in email_cc_formated_list:
-            email_parts = getaddresses([cc])[0]
+            emails_extra = record.message_ids.mapped(field_mail)
+            for email in emails_extra:
+                email_extra_formated_list.extend(email_split_and_format(email))
+        email_extra_formated_list = set(email_extra_formated_list)
+        for mail in email_extra_formated_list:
+            email_parts = getaddresses([mail])[0]
```
yajo: Have you considered this?

```suggestion
        for local, domain in getaddresses(email_extra_formated_list):
```

mail_tracking/models/mail_thread.py
```diff
@@ -30,43 +30,58 @@ def _get_failed_message_domain(self):
...
             partner_id = record.message_partner_info_from_emails(
                 [email_parts[1]])[0].get('partner_id')
             if not partner_id:
-                record._message_add_suggested_recipient(
-                    res, email=cc, reason=_('Cc'))
+                if alias_set and mail.split('@')[0] not in alias_set:
```
yajo: Beware! `sales@customer.com` is not the same as `sales@myself.com`.

See suggestion above to use `mail.catchall.domain` ICP.

mail_tracking/models/mail_thread.py
```diff
@@ -30,43 +30,58 @@ def _get_failed_message_domain(self):
...
+                    record._message_add_suggested_recipient(
+                        suggestions, email=mail, reason=reason)
             else:
                 partner = ResPartnerObj.browse(partner_id, self._prefetch)
                 record._message_add_suggested_recipient(
-                    res, partner=partner, reason=_('Cc'))
-        return res
```
yajo, line 69: No return anymore?

yajo (changes requested):
Trying to help here a little bit :blush:

mail_tracking/models/mail_thread.py
Tardo: Not only more readable, this increase the performance! thx!

mail_tracking/models/mail_message.py
Tardo: Better than Odoo eh... we see the domain :P thx

mail_tracking/models/mail_message.py
Tardo: Can't avoid this :/... 
Without filter aliases here, we can print aliases when doesn't have a partner.

Tardo:
Really thanks for all suggestions @pedrobaeza and @Yajo ...

mail_tracking/models/mail_message.py
```diff
@@ -86,6 +88,16 @@ def _get_error_description(self, tracking):
 
     def tracking_status(self):
         """Generates a complete status tracking of the messages by partner"""
+        catchall_domain = self.env['ir.config_parameter'].sudo().get_param(
+            'mail.catchall.domain')
+        db_aliases = {}
+        if catchall_domain:
+            db_aliases = set(
+                "%s@%s" % (rec['alias_name'], catchall_domain)
+                for rec in self.env['mail.alias'].search_read([
```
pedrobaeza: As said, can we cached (through a method decorated with `orm_cache` in alias model) this alias list and invalidate this cache (I think it's `self.cache.clear()` in the corresponding model) when creating/modifying one alias? This search is still hammering a lot as it's done for each record.

mail_tracking/models/mail_thread.py
```diff
@@ -30,43 +30,62 @@ def _get_failed_message_domain(self):
...
+        ensure that this information its written into the mail.message record.
         """
-        new_message = super().message_post(*args, **kwargs)
-        email_cc = kwargs.get('cc')
-        if email_cc:
-            new_message.sudo().write({
-                'email_cc': email_cc,
-            })
-        return new_message
+        kwargs.update({
```
pedrobaeza, line 36: Adding these 2 keyword arguments doesn't include them in the `mail.message` resulting record. You have to pass them by context and override `mail.message` create method for injecting them on `vals` before calling `super`.

pedrobaeza (changes requested):

pedrobaeza:
Check also runbot problem

mail_tracking/models/mail_thread.py
Tardo: This works because the parent method do this: https://github.com/odoo/odoo/blob/12.0/addons/mail/models/mail_thread.py#L2047

** Tested sending a email and works.

pedrobaeza: OK, I see, great!

Tardo:
Changes done, @pedrobaeza please review

mail_tracking/models/mail_alias.py
```diff
@@ -0,0 +1,36 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        self.clear_caches()
+        return res
+
+    @api.multi
+    def write(self, vals):
+        res = super().write(vals)
+        if 'alias_name' in vals:
```
yajo, line 26:
```suggestion
        if 'alias_name' in vals or 'alias_domain' in vals:
```

yajo (approved):

mail_tracking/models/mail_alias.py
```diff
@@ -0,0 +1,36 @@
...
+
+from odoo import api, models, tools
+
+
+class MailAlias(models.Model):
+    _inherit = 'mail.alias'
+
+    @api.model
+    @tools.ormcache('email')
+    def is_email_address_in(self, email):
```
pedrobaeza: The cached method must return the list of aliases strings, not True/False according an email, and do the comparison outside. This way, the cache is not effective, as there can be thousands of possible email comparisons

pedrobaeza (changes requested):
Inline comment and attend @Yajo's suggestion

mail_tracking/models/mail_alias.py
Tardo: 'alias_domain' It's an computed field without 'inverse' & 'search'. So... if you change these domain you don't need restart odoo instance?

Tardo:
ormcache changes donde

mail_tracking/models/mail_alias.py
yajo: Oh OK, it was just by reading this code, I didn't know that.

yajo (approved):

pedrobaeza:
Linter fails

pedrobaeza (approved):

pedrobaeza:
/ocabot merge minor

Please migrate it to v13

PR #507, opened 2020-03-04
[FIX] mail_debrand: template with original terms

chienandalu:
Depending of the context the term could come translated or in the original term.
We should consider both. Also notice that if the translated term in this module
differs from the one in the template it won't be catched.

cc @Tecnativa TT22697

Issue #441, opened 2019-10-08
mail_debrand does not work if partner has different language, v12


gr22:
If the partner has different language than the user sending the mail then `_("Powered by")` would not translate to the partner's lang but the html of the template will so the condition `if powered_by not in html:` will fail

astirpe:
I confirm that this issue occurs, introduced by https://github.com/OCA/social/commit/c706391c2e1f6fc4224a79526d0adc789a43a623#diff-171036ef0b5f4a45ad3a3b313e1fbe0a

Would be nice to restore the old behavior, that was completely removing the "Powered by" part.

pedrobaeza:
@gdgellatly you were the author of it. Can you please check?

dndoanh:
@astirpe It works if you have translation file in that lang

astirpe:
@dndoanh I didn't test it with any .po file, but if I look at [this](https://github.com/OCA/social/tree/12.0/mail_debrand/i18n) I think that you are implicitly confirming the issue for all the languages other than English, right?

gdgellatly:
@gr22 @astirpe I'm a bit confused, it seems there are differing opinions here on what the bug is.

The actual report seems to imply all that is needed is to ensure the translation is performed in the partners language, but I was pretty sure that was already passed in context.

The old way relied on XML navigation, which kind of broke when they changed all the templates for this version, and it also isn't standardised as a single template and has different formats everywhere. 

However we only use Powered By to navigate to the point in the tree, then remove the parent node. Given the name of this module is `debrand` it might be easier to find the `a` tag with `href` instead which should always start `https://www.odoo.com?utm_source=` in every language and remove one level higher

pedrobaeza:
That can work, but the other option is simply: translate in this module the text "Powered by" the same as Odoo does.

astirpe:
I perfectly understand the technical point. But in general we should also put ourselves in the shoes of non-technical users: one installs this module, expecting that the Powered By is removed, but instead it stays there. And the solution with translations is not even mentioned in the readme/configuration.

pedrobaeza:
OK, someone can propose the alternative patch that @gdgellatly mentions

gdgellatly:
@astirpe No one is disagreeing with you, clearly it is a bug that requires fixing. You seem to have the best handle on the right approach so maybe you like to propose a PR.

[...]

PR #525, opened 2020-03-20
[13.0][MIG] mail_debrand: Migration to 13.0 

ValentinVinagre:
Standard migration.


HaraldPanten (approved):
Tested in local environment and it works fine. Functional review 👍

pedrobaeza:
Is this affected by #441?

ValentinVinagre:
Hi @pedrobaeza ,
I didn't see this issue, I will test it and I will tell you something.

ValentinVinagre:
Hi @pedrobaeza ,
I tested different use cases, and all seems to work fine:

1.- I created 4 contact with different languages: English (US), Czech / Čeština, Spanish / Español, Basque / Euskara.
2.- I'm changing the active language for the user that is sending the emails, and I'm repeating steps 2.1, 2.2, and 2.3:
    2.1.- I created 1 sale for every contact, and all correct.
    2.2.- I created 1 invoice for every contact, and all correct.
    2.3.- I created 1 purchase for every contact, and all correct.

But if you want I can use the solution proposed in https://github.com/OCA/social/pull/507 that I think will solve the https://github.com/OCA/social/issues/441 .

ValentinVinagre:
Hi @chienandalu ,
The issue with translations is only in the v. 11.0-12.0? Because I tried and I don't found the problem in v.13

Thanks,

chienandalu:
Well from V12 the notification template is a qweb record so there should be no issue, I think

ValentinVinagre:
Thanks @chienandalu , so if the problem doesn't exist in v.13 I don't see the need to transfer the fix @pedrobaeza .

pedrobaeza:
OK, perfect

tarteo (approved):
Thanks!

PR #528, opened 2020-03-26
[12.0][IMP] mail_tracking: Test 'to' and named emails

Tardo:
cc @tecnativa


pedrobaeza (approved):
Fast-tracking this as they are only tests that improves coverage

pedrobaeza:
Please include them on v13 forward-port

PR #531, opened 2020-03-30
[FIX] mail_tracking: do not warn not found if already opened

### LINKED ISSUE #530: [12.0][mail_tracking] many MailTracking email not found warnings
This code snippet is polluting my logs with warning messages.

https://github.com/OCA/social/blob/8d985d8da9fa864113f87cd59a2b3173f7f89193/mail_tracking/controllers/main.py#L69-L78

How can this work if state will be marked 'opened' the first time? Each successive open will trigger above warning, because state is now 'opened' and will not match the above domain `('state', 'in', ['sent', 'delivered'])`.

 Is it intended to generate warnings each time mail is opened? Or what am I missing?

ruuter:
The state will be 'opened' after first time email is opened. Each
successive open triggered warning, because state was already 'opened'
and thus did not match domain state in (sent, delivered).

resolves #530

### DIFF
--- a/mail_tracking/controllers/main.py
+++ b/mail_tracking/controllers/main.py
@@ -68,14 +68,13 @@ class MailTrackingController(MailController):
             try:
                 tracking_email = env['mail.tracking.email'].search([
                     ('id', '=', tracking_email_id),
-                    ('state', 'in', ['sent', 'delivered']),
                     ('token', '=', token),
                 ])
-                if tracking_email:
-                    tracking_email.event_create('open', metadata)
-                else:
+                if not tracking_email:
                     _logger.warning(
                         "MailTracking email '%s' not found", tracking_email_id)
+                elif tracking_email.state in ('sent', 'delivered'):
+                    tracking_email.event_create('open', metadata)
             except Exception:
                 pass
 


pedrobaeza (approved):

mail_tracking/controllers/main.py
```diff
@@ -68,14 +68,13 @@ def mail_tracking_open(self, db, tracking_email_id, token=False, **kw):
...
-                    ('state', 'in', ['sent', 'delivered']),
                     ('token', '=', token),
                 ])
-                if tracking_email:
-                    tracking_email.event_create('open', metadata)
-                else:
+                if not tracking_email:
                     _logger.warning(
                         "MailTracking email '%s' not found", tracking_email_id)
+                elif tracking_email.state in ('sent', 'delivered'):
```
Tardo, line 76:
```suggestion
                elif tracking_email.state in {'sent', 'delivered'}:
```

Tardo (approved):
:+1: Thanks for the improvement! ~only an non-blocking comment~.

mail_tracking/controllers/main.py
ruuter: Sure, thanks for feedback, but may I ask why is this better?

Tardo: In this exactly case... tuples are better, sorry.

But in general, sets are better for search purposes:
```timeit.timeit('"two" in data', number=10000, setup='data=("one", "two")')```
```timeit.timeit('"two" in data', number=10000, setup='data={"one", "two"}')```

ruuter: Ah ok, tnx. Did not know this.

yajo (approved):

pedrobaeza:
To be forward-ported to 13.0

pedrobaeza:
Someone can do the same PR in v13?

ruuter:
@pedrobaeza Did same PR for 13.

PR #533, opened 2020-04-07
email_template_qweb: ease override of values, include defaults

simahawk:
Improvements:

1. provide hook to override rendering values
2. provide default rendering values as per Jinja templates

### DIFF
--- a/email_template_qweb/models/mail_template.py
+++ b/email_template_qweb/models/mail_template.py
@@ -1,6 +1,9 @@
 # Copyright 2016 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import api, fields, tools, models
+from odoo.addons.mail.models.mail_template import (
+    format_amount, format_date, format_tz
+)
 
 
 class MailTemplate(models.Model):
@@ -26,10 +29,9 @@ class MailTemplate(models.Model):
             if template.body_type == 'qweb' and\
                     (not fields or 'body_html' in fields):
                 for record in self.env[template.model].browse(res_id):
-                    body_html = template.body_view_id.render({
-                        'object': record,
-                        'email_template': template,
-                    })
+                    body_html = template.body_view_id.render(
+                        template._qweb_render_context(record)
+                    )
                     # Some wizards, like when sending a sales order, need this
                     # fix to display accents correctly
                     body_html = tools.ustr(body_html)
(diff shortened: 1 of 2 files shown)


email_template_qweb/models/mail_template.py
```diff
@@ -40,3 +42,36 @@ def generate_email(self, res_ids, fields=None):
...
+    def _qweb_render_context(self, record):
+        res = {}
+        if record._name == 'mail.message':
+            partner_model = self.env['res.partner']
+            # these variables are usually loaded when the notification is sent
+            # but there are some key values that that are nice to have
+            # if you want to pre-render your template in preview mode.
+            # This is particularly useful if you use the template
+            # from the mail composer, whereas the rendering is done on load.
+            # Also, you don't need to access them from `ctx`.
```
hbrunn: I believe all of this should be the docstring of this function, with a little bit of reformulation.

email_template_qweb/models/mail_template.py
```diff
@@ -40,3 +42,36 @@ def generate_email(self, res_ids, fields=None):
...
+            res = partner_model._notify_prepare_template_context(record)
+        res.update({
+            'object': record,
+            'email_template': self,
+            # Same as for Jinja rendering,
+            # taken from `mail_template.render_template`.
+            # These ease porting of old Jinja templates to qweb ones.
+            'format_date':
+                lambda date, format=False,
+                context=self._context: format_date(self.env, date, format),
```
hbrunn, line 83: I don't see why we need jinja compatibility. If you want to format a date or money, you just use the proper widget in qweb

hbrunn (changes requested):
please give an argument why not to use the qweb equivalents of what you inject here? I mean, the whole point of this module is doing mails qweb style, why do you want to bend it back to jinja?

email_template_qweb/models/mail_template.py
simahawk: the comment is for the call to partner method only. I can cook a docstring tho

email_template_qweb/models/mail_template.py
simahawk: @hbrunn as I mentioned: it eases some copy/pasting from jinja templates. It makes easier for end users to customize templates. 

I could make them optional if you want or split to another module but seems over complicated for such a small thing.

Regarding "the whole point": I disagree, qweb is not only for using widgets and such. The key point about qweb is that you can reuse templates and its syntax is much more readable.

This is just a tiny layer of compatibility between them.

simahawk: I could remove the formatters if we agree is "too much" :)

hbrunn: no, keep it the way you have it. It doesn't hurt my qweb purism and helps others, so it's a good thing to do, thank you.

But https://codecov.io/gh/OCA/social/compare/b6dc1918b4aa8670a303edff99fcedeb3391250c...3efbe62cb6dea0b91e87be7391b750cc393a2553/diff#D2-44 needs to be addressed

simahawk:
@hbrunn docstring + comments reworked. Coverage should be happy now as there's nothing to be tested on the core method call.

hbrunn (approved):

p-tombez (approved):

PR #536, opened 2020-04-16
[12.0][IMP] mail_tracking: ignore catchall

Tardo:
- Improve tests
- Ignore catchall from suggested recipients

cc @tecnativa @rafaelbn
**Needs to be ported to 13.0**

### DIFF
--- a/mail_tracking/models/mail_alias.py
+++ b/mail_tracking/models/mail_alias.py
@@ -10,9 +10,15 @@ class MailAlias(models.Model):
     @api.model
     @tools.ormcache()
     def get_aliases(self):
-        return set(x['display_name'] for x in self.search_read([
+        aliases = set(x['display_name'] for x in self.search_read([
             ('alias_name', '!=', False),
         ], ['display_name']))
+        IrConfigParamObj = self.env["ir.config_parameter"].sudo()
+        catchall = "%s@%s" % (
+            IrConfigParamObj.get_param("mail.catchall.alias"),
+            IrConfigParamObj.get_param("mail.catchall.domain"))
+        aliases.add(catchall)
+        return aliases
 
     @api.model_create_multi
     def create(self, vals_list):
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Please check Travis

Tardo:
Changes done

Tardo:
Please review @rafaelbn

mail_tracking/models/mail_alias.py
```diff
@@ -10,9 +10,15 @@ class MailAlias(models.Model):
     @api.model
     @tools.ormcache()
     def get_aliases(self):
-        return set(x['display_name'] for x in self.search_read([
+        aliases = set(x['display_name'] for x in self.search_read([
             ('alias_name', '!=', False),
         ], ['display_name']))
+        IrConfigParamObj = self.env["ir.config_parameter"].sudo()
+        catchall = "%s@%s" % (
```
pedrobaeza, line 17: The rest of the aliases don't contain the domain. Why this one does?

rafaelbn (approved):
:+1: Thanks

mail_tracking/models/mail_alias.py
Tardo: With mail.alias records we are using "display_name" that includes the catchall domain.

pedrobaeza (commented):
As it's not usual to edit/create/unlink ir.config.parameter, just clear cache always in these cases for avoiding excessive code.

mail_tracking/models/mail_alias.py
```diff
@@ -10,9 +10,15 @@ class MailAlias(models.Model):
...
+            IrConfigParamObj.get_param("mail.catchall.alias"),
+            IrConfigParamObj.get_param("mail.catchall.domain"))
+        aliases.update({catchall})
```
pedrobaeza: Isn't simpler to add the element?:
```suggestion
        aliases.add(catchall)
```

Tardo:
Changes done

pedrobaeza (approved):

pedrobaeza:
Please move it to v13:

/ocabot merge patch

PR #534, opened 2020-04-08
Fix mail_inline_css transform hook

simahawk:
The mail module offers a better hook to manipulate html: `render_post_process`.
By using this, we make sure the transform is applied properly
before html tags sanitizing happens,
which can alter - if not screw - the final result.

Finally, it simplifies code :)

Previous commit is a backport from v12 https://github.com/OCA/social/commit/0d68d9dee422befd2e3c1f48d038e1476a3c56ea

PR #549, opened 2020-06-08
[13.0][FIX] mail_inline_css: deafult value for lang

i-vyshnevska:
email.preview.wizard raises an error when [initializing values](https://github.com/odoo/odoo/blob/aba0b299243c3d9b4615e1066d641aa8360aabac/addons/mail/wizard/email_template_preview.py#L38) for lang, replacing of "body_html" field is failing as it not there yet, it will appear when [onchange](https://github.com/odoo/odoo/blob/aba0b299243c3d9b4615e1066d641aa8360aabac/addons/mail/wizard/email_template_preview.py#L47) method for 'lang' will be called

### DIFF
--- a/mail_inline_css/views/email_preview.xml
+++ b/mail_inline_css/views/email_preview.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+  <record model="ir.ui.view" id="email_template_preview_form">
+    <field name="model">mail.template</field>
+    <field name="inherit_id" ref="mail.email_template_form" />
+    <field name="arch" type="xml">
+        <!-- inject default value for lang to prevent call to generate_email method in
+        default_get of the wizard-->
+        <xpath expr="//button[@name='%(mail.wizard_email_template_preview)d']" position="attributes">
+            <attribute name="context">{'template_id':active_id, 'default_preview_lang': context['lang']}</attribute>
+        </xpath>
+
+    </field>
+  </record>
+</odoo>
--- a/mail_inline_css/__manifest__.py
+++ b/mail_inline_css/__manifest__.py
@@ -13,5 +13,6 @@
     "installable": True,
     "external_dependencies": {"python": ["premailer"]},
     "depends": ["email_template_qweb"],
+    "data": ["views/email_preview.xml"],
     "demo": ["demo/demo_template.xml", "demo/demo_mail_template.xml"],
 }


simahawk (commented):
@i-vyshnevska explanation of the issue?

@hbrunn do you remember if there was a specific reason for that check?

i-vyshnevska:
short description added, this was ASAP fix, there was a significant refactoring of template generation so previous solutions may not work as before
check may be was needed to not generate on an empty field, but this better to handl in UI

hbrunn:
the reason was to only generate what is asked - in the versions I worked on this there were a bunch of cases where `generate_email` was called ie just to generate the `To:` header, and without the check we'd then still render the body field, causing unnecessary load.

To me this change looks wrong, and I'd like to know more about which exact case this is supposed to fix.

hbrunn:
PS: it might be as simple as some code requesting the `body` field instead of `body_html`, in that case we're much better off with `not fields or 'body' in fields or 'body_html' in fields`

i-vyshnevska:
now I have this problem also in mail_inline_css
`File "/odoo/external-src/social/mail_inline_css/models/mail_template.py", line 24, in generate_email
    result["body_html"] = self._premailer_apply_transform(result["body_html"])
KeyError: 'body_html'`
changing on 'body' had no effect

simahawk:
> now I have this problem also in mail_inline_css
> `File "/odoo/external-src/social/mail_inline_css/models/mail_template.py", line 24, in generate_email result["body_html"] = self._premailer_apply_transform(result["body_html"]) KeyError: 'body_html'`
> changing on 'body' had no effect

that's because the field is not there and confirms that the check is required

i-vyshnevska:
> that's because the field is not there and confirms that the check is required

after double-checking error on mail_inline_css doesn't appear anymore, bad from my side

about "body_html", this field should be empty for qwed template,
 the error appears because of email preview wizard, on previous versions, there were no fields send in context so on method generate_email list of fields was empty and that check was skipped,
 on v13 "lang" appear in fields and this broke email previewer for qweb emails
https://github.com/odoo/odoo/blob/aba0b299243c3d9b4615e1066d641aa8360aabac/addons/mail/wizard/email_template_preview.py#L38

hbrunn:
shouldn't we end up in [this code path](https://github.com/odoo/odoo/blob/aba0b299243c3d9b4615e1066d641aa8360aabac/addons/mail/wizard/email_template_preview.py#L56) where `generate_email` is called with an empty `fields` list in the line above, thus satisfying the check? Could you post a stack trace and the value of `fields` for the case you're trying to fix?

The code you linked seems more like an argument why we want this check, it would be a waste of resources to render the whole mail just to throw it away again, because in this case we only care about the lang parameter (which is rendered by jinja2)

simahawk:
@i-vyshnevska ping :)

i-vyshnevska:
@simahawk @hbrunn I re-worked on this PR, it seems better now, the problem is only this fix needed for "email_template_qweb" and 
"mail_inline_css" but they actually independent

mail_inline_css/views/email_preview.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
```
simahawk, line 1: I don't get why this change is needed for mail_inline_css :thinking: 
Also, maybe better to open another PR because name, description and discussion are not relevant anymore.

simahawk: I see the updated description now: pls change the name and report the description into a comment in the file.

i-vyshnevska: mail_inline_css override generate_email function, if data not there it fails and this is a need for both addons, so or we introduce it on the separate addon, I put it here as this looks more basic than qweb

mail_inline_css/views/email_preview.xml
```diff
@@ -0,0 +1,15 @@
...
+<odoo>
+  <record model="ir.ui.view" id="email_template_preview_form">
+    <field name="model">mail.template</field>
+    <field name="inherit_id" ref="mail.email_template_form" />
+    <field name="arch" type="xml">
+        <!-- inject default value for lang to prevent call to generate_email method in
+        default_get of the wizard-->
+        <xpath expr="//button[@name='%(mail.wizard_email_template_preview)d']" position="attributes">
+            <attribute name="context">{'template_id':active_id, 'default_preview_lang': 'en_US'}</attribute>
```
hbrunn:
```suggestion
            <attribute name="context">{'template_id':active_id, 'default_preview_lang': context['lang']}</attribute>
```

hbrunn (changes requested):
good we talked about that, this looks like a much better solution to me

hbrunn (approved):

mail_inline_css/views/email_preview.xml
simahawk: hmm ok, tnx for the explanation. Question: if you have tmpl_qweb installed but not this, would you still have the issue?

i-vyshnevska: yes

simahawk: hmm then we I'd propose the fix to odoo core and meanwhile propose the patch for OCB.
As an alternative, could you try this?
1. port https://github.com/OCA/social/pull/534 to v13 (I realized only today that I made it on v11 :stuck_out_tongue: )
2. test if the issue is still there on mail_inline_css (it shouldn't)
3. if point 2 is happy, move the fix to _qweb only

simahawk (changes requested):
pls check my comment https://github.com/OCA/social/pull/549/files#r462769445

simahawk:
@i-vyshnevska ping on comment https://github.com/OCA/social/pull/549/#discussion_r462769445 ;)

I ported it quickly to

* v13 https://github.com/OCA/social/pull/608
* v12 https://github.com/OCA/social/pull/609

but I did no tests: can you review and test them pls?

i-vyshnevska:
refactoring in https://github.com/OCA/social/pull/608 solve the initial problem

PR #555, opened 2020-06-19
[12.0][ADD] mass_mailing_custom_unsubscribe_event: Allow to unsubscribe discretely from an event

pedrobaeza:
This addon extends the unsubscription process for allowing to unsubscribe only for an event.

Standard process includes the mail in the general blacklist instead, which can be very unconvenient.

This includes also the needed changes in the base module `mass_mailing_custom_unsubscribe`.

@Tecnativa TT24412


mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
```diff
@@ -31,3 +33,41 @@ def update_opt_out(self, email, list_ids, value):
...
+                email_fname = 'email'
+            records = model.search([
+                ('id', 'in', res_ids), (email_fname, 'ilike', email)])
+            records.write({'opt_out': value})
+            for res_id in res_ids:
+                self.env["mail.unsubscription"].create({
+                    "email": email,
+                    "mass_mailing_id": self.id,
+                    "unsubscriber_id": "%s,%d" % (
+                        self.mailing_model_real, res_id),
```
chienandalu, line 54: Looks like `res_ids` list element don't come as integers: `TypeError: %d format: a number is required, not str - - -`

chienandalu (changes requested):
Tested on customer demo database. Apart from the error commented, I'm getting two unsubscriptions:

image

And apart from those points it seems to work ok with these tests:

- email address in several event registrations
- Send mass_mailing one to event registration 1
- Unsubscribe with the unsubscribe link
- The registration gets opt_out'ed
- Send mass_mailing two to event registration 2
- The mail arrives
- Send mass mailing 3 to event registration 1
- The mail is dropped.

mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
yajo: Maybe it's `False` if empty?

chienandalu: Well, a `int(res_id)` fixed the issue in my local, although I didn't dive deep into the causes

mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
```diff
@@ -31,3 +33,41 @@ def update_opt_out(self, email, list_ids, value):
...
+        self.ensure_one()
+        model = self.env[self.mailing_model_real].with_context(
+            active_test=False)
+        if (self.mailing_model_real != "mail.mass_mailing.contact" and
+                'opt_out' in model._fields):
+            email_fname = 'email_from'
+            if 'email' in model._fields:
+                email_fname = 'email'
+            domain = safe_eval(self.mailing_domain)
+            domain += [('opt_out', '=', True)]
```
yajo: Always put safe domain first:

```suggestion
            domain = [('opt_out', '=', True)] + domain
```

mass_mailing_custom_unsubscribe/readme/CONFIGURE.rst
```diff
@@ -5,3 +5,12 @@ they are going to unsubscribe. To do it:
...
    continue.
+
+For having discrete unsubscriptions from other recipients than the mailing
+lists, you need to add a glue module that adds 2 fields in the associated
+model:
+
+- `opt_out`.
+- Either `email` or `email_from`.
+
+See `mass_mailing_custom_unsubscribe_event` for an example.
```
yajo, line 16: I think it makes sense for this module to provide an `AbstractModel` that makes these operations, so other modules such as `mass_mailing_custom_unsubscribe_event` only have to inherit it.

BTW, you're mixing syntaxes here, ans shouldn't the glue module be prefixed with the core addon it's based on? `mass_mailing_event_custom_unsubscribe`?:

```suggestion
- ``opt_out``.
- Either ``email`` or ``email_from``.

See ``mass_mailing_event_custom_unsubscribe`` for an example glue module
that adds this feature to events.
```

mass_mailing_custom_unsubscribe/readme/CONFIGURE.rst
```diff
@@ -5,3 +5,12 @@ they are going to unsubscribe. To do it:
 #. Create / edit / remove / sort as usual.
 #. If *Details required* is enabled, they will have to fill a text area to
    continue.
+
+For having discrete unsubscriptions from other recipients than the mailing
+lists, you need to add a glue module that adds 2 fields in the associated
+model:
```
yajo, line 11: "discrete" is not too easy to understand, and maybe the model already has those fields. I'd reword this paragraph like this:

```suggestion
To allow managing unsubscriptions per record for other recipients than the
mailing list contacts, the associated model needs to include these fields:
```

mass_mailing_custom_unsubscribe/readme/DESCRIPTION.rst
```diff
@@ -6,3 +6,6 @@ This addon extends the unsubscription form to let you:
   mass mailing.
 - Provide proof on why you are sending mass mailings to a given contact, as
   required by the GDPR in Europe.
+- Handle discrete unsubscriptions from other recipients that are not a mailing
```
yajo, line 9:
```suggestion
- Handle per-record unsubscriptions from other recipients that are not a mailing
```

mass_mailing_custom_unsubscribe_event/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+        "event",
+        "mass_mailing",
+    ],
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/social",
+    "data": [
+        "views/event_registration_views.xml",
+    ],
+    "license": "AGPL-3",
+    "installable": True,
```
yajo, line 17:
```suggestion
    "installable": True,
    "auto_install": True,
```

mass_mailing_custom_unsubscribe_event/models/event_registration.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class EventRegistration(models.Model):
+    _inherit = 'event.registration'
+
+    opt_out = fields.Boolean(
+        string='Opt-Out',
```
yajo, line 11: I think this field should be tracked.

```suggestion
        string='Opt-Out',
        track_visibility="onchange",
```

mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
```diff
@@ -31,3 +33,41 @@ def update_opt_out(self, email, list_ids, value):
...
+            recs = self.env[self.mailing_model_real].search(domain)
+            normalized_email = [
+                tools.email_split(c[email_fname]) for c in recs]
```
yajo: You can save a few loops here :blush: 

```suggestion
            normalized_email = (
                tools.email_split(c[email_fname]) for c in recs)
```

yajo (changes requested):

mass_mailing_custom_unsubscribe/readme/CONFIGURE.rst
pedrobaeza: The name is put that way for having all of them grouped together, as it depends on the main one, so that's the main factor. Even more, this module doesn't depend on `mass_mailing_event` on purpose, as it's not needed, as that one only adds some UI commodities.

mass_mailing_custom_unsubscribe/readme/CONFIGURE.rst
pedrobaeza: I know that for us Spanish people, it sounds weird, as we tend to translate it as `discreto`, which relates to someone cautious, polite, etc, but seeing the meaning, it seems the exact term:

https://dictionary.cambridge.org/es/diccionario/ingles/discrete
https://dictionary.cambridge.org/es/diccionario/ingles/discretely

mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
pedrobaeza: I don't see the difference changing the list by a tuple.

mass_mailing_custom_unsubscribe/readme/CONFIGURE.rst
chienandalu: It always reminds me of this Futurama joke
image

mass_mailing_custom_unsubscribe_event/models/event_registration.py
pedrobaeza: As we already have the unsubscription log, I see this as redundant and less "trackable" than the other record.

mass_mailing_custom_unsubscribe/readme/CONFIGURE.rst
pedrobaeza: I'm also not providing any AbstractModel as being only 2 fields that can collapse with existing ones, so I prefer that people defines them as their taste. It's not adding a valuable piece of reusable code, taking into account that most of it will be even rewritten for adjusting it to the target model.

mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
pedrobaeza: Fixed as core does, converting to int at the beginning.

pedrobaeza:
@chienandalu fixed the problem with the double unsubscription (not calling `super` for other mailing models), and the rest of the comments attended or replied.

chienandalu:
Great. I'll make a final functional test tomorrow at first hour

chienandalu (approved):
Tested :+1:

ernestotejeda (approved):
Code review :+1:

mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
yajo: It's not a tuple, it's a generator. For it to be a tuple, you'd have to do:

```suggestion
            normalized_email = tuple(
                tools.email_split(c[email_fname]) for c in recs)
```

pedrobaeza: OK, changed.

pedrobaeza:
@Yajo please update your review

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -43,6 +44,7 @@ def mailing(self, mailing_id, email=None, res_id=None, token="", **post):
             "Called `mailing()` with: %r",
             (mailing_id, email, res_id, token, post))
         reasons = request.env["mail.unsubscription.reason"].search([])
+        res_id = res_id and int(res_id)
```
yajo, line 47: :thinking: Humm this line makes me think... Are we actually supporting the case of `res_id=None` or `res_id=""`? If not, then I guess this should just be:

```suggestion
        res_id = int(res_id)
```

This way, we fail fast in case there's some evil `res_id` incoming.

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -53,23 +55,38 @@ def mailing(self, mailing_id, email=None, res_id=None, token="", **post):
...
+                mass_mailing.update_opt_out_other(email, [res_id], True)
+                result = request.render("mass_mailing.page_unsubscribed", {
+                    "email": email,
+                    "mailing_id": mailing_id,
+                    "res_id": res_id,
+                    "show_blacklist_button": request.env[
+                        "ir.config_parameter"
+                    ].sudo().get_param(
+                        "mass_mailing.show_blacklist_buttons"
+                    ),
```
yajo, line 71: Don't you need to cast this into bool?

```suggestion
                    "show_blacklist_button": odoo.tools.str2bool(request.env[
                        "ir.config_parameter"
                    ].sudo().get_param(
                        "mass_mailing.show_blacklist_buttons"
                    )),
```

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -53,23 +55,38 @@ def mailing(self, mailing_id, email=None, res_id=None, token="", **post):
...
                 })
+                result.qcontext.update({"reasons": reasons})
```
yajo, line 73:
```suggestion
                "reasons": reasons,
                })
```

mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
```diff
@@ -31,3 +33,42 @@ def update_opt_out(self, email, list_ids, value):
                 "action": action,
             })
         return super().update_opt_out(email, list_ids, value)
+
+    def update_opt_out_other(self, email, res_ids, value):
```
yajo, line 37: What's the need for having this method separated from `update_opt_out` instead of handling opt out in any model directly there? :thinking:

mass_mailing_custom_unsubscribe/controllers/main.py
pedrobaeza: It's not failing if False/None is passed, so this should be this way. It's copied from upstream anyway.

mass_mailing_custom_unsubscribe/controllers/main.py
pedrobaeza: This code was also copied from upstream, so the problem will also be there if so. No sense to have here extra care if in upstream this will fail.

mass_mailing_custom_unsubscribe/models/mail_mass_mailing.py
pedrobaeza: For having the possibility of extending it without needing to do all the `if` stuff, avoid calling super, etc. It's enough different concept for keeping it separated.

yajo (approved):

pedrobaeza:
Thanks for the reviews

/ocabot merge patch

yajo:
This PR broke integration tests downstream.

It seems tours are skipped here in travis: https://travis-ci.org/github/OCA/social/jobs/703446068#L1909

How to enable those?

pedrobaeza:
It talks about `websocket-client module is not installed`. Not sure if that's a Python library. It would be to install it in MQT. Anyway, do you have the fix for fixing them?

yajo:
I have no fix yet, but tours keep breaking our CIs, and I'm starting to think that the reason is they get skipped here in OCA.

pedrobaeza:
Yes, it seems so. I have trusted the green status without seeing that. But then this situation has been here for years... Can you investigate it?

yajo:
Let's see if #560 fixes it.

PR #557, opened 2020-06-24
[FIX] mail_tracking: do not slurp exceptions

yajo:
This code chunk is expected to raise. Slurping it doesn't help upstream methods that rely on these exceptions.

@Tecnativa TT24457

### DIFF
--- a/mail_tracking/models/ir_mail_server.py
+++ b/mail_tracking/models/ir_mail_server.py
@@ -80,9 +80,11 @@ class IrMailServer(models.Model):
         except Exception as e:
             if tracking_email:
                 tracking_email.smtp_error(self, smtp_server_used, e)
-        if message_id and tracking_email:
-            vals = tracking_email._tracking_sent_prepare(
-                self, smtp_server_used, message, message_id)
-            if vals:
-                self.env['mail.tracking.event'].sudo().create(vals)
+            raise
+        finally:
+            if message_id and tracking_email:
+                vals = tracking_email._tracking_sent_prepare(
+                    self, smtp_server_used, message, message_id)
+                if vals:
+                    self.env['mail.tracking.event'].sudo().create(vals)
         return message_id


pedrobaeza (changes requested):
The problem with this is that the tracking record won't be saved as an exception rolls back the DB transaction. You need to do something similar to

https://github.com/OCA/l10n-spain/blob/557fcebd9d1cd7b447cd0274edb9b73c0e713f80/l10n_es_aeat_sii/models/account_invoice.py#L1023-L1035

for having both things (the raise and the record creation)

yajo:
Thanks, fixed.

pedrobaeza:
Please check Travis

yajo:
I see. It seems the last "fix" was conceptually broken. I cannot track an email that still doesn't exist in the database because the 2nd cursor still has no data from the 1st one, which is uncommitted at this point.

I have checked and evey time this method is called in Odoo code, it is called inside a try-except block, to avoid aborting the transaction in case there's a failure. Remember that the upstream method is able to raise, and that's a feature.

So, I guess the 1st fix was OK.

pedrobaeza:
But then we won't have the tracking, so we have to choose one. Before this patch, no problem has been seen. Why is now a problem except a test that is looking specifically that raise? Isn't better to change that test?

yajo:
No, see. The test was actually revealing that this part of the code was buggy. It's not the test fault.
Actually the test revealed other things to improve in the other module code that are fixed too in https://github.com/OCA/data-protection/pull/42, but just fixing this part did make the test pass.

Just make a quick search across all odoo code and you'll see that every call to this method expects an exception and avoids the rollback. Also some of them react to the exception and apply specific behaviors that we shouldn't alter.

Finally, most chances are that these tracking events do not even have a meaning. When there's an exception here, it happens because the email couldn't be delivered to the MTA, not because the MTA couldn't deliver it to the destination. This means that the MTA couldn't track it, so all other features here wouldn't work anyway.

pedrobaeza:
OK, then revert to initial PR without creating the record (it won't be created anyway)

yajo:
> OK, then revert to initial PR

Done.

> without creating the record (it won't be created anyway)

Well.. yes, it will be created. I think there might be some misunderstanding here... Remember, since every time this method is called it's wrapped on a try-except clause, the transaction will *not* be rolled back. So the record will be created and persisted.

In the case that this wasn't protected, then the created record will be discarded, so also not creating it will make no difference.

pedrobaeza:
Then don't put it in the finally. Just add the `raise` line.

yajo:
Why? I don't understand you. :confused:

mail_tracking/models/ir_mail_server.py
```diff
@@ -80,9 +80,11 @@ def send_email(self, message, mail_server_id=None, smtp_server=None,
...
-                self, smtp_server_used, message, message_id)
-            if vals:
-                self.env['mail.tracking.event'].sudo().create(vals)
+            raise
+        finally:
+            if message_id and tracking_email:
+                vals = tracking_email._tracking_sent_prepare(
+                    self, smtp_server_used, message, message_id)
+                if vals:
+                    self.env['mail.tracking.event'].sudo().create(vals)
```
pedrobaeza, line 89:
```suggestion
            raise
        if message_id and tracking_email:
            vals = tracking_email._tracking_sent_prepare(
                self, smtp_server_used, message, message_id)
            if vals:
                self.env['mail.tracking.event'].sudo().create(vals)
```

pedrobaeza: That's to say, just add the `raise` line.

yajo: Yes, but why? What's the benefit?

pedrobaeza: Several:

- Less diff (only one line)
- Don't create something that is going to be rolled-back.
- I even doubt if this will generate other exception in that state.

yajo: > Don't create something that is going to be rolled-back.

But it is not going to be rolled back... :roll_eyes:

pedrobaeza: That happens automatically when an exception is raised.

yajo: Yes, but every time this method is called, it is done inside a try-except block that avoids the rollback.

pedrobaeza: OK

pedrobaeza (approved):

chienandalu (approved):

pedrobaeza:
Please rebase and check problems in Travis

yajo:
Main branch is broken.

pedrobaeza:
Can you please fix it? We can't merge if not.

yajo:
Fix in https://github.com/OCA/social/pull/559

PR #560, opened 2020-06-30
[12.0][FIX] mass_mailing_custom_unsubscribe: mailing contact contains `opt_out`

yajo:
So we need to explicitly opt-out (:rofl:) that model from the new alternative mechanism for unsubscribing records.

@Tecnativa TT24523

### DIFF
--- a/mass_mailing_custom_unsubscribe/controllers/main.py
+++ b/mass_mailing_custom_unsubscribe/controllers/main.py
@@ -58,7 +58,8 @@ class CustomUnsubscribe(MassMailController):
             mailing_obj = request.env['mail.mass_mailing']
             mass_mailing = mailing_obj.sudo().browse(mailing_id)
             model = mass_mailing.mailing_model_real
-            if "opt_out" in request.env[model]._fields:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@sbidoul it seems your switch to PIP has this side effect: https://github.com/OCA/social/pull/555#issuecomment-651687647

yajo:
Now it's failing as it should. Yikes, fixing this is gonna be hard... :facepalm:

pedrobaeza:
Well, I would like to know from @sbidoul why it was originally changed to PIP.

pedrobaeza:
One of the test is due to missing libraries. Maybe it's better to add support for website tours in PIP mode for avoiding that missing libraries.

yajo:
Yes, possibly https://github.com/OCA/maintainer-quality-tools/pull/672 fixes that.

pedrobaeza:
I have reused this PR for fixing the underlying problem after MQT has been applied, so letting the bot to fast-track this for having a green branch when finishes:

/ocabot merge patch

Issue #556, opened 2020-06-23
[13.0] mail_activity_board activities not removed when removing record


CasVissers-360ERP:
See video:
https://drive.google.com/file/d/1TuPwzCKDffDV-Ypd8NW29XLk1yzRqE2C/view

When mail_activity_board is installed and the underlaying record is removed the activity is not removed. When trying to open the activity it's not shown.

@pedrobaeza

pedrobaeza:
Sorry, not involved in that module.

CasVissers-360ERP:
@MiquelRForgeFlow 
@dajuayen 
Look's like a prio issue to me? I tried to solve it but wasn't able to do so.

MiquelRForgeFlow:
I cannot see the video, it says it didn't exist.

CasVissers-360ERP:
@MiquelRForgeFlow 
https://drive.google.com/file/d/1vgNLwhEvGEp0FD_kssyaYLGyVPoimwF-/view

Also tried on v12 runbot. Bug is unique to v13.

MiquelRForgeFlow:
Solved in https://github.com/OCA/social/pull/564.

PR #564, opened 2020-07-08
[13.0][FIX] mail_activity_board: unlink issue due to bad migration

MiquelRForgeFlow:
Fixes https://github.com/OCA/social/issues/556.


JordiBForgeFlow (approved):

HviorForgeFlow (approved):

lmignon (commented):
@MiquelRForgeFlow Thank you for the fix but can you add a unittest for this case.
Thanks

MiquelRForgeFlow:
> can you add a unittest for this case.

I think it's not needed, it's just a `sudo() -> with_user()`/` SUPERUSER_ID -> is_superuser()` error. It's not a bug fix, but a regression fix due to bad migration. If the migration was done correctly, nobody would have asked to add a test.

CasVissers-360ERP (approved):
Thnx! 👍 Solves the issue.

PR #568, opened 2020-07-14
[12.0][ADD] mail_notify: FCM Push notifications

kh-badep:
This module extends the native push notifications of Odoo to support all incoming messages instead of just chat messages. Moreover its real purpose is to use Firebase Cloud Messaging (FCM) to get Push notifications even if the web client is not open.

By default, push notifications on Odoo are enabled only for chat. This module removes this restriction to enable push notifications on incoming messages (inbox) and activities. It also improves the notification by adding an URL to open the related document by clicking on the notification, the document icon or the sender avatar.

Another restriction on Odoo default implementations is that it relies on bus_bus to send notifications to the web client, which means that the user does not receive a notification if he doesn't have an open browser tab on Odoo. Even worse, due to memory management changes in modern browsers like Chrome (see: https://developers.google.com/web/updates/2015/09/tab-discarding), sometimes the tab will be put on "sleep" and therefore the client will not receive notifications. This module has also an option to rely on a service worker like FCM to deliver push notifications on the background. This also works like a charm on mobile phones.


kh-badep:
Hello,

Any news as to when this PR will be reviewed?

PR #567, opened 2020-07-10
[10.0][ADD] mail_outbound_static: Allow configuration of "from" in email header

pepetreshere:
A backport of mail_outbound_static from v12 to v10, it eases the sending of emails from a fixed "from" address to comply with  DMARC conformity, use it when your emails can't be sent from Odoo due to invalid "from" address.

The module is very usfefull as [the fix in the Odoo stable version](https://github.com/odoo/odoo/pull/20911) has not being accepted yet.

Compared to v12, this module adds the feature to allow to setup the "from" address in the server conf file (usually odoo.conf), which is also a valid way for Odoo to setup outgoing email servers.

This work is also inspired on previous work on the closed PR https://github.com/OCA/social/pull/195.

PR #570, opened 2020-07-15
[10.0][FIX] travis: set premailer version fixe to last compatible with python2

etobella:
Original #567 error is probably related to this


etobella:
Maybe the error on ocabot is related to github issues:

image

PR #571, opened 2020-07-16
[12.0][FIX] mass_mailing_list_dynamic: list sync on send

chienandalu:
- We change the method to hook on (it wasn't correctly overriden anyway)
to ensure the list is resynced even if no recipients are left. For
example: we create a dynamic list with a domain expecting it to sync in
the first mass_mailing. It wouldn't sync as Odoo dismisses the mailing
considering no recipients are left thus it wouldn't get into the
`send_mail` method.

cc @Tecnativa TT24723

### DIFF
--- a/mass_mailing_list_dynamic/models/mail_mass_mailing.py
+++ b/mass_mailing_list_dynamic/models/mail_mass_mailing.py
@@ -7,7 +7,9 @@ from odoo import models
 class MassMailing(models.Model):
     _inherit = "mail.mass_mailing"
 
-    def send_mail(self):
-        """Sync dynamic lists before sending mailings to them."""
+    def get_remaining_recipients(self):
+        """When evaluating remaining recipients we must resync the list in
+           advance to avoid missing recipients due to domain change or new
+           partners fitting into the conditions"""
         self.contact_list_ids.action_sync()
-        return super().send_mail()
+        return super().get_remaining_recipients()


pedrobaeza (approved):

rafaelbn (changes requested):
Functionally tested, it doesn't work. Check my 3 minutes video:

https://drive.google.com/file/d/1W4AAzFBst66QXThTVne0ysN91HllTNVo/view?usp=sharing

rafaelbn (approved):
Synced in the moment the email is sent by Odoo, not in the moment you press the button "sent"

:+1:

rafaelbn:
We should include this in the v12 migration https://github.com/OCA/social/pull/518

chienandalu:
@rafaelbn It could be convenient to sync twice: one for the `put_in_queue` method (when the user clicks send in the mass_mailing) and and the other as it is in this PR (when the mailing is actually sent).

The reason to have it in the first one would be to give the user a chance to amend filter errors (i.e.: unexpectedly sending a mass mailing to the whole contact database!)

The reason to keep the second one, to reflect changes in the meantime (could be hours, days...) Anyway, this could lead to the same error (having a filter changed in the meantime and then sending to the wrong partners).

A third way could be to warn the user about what's going to happen: there's a dynamic list and the mail will be sent to the users fitting the criteria in the time the mail is sent...

PR #573, opened 2020-07-20
[MIG] mail_outbound_static: A backport of mail_outbound_static from v…

rquevedo:
…12 to v10, it eases the sending of emails from a fixed "from" address to comply with DMARC conformity, use it when your emails can't be sent from Odoo due to invalid "from" address.

The module is very usfefull as the fix in the Odoo stable version has not being accepted yet.

Compared to v12, this module adds the feature to allow to setup the "from" address in the server conf file (usually odoo.conf), which is also a valid way for Odoo to setup outgoing email servers.


mail_outbound_static/models/ir_mail_server.py
```diff
@@ -0,0 +1,68 @@
...
+                              odoo.tools.config['email_from'],)
+            else:
+                email_from = odoo.tools.config['email_from']
+        else:
+            # do nothing
+            pass
+
+        if email_from:
+            message.replace_header('From', email_from)
+            bounce_alias = self.env['ir.config_parameter'].get_param(
```
gurneyalex:
```suggestion
            bounce_alias = self.env['ir.config_parameter'].sudo().get_param(
```

from PR #575

gurneyalex (changes requested):

rquevedo:
@gurneyalex Could you help us to identify what is failing in codecov checks? In details links there are not relevant information about whats is wrong or missing.

hbrunn:
I'm not @gurneyalex, but if you'll take my advise too look at the [diff](https://codecov.io/gh/OCA/social/compare/340be15cfcaef1cd7ef845933d064d333c906dd3...8de7a651118934c2a65e8866cad8dc3121a9aab0/diff) tab, this shows you the code paths not covered by your tests. In fact, you don't have tests at all and just because some other modules in this repo have tests do you have coverage at all, so I suggest to add some tests of your own.

rquevedo:
Hi @hbrunn , thanks for you advice and guide. Now im realized that original module in v12 dont have any tests and im doing a migration here. Is mandatory to me, add some test, in order to my downgrade migration get merged?

hbrunn:
This depends a bit on the reviewer, I interpret the [conventions](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#it-makes-sense-to-be-picky-in-the-following-cases) such that no test can fail, and no addition may lower test coverage. Others disagree.

But then I see tests in the v12 version: https://github.com/OCA/social/tree/12.0/mail_outbound_static/tests

You should follow [the migration manual](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0#technical-method-to-migrate-a-module-from-120-to-130-branch) anyways to conserve the commits others made, and then add your changes on top of that.

PR #579, opened 2020-07-27
[12.0][FIX] mass_mailing_partner: get contextual list when importing contacts

chienandalu:
When importing contacts from csv or xls, the list is set by context. So
we must try to default to that context when setting the contact values.

cc @Tecnativa TT24782

### DIFF
--- a/mass_mailing_partner/models/mail_mass_mailing_contact.py
+++ b/mass_mailing_partner/models/mail_mass_mailing_contact.py
@@ -38,18 +38,25 @@ class MailMassMailingContact(models.Model):
             if category_ids:
                 self.tag_ids = category_ids
 
-    @api.model
-    def create(self, vals):
-        record = self.new(vals)
-        if not record.partner_id:
-            record._set_partner()
-        record._onchange_partner_mass_mailing_partner()
-        new_vals = record._convert_to_write(record._cache)
-        new_vals.update(
-            subscription_list_ids=vals.get('subscription_list_ids', False),
-            list_ids=vals.get('list_ids', False)
-        )
-        return super(MailMassMailingContact, self).create(new_vals)
+    @api.model_create_multi
+    def create(self, vals_list):
+        new_vals_list = []
+        for vals in vals_list:
+            # Ensure that defaults are loaded (e.g.: import csv or xls)
+            values_w_defaults = self.default_get(self._fields.keys())
+            values_w_defaults.update(vals)
+            record = self.new(values_w_defaults)
+            if not record.partner_id:
+                record._set_partner()
+            record._onchange_partner_mass_mailing_partner()
+            new_vals = record._convert_to_write(record._cache)
+            new_vals.update(
+                subscription_list_ids=values_w_defaults.get(
+                    'subscription_list_ids', False),
+                list_ids=values_w_defaults.get('list_ids', False)
+            )
+            new_vals_list.append(new_vals)
+        return super().create(new_vals_list)
 
     def write(self, vals):
         for contact in self:
@@ -63,7 +70,7 @@ class MailMassMailingContact(models.Model):
                 subscription_list_ids=vals.get('subscription_list_ids', False),
                 list_ids=vals.get('list_ids', False)
             )
-            super(MailMassMailingContact, contact).write(new_vals)
+            super().write(new_vals)
         return True
 
     def _get_company(self):


pedrobaeza (changes requested):
You have to initialize with defaults instead:
```python
vals = self.default(self._fields.keys())
record = self.new(vals)
...
```
Also extract the same code from both create/write.

chienandalu:
@pedrobaeza Changes done

pedrobaeza (approved):

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -40,30 +40,38 @@ def _onchange_partner_mass_mailing_partner(self):
 
     @api.model
     def create(self, vals):
-        record = self.new(vals)
+        # Ensure that defaults are loaded (e.g.: import csv or xls)
+        values_w_defaults = self.default_get(self._fields.keys())
```
yajo: I'm not saying this is wrong, but it makes me wonder... why aren't the default values used by default? :thinking: 

I thought that's what "defaults" are for.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -40,30 +40,38 @@ def _onchange_partner_mass_mailing_partner(self):
 
     @api.model
```
yajo: I see that your optimization is done specifically for importing contacts, which usually happens with a big list of records.

Have you considered taking this opportunity to migrate this method to `@api.model_create_multi`?

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -40,30 +40,38 @@ def _onchange_partner_mass_mailing_partner(self):
...
+            subscription_list_ids=values_w_defaults.get(
+                'subscription_list_ids', False),
+            list_ids=values_w_defaults.get('list_ids', False)
         )
-        return super(MailMassMailingContact, self).create(new_vals)
+        return super().create(new_vals)
 
     def write(self, vals):
+        # Ensure that defaults are loaded (e.g.: import csv or xls)
+        values_w_defaults = self.default_get(self._fields.keys())
```
yajo: I'm pretty sure this is a bad idea.

You'd be resetting every field to their defaults if it's not specified on a call to `write()`, and that's pretty common. :thinking:

mass_mailing_partner/models/mail_mass_mailing_contact.py
chienandalu: What do you think about the first approach?: https://github.com/OCA/social/commit/4f705c504b3cb855db6aafebb528dd6610c91000

mass_mailing_partner/models/mail_mass_mailing_contact.py
pedrobaeza: Yes, indeed this is something I haven't thought about. Anyway, why writing you need to check that context?

mass_mailing_partner/models/mail_mass_mailing_contact.py
pedrobaeza: The problem is that a new record doesn't apply defaults, and when converting cache to values you lose that default, as the returned dictionary contains all the fields without values, so there's no opportunity of defaults to be applied as "specifically" a value is provided. Do you get it? I still prefer this approach in create than the previous one, as it's specific for that thing, and let out any other possible default.

mass_mailing_partner/models/mail_mass_mailing_contact.py
chienandalu: Well, I'm guessing that with an import you could update existing records as well. But to be honest I didn't test if there was an issue in that case

pedrobaeza: But doing an import you don't need to change the list. You are modifying other values (like email or similar). It's better to remove this part in the update.

mass_mailing_partner/models/mail_mass_mailing_contact.py
chienandalu: Ok, now I understand also :) Let's go with it

mass_mailing_partner/models/mail_mass_mailing_contact.py
pedrobaeza: What about this?

chienandalu: Ok, done

pedrobaeza (approved):

pedrobaeza:
@Yajo is it OK for you?

rafaelbn (approved):
:+1:

PR #582, opened 2020-07-29
[FIX] 13.0 mail_outbound_static

gurneyalex:
use sudo to get ir.config_parameter, otherwise when normal users try to
send email they get AccessError


gurneyalex:
This is a forward port of #575

gurneyalex:
I'm happy if this can be fast tracked

pedrobaeza:
I would like to make you happy, but Travis is red :(

gurneyalex:
> I would like to make you happy, but Travis is red :(

I had forgotten to set up pre-commit on my local copy of oca/social... :sweat: 

travis should be fixed shortly

PR #587, opened 2020-08-03
[13.0][ADD] addons table, translation status

hbrunn:


pedrobaeza:
I personally don't like that addons table, but it's only my opinion.

skeller1 (approved):
:+1:   LGTM

gurneyalex (approved):

PR #588, opened 2020-08-03
[MIG] mail_outbound_static: A backport of mail_outbound_static from v…

rquevedo:
…12 to v10, it eases the sending of emails from a fixed "from" address to comply with DMARC conformity, use it when your emails can't be sent from Odoo due to invalid "from" address.

The module is very usfefull as the fix in the Odoo stable version has not being accepted yet.

Compared to v12, this module adds the feature to allow to setup the "from" address in the server conf file (usually odoo.conf), which is also a valid way for Odoo to setup outgoing email servers.


rquevedo:
Hi  @gurneyalex could you tell us if there are something else to do here?

gurneyalex (approved):

pepetreshere:
Hello @gurneyalex , thanks for aproving the changes. Would you please consider merging it? I've been using it in production for a few weeks without any problems.

PR #591, opened 2020-08-12
[IMP] 13.0 update fr.po file for mail_debrand module

ajaniszewska-dev:
aim of this PR is to update translation for mail_debrand module so functionality will be available also in french translation of email template.


pedrobaeza:
Please do it on Weblate instead:

https://odoo-community.org/page/translate

PR #598, opened 2020-08-28
[ADD] mail_outbound_static: domain whitelist based logic

zaoral:
Add domain whitelist based logic

### DIFF
--- a/mail_outbound_static/README.rst
+++ b/mail_outbound_static/README.rst
@@ -31,17 +31,17 @@ being appended into the proper Sender header instead. To accomplish this we:
 
 * Add a domain whitelist field in the mail server model. This one represent an
   allowed Domains list separated by commas. If there is not given SMTP server
-  it will let us to search the proper mail server to be used to sent themessages
+  it will let us to search the proper mail server to be used to send the messages
   where the message 'From' email domain match with the domain whitelist. If
-  there is not mail sever that match then will use the default mail server to
-  sent the message.
+  there is not mail server that matches then will use the default mail server to
+  send the message.
 
 * Add a Email From field that will let us to email from a specific address taking
   into account this conditions:
 
   1) If the sender domain match with the domain whitelist then the original
      message's 'From' will remain as it is and will not be changed because the
-     mail server is able to sent in the name of the sender domain.
+     mail server is able to send in the name of the sender domain.
 
   2) If the original message's 'From' does not match with the domain whitelist
      then the email From is replaced with the Email From field value.
@@ -61,12 +61,6 @@ Usage
 * Navigate to an Outbound Email Server
 * Set the `Email From` option to an email address
 * Set the `Domain Whitelist` option with the domain whitelist
-
-Known issues / Roadmap
-======================
-
-* Add validation of smtp_from field to ensure that is a valid email address
-* Add validation of domain_whitelist field to ensure that they are valid domains
 
 Bug Tracker
 ===========
--- a/mail_outbound_static/README.rst
+++ b/mail_outbound_static/README.rst
@@ -26,8 +26,29 @@ Mail Outbound Static
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
 This module brings Odoo outbound emails in to strict compliance with RFC-2822
-by allowing for a statically configured From header, with the sender's e-mail
-being appended into the proper Sender header instead.
+by allowing for a dynamically configured From header, with the sender's e-mail
+being appended into the proper Sender header instead. To accomplish this we:
+
+* Add a domain whitelist field in the mail server model. This one represent an
+  allowed Domains list separated by commas. If there is not given SMTP server
+  it will let us to search the proper mail server to be used to sent themessages
+  where the message 'From' email domain match with the domain whitelist. If
+  there is not mail sever that match then will use the default mail server to
+  sent the message.
+
+* Add a Email From field that will let us to email from a specific address taking
+  into account this conditions:
+
+  1) If the sender domain match with the domain whitelist then the original
+     message's 'From' will remain as it is and will not be changed because the
+     mail server is able to sent in the name of the sender domain.
+
+  2) If the original message's 'From' does not match with the domain whitelist
+     then the email From is replaced with the Email From field value.
+
+* Add compatibility to define the smtp information in Odoo config file. Both
+  smtp_from and smtp_whitelist_domain values will be used if there is not mail
+  server configured in the system.
 
 **Table of contents**
 
@@ -39,11 +60,13 @@ Usage
 
 * Navigate to an Outbound Email Server
 * Set the `Email From` option to an email address
+* Set the `Domain Whitelist` option with the domain whitelist
 
 Known issues / Roadmap
 ======================
 
-* Allow for domain-based whitelist that will not be manipulated
+* Add validation of smtp_from field to ensure that is a valid email address
+* Add validation of domain_whitelist field to ensure that they are valid domains
 
 Bug Tracker
 ===========
@@ -63,6 +86,7 @@ Authors
 
 * brain-tec AG
 * LasLabs
+* Adhoc SA
 
 Contributors
 ~~~~~~~~~~~~
@@ -70,6 +94,8 @@ Contributors
 * Frédéric Garbely <frederic.garbely@braintec-group.com>
 * Dave Lasley <dave@laslabs.com>
 * Lorenzo Battistini <https://github.com/eLBati>
+* Katherine Zaoral <kz@adhoc.com.ar>
+* Juan José Scarafía <jjs@adhoc.com.ar>
 
 Maintainers
 ~~~~~~~~~~~
--- a/mail_outbound_static/models/ir_mail_server.py
+++ b/mail_outbound_static/models/ir_mail_server.py
@@ -1,7 +1,10 @@
 # Copyright 2017 LasLabs Inc.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
-from odoo import api, fields, models
+from email.utils import formataddr, parseaddr
+
+from odoo import _, api, fields, models, tools
+from odoo.exceptions import ValidationError
 
 
 class IrMailServer(models.Model):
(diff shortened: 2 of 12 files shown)


mail_outbound_static/models/ir_mail_server.py
```diff
@@ -46,3 +85,26 @@ def send_email(
         return super(IrMailServer, self).send_email(
             message, mail_server_id, smtp_server, *args, **kwargs
         )
+
+    @api.model
+    def _get_mail_sever(self, mail_server_id, smtp_server, email_domain):
```
jjscarafia: May be better like this?

```
    @api.model
    def _get_mail_sever(self, mail_server_id, smtp_server, email_domain):
        """ return: 1) if defined return this one
                    2) return the one that match with the domain_whitelist
                    3) return the default available one """
        mail_server = None
        if mail_server_id:
            mail_server = self.sudo().browse(mail_server_id)
        elif not smtp_server:
            for item in self.sudo().search(
                    [("domain_whitelist", "!=", False)], order="sequence"):
                domain_whitelist = self._get_domain_whitelist(item.domain_whitelist)
                if email_domain in domain_whitelist:
                    mail_server = item
                    break
            if not mail_server:
                mail_server = self.sudo().search([], order="sequence", limit=1)
        return mail_server
```

mail_outbound_static/models/ir_mail_server.py
```diff
@@ -46,3 +85,26 @@ def send_email(
         return super(IrMailServer, self).send_email(
             message, mail_server_id, smtp_server, *args, **kwargs
         )
+
+    @api.model
```
jjscarafia, line 135: perhaps we could use ormcache because this method is going to be called many times.
Should be something like this:

```
@tools.ormcache('mail_server_id', 'smtp_server', 'email_domain')
def _get_mail_sever(self, mail_server_id, smtp_server, email_domain):
```

Then on the write/create/unlink we need to call the invalidate cache as [here](https://github.com/odoo/odoo/blob/13.0/addons/mail/models/mail_message_subtype.py#L44-L55)

mail_outbound_static/models/ir_mail_server.py
```diff
@@ -11,27 +15,62 @@ class IrMailServer(models.Model):
     smtp_from = fields.Char(
         string="Email From", help="Set this in order to email from a specific address."
     )
+    domain_whitelist = fields.Char(
+        help="Allowed domains that can use this SMTP without changing the smtp_from"
```
jjscarafia: Perhaps we should also add information on the help that this domain will be prefered if not domain specified on the outgoing email

mail_outbound_static/views/ir_mail_server_view.xml
```diff
@@ -14,6 +14,7 @@
         <field name="arch" type="xml">
             <xpath expr="//field[@name='smtp_pass']" position="after">
                 <field name="smtp_from" />
+                <field name="domain_whitelist" attrs="{'invisible': [('smtp_from', '=', False)]}"/>
```
jjscarafia: may be we could keep always visible, because actually the functionality to select a proper mailserver is still there if you just set this field (set smtp_from, set domain_whitelist, remove smtp_from)

jjscarafia (changes requested):
Thanks @zaoral 
great work! some comments..

mail_outbound_static/models/ir_mail_server.py
zaoral: @jjscarafia thank you juan, yes I have tried something similar and it was not working :( but now with the code you suggested I have update it and its working :+1:

zaoral:
Hi @eLBati @BT-fgarbely @lasley

We have changed this module to add the domain-based whitelist logic.
Please could you review and let us know your feedback?

Best Regards!

mail_outbound_static/models/ir_mail_server.py
```diff
@@ -1,36 +1,83 @@
...
-        string="Email From", help="Set this in order to email from a specific address."
+        string="Email From",
+        help="Set this in order to email from a specific address."
+        " If the original message's 'From' does not match with the domain"
+        " whitelist then it is replaced with this value. If does match with the"
+        " domain whitelist then the original message's 'From' will not change",
+    )
+    domain_whitelist = fields.Char(
+        help="Allowed Domains list separated by commas. If there is not given"
+        " SMTP server it will let us to search the proper mail server to be use"
```
eLBati: -> `to be used`

mail_outbound_static/README.rst
```diff
@@ -26,8 +26,19 @@ Mail Outbound Static
...
+
+* Add a domain whitelist field in the mail server model.
+* We dinamically select the matching mail server to be use depending on the sender domain:
+
+    1) if mail server is defined when sending then message use this one.
+    2) return the mail server that match with the whitelist domain.
+    3) if not, return the default mail server available.
+
+* Add compatibility to define the smtp information in Odoo config file. Both smtp_from and
+  smtp_whitelist_domain values will be use if there is not mail server configured in the
```
eLBati: -> `will be used`

mail_outbound_static/README.rst
```diff
@@ -26,8 +26,19 @@ Mail Outbound Static
...
-being appended into the proper Sender header instead.
+by allowing for a dynamically configured From header, with the sender's e-mail
+being appended into the proper Sender header instead. To accomplish this we:
+
+* Add a domain whitelist field in the mail server model.
+* We dinamically select the matching mail server to be use depending on the sender domain:
+
+    1) if mail server is defined when sending then message use this one.
+    2) return the mail server that match with the whitelist domain.
+    3) if not, return the default mail server available.
```
eLBati: What does "`if not`" refers to?

eLBati (changes requested):
Thanks!

zaoral:
Hi @eLBati 

Thank you for the review!
I have made the suggested changes. About the comment

> What does "if not" refers to?

I was referencing to bullet (2). I changed it to "if not match" please let me know if this way is better or if you have any suggestion :)

Best Regards

mail_outbound_static/README.rst
```diff
@@ -26,8 +26,19 @@ Mail Outbound Static
...
-being appended into the proper Sender header instead.
+by allowing for a dynamically configured From header, with the sender's e-mail
+being appended into the proper Sender header instead. To accomplish this we:
+
+* Add a domain whitelist field in the mail server model.
+* We dinamically select the matching mail server to be use depending on the sender domain:
+
+    1) if mail server is defined when sending then message use this one.
+    2) return the mail server that match with the whitelist domain.
+    3) if not match, return the default mail server available.
```
eLBati: I was able to understand how the module works after reading the help messages of `smtp_from` and `domain_whitelist` fields. These steps are a bit confusing for me. Maybe they could be written taking the cue from those help messages.
What do you think?

zaoral: Hi @eLBati 
Yes, I am totally agree.

I have updated the README based of the help of the fields.
Please could you do a new review of the README?

mail_outbound_static/README.rst
```diff
@@ -26,8 +26,29 @@ Mail Outbound Static
...
 
 This module brings Odoo outbound emails in to strict compliance with RFC-2822
-by allowing for a statically configured From header, with the sender's e-mail
-being appended into the proper Sender header instead.
+by allowing for a dynamically configured From header, with the sender's e-mail
+being appended into the proper Sender header instead. To accomplish this we:
+
+* Add a domain whitelist field in the mail server model. This one represent an
+  allowed Domains list separated by commas. If there is not given SMTP server
+  it will let us to search the proper mail server to be used to sent themessages
```
eLBati: -> `to be used to send the messages`

mail_outbound_static/README.rst
```diff
@@ -26,8 +26,29 @@ Mail Outbound Static
...
+  where the message 'From' email domain match with the domain whitelist. If
```
eLBati, line 35: -> `matches`

mail_outbound_static/README.rst
```diff
@@ -26,8 +26,29 @@ Mail Outbound Static
...
+  there is not mail sever that match then will use the default mail server to
```
eLBati: -> `matches`

mail_outbound_static/README.rst
```diff
@@ -26,8 +26,29 @@ Mail Outbound Static
...
+  sent the message.
```
eLBati: -> `send`

mail_outbound_static/README.rst
```diff
@@ -26,8 +26,29 @@ Mail Outbound Static
...
+
+* Add a Email From field that will let us to email from a specific address taking
+  into account this conditions:
+
+  1) If the sender domain match with the domain whitelist then the original
+     message's 'From' will remain as it is and will not be changed because the
+     mail server is able to sent in the name of the sender domain.
```
eLBati: -> `is able to send`

eLBati (changes requested):
Thanks

zaoral:
HI @eLBati 
Thanks to you for the review.
I have updated the documentation with the misspelling error suggestion you made.

You can check it in the last commit https://github.com/OCA/social/pull/598/commits/542a0b90c0577fe99a3d3216cd90d5146db90981 where I also added more unit test and add validation of the email from and domain whitelist fields to make the github checks codecov patch/project pass.

eLBati (approved):
Thanks

zaoral:
Hi @Yajo 

Please could you review this PR?
Best Regards!

jjscarafia (approved):
Using in production, working great. Thanks!

yajo:
Sorry, I'm not the best one to review this as I don't use it.

We use a local mail relay that applies [SRS](https://en.wikipedia.org/wiki/Sender_Rewriting_Scheme) automatically when needed.

jjscarafia:
Hi @Yajo 
Thanks for getting into this. I'm not an specialist on this regards, reading about SRS it says "without breaking the Sender Policy Framework (SPF)".
On our case we use mailgun and with mailgun the SPF is always align, I guess it do that (SRS), but the problem we solve with this module is to be aligned on  DMARC. Let me give an example:
* I've a milgun smtp server "mydomain.comar"
* If I sent emails from odoo they will be juan@mydomain.com
* If, a customer replies to odoo (for eg. pedro@mycustomer.com) and there are other followers that receives email, odoo will send emails with from "pedro@mycustomer.com" and DMARC won't be aligned. With this module, on this specific situation, the from will be change to something like "Pedro <notifications@mydomain.com>"

You still believe it's not interesting? Which is your solution on this regards?
Thanks!

yajo:
> You still believe it's not interesting? Which is your solution on this regards?

Well it can be for whoever doesn't use or like SRS, but SRS is good enough for us. :shrug: 

In the case you say, `pedro@mycustomer.com` would be rewritten to something similar to `SRS0=AhgY=FA=mycustomer.com=pedro@mydomain.comar`, thus validating against any SRS/DKIM/DMARC rules. Still `mydomain.com` can be included in the whitelist if your policies allow that one, to avoid rewriting it.

jjscarafia (approved):
LGTM. Tested on many customers

jjscarafia:
Thanks @Yajo 
We'll take a look to that approach.

I believe that the PR could be merged as the module already exists en OCA and this PR improves that module.

Thanks!

rafaelbn (approved):
:+1: LGTM

jjscarafia:
Thanks @rafaelbn !

COSTLAND:
Thanks for this interesting function.
What's the best use case when using this whitelist based logic?
- By default the outgoing email will use the mail.catchall.domain as domain suffix
- Seems no easy way for odoo email sending from mutli-domain setup, so is this function complements another module's function?

- I'm trying to understand if this module is part of solution for Odoo handling the multi-domain email sending and receiving.

Thanks for your time.

jjscarafia:
hi @HYROPES . No, this module does nothing related to multi-domain sending. Use case with this module.
I've my database with catchall domain @mycompany.com
I've two users:
a) user1@mycompany.com
b) user2@subsidiary.com

If I've two smtp servers (one for @mycompany.com and another for @subsidiariy.com), before this module, all mails where going to be send by the "first" smtp server and with a possible infraction of dmarc.

With this module, if user1 sends an email, the smtp server used would be @mycompany.com and if user2 sends an email the smtp server used would be @subsidary.com, both mails aligned on dmarc

The reply to / catchall email will be the same on both cases (catchall@mycompany.com), no change here.

It's clearer now?

COSTLAND:
Hi @jjscarafia Thanks for the reply, it's all make sense now. 
However, I do have a hard time to get this whitelist function working.
https://github.com/OCA/social/issues/653

ossimantylahti:
Big hand for the creator of this PR! This would fix Odoo e-mail spoofing. If you need any help on testing, documenting, debugging etc.  do not hesitate to ask.

PR #600, opened 2020-08-28
[13.0] [MIG] mail_attach_existing_attachment_account

Rad0van:
Standard migration


mail_attach_existing_attachment_account/wizard/account_invoice_send_view.xml
```diff
@@ -0,0 +1,23 @@
...
+<odoo>
+    <record id="account_invoice_send_wizard_form" model="ir.ui.view">
+        <field name="name">Send Invoice</field>
+        <field name="model">account.invoice.send</field>
+        <field name="inherit_id" ref="account.account_invoice_send_wizard_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='attachment_ids']" position="after">
+                <field name="model" invisible="1"/>
+                <field name="res_id" invisible="1"/>
+                <field name="can_attach_attachment" invisible="1"/>
```
gaikaz, line 15: I'm not sure why, but it looks like `can_attach_attachment` doesn't get set to `True` and so the checkboxes don't show up.
Tested on runbot and on a local machine.
If I remove the `invisible="1"` here and check the checkbox, the attachment checkboxes appear.
Am I missing something?

victoralmau:
Superseed on: https://github.com/OCA/social/pull/797

PR #604, opened 2020-09-01
[12.0][ADD] mail_preview_base & mail_preview_audio

etobella:
@pedrobaeza as promised, I created the module we talked about on OCA/dms#15

This PR adds two modules:

`mail_preview_base` creates a unified structure in order to add previews and creates a new field widget
`mail_preview_audio` is an example of how to add a new kind of file and allows to preview audio files

In order to test the audio, it is easy, we need to go to a thread and add an audio file, now we can preview it.


pedrobaeza (approved):
Tried and working properly. Not sure about the base module name, as there's no "field" involved here. Maybe it can be called `mail_preview_base`.

etobella:
was for the widget, but I will change for base :wink:

etobella:
Just a comment, the `mail_preview_base` must be LGPL in order to keep the compatibility to dms, as it was original LGPL

pedrobaeza:
👍

mail_preview_audio/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Preview audio files',
+    'summary': """
+        Allow to preview audio files""",
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'Creu Blanca,Odoo Community Association (OCA)',
+    'website': 'www.creublanca.es',
```
jarroyomorales: Website should be https://github.com/OCA/social/

jarroyomorales (approved):
Functional review and works as expected! Also did a minor comment.

PR #608, opened 2020-09-08
[13.0][FWD] mail_inline_css: fix transform hook

simahawk:
The mail module offers a better hook to manipulate html: `render_post_process`.
By using this, we make sure the transform is applied properly
before html tags sanitizing happens,
which can alter - if not screw - the final result.

Finally, it simplifies code :)

FWD PORT of https://github.com/OCA/social/pull/534


leemannd (approved):
Seems to be legit as it is the same code I saw in V11 (and that we are using).

I will try to test it this week.

vrenaville (approved):

santostelmo (approved):
LGTM! Tested on my local.

PR #609, opened 2020-09-08
[12.0][FWD] mail_inline_css: fix transform hook

simahawk:
The mail module offers a better hook to manipulate html: `render_post_process`.
By using this, we make sure the transform is applied properly
before html tags sanitizing happens,
which can alter - if not screw - the final result.

Finally, it simplifies code :)

FWD PORT of https://github.com/OCA/social/pull/534


len-foss (approved):
LGTM

simahawk:
fix on https://github.com/OCA/social/pull/892 getting merged

PR #614, opened 2020-09-28
[13.0][MIG] mail_preview_base & mail_preview_audio

jarroyomorales:
Standard migration of mail_preview_base and mail_preview_audio

@etobella @pedrobaeza


etobella (changes requested):
Only working when debug=assets is set, please review

etobella (approved):
:thinking: Now it is working without assets, seems like something is kept on memory until the debug assets is set :dizzy_face:

jarroyomorales:
@etobella Javascript was not loading the files always in the same order and was causing an error in some cases. It should be fixed now.
I don't know if this is necessary in v12, I have never had such problem but i may cherry-pick these commits just in case.

etobella:
@jarroyomorales I have found the same issue on 12.0, sometimes, it is not loading properly. Can you make the cherry-pick please

jarroyomorales:
@etobella here it is https://github.com/OCA/social/pull/615

LoisRForgeFlow (approved):
Works nicely! :+1:

PR #617, opened 2020-09-30
[13.0][UPD] Eficent -> ForgeFlow

MiquelRForgeFlow:
Update company name.

### DIFF
--- a/base_search_mail_content/__init__.py
+++ b/base_search_mail_content/__init__.py
@@ -1,5 +1,5 @@
-# Copyright 2016 Eficent Business and IT Consulting Services S.L.
-#   (http://www.eficent.com)
+# Copyright 2016 ForgeFlow S.L.
+#   (http://www.forgeflow.com)
 # Copyright 2016 Serpent Consulting Services Pvt. Ltd.
 #   (<http://www.serpentcs.com>)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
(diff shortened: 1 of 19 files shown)


mail_debrand/README.rst
```diff
@@ -20,7 +20,7 @@ Mail Debrand
     :target: https://runbot.odoo-community.org/runbot/205/13.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| 
+|badge1| |badge2| |badge3| |badge4|
```
pedrobaeza: You shouldn't touch these files manually...

JordiBForgeFlow: @pedrobaeza agreed. But for all versions? Perhaps in older versions the bot does not auto generate the readme. Not sure.

pedrobaeza: Yes, everywhere is updated. If edited manually, you must removed the automatic trim of trailing whitespaces in the IDE.

MiquelRForgeFlow: @pedrobaeza A new .rst file will be autogenerated independently if I touch the file or not. So, why bother if I touch it or not?

pedrobaeza: Because it generates a lot of diff noise.

MiquelRForgeFlow: The diff noise will exist independently of what I do. We can't avoid the noise.

HviorForgeFlow (approved):

HviorForgeFlow (approved):

mail_debrand/README.rst
MiquelRForgeFlow: @pedrobaeza the diff noise was removed. could we merge this?

PR #626, opened 2020-10-09
[FIX] mass_mailing_custom_unsubscribe: Pivot error

victoralmau:
Solved error in pivot view.

@Tecnativa TT26101

### DIFF
--- a/mass_mailing_custom_unsubscribe/readme/CONTRIBUTORS.rst
+++ b/mass_mailing_custom_unsubscribe/readme/CONTRIBUTORS.rst
@@ -6,3 +6,4 @@
   * David Vidal
   * Ernesto Tejeda
   * Pedro M. Baeza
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

mail_debrand/i18n/it.po
```diff
@@ -6,38 +6,40 @@ msgid ""
 msgstr ""
```
pedrobaeza: Why is this change here?

victoralmau: Very strange, obviously .po file not necessary to modify. I will check my commit again and solve it.

pedrobaeza:
Travis is red

victoralmau:
> Travis is red

Locked by: https://github.com/OCA/social/pull/630

PR #631, opened 2020-10-23
[FIX] re.search will return None when nothing found. So we have to check for this

wpichler:


yajo (approved):

yajo:
I "fast" track it as it is pretty obvious.

pedrobaeza:
Let's call it correctly :wink: 

/ocabot merge patch

pedrobaeza:
Is this needed for v13 as well?

wpichler:
> Is this needed for v13 as well?

just looked at v13 - it is also needed there...

pedrobaeza:
Cherry-pick to v13 in 05905e50c383e0ebe520e4f3a941853811d47495

PR #637, opened 2020-11-02
[13.0] [FIX] mail_debrand: Fixed travis error.

Chandresh-SerpentCS:


mail_debrand/__manifest__.py
```diff
@@ -14,6 +14,5 @@
     "license": "AGPL-3",
     "installable": True,
     "depends": ["mail"],
-    "development_status": "Stable",
```
pedrobaeza, line 17: Please change it to `Production/Stable` instead

Chandresh-SerpentCS: @pedrobaeza You mean `Production/Stable` would be a key?

pedrobaeza: It would be the proper value for the key `development_status`.

Chandresh-SerpentCS: Ok, I will modify the value.

bodedra (approved):
LG :+1:

bodedra:
@pedrobaeza Can we fast-track this?

pedrobaeza (approved):

PR #633, opened 2020-10-26
[12.0][FIX] mass_mailing_partner: partner merge

victoralmau:
Solved error related to merge contacts (in update records related to partners, exists multiple mailing contact related to same partner).

**Reproduce error**
Create partner A > Add to mailing list A
Create partner B > Add to mailing list A + B
Merge partner A + B into A

**Expected result**:
Record related to Partner A > Mailing list A + B

Result (with errors):
Record related to Partner A > Mailing list A
Record related to Partner A > Mailing list A + B

@Tecnativa TT26380

PR #641, opened 2020-11-09
[13.0][FIX] mass_mailing_partner: partner merge

victoralmau:
Related to https://github.com/OCA/social/pull/633 in V13

@Tecnativa TT26380

### DIFF
--- a/mass_mailing_partner/wizard/partner_merge.py
+++ b/mass_mailing_partner/wizard/partner_merge.py
@@ -1,0 +1,25 @@
+# Copyright 2020 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class BasePartnerMergeAutomaticWizard(models.TransientModel):
+    _inherit = "base.partner.merge.automatic.wizard"
+
+    def _merge(self, partner_ids, dst_partner=None, extra_checks=True):
+        if dst_partner:
+            contacts = self.env["mailing.contact"].search(
+                [("partner_id", "in", partner_ids)]
+            )
+            if contacts:
+                contacts = contacts.sorted(
+                    lambda x: 1 if x.partner_id == dst_partner else 0
+                )
+                list_ids = contacts.mapped("list_ids").ids
+                contacts[1:].unlink()
+                contacts[0].partner_id = dst_partner
+                contacts[0].list_ids = [(4, x) for x in list_ids]
+        return super()._merge(
+            partner_ids, dst_partner=dst_partner, extra_checks=extra_checks
+        )
--- a/mass_mailing_partner/readme/CONTRIBUTORS.rst
+++ b/mass_mailing_partner/readme/CONTRIBUTORS.rst
@@ -9,5 +9,6 @@
     * Ernesto Tejeda
     * Victor M.M. Torres
     * Manuel Calero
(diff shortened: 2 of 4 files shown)


pedrobaeza:
Check 12.0 PR comment

pedrobaeza (approved):

joao-p-marques (approved):

mass_mailing_partner/wizard/partner_merge.py
```diff
@@ -0,0 +1,24 @@
...
+class BasePartnerMergeAutomaticWizard(models.TransientModel):
+    _inherit = "base.partner.merge.automatic.wizard"
+
+    def _merge(self, partner_ids, dst_partner=None, extra_checks=True):
+        if dst_partner:
+            contact_ids = self.env["mailing.contact"].search(
+                [("partner_id", "in", partner_ids)]
+            )
+            if contact_ids:
+                contact_ids.sorted(lambda x: 1 if x.partner_id == dst_partner else 0)
```
pedrobaeza: I'm thinking... isn't this returning the sorted result as returned instead of sorting inline? If so, you need to:

```suggestion
                contacts = contact_ids.sorted(lambda x: 1 if x.partner_id == dst_partner else 0)
```

And please, always name properly the variables. It's `contacts`, not `contact_ids`.

pedrobaeza (changes requested):

mass_mailing_partner/wizard/partner_merge.py
```diff
@@ -0,0 +1,24 @@
...
+
+from odoo import models
+
+
+class BasePartnerMergeAutomaticWizard(models.TransientModel):
+    _inherit = "base.partner.merge.automatic.wizard"
+
+    def _merge(self, partner_ids, dst_partner=None, extra_checks=True):
+        if dst_partner:
+            contact_ids = self.env["mailing.contact"].search(
```
pedrobaeza: This is already `contacts` as well, as it's returning a recordset.

mass_mailing_partner/wizard/partner_merge.py
```diff
@@ -0,0 +1,24 @@
...
+    def _merge(self, partner_ids, dst_partner=None, extra_checks=True):
+        if dst_partner:
+            contacts = self.env["mailing.contact"].search(
+                [("partner_id", "in", partner_ids)]
+            )
+            if contacts:
+                contacts = contacts.sorted(
+                    lambda x: 1 if x.partner_id == dst_partner else 0
+                )
+                contacts[1:].unlink()
```
pedrobaeza, line 20: You are doing this in the wrong order: removing them first, then next call to `mapped` will fail.

victoralmau:
Renamed variables now, unlink move to before (it's necessary to unlink before to assign partner_id to prevent constraint and mapped list ids need to assign in variable too).

I'm thinking need to apply to in V12, what do you think about it?

pedrobaeza:
@victoralmau you can't unlink first and then use that recordset later. You will have to save values in an intermediate variable if so.

victoralmau:
> @victoralmau you can't unlink first and then use that recordset later. You will have to save values in an intermediate variable if so.

Yes, changed now.

pedrobaeza (approved):

pedrobaeza:
12.0 synced with b3fbb0058da07ca9086fe6ef4dbbcd99592db2a3

PR #650, opened 2020-12-22
[ADD] mail_notification_custom_subject: Apply a custom subject to mail notifications

joao-p-marques:
@Tecnativa
TT27191

ping @pedrobaeza


mail_notification_custom_subject/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "summary": "Apply a custom subject to mail notifications",
+    "version": "12.0.1.0.0",
+    "category": "Social Network",
+    "website": "https://github.com/OCA/social/",
+    "author": "Tecnativa, Eficent, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": [
+        "mail",
+        "account", # needed for tests
```
victoralmau:
```suggestion
```
Need to user base models (`res.partner` model for example).

mail_notification_custom_subject/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2020 Tecnativa - João Marques
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Mail Notification Custom Subject",
+    "summary": "Apply a custom subject to mail notifications",
+    "version": "12.0.1.0.0",
+    "category": "Social Network",
+    "website": "https://github.com/OCA/social/",
+    "author": "Tecnativa, Eficent, Odoo Community Association (OCA)",
```
victoralmau:
```suggestion
    "author": "Tecnativa, Odoo Community Association (OCA)",
```

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -0,0 +1,32 @@
...
+    ):
+        vals = super(MailThread, self).message_post(
+            body=body, subject=subject,
+            message_type=message_type, subtype=subtype,
+            parent_id=parent_id, attachments=attachments,
+            notif_layout=notif_layout, add_sign=add_sign, model_description=model_description,
+            mail_auto_delete=mail_auto_delete, **kwargs
+        )
+        mail_template = self.env["mail.message.custom.subject"].search([
+            ("model_id.model", "=", self._name),
```
victoralmau:
```suggestion
            ("model_id.model", "=", vals['model']),
```

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -0,0 +1,32 @@
...
+            ("subtype_ids", "=", vals["subtype_id"].id),
+        ])
+        subject_override = mail_template.render_subject_template(self._name, self.id)
```
victoralmau:
```suggestion
        subject_override = mail_template.render_subject_template(vals['model'], vals['res_id'])
```

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,158 @@
...
+        help="Message subtypes followed, meaning subtypes that will be pushed onto the user's Wall")
+    subject_template = fields.Char(
+        string='Subject Template', required=True,
+        help="Subject (placeholders may be used here)"
+    )
+    replace = fields.Boolean(
+        string="Replace", default=True,
+        help="Whether to replace or append to other templates that apply to a"
+        " given context"
+    )
```
victoralmau: Append allways after right?
Maybe need to change `replace` field to `position` (select): replace/append_after/append_before

What do you think?

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
```diff
@@ -0,0 +1,108 @@
...
+            'type': 'in_invoice',
+            'invoice_line_ids': [(0, 0, {
+                'product_id': product.id,
+                'quantity': 1.0,
+                'price_unit': 100.0,
+                'name': 'product that cost 100',
+                'account_id': self.invoice_line_account,
+            })]
+        })
+        self.invoice_2.number = 'INV/TEST'
```
victoralmau:
```suggestion
```
Need to remove everithing about `account` addon and use `res.partner` model

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
```diff
@@ -0,0 +1,108 @@
...
+        # Delete current DB records for testing
+        self.env["mail.message.custom.subject"].search([]).unlink()
+        self.template = self.env["mail.message.custom.subject"].create({
+            "name": "Test template 1",
+            "model_id": self.env.ref("account.model_account_invoice").id,
```
victoralmau:
```suggestion
            "model_id": self.env.ref("base.model_res_partner").id,
```

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
```diff
@@ -0,0 +1,108 @@
...
+            "subtype_ids": [(6, 0, [
+                self.env.ref('mail.mt_comment').id,
+            ])],
+            "subject_template": "${object.number or 'n/a'} and something more",
```
victoralmau:
```suggestion
            "subject_template": "${object.name or 'n/a'} and something more",
```

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
```diff
@@ -0,0 +1,108 @@
...
+            "subtype_ids": [(6, 0, [
+                self.env.ref('mail.mt_comment').id,
+            ])],
+            "subject_template": "${' and yet something else'}",
+            "replace": False,
+        })
+        # Send message in invoice
+        mail_message_2 = self.invoice_1.message_post(body='Test', subtype='mail.mt_comment')
+        # Get message and check subject
+        self.assertEquals(mail_message_2.subject, "n/a and something different and yet something else")
```
victoralmau: Test with partner messages only.

victoralmau (changes requested):
Need to add .pot files

mail_notification_custom_subject/models/mail_thread.py
joao-p-marques: I ended up not applying this because, after the latest re-factor, I don't have `vals` available, and `self._name` seems like the best option.

mail_notification_custom_subject/models/mail_thread.py
joao-p-marques: Same here as above.

mail_notification_custom_subject/models/mail_message_custom_subject.py
joao-p-marques: That may be an interesting option. @pedrobaeza WDYT?

pedrobaeza: OK

joao-p-marques:
Changes made @victoralmau 

> Need to add .pot files

I think this is added automatically by OCA Bot, right @pedrobaeza?

pedrobaeza:
POT is added automatically, but please get used to provide such file as side effect of the most important one: the Spanish translation for new modules, for having them translated since the beginning.

pedrobaeza:
Red Travis

joao-p-marques:
Ok, thanks for the explanation. I will add it and check travis.

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,257 @@
...
+        :param str model: model name of the document record this mail is related to.
+        :param int res_ids: list of ids of document records those mails are related to.
+        """
+        multi_mode = True
+        if isinstance(res_ids, pycompat.integer_types):
+            multi_mode = False
+            res_ids = [res_ids]
+
+        results = dict.fromkeys(res_ids, u"")
+
```
victoralmau: ### 
```suggestion
```

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,257 @@
...
+            mako_env = (
+                mako_safe_template_env
+                if self.env.context.get("safe")
+                else mako_template_env
+            )
+            template = mako_env.from_string(tools.ustr(template_txt))
+        except Exception:
+            _logger.info("Failed to load template %r", template_txt, exc_info=True)
+            return multi_mode and results or results[res_ids[0]]
+
```
victoralmau:
```suggestion
```

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,257 @@
...
+                )
+                raise UserError(
+                    _("Failed to render template %r using values %r")
+                    % (template, variables)
+                    + "\n\n%s: %s" % (type(e).__name__, str(e))
+                )
+            if render_result == u"False":
+                render_result = u""
+            results[res_id] = render_result
+
```
victoralmau:
```suggestion
```

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,257 @@
...
+        if post_process:
+            for res_id, result in results.items():
+                results[res_id] = self.render_post_process(result)
+
```
victoralmau:
```suggestion
```

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -0,0 +1,40 @@
...
+                subtype = 'mail.%s' % subtype
+            subtype_id = self.env['ir.model.data'].xmlid_to_res_id(subtype)
+        mail_template = self.env["mail.message.custom.subject"].search([
+            ("model_id.model", "=", self._name),
+            ("subtype_ids", "=", subtype_id),
+        ])
+        subject_override = mail_template.render_subject_template(
+            self._name, self.id
+        )
+        subject = subject_override or subject
```
victoralmau:
```suggestion
        if mail_template:
            subject_override = mail_template.render_subject_template(
                self._name, self.id
            )
            subject = subject_override or subject
```

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
```diff
@@ -0,0 +1,122 @@
...
+            }
+        )
+        self.partner_2 = self.env["res.partner"].create(
+            {
+                "name": "Test partner 2",
+                "supplier": True,
+            }
+        )
+        # Delete current DB records for testing
+        self.env["mail.message.custom.subject"].search([]).unlink()
```
victoralmau:
```suggestion
```

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+
+    def test_email_subject_template_overrides(self):
+        # Send message in invoice
+        mail_message_1 = self.partner_1.message_post(
+            body="Test", subtype="mail.mt_comment"
+        )
+        # Get message and check subject
+        self.assertEquals(mail_message_1.subject, "Test partner 1 and something more")
+
```
victoralmau, line 48:
```suggestion
```

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
```diff
@@ -0,0 +1,122 @@
...
+        # Send message in invoice 2
+        mail_message_2 = self.partner_2.message_post(
+            body="Test", subtype="mail.mt_comment"
+        )
+        # Get message and check subject
+        self.assertEquals(mail_message_2.subject, "Test partner 2 and something more")
+
```
victoralmau, line 55:
```suggestion
```

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+        # Send message in invoice
+        mail_message_1 = self.partner_1.message_post(
+            body="Test", subtype="mail.mt_comment"
+        )
+        # Get message and check subject
+        self.assertEquals(
+            mail_message_1.subject, "Test partner 1 and something different"
+        )
+
```
victoralmau:
```suggestion
```

mail_notification_custom_subject/views/mail_notification_custom_subject_views.xml
```diff
@@ -0,0 +1,50 @@
...
+                            <field name="subject_template" placeholder="Subject (placeholders may be used here)"/>
+                            <field name="model_id"/>
+                            <field name="subtype_ids"/>
+                            <field name="position"/>
+                        </group>
+                    </sheet>
+                </form>
+            </field>
+        </record>
+
```
victoralmau, line 24:
```suggestion
```

mail_notification_custom_subject/views/mail_notification_custom_subject_views.xml
```diff
@@ -0,0 +1,50 @@
...
+            <field name="arch" type="xml">
+                <tree string="Templates">
+                    <field name="name"/>
+                    <field name="model_id"/>
+                    <field name="subtype_ids"/>
+                    <field name="subject_template"/>
+                </tree>
+            </field>
+        </record>
+
```
victoralmau, line 37:
```suggestion
```

mail_notification_custom_subject/views/mail_notification_custom_subject_views.xml
```diff
@@ -0,0 +1,50 @@
...
+        <record model="ir.actions.act_window" id="action_mail_notification_custom_subject_tree_all">
+            <field name="name">Subject Replacement Templates</field>
+            <field name="res_model">mail.message.custom.subject</field>
+            <field name="view_type">form</field>
+            <field name="view_mode">form,tree</field>
+            <field name="view_id" ref="mail_notification_custom_subject_tree" />
+        </record>
+
```
victoralmau, line 45:
```suggestion
```

mail_notification_custom_subject/views/mail_notification_custom_subject_views.xml
```diff
@@ -0,0 +1,50 @@
...
+        <menuitem id="menu_mail_notification_custom_subject" parent="base.menu_email" action="action_mail_notification_custom_subject_tree_all"
+                  sequence="21"/>
+
```
victoralmau, line 48:
```suggestion
```

victoralmau (changes requested):
Please, remove empty lines and other changes.

mail_notification_custom_subject/models/mail_thread.py
joao-p-marques: Thank you! :+1: 
I don't think it would cause problems, though, as the search would return an empty recordset, which [will then be iterated over](https://github.com/OCA/social/pull/650/files#diff-9c0df1d72822413b9e7331e329dd186ca0093c700694cabe007f8def001bf5baR172), right?
Just a curiosity :smile: Change made :+1:

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
joao-p-marques: I was using this to get correct test results in case there are some records presents in the database beforehand that would produce different results. 
This transaction is then rolled back at the end of the test, as it is a `TransactionCase`, so it wouldn't affect the real DB. 
What is the correct way to do it, then?

pedrobaeza: This is risky indeed. Perform a search before doing `message_post` and after, and then the difference will be the new created message.

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,243 @@
...
+        ],
+        string="Position",
+        default="replace",
+        help="Whether to replace, append at beggining or append at end to other"
+        " templates that apply to a given context",
+    )
+
+    @api.multi
+    def render_subject_template(self, model, res_id):
+        subject_override = ""
```
victoralmau:
```suggestion
        subject_override = False
```

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
joao-p-marques: Thanks. I ended up deleting these lines, as suggested by @victoralmau as I misunderstood the problem :sweat_smile:

victoralmau (changes requested):

mail_notification_custom_subject/models/mail_message_custom_subject.py
joao-p-marques: Thanks. However, I can't set this to `False` as I later need to append other strings to this. What I can do instead is to change the return statement to:
```python
return subject_override or False
```
WDYT?

joao-p-marques:
All changes done @victoralmau

victoralmau (approved):

mail_notification_custom_subject/README.rst
```diff
@@ -0,0 +1,99 @@
...
+    :target: https://translation.odoo-community.org/projects/social-12-0/social-12-0-mail_notification_custom_subject
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/205/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module allows you to specify templates to override the emails sent by Odoo
+in the chatter.
```
pedrobaeza:
```suggestion
This module allows you to specify templates to override the subject on the notification emails
sent by Odoo.
```

mail_notification_custom_subject/README.rst
```diff
@@ -0,0 +1,99 @@
...
+=====
+
+To use this module, you need to:
+
+* Activate access to **Technical Features** (debug mode).
+* Go to **Settings > Technical > Email > Subject Replacement Templates**
+* Create a new template.
+
+    * The field **Subject Template** accepts `Jinjah<ttps://jinja.palletsprojects.com/en/2.11.x/>` expressions.
+    * The field **Replace** specifies if the template should replace existing templates or append to them.
```
pedrobaeza:
```suggestion
    * The field **Replace** specifies if the template should replace existing content or append to it.
```

pedrobaeza: Missing model definition

pedrobaeza: Till here should go on CONFIGURE.rst, not USAGE.rst

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,243 @@
...
+from werkzeug import urls
+
+from odoo import _, api, fields, models, tools
+from odoo.exceptions import UserError
+from odoo.tools import pycompat
+
+_logger = logging.getLogger(__name__)
+
+
+def format_date(env, date, pattern=False):
```
pedrobaeza: Why redefining all these methods? Can't you import them?
This is linked with the following comment

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,243 @@
...
+            if template.position == "replace":
+                subject_override = rendered_subject_templates
+            elif template.position == "append_before":
+                subject_override = rendered_subject_templates + subject_override
+            elif template.position == "append_after":
+                subject_override += rendered_subject_templates
+        return subject_override or False
+
+    @api.multi
+    def _render_template(self, template_txt, model, res_ids, post_process=False):
```
pedrobaeza: Aren't you able to call existing methods in mail.template or similar to avoid to rewrite all this stuff?

mail_notification_custom_subject/models/mail_message_custom_subject.py
pedrobaeza: This is not taking into account existing "core" subject. IMO this should be done on `mail.thread`, not here.

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -0,0 +1,41 @@
...
+        mail_template = self.env["mail.message.custom.subject"].search([
+            ("model_id.model", "=", self._name),
+            ("subtype_ids", "=", subtype_id),
+        ])
+        if mail_template:
+            subject_override = mail_template.render_subject_template(
+                self._name, self.id
+            )
+            subject = subject_override or subject
+        return super(MailThread, self).message_post(
```
pedrobaeza:
```suggestion
        return super().message_post(
```

mail_notification_custom_subject/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* Tecnativa <https://www.tecnativa.com>
+
+    * João Marques
```
pedrobaeza, line 4: Please include me as designer

mail_notification_custom_subject/static/description/icon.svg
```diff
@@ -0,0 +1,248 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
pedrobaeza: This icon is not correct (it's from `mail_debrand`). Adapt it or remove it.

pedrobaeza (changes requested):

joao-p-marques:
Thanks for the suggestions @pedrobaeza!
Changes made

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class MailMessageCustomSubject(models.Model):
+    _name = "mail.message.custom.subject"
+    _description = "Mail Message Custom Subject"
+
+    name = fields.Char(string="Template Name")
+    model_id = fields.Many2one(
+        comodel_name="ir.model",
+        string="Models",
```
pedrobaeza:
```suggestion
        string="Model",
```

mail_notification_custom_subject/README.rst
```diff
@@ -0,0 +1,106 @@
...
+
+Configuration
+=============
+
+
+* Activate access to **Technical Features** (debug mode).
+* Go to **Settings > Technical > Email > Subject Replacement Templates**
+* Create a new template.
+
+    * The field **Models** specifies the models to which the subject template should apply in the emails sent by Odoo
```
pedrobaeza, line 44:
```suggestion
    * The field **Model** specifies the model to which the subject template should apply in the notification emails sent by Odoo.
```

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -0,0 +1,49 @@
...
+        parent_id=False, attachments=None,
+        notif_layout=False, add_sign=True, model_description=False,
+        mail_auto_delete=True, **kwargs
+    ):
+        subtype_id = kwargs.get('subtype_id', False)
+        if not subtype_id:
+            subtype = subtype or 'mt_note'
+            if '.' not in subtype:
+                subtype = 'mail.%s' % subtype
+            subtype_id = self.env['ir.model.data'].xmlid_to_res_id(subtype)
```
pedrobaeza: Don't fail if invalid:
```suggestion
            subtype_id = self.env['ir.model.data'].xmlid_to_res_id(subtype, raise_if_not_found=False)
        if subtype_id:
            ...
```

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -0,0 +1,49 @@
...
+        mail_templates = self.env["mail.message.custom.subject"].search([
```
pedrobaeza: Use proper variable names
```suggestion
        custom_subjects = self.env["mail.message.custom.subject"].search([
```

pedrobaeza (changes requested):

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -0,0 +1,49 @@
...
+            ("model_id.model", "=", self._name),
+            ("subtype_ids", "=", subtype_id),
+        ])
+        for template in mail_templates:
+            rendered_subject_template = self.env["mail.template"]._render_template(
+                template_txt=template.subject_template,
+                model=self._name, res_ids=self.id
+            )
```
pedrobaeza: I'm thinking that we shouldn't interrupt by any chance the notification sending:
```suggestion
            try:
                rendered_subject_template = self.env["mail.template"]._render_template(
                    template_txt=template.subject_template,
                    model=self._name, res_ids=self.id
                )
            except Exception:
                rendered_subject_template = False
```

joao-p-marques: Ok, thanks! I just made the change. 
However, can't that make it harder to debug if a subject template is well constructed or not? The `_render_template` function throws some helpful exceptions for that (https://github.com/odoo/odoo/blob/fd56023d26dcebb1503fe8d4ece6d6125dfc3e90/addons/mail/models/mail_template.py#L338). Should I catch them and do something with them?

pedrobaeza: You would already have the log entry, so no, nothing should block the notification.

joao-p-marques: Ok, nice! :+1:

pedrobaeza:
And the rest of the asked changes?

joao-p-marques:
Sorry, I missed those earlier. Changes made.
I also added 1 more check bellow to make sure we don't get `subject = None` when trying to append to an existing one.

pedrobaeza:
Travis is red. Please also get full coverage.

joao-p-marques:
The latest commit should improve test coverage. Everything else is done, I think @pedrobaeza

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -0,0 +1,68 @@
...
+        subtype=None,
+        parent_id=False,
+        attachments=None,
+        notif_layout=False,
+        add_sign=True,
+        model_description=False,
+        mail_auto_delete=True,
+        **kwargs
+    ):
+        new_message = super().message_post(
```
pedrobaeza: If you call super before, the mail sending is done in this stage AFAIK

joao-p-marques:
From my testing the email is only sent after I return the new message 
at the end.

mail_notification_custom_subject/models/mail_thread.py
joao-p-marques: From my testing the email is only sent after I return the new message
at the end.

pedrobaeza (approved):

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -0,0 +1,75 @@
...
+                )._get_record_name({})
+            for template in custom_subjects:
+                try:
+                    rendered_subject_template = self.env[
+                        "mail.template"
+                    ]._render_template(
+                        template_txt=template.subject_template,
+                        model=self._name,
+                        res_ids=self.id,
+                    )
```
joao-p-marques, line 54: Should we still check that `rendered_subject_template` has something?
```suggestion
                    )
                    if not rendered_subject_template:
                        continue
```

joao-p-marques (commented):
Thank you for the changes @pedrobaeza :smiley: 

I just have a small comment/doubt.

mail_notification_custom_subject/models/mail_thread.py
pedrobaeza: No being everything inside try

joao-p-marques: Ok, seems good then :+1:

PR #652, opened 2021-01-07
[MIG] mass_mailing_unique migration to 13.0

vince-dynapps:


mass_mailing_unique/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 {
     "name": "Unique records for mass mailing",
     "summary": "Avoids duplicate mailing lists and contacts",
-    "version": "12.0.1.0.1",
+    "version": "13.0.1.0.1",
```
atchuthan, line 9: 13.0.1.0.1 => 13.0.1.0.0

atchuthan (changes requested):
Please do not change the model name directly or provide a data migration such that we do not lose the data when migrating the DB of customers from V12 to v13

If other OCA members think this change in the model name is acceptable, then I will change the PR review afterwards

atchuthan:
Also, you might want to run pre-commit to make Travis green
For more info: https://travis-ci.com/github/OCA/social/jobs/469475536

PR #657, opened 2021-01-15
[14.0][ADD] mail_partial_autodelete

kevinkhao:


sebastienbeau:
@kevinkhao can you check why test fail?

mail_partial_autodelete/models/mail_mail.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 Akretion (http://www.akretion.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class MailMail(models.Model):
+    _inherit = "mail.mail"
+
+    def unlink(self):
```
sebastienbeau, line 14: It will ve better to inherit def _send 

https://github.com/odoo/odoo/blob/7368cbfcb477aaf156c1c5ad93c717720e048a23/addons/mail/models/mail_mail.py#L274

And pass a key in the context "skip_unlink"

Then we inherit the def unlink and skip it if we have this key in the context

mail_partial_autodelete/models/mail_mail.py
```diff
@@ -0,0 +1,20 @@
...
+        self = self.with_context(autodelete_skip_unlink=True)
+        super()._send(auto_commit, raise_exception, smtp_session)
+
+    def unlink(self):
+        to_unlink = self
+        if self.env.context.get("autodelete_skip_unlink"):
+            to_purge = self.filtered(lambda r: r.auto_delete)
+            to_purge.write({"body_html": "", "body": ""})
+            to_unlink -= to_purge
+        super(MailMail, to_unlink).unlink()
```
sebastienbeau:
```suggestion
        if self.env.context.get("autodelete_skip_unlink"):
            return self.write({"body_html": "", "body": ""})
        else:
            return super().unlink()
```
unlink will be called only if it's autodelete so filtering on autodelete is not necessary

mail_partial_autodelete/models/mail_mail.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import models
+
+
+class MailMail(models.Model):
+    _inherit = "mail.mail"
+
+    def _send(self, auto_commit=False, raise_exception=False, smtp_session=None):
+        self = self.with_context(autodelete_skip_unlink=True)
+        super()._send(auto_commit, raise_exception, smtp_session)
```
sebastienbeau:
```suggestion
        return super()._send(auto_commit, raise_exception, smtp_session)
```

sebastienbeau:
@kevinkhao can you rebase?

PR #658, opened 2021-01-21
[12.0][MIG] mail_edit

Kiplangatdan:


mail_edit/security/mail_edit_security.xml
```diff
@@ -1,27 +1,23 @@
...
-            <field name="name">Edit own mail messages</field>
-            <field name="comment">Members of this group are able to edit mail messages.</field>
-            <field name="category_id" ref="base.module_category_hidden"/>
-            <field name="users" eval="[(4, ref('base.user_root'))]"/>
-        </record>
+    <record model="res.groups" id="group_mail_edit_mover">
+        <field name="name">Move own mail messages</field>
+        <field name="comment">Members of this group are able to move mails.</field>
+        <field name="category_id" ref="base.module_category_hidden"/>
+        <field name="users" eval="[(4, ref('base.user_root'))]"/>
```
thomaspaulb: Upon installing the module I saw that `admin` user does not have these rights by default, I think it should now be `base.user_admin` since `user_root` has become the technical user (Odoo Bot).

mail_edit/security/mail_edit_security.xml
```diff
@@ -1,27 +1,23 @@
 <?xml version="1.0" encoding="utf-8"?>
-<openerp>
-    <data noupdate="1">
+<odoo noupdate="1">
+    <record model="res.groups" id="group_mail_edit_editor" >
+        <field name="name">Edit own mail messages</field>
+        <field name="comment">Members of this group are able to edit mail messages.</field>
+        <field name="category_id" ref="base.module_category_hidden"/>
+        <field name="users" eval="[(4, ref('base.user_root'))]"/>
```
thomaspaulb: Same `base.user_admin` here.

mail_edit/readme/USAGE.rst
```diff
@@ -0,0 +1,7 @@
+To use this module, you need to:
+
+* Go to *Message* *Chatter* in any model, click the Edit (pen) Icon to open up
+  the 'Edit or Move' wizard.
+* To edit, edit the Mail as necessary and click Save.
```
thomaspaulb, line 5: I've noticed an issue that was already there in 8.0: if you just want to edit a message but you also have 'move' rights, the "destination object" field is mandatory and you're basically forced to move the message elsewhere. I think that `destination_object_id` should [not be required in the view](https://github.com/OCA/social/blob/8.0/mail_edit/views/compose_message.xml#L19).

mail_edit/static/src/js/mail_edit_old.js
```diff
@@ -0,0 +1,50 @@
+/* © 2014-2017 Sunflower IT <www.sunflowerweb.nl>*/
```
thomaspaulb: This 'old' file should be removed.

mail_edit/static/src/js/mail_edit.js
```diff
@@ -0,0 +1,75 @@
...
+            })
+            .done(function (action) {
+                action.res_id = msg_id;
+                self.do_action(action, {
+                    on_close: function () {
+                        location.reload();
+                        //self.trigger_up('refresh_on_fly');
+                        //this.reload.bind(this)
+                        //self.trigger_up('reload', { keepChanges: true });
+                         //self.trigger.bind(self, 'need_refresh')
```
thomaspaulb: I'll take a look at this.

thomaspaulb (changes requested):

mail_edit/static/src/js/mail_edit.js
thomaspaulb: @Kiplangatdan That was more work than I thought, Odoo has many routines for live adding of new messages, but not for updating existing messages alas. Can you handle the remaining points, then I think we're good to go.

Kiplangatdan: @thomaspaulb Reload is now even better than before! 
We can now push it for review.

thomaspaulb:
@hbrunn This is ready as a test version.

hbrunn:
I just created a note on http://3461265-658-0932be.runbot1-2.odoo-community.org/web?#id=26&action=125&model=res.partner&view_type=form&menu_id=94 (just some partner) as user demo but can't edit anything. Is this expected?

thomaspaulb:
@hbrunn There are security groups to enable, either to edit your own notes only or to edit other people's notes as well.

hbrunn:
then the button shouldn't be there too if the demo user doesn't have the correct group, should it? As the demo user probably needs the group for tests anyways, I suggest to add them to the edit group in demo data

hbrunn:
screenshot:

image

mail_edit/models/compose_message.py
```diff
@@ -0,0 +1,51 @@
...
+        "Destination object",
+        help="Object where the message will be moved to")
+
+    @api.model
+    def _message_read_dict_postprocess(self, messages, message_tree):
+        res = super(MailMessage, self)._message_read_dict_postprocess(
+            messages, message_tree)
+        for message_dict in messages:
+            # Check if current user is a superuser
+            if self.env.user.has_group('mail_edit.group_mail_edit_superuser'):
```
ntsirintanis, line 47: Why is this needed?

Kiplangatdan: @ntsirintanis Edit/delete buttons will be [shown](https://github.com/sunflowerit/social/blob/12.0-mig-mail-edit/mail_edit/static/src/xml/mail_edit.xml#L5) based on these rights.

mail_edit/static/src/js/tour.js
```diff
@@ -0,0 +1,56 @@
...
+}, {
+    trigger: '.o_mail_discuss .o_thread_delete_icon .o_mail_delete:last',
+    content: _t("Click the Delete Icon on the last message"),
+    position: "right",
+}, {
+    trigger: 'button span:contains(Ok)',
+    extra_trigger: '.modal-footer',
+    content: _t("Confirm Deletion"),
+    run: 'click',
+}]);
```
thomaspaulb, line 65: This tour runs well, but to make a test out of it and have it called by Travis, you'll need to add something like this:

https://github.com/OCA/server-tools/blob/11.0/profiler/tests/test_profiling_ui.py

thomaspaulb (changes requested):

mail_edit/static/src/js/tour.js
thomaspaulb: With login="demo", probably.

thomaspaulb (approved):
Code, test and functional OK.

mail_edit/static/src/js/tour.js
```diff
@@ -0,0 +1,56 @@
+odoo.define('mail_edit.tour', function (require) {
+"use strict";
+
+var core = require('web.core');
+var tour = require('web_tour.tour');
+
+var _t = core._t;
+
+tour.register('mail_edit_tour', {
+    url: "/web",
```
thomaspaulb, line 10: Travis is running into this:

image

Could it be that we have to add the `test` and `wait_for` lines? They seem to be all around in Odoo 12.0

image

thomaspaulb (changes requested):

marylla:
@Kiplangatdan Can you check the test please? I really would like to this feature in version 12. :)

mail_edit/tests/test_ui.py
```diff
@@ -0,0 +1,15 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
+
+import odoo.tests
+
+
+@odoo.tests.tagged("post_install", "-at_install")
+class TestMailEdit(odoo.tests.HttpCase):
+    def test_tour(self):
+        self.phantom_js(
+            "/",
```
hbrunn: this should be '/web', otherwise you'll end up on the homepage, not the backend

Kiplangatdan: @hbrunn I have made the fix and did a rebase. Waiting for the runbot.

mail_edit/static/src/js/tour.js
```diff
@@ -0,0 +1,56 @@
...
+    content: _t("Add some text message to the composer chat box"),
+    position: "top",
+    width: 350,
+    run: function (actions) {
+        actions.text("Sample Message to be edited & Deleted!", this.$anchor);
+    },
+}, {
+    trigger: '.o_mail_discuss .o_thread_composer .o_composer_button_send',
+    content: _t("Post your message on the thread"),
+    position: "top",
```
thomaspaulb: The test now runs, but looking at the HTML output it gives, it seems that the screen that opens after this step does not have an element with `o_mail_info`, and thus also no `.o_thread_edit_icon .o_mail_edit`, and so the next step fails.

clementmbr:
Hi @Kiplangatdan @thomaspaulb @marylla @ntsirintanis 
I'm wondering if anyone is using this module in production. Is the module ready to be used event if the test are not working ? I might be interested to port in in v14...

Thanks!

thomaspaulb:
@clementmbr Yes it's being used.

hbrunn:
indeed it is. I don't believe you can recycle a lot of code here, because the js part changed quite a bit between 12 and 14

Martronic-SA:
Tested on V12 and working as expected!! :+1:

thomaspaulb:
@OCA/social-maintainers can this be reopened?

thomaspaulb:
Sorry, nevermind - I'll open a new MR for 14.0.

PR #659, opened 2021-01-21
[13.0] [FIX] mail_tracking

nmadhoc:
When a user try to resend message gets an error stating that it's only allowed for users who belong to the Administration/Settings group


pedrobaeza (approved):

jjscarafia (approved):
+1 functional review + technical review

bodedra (approved):
LG :+1:

PR #624, opened 2020-10-07
[FIX] mail_tracking: Discuss failed messages: Button and domain

Tardo:
Fixes the domain to found the active user in the message destinations.

It also fixes the button visibility.

cc @tecnativa

PR #665, opened 2021-02-01
[13.0][IMP] mail_tracking: Discuss failed messages: Button and domain

Tardo:
fw-port of https://github.com/OCA/social/pull/624

cc @tecnativa


pedrobaeza:
When I press the button "Set all as reviewed", no refresh is performed, and the counter is still there. Refreshing browser, the fail message is gone though. Can you please check it?

pedrobaeza (approved):
Check that the problem happens on runbot where longpolling seems to not work properly, so fast-tracking this as equal to v12

PR #671, opened 2021-02-16
[13.0] [MIG] mail_notification_custom_subject: Migration to v13.0

CarlosRoca13:
cc @Tecnativa TT28063

Please @joao-p-marques @pedrobaeza review this.


mail_notification_custom_subject/i18n/mail_notification_custom_subject.pot
```diff
@@ -128,7 +130,29 @@ msgid "Templates"
...
+#: code:addons/mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py:0
+#, python-format
+msgid "Test"
+msgstr ""
+
+#. module: mail_notification_custom_subject
+#: code:addons/mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py:0
+#, python-format
+msgid "Test partner 1"
 msgstr ""
```
joao-p-marques: Why did this strings get added to the translation files? They are only used in tests...

joao-p-marques (commented):
LGTM Overall :+1: 

Just some comments regarding the translation files

mail_notification_custom_subject/i18n/mail_notification_custom_subject.pot
CarlosRoca13: These translations were added because precommit force me to set `_( )` to each term. When I extract the pot file from Odoo, the terms were added

joao-p-marques: Mmm ok :thinking: Although it is weird that pre-commit makes you do that... I guess it isn't a huge problem, anyway...

joao-p-marques (approved):

mail_notification_custom_subject/i18n/mail_notification_custom_subject.pot
CarlosRoca13: I will avoid the error by pylint: disable and remove the translation

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,36 @@
...
+    subject_template = fields.Char(
+        string="Subject Template",
+        required=True,
+        help="Subject (placeholders may be used here)",
+    )
+    position = fields.Selection(
+        selection=[
+            ("append_before", _("Append Before")),
+            ("append_after", _("Append After")),
+            ("replace", _("Replace")),
```
yajo: This is a bug:

```suggestion
            ("append_before", "Append Before"),
            ("append_after", "Append After"),
            ("replace", "Replace"),
```

Never use `_()` outside of a method code. Otherwise, the translation is done without a real context, and becomes English always.

yajo (changes requested):

CarlosRoca13:
@Yajo changes done

mail_notification_custom_subject/models/mail_message_custom_subject.py
joao-p-marques: Should you move the declaration of the selection fields to a method, then? @CarlosRoca13

CarlosRoca13: I think that's not necessary. @joao-p-marques

pedrobaeza: Selection values inside a selection field are already translatable and handled by Odoo Framework without anything extra.

mail_notification_custom_subject/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,8 @@
+
+* Activate access to **Technical Features** (debug mode).
+* Go to **Settings > Technical > Email > Subject Replacement Templates**
+* Create a new template.
+
+    * The field **Model** specifies the model to which the subject template should apply in the notification emails sent by Odoo.
+    * The field **Subject Template** accepts `Jinjah<ttps://jinja.palletsprojects.com/en/2.11.x/>` expressions.
```
yajo:
```suggestion
    * The field **Subject Template** accepts `Jinja <https://jinja.palletsprojects.com/en/2.11.x/>`__ expressions.
```

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,36 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class MailMessageCustomSubject(models.Model):
+    _name = "mail.message.custom.subject"
+    _description = "Mail Message Custom Subject"
+
+    name = fields.Char(string="Template Name")
```
yajo, line 11: Does the name field have a meaning here? If not:

```suggestion
    _description = "Mail Message Custom Subject"
    _rec_name = "subject_template"
    _order = "subject_template"

```

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -0,0 +1,36 @@
...
+        string="Model",
+        required=True,
+        help="Model where this template applies",
+    )
+    subtype_ids = fields.Many2many(
+        comodel_name="mail.message.subtype", string="Applied Subtypes", required=True,
+    )
+    subject_template = fields.Char(
+        string="Subject Template",
+        required=True,
```
yajo, line 23: It seems logical to me that this field should be translatable:

```suggestion
        required=True,
        translate=True,
```

Also, usually you'll need to define the lang as a template, to be able to use expressions such as `${object.partner_id.lang or user.lang}"

pedrobaeza: This is intended for putting an expression, so making that expression translatable will over-complicate things. We don't have such use case, so I would keep it as is.

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
```diff
@@ -0,0 +1,167 @@
...
+        # Get message and check subject
+        self.assertEquals(mail_message_1.subject, "Test and something more")
+
+    def test_bad_template_does_not_break(self):
+        self.env["mail.message.custom.subject"].create(
+            {
+                "name": "Test bad template 1",
+                "model_id": self.env.ref("base.model_res_partner").id,
+                "subtype_ids": [(6, 0, [self.env.ref("mail.mt_comment").id])],
+                "subject_template": "${obaject.number_a} and something",
```
carlosdauden, line 148:
```suggestion
                "subject_template": "${object.number_a} and something",
```

carlosdauden (changes requested):

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
joao-p-marques: This is intended @carlosdauden 
The tests asserts that a badly formed template will not render, but also won't block the sending of emails.

mail_notification_custom_subject/models/mail_message_custom_subject.py
CarlosRoca13: I think that's necessary for identify the message template. What do you think @joao-p-marques ?

joao-p-marques: It is not strictly necessary, but it can help the UX IMHO, as it helps to identify each template more easily.

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
carlosdauden: You're right, sorry...

CarlosRoca13:
Can you review it again @carlosdauden @Yajo ?? 😄

mail_notification_custom_subject/models/mail_message_custom_subject.py
yajo: > We don't have such use case

🤔 I think I don't understand what you mean. You don't have somebody whose native language isn't English?

I mean: I know people that systematically deletes mails with English subjects because they think they're spam.

We can leave this untranslatable if you prefer, but I feel like sooner or later we'll have to add translations.

mail_notification_custom_subject/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,8 @@
+
+* Activate access to **Technical Features** (debug mode).
+* Go to **Settings > Technical > Email > Subject Replacement Templates**
+* Create a new template.
+
+    * The field **Model** specifies the model to which the subject template should apply in the notification emails sent by Odoo.
+    * The field **Subject Template** accepts `Jinja <https://jinja.palletsprojects.com/en/2.11.x/>`__ expressions.
```
yajo, line 7: Does this use normal Jinja2 syntax, or Odoo-specific syntax? I think it would be good to specify it. Odoo syntax has some limitations, as you see in https://github.com/odoo/odoo/issues/45611 and https://github.com/odoo/odoo/pull/67868.

mail_notification_custom_subject/models/mail_message_custom_subject.py
pedrobaeza: Yes, we are leaving this for now. We don't have multi-language users. It can go to the ROADMAP.

mail_notification_custom_subject/readme/CONFIGURE.rst
pedrobaeza: It's not the same. This is one line and raw text, not HTML, so no possible problems like the ones you point. It's pure Jinja.

yajo (approved):

PR #674, opened 2021-02-16
[11.0] fix missing mail_id by adding ondelete cascade

agent-z28:


agent-z28:
Wrong approach, since all unittest seemed to fail

PR #683, opened 2021-03-01
[13.0][FIX] mail_tracking: Store cc

Tardo:
Fixes an error when saving "mail_cc" and "mail_to" values in the message.

cc @tecnativa TT28517

### DIFF
--- a/mail_tracking/models/mail_thread.py
+++ b/mail_tracking/models/mail_thread.py
@@ -29,17 +29,20 @@ class MailThread(models.AbstractModel):
             ("mail_tracking_ids.state", "in", list(failed_states)),
         ]
 
-    @api.returns("self", lambda value: value.id)
-    def message_post(self, *args, **kwargs):
+    @api.model
+    def _message_route_process(self, message, message_dict, routes):
         """Adds CC recipient to the message.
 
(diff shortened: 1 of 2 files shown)


mail_tracking/models/mail_thread.py
```diff
@@ -29,17 +29,24 @@ def _get_failed_message_domain(self):
...
         Because Odoo implementation avoid store 'from, to, cc' recipients we
         ensure that this information its written into the mail.message record.
         """
-        kwargs.update(
-            {"email_cc": kwargs.get("cc", False), "email_to": kwargs.get("to", False)}
+        thread_id = super()._message_route_process(message, message_dict, routes)
+        message = self.env["mail.message"].search(
+            [("message_id", "=", message_dict.get("message_id"))]
+        )
+        message.write(
```
pedrobaeza: Can we avoid this double writing? This one is critical due to the high volume of the messages, so we shouldn't write several times and do it at message creation.

pedrobaeza (changes requested):
Check also Travis

Tardo:
Thanks! changes done :)

pedrobaeza (approved):

yajo (approved):

PR #689, opened 2021-03-16
[12.0][FIX] mass_mailing_list_dynamic: Prevent ValidationError when merge contacts

victoralmau:
Prevent ValidationError when merge contacts wizard related to dynamic and full sync list.
Similar to https://github.com/OCA/social/pull/633

Need to be fixed in 13.0 too.

Please @pedrobaeza and @chienandalu  can you review it?

@Tecnativa TT28777

### DIFF
--- a/mass_mailing_list_dynamic/models/mail_mass_mailing_contact.py
+++ b/mass_mailing_list_dynamic/models/mail_mass_mailing_contact.py
@@ -11,7 +11,10 @@ class MassMailingContact(models.Model):
     _inherit = "mail.mass_mailing.contact"
 
     def _check_dynamic_full_sync_list(self, mailing_list):
-        if mailing_list.dynamic and mailing_list.sync_method == "full":
+        if (
+            mailing_list.dynamic and mailing_list.sync_method == "full"
+            and not self.env.context.get("bypass_dynamic_list_check")
+        ):
             raise ValidationError(_(
                 "Cannot edit manually contacts in a fully "
                 "synchronized list. Change its sync method or execute "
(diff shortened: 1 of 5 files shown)


chienandalu (approved):

mass_mailing_list_dynamic/wizards/partner_merge.py
```diff
@@ -0,0 +1,17 @@
...
+from odoo import models
+
+
+class BasePartnerMergeAutomaticWizard(models.TransientModel):
+    _inherit = "base.partner.merge.automatic.wizard"
+
+    def _merge(self, partner_ids, dst_partner=None, extra_checks=True):
+        return super(
+            BasePartnerMergeAutomaticWizard, self.with_context(
+                partner_merge_wizard=True
```
pedrobaeza: Isn't there already any context to use?

pedrobaeza (commented):
Why not avoid the exception condition in advance? Or is not possible? If not, put a generic context key like `bypass_dynamic_list_check` for being used by any other dependent module.

mass_mailing_list_dynamic/wizards/partner_merge.py
victoralmau: Not exists any value in context to re-use it here, changed to `bypass_dynamic_list_check`.

mass_mailing_list_dynamic/models/mail_mass_mailing_contact.py
```diff
@@ -11,7 +11,10 @@ class MassMailingContact(models.Model):
     _inherit = "mail.mass_mailing.contact"
 
     def _check_dynamic_full_sync_list(self, mailing_list):
-        if mailing_list.dynamic and mailing_list.sync_method == "full":
+        if (
+            mailing_list.dynamic and mailing_list.sync_method == "full"
+            and "bypass_dynamic_list_check" not in self.env.context
```
pedrobaeza: You should explicitly check the value of the context

```suggestion
            and self.env.context.get("bypass_dynamic_list_check")
```

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

I suppose this is needed in v13 as well

PR #695, opened 2021-04-06
[12.0][ADD] mail_activity_redirection

ypapouin:
This module implements the capability to redirect activities to specific users


ypapouin:
I don't know why ci/runbot is failing, any idea ?

PR #662, opened 2021-01-22
[MIG] mail_activity_team: Migration to 14.0

SodexisTeam:
Depends on https://github.com/OCA/social/pull/661

@stephankeller

PR #697, opened 2021-04-15
[FIX] mail_activity_team: scheduled activity

chienandalu:
When an activity is scheduled due to automatic processes the onchanges
won't adapt the proper user team, so we could have a mismatch that
provokes an exceptions due to the designed constaints

fw-port from https://github.com/OCA/social/commit/6a96684e50a12b067d03eb7da381a5b812e1fa04

cc @Tecnativa TT28781

### DIFF
--- a/mail_activity_team/models/__init__.py
+++ b/mail_activity_team/models/__init__.py
@@ -1,3 +1,4 @@
 from . import mail_activity_team
 from . import mail_activity
+from . import mail_activity_mixin
 from . import res_users
--- a/mail_activity_team/models/mail_activity_mixin.py
+++ b/mail_activity_team/models/mail_activity_mixin.py
@@ -1,0 +1,32 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import models
+
+
+class MailActivityMixin(models.AbstractModel):
+    _inherit = "mail.activity.mixin"
+
+    def activity_schedule(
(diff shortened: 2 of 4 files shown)


chienandalu:
@MiquelRForgeFlow I detected during this fw port that these changes are missing from https://github.com/OCA/social/commit/c3b3dffe97aacfc4a6473ca392a4bef0ad3e29c1

Is it right or somebody forgot to fw port them?

MiquelRForgeFlow:
Yeah, it seems someone forgot.

MiquelRForgeFlow (approved):

pedrobaeza:
Are you including here the forgotten commit then?

chienandalu:
Sorry, I didn't put the right commit. Is this one: https://github.com/OCA/social/commit/c3b3dffe97aacfc4a6473ca392a4bef0ad3e29c1 

It's a large feature. I don't know if we would run into migration issues...

MiquelRForgeFlow:
I would include it. It will not cause migration issues.

chienandalu:
Not that I want to skulk from include it, but it has quite a history that comes along with it: https://github.com/OCA/social/commits/12.0/mail_activity_team/static/src/js

MiquelRForgeFlow:
I would squash them. @jarroyomorales ok?

rafaelbn (approved):

MiquelRForgeFlow:
Should this be forward ported?

pedrobaeza:
https://github.com/OCA/social/pull/662#issuecomment-855428743

PR #703, opened 2021-04-22
[13.0][add][mail_partner_opt_out]

JordiBForgeFlow:
This module adds the capability for a user to add a partner's email address to the blackmail list so that she will not receive any emails from mass mailing campaigns.

Usage
--------
Go to a partner. Next to the partner's email address you will see the
checkbox 'Opt Out'. Once you press it the partner's email address will be
blacklisted and the checkbox will be set. If you uncheck if the email address
will be removed from the blacklist.

You can also search for partners using the opt out as search criteria.

One of the benefits of this approach is that it keeps the field 'opt_out' from previous version. If a company used this field in a past mass completed mail campaign during migration the system will respect the domain that was used. 

image

### DIFF
--- a/mail_partner_opt_out/__manifest__.py
+++ b/mail_partner_opt_out/__manifest__.py
@@ -1,0 +1,15 @@
+# Copyright 2021 ForgeFlow S.L.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+{
+    "name": "Mail Partner Opt Out",
+    "summary": "Add the partner's email to the blackmailed list",
+    "version": "13.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Social Network",
+    "website": "https://github.com/OCA/social",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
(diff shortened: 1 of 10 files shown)


JordiBForgeFlow:
@ForgeFlow @JordiMForgeFlow @AdriaGForgeFlow @LoisRForgeFlow

AaronHForgeFlow (approved):
:+1: code review LGTM and it makes sense to keep the feature.

mail_partner_opt_out/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2018 ForgeFlow S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
JordiMForgeFlow:
```suggestion
# Copyright 2021 ForgeFlow S.L.
# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
```

JordiMForgeFlow (commented):
Code review + functional testing 👍🏼

MiquelRForgeFlow (approved):

Issue #709, opened 2021-05-11
[14.0] mail_send_copy to user or sender ?


petrus-v:
Hi there,

As far I've experimented and [understood reading source code](https://github.com/OCA/social/blob/14.0/mail_send_copy/models/ir_mail_server.py#L20), at the time writing copy is sent to the sender.

In case we are using mail template, sender can be different from the user ([likes on the default account move email template](https://github.com/OCA/OCB/blob/14.0/addons/account/data/mail_template_data.xml#L11) if a saler is defined odoo will picked him as email sender).  

On the account move the saler (email sender) is automatically following the document. Using this module the saler receives 2 emails (one from this module as bcc and an other one from the following system as long an other user is clicking on the send button using the default template).

I would like to purpose a PR to change that behavior to use the user email instead of the sender.

What do you think ? I'm wondering if there are some cases where email are send using `sudo` ? (in this case this change wouldn't be a good idea)

petrus-v:
@legalsylvain as author of this module do you have any suggestion ?

petrus-v:
I'm closing this issue has #710 has been merged

PR #710, opened 2021-05-12
[14.0] mail_send_copy: description wording

petrus-v:


petrus-v:
This is an intent to fix #709, at least clarify the module promise!

@legalsylvain as discussed here is a suggestion !

As told, in the mean time I've figured out that the default Odoo behavior makes no sense to send mail from an other user (the sale user) than the current user (the account user) so I've changed email template to avoid:

* sender receiving email twice
* user do not receive any copy

regards,

legalsylvain (approved):
Makes sense.

thanks for improving the description.

legalsylvain:
CC :@quentinDupont

quentinDupont (approved):

PR #720, opened 2021-05-28
14.0

cbssolutions:
fix regex expresion was not escapig (odoo )odoo


pedrobaeza:
Please clean up the PR and give a proper PR and commit name

flotho:
ping @alexteodor , thanks for your proposal.
Is it possible to fulfill @pedrobaeza request, we'll probably have positive reviews afterwards.
Thanks for contributing.

cbssolutions:
the commit does nothing usefull.

in oca is too much formalism and is stopping the progress.

1. a good code is with comments, with help in fields..  and code that can
be read by others with ease.  This can not happen because:
-the 80 character line limit (a normal row of comments/help is transformed
in 10). here we can put alo the xml that are destroyed by this 1990 coding
rule
-the limit of function size (too complex)
2. tests are good but this shouldn't be mandatory because if it passes the
test does not mean that the code is functioning.
....

pedrobaeza:
A PR with two merge commits coming from an incorrect git usage is not acceptable neither on OCA, Odoo or any other open source project.

pedrobaeza:
Superseded by #770

PR #721, opened 2021-05-31
[14.0] [NEW] mail_allow_portal_internal_note: allow internal note to portal user.

Chandresh-OSI:
To Access an Internal Note to Portal User.
CC @dreispt

### DIFF
--- a/mail_allow_portal_internal_note/README.rst
+++ b/mail_allow_portal_internal_note/README.rst
@@ -1,0 +1,1 @@
+To be generated
--- a/mail_allow_portal_internal_note/__init__.py
+++ b/mail_allow_portal_internal_note/__init__.py
@@ -1,0 +1,5 @@
+# Copyright 2021 Open Source Integrators
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import controllers
+from . import models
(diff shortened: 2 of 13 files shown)


pedrobaeza:
Please don't plurals in module name

Chandresh-OSI:
> Please don't plurals in module name

Updated.

mail_allow_portal_internal_note/models/mail_message.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "mail.message"
+
+    def _get_search_domain_share(self):
+        if (
+            self.user_has_groups("base.group_portal")
+            and self.env.user.special_portal_users
+        ):
+            return []
+        else:
+            return super(Message, self)._get_search_domain_share()
```
dreispt: Make this based on having a Group, instead of a checkbox.
Also, it is best to always call super(), to avoid breaking the inheritance chain.
```suggestion
        domain = super(Message, self)._get_search_domain_share()
        if self.user_has_groups("mail_allow_portal_internal_note.see_internal_notes"):
            domain = []
        return domain
```

dreispt (approved):

patrickrwilson (approved):
LGTM

PR #728, opened 2021-06-16
[13.0][ADD] mail_template_multi_attachment

acsonefho:
This module allows to add more than 1 attachment (limitation of Odoo standard) to be generated during the email template render. These attachments are sent to the customer with the email's body


gurneyalex (approved):

tarteo (approved):
Looks good to me. Just  mail_template_multi_report was the previous name of this module. It looks this is a new module instead of a migration fo the previous module, but the modules fuctionalities are the same.

acsonefho:
> Looks good to me. Just mail_template_multi_report was the previous name of this module. It looks this is a new module instead of a migration fo the previous module, but the modules fuctionalities are the same.

@tarteo I didn't know about the existing module. So yes it's a new one.
I see your PR to improve this module but I prefer to merge this PR first and then add yours (as it's an improvement).

tarteo:
@acsonefho No problem, thanks for your work. I'll create a PR with the extra functionality when this one gets merged. :+1:

PR #729, opened 2021-06-17
[FIX] mail_activity_team: archived users

chienandalu:
- If a user is archived but there's an automatic activity, the team
check wouldn't locate him. We have to ensure the context.
- When the automatic activity gets a user with no team, we should set
the team to a void one as well.

cc @Tecnativa TT30470

please review @sergio-teruel @MiquelRForgeFlow

### DIFF
--- a/mail_activity_team/models/mail_activity.py
+++ b/mail_activity_team/models/mail_activity.py
@@ -66,10 +66,16 @@ class MailActivity(models.Model):
             # (for example with exception type activity, with no backorder).
             # SUPERUSER is inactive and then even if you add it
             # to member_ids it's not taken account
-            # To not be blocked we must add it to constraint condition
-            if activity.user_id.id != SUPERUSER_ID and activity.team_id and \
-                    activity.user_id and \
-                    activity.user_id not in activity.team_id.member_ids:
+            # To not be blocked we must add it to constraint condition.
+            # We must consider also users that could be archived but come from
+            # an automatic scheduled activity
+            if (
+                activity.user_id.id != SUPERUSER_ID
+                and activity.team_id and activity.user_id
+                and activity.user_id not in activity.team_id.with_context(
+                    active_test=False
+                ).member_ids
+            ):
                 raise ValidationError(
                     _('The assigned user is not member of the team.'))
 
--- a/mail_activity_team/models/mail_activity_mixin.py
+++ b/mail_activity_team/models/mail_activity_mixin.py
@@ -37,8 +37,8 @@ class MailActivityMixin(models.AbstractModel):
             team = self.env["mail.activity"].with_context(
                 default_res_model=self._name,
             )._get_default_team_id(user_id=user_id)
-            if team:
-                act_values.update({"team_id": team.id})
+            # Even if it comes empty, we don't want to mismatch the user's team
+            act_values.update({"team_id": team.id})
         return super().activity_schedule(
             act_type_xmlid=act_type_xmlid, date_deadline=date_deadline,
             summary=summary, note=note, **act_values


sergio-teruel (approved):

mail_activity_team/models/mail_activity.py
```diff
@@ -66,10 +66,16 @@ def _check_team_and_user(self):
...
-                    activity.user_id not in activity.team_id.member_ids:
+            # To not be blocked we must add it to constraint condition.
+            # We must consider also users that could be archived but come from
+            # an automatic scheduled activity
+            if (
+                activity.user_id.id != SUPERUSER_ID
+                and activity.team_id and activity.user_id
+                and activity.user_id not in activity.team_id.with_context(
+                    active_test=False
+                ).member_ids
```
MiquelRForgeFlow, line 77: I don't understand this context. member_ids is a non computed many2many, and thus member_ids is fixed, not depends on the context AFAIK. maybe I am wrong, but I am skeptic that this works :S have you tested it?

mail_activity_team/models/mail_activity_mixin.py
```diff
@@ -37,8 +37,8 @@ def activity_schedule(
             team = self.env["mail.activity"].with_context(
                 default_res_model=self._name,
             )._get_default_team_id(user_id=user_id)
-            if team:
-                act_values.update({"team_id": team.id})
+            # Even if it comes empty, we don't want to mismatch the user's team
```
MiquelRForgeFlow, line 40: how can you mismatch the user's team if you are taking the user's default team?

mail_activity_team/models/mail_activity_mixin.py
```diff
@@ -37,8 +37,8 @@ def activity_schedule(
...
+            act_values.update({"team_id": team.id})
```
MiquelRForgeFlow, line 41:
```suggestion
            act_values.update(team_id=team.id)
```
?

mail_activity_team/models/mail_activity.py
chienandalu: Sure I did :sweat_smile: Without the `active_test` context, the user isn't found. I thought the same as you anyway.

mail_activity_team/models/mail_activity_mixin.py
chienandalu: If a user doesn't have a team, the team_id should be False, not the one coming from the user that started the action

mail_activity_team/models/mail_activity.py
MiquelRForgeFlow: Good to know haha

mail_activity_team/models/mail_activity_mixin.py
chienandalu: AFAIK there's no difference, isn't it?

MiquelRForgeFlow: ah, nothing. an update doesn't rewrite the context. so forget this comment

mail_activity_team/models/mail_activity_mixin.py
MiquelRForgeFlow: ok, makes sense.

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
you can minimize my comments haha

runbot is already red in 12.0 branch, so it's not a problem of this PR.

@pedrobaeza ready to merge

chienandalu:
Thanks, @MiquelRForgeFlow !

pedrobaeza (approved):

PR #731, opened 2021-06-23
[MIG] mail_tracking: Migration to 14.0

angelJuaGar:

PR #733, opened 2021-06-25
[14.0][MIG] mail_tracking

SimoneVagile:


pedrobaeza:
There's already #731

SimoneVagile:
> There's already #731

@pedrobaeza I just noticed that.
I already fixed most of testing errors like `AttributeError: 'tuple' object has no attribute 'startswith'`.
May we keep this?

Thanks

pedrobaeza:
Priority is set based on proposition date, so for now you should propose the changes in the other PR that has more priority, and if they don't answer, then we can supersede the other pull request with yours.

PR #613, opened 2020-09-17
[13.0][FIX] mail_debrand: support all languages

astirpe:
Fixes https://github.com/OCA/social/issues/607 (for V13)

Should be backported to fix https://github.com/OCA/social/issues/441 (for V12)

Issue #725, opened 2021-06-14
[12.0] ci/runbot fail with an error about phantomjs-prebuilt: Phantom installation failed TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined


ypapouin:
All current PR against 12.0 fail because of this error:

```
[0m[37;40mnpm[0m [0m[31;40mERR![0m [0m[35mcode[0m 1
[0m[37;40mnpm[0m [0m[31;40mERR![0m [0m[35mpath[0m /home/odoo/build/OCA/social/node_modules/phantomjs-prebuilt
[0m[37;40mnpm[0m [0m[31;40mERR![0m[35m[0m command failed
[0m[37;40mnpm[0m [0m[31;40mERR![0m [0m[35mcommand[0m sh -c node install.js
[0m[37;40mnpm[0m [0m[31;40mERR![0m[35m[0m Found PhantomJS at /usr/bin/phantomjs ...verifying
[0m[37;40mnpm[0m [0m[31;40mERR![0m[35m[0m PhantomJS detected, but wrong version 1.9.0 @ /usr/bin/phantomjs.
[0m[37;40mnpm[0m [0m[31;40mERR![0m[35m[0m Phantom installation failed TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
[0m[37;40mnpm[0m [0m[31;40mERR![0m[35m[0m     at new NodeError (node:internal/errors:329:5)
[0m[37;40mnpm[0m [0m[31;40mERR![0m[35m[0m     at validateString (node:internal/validators:129:11)
[0m[37;40mnpm[0m [0m[31;40mERR![0m[35m[0m     at Object.join (node:path:1081:7)
[0m[37;40mnpm[0m [0m[31;40mERR![0m[35m[0m     at findSuitableTempDirectory (/home/odoo/build/OCA/social/node_modules/phantomjs-prebuilt/install.js:126:30)
[0m[37;40mnpm[0m [0m[31;40mERR![0m[35m[0m     at /home/odoo/build/OCA/social/node_modules/phantomjs-prebuilt/install.js:441:19
[0m[37;40mnpm[0m [0m[31;40mERR![0m[35m[0m     at nextTickCallback (/home/odoo/build/OCA/social/node_modules/kew/kew.js:47:28)
[...]

PR #736, opened 2021-07-02
[12.0][FIX] mail_debrand: support all languages

SimoneVagile:
Fixes [https://github.com/OCA/social/issues/441](https://github.com/OCA/social/issues/441) (V12)

Backporting of [https://github.com/OCA/social/pull/613](https://github.com/OCA/social/pull/613)

### DIFF
--- a/mail_debrand/models/mail_template.py
+++ b/mail_debrand/models/mail_template.py
@@ -11,10 +11,10 @@ class MailTemplate(models.Model):
     _inherit = "mail.template"
 
     @api.model
-    def _debrand_translated_words(self):
+    def _debrand_translated_words(self, lang):
         def _get_translated(word):
             return self.env["ir.translation"]._get_source(
-                "ir.ui.view,arch_db", "model_terms", self.env.lang, word
+                "ir.ui.view,arch_db", "model_terms", lang, word
             )
 
         odoo_word = _get_translated("Odoo") or _("Odoo")
@@ -23,8 +23,10 @@ class MailTemplate(models.Model):
         return odoo_word, powered_by, using_word
 
     @api.model
-    def _debrand_body(self, html):
-        odoo_word, powered_by, using_word = self._debrand_translated_words()
+    def _debrand_body(self, html, lang=False):
+        if not lang:
+            lang = self.env.lang
+        odoo_word, powered_by, using_word = self._debrand_translated_words(lang)
         html = re.sub(using_word + "(.*)[\r\n]*(.*)>" + odoo_word + r"</a>", "", html)
         if powered_by not in html:
             return html
--- a/mail_debrand/models/mail_template.py
+++ b/mail_debrand/models/mail_template.py
@@ -1,5 +1,6 @@
 # Copyright 2019 O4SB - Graeme Gellatly
 # Copyright 2019 Tecnativa - Ernesto Tejeda
+# Copyright 2020 Onestein - Andrea Stirpe
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from lxml import html as htmltree
 import re
@@ -10,13 +11,21 @@ class MailTemplate(models.Model):
     _inherit = "mail.template"
 
     @api.model
+    def _debrand_translated_words(self):
+        def _get_translated(word):
+            return self.env["ir.translation"]._get_source(
+                "ir.ui.view,arch_db", "model_terms", self.env.lang, word
+            )
+
+        odoo_word = _get_translated("Odoo") or _("Odoo")
+        powered_by = _get_translated("Powered by") or _("Powered by")
+        using_word = _get_translated("using") or _("using")
+        return odoo_word, powered_by, using_word
+
+    @api.model
     def _debrand_body(self, html):
-        using_word = _('using')
-        odoo_word = _('Odoo')
-        html = re.sub(
-            using_word + "(.*)[\r\n]*(.*)>" + odoo_word + r"</a>", "", html,
-        )
-        powered_by = _("Powered by")
+        odoo_word, powered_by, using_word = self._debrand_translated_words()
+        html = re.sub(using_word + "(.*)[\r\n]*(.*)>" + odoo_word + r"</a>", "", html)
         if powered_by not in html:
             return html
         root = htmltree.fromstring(html)
--- a/mail_debrand/models/mail_thread.py
+++ b/mail_debrand/models/mail_thread.py
@@ -10,5 +10,16 @@ class MailThread(models.AbstractModel):
 
     def _replace_local_links(self, html, base_url=None):
         html = super()._replace_local_links(html, base_url=base_url)
-        html_debranded = self.env["mail.template"]._debrand_body(html)
+        lang = False
+        if {'default_template_id', 'default_model', 'default_res_id'} \
+                <= self.env.context.keys():
+            template = self.env["mail.template"].browse(
+                self.env.context['default_template_id'])
+            if template.lang:
+                lang = template._render_template(template.lang,
+                                                 self.env.context['default_model'],
+                                                 self.env.context['default_res_id'])
+            elif template._context.get('lang', False):
+                lang = template._context.get('lang')
+        html_debranded = self.env["mail.template"]._debrand_body(html, lang)
         return html_debranded
--- a/mail_debrand/tests/test_mail_debrand.py
+++ b/mail_debrand/tests/test_mail_debrand.py
@@ -1,25 +1,38 @@
 # Copyright 2017 Tecnativa - Pedro M. Baeza
+# Copyright 2020 Onestein - Andrea Stirpe
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo.tests import common
+from odoo.tools.misc import mute_logger
 
 
 class TestMailDebrand(common.TransactionCase):
     def setUp(self):
         super().setUp()
-        self.default_arch = self.env.ref(
-            'mail.message_notification_email'
-        ).arch
-        self.paynow_arch = self.env.ref(
-            'mail.mail_notification_paynow'
-        ).arch
+        self.default_arch = self.env.ref("mail.message_notification_email").arch
+        self.paynow_arch = self.env.ref("mail.mail_notification_paynow").arch
 
     def test_default_debrand(self):
-        self.assertIn('using', self.default_arch)
+        self.assertIn("using", self.default_arch)
         res = self.env["mail.template"]._debrand_body(self.default_arch)
-        self.assertNotIn('using', res)
+        self.assertNotIn("using", res)
 
     def test_paynow_debrand(self):
-        self.assertIn('Powered by', self.paynow_arch)
+        self.assertIn("Powered by", self.paynow_arch)
         res = self.env["mail.template"]._debrand_body(self.paynow_arch)
-        self.assertNotIn('Powered by', res)
+        self.assertNotIn("Powered by", res)
+
+    def test_lang_paynow_debrand(self):
+        with mute_logger("odoo.addons.base.models.ir_translation"):
+            self.env["base.language.install"].create(
+                {"lang": "nl_NL", "overwrite": True}
+            ).lang_install()
+        with mute_logger("odoo.tools.translate"):
+            self.env["base.update.translations"].create({"lang": "nl_NL"}).act_update()
+
+        ctx = dict(lang="nl_NL")
+        paynow_template = self.env.ref("mail.mail_notification_paynow")
+        paynow_arch = paynow_template.with_context(ctx).arch
+        self.assertIn("Aangeboden door", paynow_arch)
+        res = self.env["mail.template"].with_context(ctx)._debrand_body(paynow_arch)
+        self.assertNotIn("Aangeboden door", res)
--- a/mail_debrand/tests/test_mail_debrand.py
+++ b/mail_debrand/tests/test_mail_debrand.py
@@ -2,15 +2,45 @@
 # Copyright 2020 Onestein - Andrea Stirpe
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo.tests import common
+from odoo.tests.common import SavepointCase
 from odoo.tools.misc import mute_logger
 
 
-class TestMailDebrand(common.TransactionCase):
-    def setUp(self):
-        super().setUp()
-        self.default_arch = self.env.ref("mail.message_notification_email").arch
-        self.paynow_arch = self.env.ref("mail.mail_notification_paynow").arch
+class TestMailDebrand(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.other_lang = "it_IT"
+        cls.get_translated_paynow_arch_term()
+
+        cls.default_arch = cls.env.ref(
+            'mail.message_notification_email'
+        ).arch
+        cls.paynow_arch = cls.env.ref(
+            'mail.mail_notification_paynow'
+        ).arch
+
+    @classmethod
+    def get_translated_paynow_arch_term(cls):
+        with mute_logger("odoo.tools.translate"):
+            cls.env["res.lang"].load_lang(cls.other_lang)
+            cls.env["base.update.translations"].create(
+                {"lang": cls.other_lang}
+            ).act_update()
+        with mute_logger("odoo.addons.base.models.ir_translation"):
+            cls.env["base.language.install"].create(
+                {"lang": cls.other_lang, "overwrite": True}
+            ).lang_install()
+
+        paynow_template = cls.env.ref("mail.mail_notification_paynow")
+        domain = [
+            ("lang", "=", cls.other_lang),
+            ("src", "like", "Powered by"),
+            ("res_id", "=", paynow_template.id)
+        ]
+        cls.pb_term = cls.env["ir.translation"].search(domain).value
+        cls.paynow_arch_translated = paynow_template.with_context(
+            lang=cls.other_lang).arch
 
     def test_default_debrand(self):
         self.assertIn("using", self.default_arch)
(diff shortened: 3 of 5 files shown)


SimoRubi (changes requested):
Code review and tried locally, code changes are ok for me.

Please fix runbot: it is ❌ because of https://github.com/OCA/social/issues/725

SimoRubi (changes requested):
I tried the following in runbot:

1. In Settings > Translations > Load a translation, load the italian language
2. Install sale_management
3. Create a customer having italian language (keep admin's language to English)
4. Create a sale order for the italian customer and click on "Send by email"
5. Confirm the popup
6. Activate debug mode and check the sent email in Settings > Technical > Email > Emails

**Actual behavior**
The email has the "Fornito da Odoo" brand:
image

**Desired behavior**
The email does not have the "Fornito da Odoo" brand

Can you check?
Or maybe this fix is made to fix some other behavior?

SimoneVagile:
> I tried the following in runbot:
> 
>     1. In Settings > Translations > Load a translation, load the italian language
> 
>     2. Install sale_management
> 
>     3. Create a customer having italian language (keep admin's language to English)
> 
>     4. Create a sale order for the italian customer and click on "Send by email"
> 
>     5. Confirm the popup
> 
>     6. Activate debug mode and check the sent email in Settings > Technical > Email > Emails
> 
> 
> **Actual behavior**
> The email has the "Fornito da Odoo" brand:
> image
> 
> **Desired behavior**
> The email does not have the "Fornito da Odoo" brand
> 
> Can you check?
> Or maybe this fix is made to fix some other behavior?

I changed html_debranded: in message body does not compare `Powered by` tag, 'cause it's added in method `render` of QWeb model

mail_debrand/models/ir_qweb.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class IrQWeb(models.AbstractModel, QWeb):
+    _inherit = 'ir.qweb'
+
+    def render(self, template, values=None, **options):
+        result = super(IrQWeb, self).render(template, values=values, **options)
+        lang = values.get("lang", False)
+        message = values.get("message", False)
+        if lang and message:
```
SimoRubi: Why are you checking for `message` if then you always debrand all the `result`?

Moreover, `values` is an arbitrary dictionary passed to a mail template so we are not sure that language's variable is named `lang`; is it possible to get directly the language used for rendering?
From https://github.com/odoo/odoo/blob/2be47637dd5cc48ca561ff85cf2d359dde0829ab/odoo/addons/base/models/qweb.py#L1506-L1554 looks like the language will be in `self`'s context.

Can you add a test for the use-case you have fixed?

mail_debrand/tests/test_mail_debrand.py
```diff
@@ -23,3 +24,18 @@ def test_paynow_debrand(self):
...
+            self.env["base.language.install"].create(
+                {"lang": "nl_NL", "overwrite": True}
+            ).lang_install()
+        with mute_logger("odoo.tools.translate"):
+            self.env["base.update.translations"].create({"lang": "nl_NL"}).act_update()
+
+        ctx = dict(lang="nl_NL")
+        paynow_template = self.env.ref("mail.mail_notification_paynow")
+        paynow_arch = paynow_template.with_context(ctx).arch
+        self.assertIn("Aangeboden door", paynow_arch)
```
SimoRubi: This test depends on a translation decided from whoever translates Odoo/OCA modules so looks to me like it will easily break as soon as someone changes the translation.
Can you add a translation in the test and use that one instead?

SimoRubi (changes requested):
Code review

mail_debrand/models/ir_qweb.py
SimoneVagile: > Why are you checking for `message` if then you always debrand all the `result`?
> 
> Moreover, `values` is an arbitrary dictionary passed to a mail template so we are not sure that language's variable is named `lang`; is it possible to get directly the language used for rendering?
> From https://github.com/odoo/odoo/blob/2be47637dd5cc48ca561ff85cf2d359dde0829ab/odoo/addons/base/models/qweb.py#L1506-L1554 looks like the language will be in `self`'s context.
> 

I have to be sure if `render` method of IrQWeb is called after `_notify`, just before an email sending.
From [https://github.com/odoo/odoo/blob/12.0/addons/mail/models/res_partner.py#L37-L86](https://github.com/odoo/odoo/blob/12.0/addons/mail/models/res_partner.py#L37-L86) you have both `message` and `lang`, and that's when you want to debrand (`lang` can be explicitly set in `mail.template`).

> Can you add a test for the use-case you have fixed?

WIP ;)

mail_debrand/tests/test_mail_debrand.py
```diff
@@ -33,9 +33,35 @@ def test_lang_paynow_debrand(self):
         with mute_logger("odoo.tools.translate"):
             self.env["base.update.translations"].create({"lang": "nl_NL"}).act_update()
 
-        ctx = dict(lang="nl_NL")
         paynow_template = self.env.ref("mail.mail_notification_paynow")
+        domain = [
```
SimoRubi, line 36: Why is this commit `[FIX] new test` containing also fixes to an old test?

mail_debrand/models/ir_qweb.py
SimoRubi: > I have to be sure if `render` method of IrQWeb is called after `_notify`, just before an email sending.
> From https://github.com/odoo/odoo/blob/12.0/addons/mail/models/res_partner.py#L37-L86 you have both `message` and `lang`, and that's when you want to debrand (`lang` can be explicitly set in `mail.template`).

Checking for `lang` and `message` can be good for the specific case when a message is sent to a partner because the method you have linked `_notify_prepare_template_context` is called _just_ before `render` in `_notify`: https://github.com/odoo/odoo/blob/5d40023eaf9b3c25941429e7fac266722780846d/addons/mail/models/res_partner.py#L107-L136

The problem is that `render` can be called in other situations too where there is no `lang` or `message` in `values`, the debranding should happen in those cases too.


> > Can you add a test for the use-case you have fixed?

> WIP ;)

The new test passes just the same even without 89e2d1c, can you add a test that fails without your fix and passes with it?

SimoRubi (changes requested):

mail_debrand/models/ir_qweb.py
SimoneVagile: > Checking for lang and message can be good for the specific case when a message is sent to a partner because the method you have linked _notify_prepare_template_context is called just before render in _notify: https://github.com/odoo/odoo/blob/5d40023eaf9b3c25941429e7fac266722780846d/addons/mail/models/res_partner.py#L107-L136
> 
> The problem is that render can be called in other situations too where there is no lang or message in values, the debranding should happen in those cases too.

This module debrands only `mail.message` html body (that includes mails, internal messages and notes), and this is why I wrote the condition. When `_notify_prepare_template_context` is called, I'm sure that in `render`'s `value` are stored a `mail.message` and `lang`. Why should it include other cases? There are other situations where a `mail.message` is created outside `_notify`?

mail_debrand/tests/test_mail_debrand.py
SimoneVagile: Corrected

mail_debrand/tests/test_mail_debrand.py
SimoneVagile: Changed

mail_debrand/models/ir_qweb.py
SimoRubi: > Why should it include other cases? There are other situations where a mail.message is created outside _notify?

Checking for `lang` and `message` is a very weak condition to be sure that an email is being sent:

- if `lang` and `message` are in `values`, it does not imply that an email is being sent: we might be rendering any QWeb template that needs those values and in that case we do not want any substitution to be performed by this module.
- if an email is being sent, it does not imply that `lang` and `message` are in `values`: anyone can create a `mail.mail` record, render a QWeb template for its body and call its `send` method

> The new test passes just the same even without 89e2d1c, can you add a test that fails without your fix and passes with it?

SimoRubi (changes requested):

mail_debrand/models/ir_qweb.py
```diff
@@ -0,0 +1,21 @@
...
+    def render(self, template, values=None, **options):
+        result = super(IrQWeb, self).render(template, values=values, **options)
+        lang = values.get("lang", False)
+        is_mail = False
+        # search for mail.message object in values
+        for value in values.values():
+            if isinstance(value, models.Model) and value._name == "mail.message":
+                is_mail = True
+                break
+        if lang and is_mail:
```
SimoRubi: Checking for `lang` and any `mail.message` record in values is a very weak condition to be sure that an email is being sent for basically the same reasons I wrote in https://github.com/OCA/social/pull/736#discussion_r694986881

mail_debrand/tests/test_mail_debrand.py
```diff
@@ -63,5 +76,5 @@ def test_lang_paynow_debrand_partner(self):
         mail = self.env['mail.mail'].search(
             [('mail_message_id', '=', msg.id)]
         )
-        self.assertNotIn('Powered by', mail.body_html)
```
SimoRubi: The test was correctly looking for `'Powered by'` string because this specific test checks the behavior described in https://github.com/OCA/social/pull/736#pullrequestreview-699620680, please restore this assert

SimoRubi (changes requested):
Please also check failing pre-commit hook.

SimoneVagile:
@SimoRubi I restored the only backporting, plus tests fix.

I removed the `ir.qweb` inheriting because I didn't replicate the case where it doesn't work (maybe it resolved already).

SimoRubi (changes requested):
https://github.com/OCA/social/pull/736#pullrequestreview-699620680 is happening again, here the email I just generated (locally because currently there is no runbot in this PR):
image

mail_debrand/models/mail_template.py
```diff
@@ -10,13 +10,23 @@ class MailTemplate(models.Model):
...
 
     @api.model
-    def _debrand_body(self, html):
-        using_word = _('using')
-        odoo_word = _('Odoo')
-        html = re.sub(
-            using_word + "(.*)[\r\n]*(.*)>" + odoo_word + r"</a>", "", html,
-        )
-        powered_by = _("Powered by")
+    def _debrand_translated_words(self, lang):
```
SimoRubi, line 14: Could you please check why here we don't have anymore the copyright edits that were in https://github.com/OCA/social/pull/613?
The same happens for the author field in manifest

mail_debrand/models/mail_thread.py
```diff
@@ -10,5 +10,13 @@ class MailThread(models.AbstractModel):
...
         html = super()._replace_local_links(html, base_url=base_url)
-        html_debranded = self.env["mail.template"]._debrand_body(html)
+        lang = False
+        if {'default_template_id', 'default_model', 'default_res_id'} \
+                <= self.env.context.keys():
+            template = self.env["mail.template"].browse(
+                self.env.context['default_template_id'])
+            lang = template._render_template(template.lang,
+                                             self.env.context['default_model'],
+                                             self.env.context['default_res_id'])
```
SimoRubi: Trying this I noticed that the correct `lang` is already in the context because `generate_email` puts it there: https://github.com/odoo/odoo/blob/2be47637dd5cc48ca561ff85cf2d359dde0829ab/addons/mail/models/mail_template.py#L443
so looks like there is no need to look for these values being in `context`, find the template and call render.
This would be much cleaner if we could just use the language that is already in the context, what do you think?

If any test has a problem with that, that's probably because the test does not emulate correctly what really happens.

mail_debrand/tests/test_mail_debrand.py
```diff
@@ -2,18 +2,42 @@
...
 
 from odoo.tests import common
+from odoo.tools.misc import mute_logger
 
 
 class TestMailDebrand(common.TransactionCase):
     def setUp(self):
         super().setUp()
+        self.other_lang = "it_IT"
+        self.env["res.lang"].load_lang(self.other_lang)
```
SimoRubi: Loading a language is quite expensive and here it is done again and again for each test because we are in `TransactionCase` subclass, could you please avoid reloading the language for each test?

You might switch to `SavepointCase` or just load the language for the test that needs it if it's just one.

SimoRubi (changes requested):

mail_debrand/models/mail_template.py
SimoneVagile: Commit I cherry-picked was outdated, now its the right one.
I took the opportunity for a rebase with `12.0` branch.

mail_debrand/tests/test_mail_debrand.py
SimoneVagile: Moved the `load_lang` method in `get_translated_paynow_arch_term`, so it'll load the language when needed.

mail_debrand/models/mail_thread.py
SimoneVagile: This is why the test you suggested me days ago went wrong: [https://github.com/OCA/social/pull/736#pullrequestreview-699620680](https://github.com/OCA/social/pull/736#pullrequestreview-699620680).
When execution reach that point, you have:

- template.lang = '${object.partner_id.lang}'  ==> (in this case, 'it_IT')
- template._context.get('lang') = 'en_US'

In order to debrand correctly the mail, you must translate the terms `using` and `Powered by` with the language set in the `mail.template` record if exists, otherwise the language in context.
This is why I did this correction: [https://github.com/OCA/social/pull/736/commits/d4a5e288695f47ae5cf939bd29b66286b7cd679f#diff-3166b8d109f4a2e1171211a98d2ed169012a6c6e9cf59473dc8f6d72d318eae8L17](https://github.com/OCA/social/pull/736/commits/d4a5e288695f47ae5cf939bd29b66286b7cd679f#diff-3166b8d109f4a2e1171211a98d2ed169012a6c6e9cf59473dc8f6d72d318eae8L17).

SimoneVagile:
Runbot CI is missing, can any maintainer check for this? Thanks

SimoRubi:
@OCA/social-maintainers ^

mail_debrand/models/mail_thread.py
SimoRubi: I have checked that the first time the message body is parsed it has the correct language in the context because the body is generated by the method containing https://github.com/odoo/odoo/blob/2be47637dd5cc48ca561ff85cf2d359dde0829ab/addons/mail/models/mail_template.py#L443.
However, the layout is generated later by https://github.com/odoo/odoo/blob/00f9bee5804ffafa71e1fb5ae90f5e4c8fe2451e/addons/mail/models/res_partner.py#L136 and here you can't reach the template's context from `_replace_local_links` (let's exclude a recomputation using `_notify_classify_recipients_on_records` and `_notify_prepare_template_context` to keep it simple).

The values you are looking for in the context are there if we are using the composer because they are set in https://github.com/odoo/odoo/blob/899ae3e0aadba2adae11f5a58743022397a77ba2/addons/mail/models/mail_thread.py#L2163-L2165 so this looks like an acceptable implementation if we are using the composer to send an email; it would probably still fail if we are not using the composer with a layout that includes the "Powered by" but this is not the purpose of this PR.

As an additional proof that is an acceptable implementation, this is almost the same that already happens in https://github.com/odoo/odoo/blob/00f9bee5804ffafa71e1fb5ae90f5e4c8fe2451e/addons/mail/models/res_partner.py#L55-L61 but in this PR's implementation we are missing a couple of things:
- check if the template has a lang to be parsed before parsing it
- fallback on the `lang` in the context if nothing is found

Can you add them here too? Otherwise we would have a different behavior from what happens in the core.

mail_debrand/tests/test_mail_debrand.py
```diff
@@ -1,25 +1,89 @@
...
         self.default_arch = self.env.ref(
             'mail.message_notification_email'
         ).arch
         self.paynow_arch = self.env.ref(
             'mail.mail_notification_paynow'
         ).arch
 
+    def get_translated_paynow_arch_term(self):
+        with mute_logger("odoo.tools.translate"):
+            self.env["res.lang"].load_lang(self.other_lang)
```
SimoRubi: > Loading a language is quite expensive and here it is done again and again for each test because we are in TransactionCase subclass, could you please avoid reloading the language for each test?
>
>You might switch to SavepointCase or just load the language for the test that needs it if it's just one.
> > Moved the load_lang method in get_translated_paynow_arch_term, so it'll load the language when needed.

(from https://github.com/OCA/social/pull/736#discussion_r733544947)
This still loads the language many times and, as I said, it is an expensive operation that can slow down the tests, why not loading the language only once?

SimoRubi (changes requested):
Please also give a more meaningful message to the last commit as suggested in https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message or squash it in the previous one, right now it is `[FIX] tests`

mail_debrand/tests/test_mail_debrand.py
```diff
@@ -1,25 +1,91 @@
...
-class TestMailDebrand(common.TransactionCase):
-    def setUp(self):
-        super().setUp()
-        self.default_arch = self.env.ref(
+class TestMailDebrand(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.other_lang = "it_IT"
+        cls.get_translated_paynow_arch_term(cls)
```
SimoRubi: If a method is called from a `classmethod`, the called method should be a `classmethod` too, unless you are calling it for an instance of the class.

More simply: calling `cls.method(cls)` is bad, just decorate `method` with `classmethod` and call it in the usual way `cls.method`.

mail_debrand/tests/test_mail_debrand.py
```diff
@@ -1,25 +1,91 @@
...
+
+        cls.default_arch = cls.env.ref(
             'mail.message_notification_email'
         ).arch
-        self.paynow_arch = self.env.ref(
+        cls.paynow_arch = cls.env.ref(
             'mail.mail_notification_paynow'
         ).arch
 
+    def get_translated_paynow_arch_term(self):
```
SimoRubi: `self` here is not an instance but a class type, see other comment about class methods

mail_debrand/models/mail_thread.py
```diff
@@ -10,5 +10,14 @@ class MailThread(models.AbstractModel):
...
     def _replace_local_links(self, html, base_url=None):
         html = super()._replace_local_links(html, base_url=base_url)
-        html_debranded = self.env["mail.template"]._debrand_body(html)
+        lang = False
+        if {'default_template_id', 'default_model', 'default_res_id'} \
+                <= self.env.context.keys():
+            template = self.env["mail.template"].browse(
+                self.env.context['default_template_id'])
+            template_lang = template.lang or template._context.get('lang')
+            lang = template._render_template(template_lang,
```
SimoRubi: Please only render the language if it has a value as suggested in https://github.com/OCA/social/pull/736#discussion_r778752727.
Moreover, why is this rendering the language coming from the context?

SimoRubi (changes requested):
> Please also give a more meaningful message to the last commit as suggested in https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message or squash it in the previous one, right now it is `[FIX] tests`

mail_debrand/models/mail_thread.py
SimoneVagile: From [https://github.com/OCA/social/pull/736#discussion_r733688343](https://github.com/OCA/social/pull/736#discussion_r733688343):

> 
In order to debrand correctly the mail, you must translate the terms `using` and `Powered by` with the language set in the mail.template record **if exists, otherwise the language in context**.

See also [https://github.com/odoo/odoo/blob/2be47637dd5cc48ca561ff85cf2d359dde0829ab/addons/mail/models/mail_template.py#L443](https://github.com/odoo/odoo/blob/2be47637dd5cc48ca561ff85cf2d359dde0829ab/addons/mail/models/mail_template.py#L443)

mail_debrand/models/mail_template.py
SimoneVagile: This can be resolved?

mail_debrand/tests/test_mail_debrand.py
SimoneVagile: This can be resolved?

mail_debrand/models/mail_thread.py
```diff
@@ -10,5 +10,15 @@ class MailThread(models.AbstractModel):
...
         html = super()._replace_local_links(html, base_url=base_url)
-        html_debranded = self.env["mail.template"]._debrand_body(html)
+        lang = False
+        if {'default_template_id', 'default_model', 'default_res_id'} \
+                <= self.env.context.keys():
+            template = self.env["mail.template"].browse(
+                self.env.context['default_template_id'])
+            template_lang = template.lang or template._context.get('lang')
+            if template_lang:
+                lang = template._render_template(template_lang,
```
SimoRubi: > From [#736 (comment)](https://github.com/OCA/social/pull/736#discussion_r733688343):
> 
> > 
> 
> In order to debrand correctly the mail, you must translate the terms `using` and `Powered by` with the language set in the mail.template record **if exists, otherwise the language in context**.
> 
> See also https://github.com/odoo/odoo/blob/2be47637dd5cc48ca561ff85cf2d359dde0829ab/addons/mail/models/mail_template.py#L443

(from https://github.com/OCA/social/pull/736#discussion_r785845002)

I see, but I think the value in the context is `en_US` or `it_IT` or something similar, so looks to me like there is no need to call `_render_template` for it, that's why I asked:
> why is this rendering the language coming from the context?

and I don't understand how https://github.com/OCA/social/pull/736#discussion_r785845002 can answer that question, could you please rephrase?

SimoRubi (approved):

mail_debrand/models/mail_thread.py
SimoneVagile: Got it.

Now it values `lang` variable directly with the value in the context if `template.lang` is `False`.

SimoRubi (approved):

rafaelbn (approved):

rafaelbn:
Hello @SimoneVagile @SimoRubi could you please check Travis 🔴 ? Thanks!

PR #741, opened 2021-07-09
[FIX] mail_tracking: reduce spammy score

### LINKED ISSUE #701: [12.0] mail_tracking: SpamAssassin flag 'RAND_MKTG_HEADER'
Hello,
Recently some invoice emails that I have been sending through Odoo were not reaching the recipients.They were always to the same customers, and if you sent with the same email account but from Thunderbird for example, it would arrive perfectly.

I have tested with mail-tester to check mail reputation and due to a recent update of SpamAssasin, the headers for Odoo mail tracking detects them as SPAM and drops almost 3 points. This increases the probability of getting into the inboxes of some customers.
imagen

imagen

It appears that if SpamAssasin finds several headers that start with "X-..." and have keywords like "Tracking" it lowers the email's score and may cause problems for certain recipients.
https://bz.apache.org/SpamAssassin/show_bug.cgi?id=7888

If I uninstall the mail_tracking module my score is 10/10.

Has anyone else experienced this?
or any ideas on how to fix this, mail_tracking is a very powerful tool and I would hate to lose it.
Thanks,

yajo:

@Tecnativa TT29674

### DIFF
--- a/mail_tracking/models/ir_mail_server.py
+++ b/mail_tracking/models/ir_mail_server.py
@@ -13,10 +13,8 @@ class IrMailServer(models.Model):
     def _tracking_headers_add(self, tracking_email_id, headers):
         """Allow other addons to add its own tracking SMTP headers"""
         headers = headers or {}
-        headers["X-Odoo-Database"] = (
-            getattr(threading.currentThread(), "dbname", None),
-        )
-        headers["X-Odoo-Tracking-ID"] = tracking_email_id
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

Tardo (approved):

yajo:
Tested and it works as expected: 
image

pedrobaeza:
Cherry-picked for v14 in 6480cf888a0674d518769bce517c031b26045efa

PR #712, opened 2021-05-18
[14.0][FIX] mail_debrand: pass original args to super()

ryanc-me:
Previously, the kwargs `engine`, `add_context`, and `post_process` were not passed up to the super() call. This casued issues when the original caller specified those arguments.

For example, see the Digest module: https://github.com/odoo/odoo/blob/14.0/addons/digest/models/digest.py#L118

Prior to this commit, Digest emails would be blank, only containing the text _mail.digest_mail_main_.

PR #749, opened 2021-07-22
[14.0] mail_debrand: pass original args to super()

ioans73:
This PR superseds #712 ping @ryanc-me

@pedrobaeza's proposed changes have been implemented and the original PR's conflicts have been fixed

### DIFF
--- a/mail_debrand/models/mail_mail.py
+++ b/mail_debrand/models/mail_mail.py
@@ -1,0 +1,24 @@
+# Copyright 2019 O4SB - Graeme Gellatly
+# Copyright 2019 Tecnativa - Ernesto Tejeda
+# Copyright 2020 Onestein - Andrea Stirpe
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class MailMail(models.AbstractModel):
+    _inherit = "mail.mail"
+
+    # in messages from objects is adding using Odoo that we are going to remove
+
+    @api.model_create_multi
+    def create(self, values_list):
+        for index, _value in enumerate(values_list):
+            values_list[index]["body_html"] = self.env[
+                "mail.render.mixin"
+            ].remove_href_odoo(
+                values_list[index].get("body_html", ""),
+                remove_parent=0,
+                remove_before=1,
+            )
+
+        return super().create(values_list)
--- a/mail_debrand/models/mail_mail.py
+++ b/mail_debrand/models/mail_mail.py
@@ -2,23 +2,14 @@
 # Copyright 2019 Tecnativa - Ernesto Tejeda
 # Copyright 2020 Onestein - Andrea Stirpe
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import api, models
+from odoo import models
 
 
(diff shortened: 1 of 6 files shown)


fuentes73 (approved):

mail_debrand/models/mail_mail.py
```diff
@@ -0,0 +1,24 @@
...
+# Copyright 2019 Tecnativa - Ernesto Tejeda
+# Copyright 2020 Onestein - Andrea Stirpe
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class MailMail(models.AbstractModel):
+    _inherit = "mail.mail"
+
+    # in messages from objects is adding using Odoo that we are going to remove
```
pedrobaeza: This should be a docstring of the method, not a comment in the middle.

mail_debrand/models/mail_mail.py
```diff
@@ -0,0 +1,24 @@
...
+    def create(self, values_list):
+        for index, _value in enumerate(values_list):
+            values_list[index]["body_html"] = self.env[
+                "mail.render.mixin"
+            ].remove_href_odoo(
+                values_list[index].get("body_html", ""),
+                remove_parent=0,
+                remove_before=1,
+            )
+
```
pedrobaeza: Don't add empty lines inside a method.

mail_debrand/models/mail_mail.py
```diff
@@ -0,0 +1,24 @@
...
+class MailMail(models.AbstractModel):
+    _inherit = "mail.mail"
+
+    # in messages from objects is adding using Odoo that we are going to remove
+
+    @api.model_create_multi
+    def create(self, values_list):
+        for index, _value in enumerate(values_list):
+            values_list[index]["body_html"] = self.env[
```
pedrobaeza: You don't need to enumerate:
```suggestion
        for vals in values_list:
            vals["body_html"] = self.env[
```

pedrobaeza (changes requested):
Can we avoid to process each of the created mail with this filter? Why not intercepting the rendering of the template that injects the value?

ioans73:
> Can we avoid to process each of the created mail with this filter? Why not intercepting the rendering of the template that injects the value?

Your proposal seems good to me, but I believe the correct thing would be to propose those changes in another PR, right? (at least the part of modifying the way of deleting the text, so that it is not done in the create() method) In order to unblock this and fix the detected bug as soon as possible
What do you think @pedrobaeza ?

pedrobaeza:
I don't think so, as we are trying to "fix" this module, and the proper fix is that one. Note that this module has been "broken" just in this version.

ioans73:
Ok, we will try to implement the proposal, although I don't entirely agree

ioans73:
@pedrobaeza In the end we've chosen to inherit the `_send_prepare_body()` function, as it's prepared precisely for these type of changes

pedrobaeza:
Thanks for the efforts. Do we need then to override now `_render_template` method?

ioans73:
According to our testing, yes, as not all mails pass through `_send_prepare_body()`

pedrobaeza (approved):
Thanks for all

PR #751, opened 2021-07-23
[14.0][MIG] mass_mailing_custom_unsubscribe: Migration to 14.0

ernestotejeda:
Cc @Tecnativa TT30288


pedrobaeza:
Travis is red

mass_mailing_custom_unsubscribe/static/src/js/contact.tour.js
```diff
@@ -80,16 +77,15 @@ odoo.define("mass_mailing_custom_unsubscribe.contact_tour", function (require) {
                 content: "Subscribe again to list 0",
                 // eslint-disable-next-line no-multi-str
                 trigger:
-                    "body:not(:has(#unsubscribe_form #custom_div_feedback\
-                            :visible)):has(.alert-success) li:contains('test list 0') \
-                            input:not(:checked)",
+                    "body:not(:has(#unsubscribe_form #custom_div_feedback" +
+                    ":visible)):has(.alert-success) li:contains('test list 0') " +
```
yajo: Are you sure this change is OK?

- `#custom_div_feedback :visible` means "find anything visible inside an element with ID custom_div_feedback
- `#custom_div_feedback:visible` means "find any visible element with ID custom_div_feedback

Maybe it was wrong before, but in any case it seems like the typical poisoned change. Please double-check it.

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -5,46 +5,46 @@
  * to extend it as it is currently designed. It is almost a copy+paste from
  * upstream, to allow easier version/patch updates, so linter is disabled. */
```
yajo: As you see here, this is almost a copy of upstream (find it in https://github.com/odoo/odoo/blob/14.0/addons/mass_mailing/static/src/js/unsubscribe.js).

In these specific cases, I don't recommend modernizing the code too much. This way, we can keep it more similar to upstream, and it's easier to extract the diff to know exactly what has changed.

Also you can see that one line below, we disable ESLint (and that's the reason: to avoid linter failures because we want the file to be "badly formatted" for the sake of being closer to upstream code).

So, although I appreciate your effort on modernizing this file's code, I have to ask you to leave it as it was please. More specifically: copy the v14 version of this file, paste it, and keep the important part of the diff.

Another option you have is to try to do it without replacing the file, by overriding something. For example, in the qweb template you can replace any element ID where we need a different logic, so we only need to override that part, and the JS code that attaches events here doesn't trigger because the element with the old ID isn't found. Or to remove the events that upstream attached and replace it with yours.

mass_mailing_custom_unsubscribe/tests/test_ui.py
```diff
@@ -57,10 +56,6 @@ def setUp(self):
             </div>
         """
 
-    def tearDown(self):
-        del self.email, self.lists, self.contact, self.mailing, self.url
```
yajo: You still need to delete `self.url` after each test.

yajo (changes requested):
Please check Travis. Some comments below.

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
ernestotejeda: Thanks @Yajo , you're right, I shouldn't have modernize the file since it's almost a copy of Upstream. As you suggested, I copied the code from version 14.0 of the upstream file and kept the important parts of the diff and I added ``// prettier-ignore`` comment to avoid re-formating by Prettier. Now we can clearly see the modifications to the original code.

mass_mailing_custom_unsubscribe/tests/test_ui.py
ernestotejeda: done

mass_mailing_custom_unsubscribe/static/src/js/contact.tour.js
ernestotejeda: well, I left it as it was in v13 and I checked it again and it seems to me that it is right as it is.

ernestotejeda:
Travis is green.
@Yajo changes done.
ping @Tardo @CarlosRoca13

mass_mailing_custom_unsubscribe/tests/test_ui.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def extract_url(self, mail, *args, **kwargs):
+        url = mail.mailing_id._get_unsubscribe_url(self.email, mail.res_id)
+        self.assertTrue(urls.url_parse(url).decode_query().get("token"))
+        self.assertTrue(url.startswith(self.domain))
+        self.url = url.replace(self.domain, "", 1)
+        return True
+
+    def setUp(self):
+        super(UICase, self).setUp()
```
CarlosRoca13:
```suggestion
        super().setUp()
```

mass_mailing_custom_unsubscribe/tests/test_ui.py
```diff
@@ -0,0 +1,162 @@
...
+            <div>
+                <a href="/unsubscribe_from_list">
+                    This link should get the unsubscription URL
+                </a>
+            </div>
+        """
+
+    def tearDown(self):
+        del self.email, self.lists, self.contact, self.mailing, self.url
+        super(UICase, self).tearDown()
```
CarlosRoca13:
```suggestion
        super().tearDown()
```

CarlosRoca13 (approved):
LGTM just 2 minor comments

ernestotejeda:
@CarlosRoca13 changes done

pedrobaeza:
@Yajo please check

mass_mailing_custom_unsubscribe/static/src/js/contact.tour.js
```diff
@@ -78,18 +75,20 @@ odoo.define("mass_mailing_custom_unsubscribe.contact_tour", function (require) {
             },
             {
                 content: "Subscribe again to list 0",
-                // eslint-disable-next-line no-multi-str
+                /* eslint-disable no-multi-str */
```
Tardo: Idk what is best... ignore no-multi-str or ignore max-length ... :/ for me, ignore "max-length" to better readability sounds better.

ernestotejeda: In this case Eslint shows the warning (no-multi-str) because of the way the text is divided into several lines. If I ignore max-length it will still show the warning (no-multi-str)

pedrobaeza: But the idea is to change the warning and join the multi line in one.

ernestotejeda:
Thanks @Tardo , I join the multi line in one line for better readability and fortunately it's not necessary to disable any eslint rules. check it out.

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -3,35 +3,37 @@
 
 /* This JS module replaces core AJAX submission because it is impossible
  * to extend it as it is currently designed. It is almost a copy+paste from
- * upstream, to allow easier version/patch updates, so linter is disabled. */
+ * upstream, to allow easier version/patch updates, so linter is disabled
+ * and prettier is ignore. */
 /* eslint-disable */
+// prettier-ignore
```
yajo: This might be a little nitpicking, but could you add this little part in a commit _before_ passing prettier to it? Otherwise we have a commit passing prettier to this file, and another commit undoing it, and generates a big diff that makes it difficult to see what has really changed here downstream.

yajo (approved):

mass_mailing_custom_unsubscribe/controllers/main.py
```diff
@@ -0,0 +1,159 @@
...
+    @route()
+    def mailing(self, mailing_id, email=None, res_id=None, token="", **post):
+        """Ask/save unsubscription reason."""
+        _logger.debug(
+            "Called `mailing()` with: %r", (mailing_id, email, res_id, token, post)
+        )
+        reasons = request.env["mail.unsubscription.reason"].search([])
+        if not res_id:
+            res_id = "0"
+        res_id = res_id and int(res_id)
```
Tardo, line 51: what is the reason for using "0" and not 0?
```suggestion
        res_id = res_id or 0
```
Or better yet, set the default value to zero directly...

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,274 @@
...
+});
+
+// prettier-ignore
+function toggle_opt_out_section(value) {
+    var result = !value;
+    $("#div_opt_out").find('*').attr('disabled',result);
+    $("#button_add_blacklist").attr('disabled', false);
+    $("#button_remove_blacklist").attr('disabled', false);
+    if (value) { $('[name="button_subscription"]').addClass('clickable');  }
+    else { $('[name="button_subscription"]').removeClass('clickable'); }
```
Tardo, line 273:
```suggestion
    $('[name="button_subscription"]').toggleClass('clickable', value);
```

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,274 @@
...
+                    }
+                    $('#button_add_blacklist').show();
+                    $('#button_remove_blacklist').hide();
+                    $('#unsubscribed_info').hide();
+                }
+            })
+            .guardedCatch(function () {
+                $('#subscription_info').html(_t('An error occured. Please try again later or contact us.'));
+                $('#info_state').removeClass('invisible');
+                $info_state.removeClass('alert-success').removeClass('alert-info').removeClass('alert-warning').addClass('alert-error');
```
Tardo, line 261:
```suggestion
                $info_state.removeClass('alert-success alert-info alert-warning').addClass('alert-error');
```
**Check other places with similar "call chain"...

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,274 @@
...
+            });
+    });
+
+    $('#button_remove_blacklist').click(function (e) {
+        e.preventDefault();
+
+        ajax.jsonRpc('/mailing/blacklist/remove', 'call', {'email': email, 'mailing_id': mailing_id, 'res_id': res_id, 'token': token})
+            .then(function (result) {
+                if (result == 'unauthorized'){
```
Tardo, line 234:
```suggestion
                if (result === 'unauthorized'){
```

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,274 @@
...
+/* This JS module replaces core AJAX submission because it is impossible
+ * to extend it as it is currently designed. It is almost a copy+paste from
+ * upstream, to allow easier version/patch updates, so linter is disabled
+ * and prettier is ignore. */
+/* eslint-disable */
+// prettier-ignore
+odoo.define("mass_mailing_custom_unsubscribe.unsubscribe", function (require) {
+    "use strict";
+
+    var session = require('web.session');
```
Tardo, line 13: Maybe can update this file to ES6

Tardo (changes requested):

pedrobaeza:
@Tardo this was done for keeping as similar as possible upstream code, so better to not do that kind of review. @ernestotejeda you should put an initial comment about that on the file for avoiding this confusion.

mass_mailing_custom_unsubscribe/static/src/js/unsubscribe.js
```diff
@@ -0,0 +1,274 @@
+/* Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+ * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+
+/* This JS module replaces core AJAX submission because it is impossible
+ * to extend it as it is currently designed. It is almost a copy+paste from
+ * upstream, to allow easier version/patch updates, so linter is disabled
+ * and prettier is ignore. */
```
yajo, line 7: @Tardo I'll mark as resolved some of your comments because of what is explained here (btw @pedrobaeza you can see the comment is already in place).

pedrobaeza:
Merging then:

/ocabot merge nobump

PR #621, opened 2020-10-05
[ADD] mail_auto_subscribe

nmadhoc:

PR #754, opened 2021-08-09
[14.0][ADD] mail_autosubscribe

ivantodorovich:
This module allows you to configure partners that will be automatically in copy of their company's business documents.

For example, you can configure an accountant to be in copy of all invoices sent for a given commercial partner, regardless of the invoicing address.

image

**Configuration**

Go to Configuration > Technical > Automation > Autosubscribe Models and configure the models for which you want the feature to work.

Then, on each partner, you can check the company documents subscriptions in the field `In copy of`.

### DIFF
--- a/mail_autosubscribe/models/models.py
+++ b/mail_autosubscribe/models/models.py
@@ -1,0 +1,37 @@
+# Copyright 2021 Camptocamp (http://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class BaseModel(models.AbstractModel):
+    _inherit = "base"
+
+    @api.model
+    def _message_get_autosubscribe_followers_domain(self, partners):
+        return [
+            ("id", "child_of", partners.commercial_partner_id.ids),
+            ("mail_autosubscribe_ids.model", "=", self._name),
+        ]
+
+    @api.model
+    def _message_get_autosubscribe_followers(self, partners):
+        domain = self._message_get_autosubscribe_followers_domain(partners)
+        return self.env["res.partner"].sudo().search(domain)
+
+    def _message_get_default_recipients(self):
+        # Overload to include auto follow document partners in the composer
+        # Note: This only works if the template is configured with 'Default recipients'
+        res = super()._message_get_default_recipients()
+        if self.env.context.get("no_autosubscribe_followers"):
+            return res
+        for rec in self:
+            partner_ids = res[rec.id]["partner_ids"]
+            partners = self.env["res.partner"].sudo().browse(partner_ids)
+            followers = rec._message_get_autosubscribe_followers(partners)
+            follower_ids = [
+                follower.id for follower in followers if follower not in partners
+            ]
+            partner_ids += follower_ids
+        return res
--- a/mail_autosubscribe/tests/models/fake_order.py
+++ b/mail_autosubscribe/tests/models/fake_order.py
@@ -1,0 +1,13 @@
+# Copyright 2021 Camptocamp (http://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class FakeOrder(models.Model):
+    _name = "fake.order"
+    _inherit = "mail.thread"
+    _description = "Fake sale.order like model"
+
+    partner_id = fields.Many2one("res.partner", required=True)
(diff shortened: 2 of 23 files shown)


ivantodorovich:
@jjscarafia I think we talked about this at some point, maybe you find it interesting :)

jjscarafia:
Hi @ivantodorovich 
Thanks for pinging.
We've developed a similar module but using partner tags.
The PR don't have a nice description, the person that was working on that leaves the company but we're using this module on many customers https://github.com/OCA/social/pull/621/

Any thoughts? I will also take a look to this one. 

Thanks!

ivantodorovich:
> Hi @ivantodorovich
> Thanks for pinging.
> We've developed a similar module but using partner tags.
> The PR don't have a nice description, the person that was working on that leaves the company but we're using this module on many customers #621
> 
> Any thoughts? I will also take a look to this one.
> 
> Thanks!


I see! Actually it's pretty similar indeed, only here we have this many2many_checkboxes to handle it (I might be biased, but I prefer it over tags)

image

We both chose to overload `message_subscribe`, but I see in the other PR that there's no special handling for the message composer. I found it was necessary in order to work properly with the sales.order workflow, for example.. where the customer is only subscribed upon order confirmation, and if you send a quotation before confirming the order and don't handle this case, the "autosubscriber followers" won't be in copy.

simahawk:
personally I think this solution is more complete and clean. The one in #621 could be rebased on this one by extending the lookup of the partners by category.

mail_autosubscribe/models/models.py
```diff
@@ -0,0 +1,38 @@
...
+class BaseModel(models.AbstractModel):
+    _inherit = "base"
+
+    @api.model
+    def _message_get_autosubscribe_followers(self, partners):
+        return (
+            self.env["res.partner"]
+            .sudo()
+            .search(
+                [
```
simahawk: move this domain out to an helper method so that it can be easily overridden by extending modules like on #621 (to be renamed as `mail_autosubscribe_partner_category`)

mail_autosubscribe/tests/models/fake_order.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2021 Camptocamp (http://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class FakeOrder(models.Model):
```
simahawk, line 8: res.partner already inherits from mail.thread, you could maybe work w/ it for tests?
Not a blocker.

mail_autosubscribe/views/res_partner.xml
```diff
@@ -0,0 +1,35 @@
...
+<odoo>
+
+    <record id="view_partner_form" model="ir.ui.view">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='email']/parent::div" position="after">
+                <field
+                    name="mail_autosubscribe_ids"
+                    string="In copy of"
```
simahawk, line 16:
```suggestion
                    string="In copy for"
```
maybe?

jjscarafia:
Hi! yes, the main difference (aside code style / bugs) is the approach of using a new model vs re-using partner category.
We take the approach of using partner category because:
* we avoid a new module
* partner categories / tags are already used to classify partners on some criteria, so only by classifying you could also allow to make him auto-follow related documents

I don't see a real value of having both modules, so keeping one is enough for me. If you believe the new model approach is better, no problem for me.

Thanks!

mail_autosubscribe/models/models.py
ivantodorovich: done :+1:

mail_autosubscribe/tests/models/fake_order.py
ivantodorovich: I thought so but I figured this makes for a more real-case use case

mail_autosubscribe/views/res_partner.xml
ivantodorovich: My english instinct is telling me it's _In copy of_ :man_shrugging:

simahawk (approved):

rousseldenis (approved):
LGTM

jjscarafia:
Hi @ivantodorovich @simahawk 
We've start using this module instead of https://github.com/OCA/social/pull/621/

I now realise that there was another difference. On our proposal we allow to add many models on one auto subscribe rule (partner category on our side).

What do you think about this? If it seems a good improvement we can add it when migrating to v15. 

One use case could be:
* create an auto-subscribe rull called "Invoicing & and payments" or "Accounting docs" or "Administrative"
* choose as models "account.move" and "account.payment"
* then, on the setup of the partner, the user can think more on a "user function" rather than models, so he can check the "Accounting Docs" instead of choosing many options.


Any thoughts?

ivantodorovich:
Yes I think it makes sense :+1:

jjscarafia:
Thanks @ivantodorovich 
We'll do it while migrating to v15. Thanks!

PR #757, opened 2021-08-12
[MIG] mail_notification_custom_subject: Migration to 14.0

naglis:
Most notable changes:
- `mail.thread` `message_post` and `mail.template` `_render_template` signatures changed;
- Changed `mail.message.custom.subject` `model_id` `ondelete` attribute to
  `cascade` (the default for required fields is `restrict`) as you now cannot have `ondelete="restrict"` for IR models: https://github.com/odoo/odoo/blob/f964b761eae32016e36dd9d39cc2c35fc2270eba/odoo/fields.py#L2601-L2606:
    > `Field model_id of model mail.message.custom.subject is defined as ondelete='restrict' while having ir.model as comodel, the 'restrict' mode is not supported for this type of field as comodel.`
- (_Tests_) `assertEqual` instead of `assertEquals` (deprecated);
- (_Tests_) `SavepointCase` + `tracking_disable=True` context to reduce the number of queries.


hbrunn (approved):

pedrobaeza (approved):
Can you please squash the administrative commits into the migration one?

Selección_019

naglis:
> Can you please squash the administrative commits into the migration one?

@pedrobaeza, it is done.

Issue #750, opened 2021-07-23
[12.0][mail_activity_done] - CRM Pipeline Progress Bar issue


kurtbinspired:
_read_progress_bar() function on mail.activity.mixin (odoo/addons/mail/models/mail_activity.py) has been refactored to be a SQL statement for performance:
```
SELECT 1 AS id, count(*) AS "__count", {fields}
            FROM {from_clause}
            JOIN (
                SELECT res_id,
                CASE
                    WHEN min(date_deadline - (now() AT TIME ZONE COALESCE(res_partner.tz, %s))::date) > 0 THEN 'planned'
                    WHEN min(date_deadline - (now() AT TIME ZONE COALESCE(res_partner.tz, %s))::date) < 0 THEN 'overdue'
                    WHEN min(date_deadline - (now() AT TIME ZONE COALESCE(res_partner.tz, %s))::date) = 0 THEN 'today'
                    ELSE null
                END AS activity_state
                FROM mail_activity
                JOIN res_users ON (res_users.id = mail_activity.user_id)
                JOIN res_partner ON (res_partner.id = res_users.partner_id)
                WHERE res_model = '{model}'
                GROUP BY res_id
            ) AS "_last_activity_state" ON ("{table}".id = "_last_activity_state".res_id)
            WHERE {where_clause}
            GROUP BY {group_by}
```
            
this conflicts with the 'done' and 'active' fields added by mail_activity_done; 'done' activities are being included in the JOIN, causing the progress bar to show a large number of opportunities/leads as 'overdue'.
In order to work, Odoo's SQL statement would need to be to:
```
SELECT 1 AS id, count(*) AS "__count", {fields}
            FROM {from_clause}
            JOIN (
                SELECT res_id,
                CASE
                    WHEN min(date_deadline - (now() AT TIME ZONE COALESCE(res_partner.tz, %s))::date) > 0 THEN 'planned'
                    WHEN min(date_deadline - (now() AT TIME ZONE COALESCE(res_partner.tz, %s))::date) < 0 THEN 'overdue'
                    WHEN min(date_deadline - (now() AT TIME ZONE COALESCE(res_partner.tz, %s))::date) = 0 THEN 'today'
                    ELSE null
                END AS activity_state
                FROM mail_activity
                JOIN res_users ON (res_users.id = mail_activity.user_id)
                JOIN res_partner ON (res_partner.id = res_users.partner_id)
                WHERE res_model = '{model}'
                    AND done = 'false'
[...]

PR #758, opened 2021-08-13
[14.0][MIG]mail_activity_done

manuelregidor:
Standard migration


manuelregidor:
@HaraldPanten @ValentinVinagre

mail_activity_done/models/mail_activity.py
```diff
@@ -9,16 +9,14 @@ class MailActivity(models.Model):
 
     active = fields.Boolean(default=True)
     done = fields.Boolean(default=False)
-    state = fields.Selection(selection_add=[("done", "Done")], compute="_compute_state")
-    date_done = fields.Date(
-        "Completed Date",
-        index=True,
-        readonly=True,
+    state = fields.Selection(
+        selection_add=[("done", "Done")], compute="_compute_state", store=True
```
CasVissers-360ERP: Why did you make the field stored?

CasVissers-360ERP:
@manuelregidor See #750 Maybe not doable to fix this in a clean way without asking Odoo for an option to inherit this method.

mail_activity_done/models/mail_activity.py
manuelregidor: Hi @CasVissers-360ERP I made it stored as you get an error with filter "Completed Activities" in mail_activity_view_search  when field state is not stored. Any suggestion? Thank you.

CasVissers-360ERP: Nop! Only wondering why this was not an issue in the v13 version?

CasVissers-360ERP: @manuelregidor In addition this change breaks the standard method of Odoo since the state is based on the date of the activity. When you open the list view you want the state to be computed based on the current date, the current depends don't solve this.

Rad0van:
> @manuelregidor See #750 Maybe not doable to fix this in a clean way without asking Odoo for an option to inherit this method.

@CasVissers-360ERP is there a dirty way of doing this in the meantime? And has someone asked Odoo for allowing to inherit the method? Thank you very much.

Rad0van:
@CasVissers-360ERP also why is there hesitation of using dirty ways when this module relies on monkey patching MailActivity._action_done in post_load hook? Sorry for stupid questions but trying to learn...

AaronHForgeFlow (commented):
Another option may be to add a boolean that indicates the activity is done.

BTW you have to rebase this PR

mail_activity_done/models/mail_activity.py
Rad0van: @CasVissers-360ERP was looking on why it was not an issue in 13.0 The thing is validation was changed in 14.0:
postprocess method in ir_ui_view calls _validate_attrs which wasn't there in 13.0 and has the following code;

https://github.com/odoo/odoo/blob/63500d6357f6f27724d1cda44db071aba968515a/odoo/addons/base/models/ir_ui_view.py#L1292-L1294

That _get_server_domain_variables is the culprit as it states:
https://github.com/odoo/odoo/blob/63500d6357f6f27724d1cda44db071aba968515a/odoo/addons/base/models/ir_ui_view.py#L1528-L1532

And the definition of that `_description_searchable` is:
https://github.com/odoo/odoo/blob/63500d6357f6f27724d1cda44db071aba968515a/odoo/fields.py#L687-L689

So that means for this to be working without the state field being stored we need to provide search method.

Rad0van: Update: adding search method on the `state` fields makes it work. At least it seems so. And the method doesn;t even seems to be called. Ever.

Rad0van:
```suggestion
        selection_add=[("done", "Done")], compute="_compute_state", search="_search_state"
```

mail_activity_done/models/mail_activity.py
```diff
@@ -0,0 +1,29 @@
...
+    )
+    date_done = fields.Date("Completed Date", index=True, readonly=True)
+
+    @api.depends("date_deadline", "done")
+    def _compute_state(self):
+        super()._compute_state()
+        for record in self.filtered(lambda activity: activity.done):
+            record.state = "done"
+
```
Rad0van, line 24: This will make it work:

```suggestion

    def _search_state(self, operator, operand):
        # This is never called even
        return []
```

Rad0van (changes requested):
Suggested changes to make it work.

Rad0van:
@manuelregidor after a bit of toying with this I found out my previous statement wasn't true. The search method actually gets called when there is an attempt to filter by state. So in order for this to work correctly an implementation needs to be done.

mail_activity_done/models/mail_activity.py
```diff
@@ -0,0 +1,35 @@
...
+    def _search_state(self, operator, operand):
+        # This is never called even
+        return []
```
Rad0van, line 27:
```suggestion
    def _search_state(self, operator, operand):
        if operand == False:
            # checking for is (not) set
            if operator == '=':
                # is not set - never happens actually so we create impossible domain
                return [('id', '=', False)]
            else:
                # is set - always - return empty domain
                return []
        else:
            # checking for value
            if operand == 'done':
                return ['&', ('done', operator, True), ('active', '=', False)]
            else:
                if operator == '=':
                    return ['&', ('date_deadline', {'today': '=', 'overdue': '<', 'planned': '>'}[operand], fields.Date.today()), ('done', '=', False)]
                else:
                    return ['|', ('date_deadline', {'today': '!=', 'overdue': '>=', 'planned': '<='}[operand], fields.Date.today()), ('done', '=', True)]
```

Rad0van (changes requested):
Here's update on search method.

mail_activity_done/models/mail_activity.py
Rad0van:
```suggestion
        if operand == False:
            # checking for is (not) set
            if operator == '=':
                # is not set - never happens actually so we create impossible domain
                return [('id', '=', False)]
            else:
                # is set - always - return empty domain
                return []
        else:
            # checking for value
            if operand == 'done':
                if operator == '=':
                    return ['&', ('done', operator, True), ('active', '=', False)]
                else:
                    return ['&', ('done', operator, False), ('active', '=', True)]
            else:
                if operator == '=':
                    return ['&', ('date_deadline', {'today': '=', 'overdue': '<', 'planned': '>'}[operand], fields.Date.today()), ('done', '=', False)]
                else:
                    return ['|', ('date_deadline', {'today': '!=', 'overdue': '>=', 'planned': '<='}[operand], fields.Date.today()), ('done', '=', True)]
```

Rad0van (changes requested):
Another attempt.

mail_activity_done/models/mail_activity.py
AaronHForgeFlow: @manuelregidor can you try this?

Rad0van:
@manuelregidor I guess you could add me to contributors as I have done some work here ;-)

JoanMForgeFlow (approved):

scigghia (approved):

HaraldPanten (approved):
@manuelregidor you can add yourself as contributor as well 👍

AaronHForgeFlow (approved):

JoanMForgeFlow (commented):
Could we do merge please?

marcelsavegnago (approved):
LGTM

AaronHForgeFlow:
@Rad0van can you approve here? :)

Rad0van (approved):
LGTM

HaraldPanten:
@pedrobaeza @simahawk Could you merge?

THX!

AaronHForgeFlow:
@OCA/social-maintainers ready to merge

ValentinVinagre (approved):
LGTM 👍🏻

PR #759, opened 2021-08-13
[12.0] mail_tracking: avoid common users accessing to general trackings

chienandalu:
A user accessing to all trackings with no restrictions could view some sensible informations contained in the mail tracking mail subject. For example, employees leave causes.

Refine which mail tracking record is allowed to see. We can rely on the ACLs of the related records so we minimize inconvenient message subject leaks.

A regular user can read these mail.tracking.email records:

- Those with a linked mail.message that the user can read.
- Those with a linked mail.mail that the user can read.
- Those with no message/mail link but a linked partner that the user can read.
- Those with no linked records.

TT31399

cc @Tecnativa

### DIFF
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -8,7 +8,8 @@ import re
 import uuid
 from datetime import datetime
 
-from odoo import models, api, fields, tools
+from odoo import _, models, api, fields, tools
+from odoo.exceptions import AccessError
 import odoo.addons.decimal_precision as dp
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


pedrobaeza:
@chienandalu what happened with this?

chienandalu:
I have to finish it, yes

mail_tracking/static/src/js/mail_tracking.js
```diff
@@ -140,24 +141,28 @@ odoo.define('mail_tracking.partner_tracking', function (require) {
...
-                view_type: 'form',
-                view_mode: 'form',
-                res_model: 'mail.tracking.email',
-                views: [[false, 'form']],
-                target: 'new',
-                res_id: tracking_email_id,
-            };
-            this.do_action(action);
+            // Only admins can check advanced tracking stuff
+            session.user_has_group('base.group_system').then(has_group => {
```
pedrobaeza: Why this limitation? It's very useful for people to check the tracking lines.

chienandalu: I took a new approach dinamically evaluating acls the same way mail.message does relaying on the related records permissions to filter the trackings a user can see.

pedrobaeza:
Please check CI

chienandalu:
Fixed

mail_tracking/models/res_partner.py
```diff
@@ -26,7 +26,7 @@ def _compute_tracking_emails_count(self):
         for partner in self:
             count = 0
             if partner.email:
-                count = self.env['mail.tracking.email'].search_count([
+                count = len(self.env['mail.tracking.email']._search([
```
pedrobaeza, line 33: Isn't this going to make very slow the partner view loading? What about simply not putting the number on the smart-button?

pedrobaeza:
Can you display any banner saying that the shown records may not be all depending on the access rules?

chienandalu:
> Can you display any banner saying that the shown records may not be all depending on the access rules?

Can it be done in the tree view or do we have to use a `banner_route`?

pedrobaeza:
I can't say, sorry

mail_tracking/models/res_partner.py
chienandalu: Ok, I took the middle path: admins can see the number (they skip ACLs) but regular internal users don't

mail_tracking/models/mail_tracking_email.py
```diff
@@ -117,11 +118,83 @@ def write(self, vals):
...
+            + " - ({} {}, {} {})".format(
+                _("Records:"),
+                list(disallowed_ids),
+                _("User:"), self._uid
+            )
+        )
+
+    def read(self, fields=None, load="_classic_read"):
+        """Override to explicitely call check_access_rule, that is not called
+        by the ORM. It instead directly fetches ir.rules and apply them."""
```
pedrobaeza:
```suggestion
        """Override to explicitly call check_access_rule, that is not called
        by the ORM. It instead directly fetches ir.rules and apply them.
        """
```

pedrobaeza (approved):

mail_tracking/models/mail_tracking_email.py
chienandalu: Done

carlosdauden (approved):

pedrobaeza:
Well, let's fw-port it to v13 and up and let this one as is.

PR #760, opened 2021-08-13
[FIX] mail_tracking: Added restriction on email search.

### LINKED ISSUE #752: [14.0] mail_tracking - Contact > 'MailTracking emails' is searching for email using contains instead of equal
Pre-Condition: Your database should have enough data so you can see test result and emails sent to those contacts. 

Steps to Reproduce:
1. Go to a Contact
2. Go to 'MailTracking emails' menu

Issue: The 'MailTracking emails' searches for the main email from the entire database for all emails related to the given contact. 
It uses contains for search instead of equal. 

For example: Let's say contact's main email = saman@gmail.com
Database has other emails as saman.nisar@gmail.com , saman.shahid@gmail.com, saman.sameer@gmail.com, samankhan@yahoo.com, chsamanahmed@hotmail.com   (Random emails used for examples)

In this case the contact's mail tracking email shows  Emails from all above email address as related to the given contact. 
But in reality those emails belong to other contact and not this contact. 

System uses "saman" , "gmail" and search for all emails with "saman" and "gmail" in the email.   

Please fix and use the equal value for searching so given contact only displays emails related to that contact ONLY.

angelJuaGar:
A filter_domain was added in the sender and recipient fields to be able to handle a more specific filter when using mail tracking.

Fix https://github.com/OCA/social/issues/752

### DIFF
--- a/mail_tracking/views/mail_tracking_email_view.xml
+++ b/mail_tracking/views/mail_tracking_email_view.xml
@@ -99,8 +99,16 @@
                     string="Email"
                     filter_domain="['|', ('sender', 'ilike', self), ('recipient', 'ilike', self)]"
                 />
-            <field name="sender" string="Sender" />
-            <field name="recipient" string="Recipient" />
+            <field
+                    name="sender"
+                    string="Sender"
+                    filter_domain="[('sender', '=', self)]"
+                />
+            <field
+                    name="recipient"
+                    string="Recipient"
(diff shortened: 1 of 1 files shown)


fernandahf:
LGTM

cc @moylop260

fernandahf (approved):

moylop260 (approved):
👍

PR #762, opened 2021-08-18
[13.0] [FWD] [ADD] mail_autosubscribe: New Module

nicomacr:


nicomacr:
@ivantodorovich i've backported this module from v14, i've problems with 2 tests cases [here](https://app.travis-ci.com/github/OCA/social/jobs/532859123)
I tried in local and works fine, any thought about it?

ivantodorovich:
> @ivantodorovich i've backported this module from v14, i've problems with 2 tests cases [here](https://app.travis-ci.com/github/OCA/social/jobs/532859123)
> I tried in local and works fine, any thought about it?

Hmmm, you mean tests are working locally?
If that's the case there may be a conflict with another module in this repo. It's hard to say just by looking at it. Check locally if tests are still passing with other modules installed on the test db, specially ones that suggest messing with followers and that are not yet migrated to 14.0, like this one for instance: https://github.com/OCA/social/tree/13.0/mail_restrict_follower_selection

nicomacr:
@ivantodorovich yes you're right, the module https://github.com/OCA/social/tree/13.0/mail_restrict_follower_selection it's the cause of the fail of the test cases.
This module add a parameter with a domain who are filter by this domain when add a follower.
I've made some changes for solve this and test in locally and works.

ivantodorovich:
I'm not a fan of changing this module's tests to adapt for the fact `mail_restrict_follower_selection` is messing with add followers

One way of dealing with this would be to have travis exclude `mail_restrict_follower_selection` from the general tests and runs it in a separate build. If I'm not mistaken I think it's possible with the `EXCLUDE` env var set in `travis.yaml`, but I don't have enough experience to say exactly how 

ping @OCA/social-maintainers  ?

hbrunn:
yes, better test those modules apart then. You can find an example in https://github.com/OCA/server-tools/blob/12.0/.travis.yml#L21

mail_autosubscribe/tests/test_mail_autosubscribe.py
```diff
@@ -54,7 +54,9 @@ def tearDownClass(cls):
     def test_message_subscribe(self):
         """Test autosubscribe on a basic workflow"""
         self.assertFalse(self.order.message_partner_ids, "No subscribers yet")
-        self.order.message_subscribe([self.order.partner_id.id])
+        self.order.with_context(no_restrict_follower=True).message_subscribe(
```
ivantodorovich: Since conflicting modules are tested separately now, this context key can be removed :)
(same for the other one a few lines below)

ivantodorovich (approved):

nicomacr:
@hbrunn i've made the changes as you said, but the test still failed, whats it's wrong?

.travis.yml
```diff
@@ -28,6 +28,18 @@ env:
...
+    - TESTS="1" ODOO_REPO="OCA/OCB"
+      EXCLUDE="mail_restrict_follower_selection,mail_autosubscribe"
+    - TESTS="1" ODOO_REPO="OCA/OCB" INCLUDE="mail_restrict_follower_selection"
+    - TESTS="1" ODOO_REPO="OCA/OCB" INCLUDE="mail_autosubscribe"
+    - TESTS="1" ODOO_REPO="odoo/odoo"
+      EXCLUDE="mail_restrict_follower_selection,mail_autosubscribe" MAKEPOT="1"
+    - TESTS="1" ODOO_REPO="odoo/odoo" INCLUDE="mail_restrict_follower_selection"
+      MAKEPOT="1"
+    - TESTS="1" ODOO_REPO="odoo/odoo" INCLUDE="mail_autosubscribe" MAKEPOT="1"
+
```
ivantodorovich:
```suggestion
```

I think instead of doing this in `matrix` (as in the provided 12.0 example) it has to be done in `jobs`, because travis is configured differently in 12.0.

Try to do like this instead (replacing the current `jobs` config with this):


```yml
jobs:
  include:
    # Test separately: mail_restrict_follower_selection
    - stage: test
      env:
        - TESTS=1 ODOO_REPO="odoo/odoo" MAKEPOT=1 INCLUDE="mail_restrict_follower_selection"
    - stage: test
      env:
        - TESTS=1 ODOO_REPO="OCA/OCB" INCLUDE="mail_restrict_follower_selection"
    # Test all other addons together
    - stage: test
      env:
        - TESTS=1 ODOO_REPO="odoo/odoo" MAKEPOT=1 EXCLUDE="mail_restrict_follower_selection"
    - stage: test
      env:
        - TESTS=1 ODOO_REPO="OCA/OCB" EXCLUDE="mail_restrict_follower_selection"
```

Based on https://github.com/OCA/oca-addons-repo-template/blob/master/.travis.yml.jinja


Note that on this example I'm only excluding `mail_restrict_follower_selection` from the general tests, because it's the only one doing blocking things IMO (ie: preventing subscribing partners). `mail_autosubscribe` is not doing that and should play nicely with others.

nicomacr: Thanks @ivantodorovich i thinks the same as you about to  only exclude `mail_restrict_follower_selection`.

hbrunn:
@nicomacr the tests themselves are fine now, thanks so much for your efforts.

What's screaming now is coverage, I used to be picky about that but lately got a bit more flexible on it. Fine for me and thanks

hbrunn (approved):

HaraldPanten (approved):

PR #766, opened 2021-09-09
[14.0][ADD] mass_mailing_contact_unique

ivantodorovich:
Prevent duplicate email addresses in mass mailing contacts.

### DIFF
--- a/mass_mailing_contact_unique/__init__.py
+++ b/mass_mailing_contact_unique/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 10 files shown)


pedrobaeza:
Isn't this the same as https://github.com/OCA/social/tree/12.0/mass_mailing_unique ?

ivantodorovich:
> Isn't this the same as https://github.com/OCA/social/tree/12.0/mass_mailing_unique ?

:open_mouth: indeed! I didn't see it before, thanks!

I'll migrate it instead!

it's doing slightly a bit more than what I need though. What do you think if I split it in two: `mass_mailing_contact_unique` and `mass_mailing_list_unique` ?

pedrobaeza:
Do you really think the split is needed? It's pretty basic...

ivantodorovich:
> Do you really think the split is needed? It's pretty basic...

They're 2 different requirements IMO
But ok I keep it as is, it doesn't hurt anyways

Closing this, superseeded by https://github.com/OCA/social/pull/773

PR #767, opened 2021-09-09
[14.0][ADD] mass_mailing_contact_partner

ivantodorovich:
This module links mailing.contacts with res.partners.

This module is different than `mass_mailing_partner` because the relationship is in the opposite direction.
The same mailing.contact can be linked to multiple partners, if they share the same email address.

It's recommended to use it with `mass_mailing_unique` module (https://github.com/OCA/social/pull/773).

### DIFF
--- a/mass_mailing_contact_partner/models/res_partner.py
+++ b/mass_mailing_contact_partner/models/res_partner.py
@@ -1,0 +1,37 @@
+# Copyright 2021 Camptocamp (http://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    mailing_contact_id = fields.Many2one(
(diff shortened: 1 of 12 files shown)


damdam-s (approved):

mass_mailing_contact_partner/models/res_partner.py
```diff
@@ -0,0 +1,37 @@
...
+        compute="_compute_mailing_contact_id",
+        store=True,
+    )
+
+    @api.depends("email_normalized")
+    def _compute_mailing_contact_id(self):
+        # Unique list of normalized email addresses
+        emails = {m for m in self.mapped("email_normalized") if m}
+        if not emails:
+            self.mailing_contact_id = False
```
simahawk, line 24:
```suggestion
            self.update({"mailing_contact_id": False})
```

simahawk (approved):

mass_mailing_contact_partner/models/res_partner.py
ivantodorovich: Is there a reason for this? Same result but less readable IMO 🤔

simahawk: nvm :)

PR #714, opened 2021-05-18
[14.0][IMP] mail_debrand: Ensure links to users website are not stripped

ryanc-me:
See #713 for more details.

The original regexes were naively searching for domains that matched `odoo.com`. This included Odoo.sh/Odoo Online users (`my-company.odoo.com`). The final-pass regex was also matching text snippets it shouldn't (e.g. `https://odoo.mycompany.com` -> `https:/.mycompany.com`, note the missing `/` because `/odoo` was matched).

The new regexes:
 - [Search the doc to find any `href="odoo.com"`-like snippets](https://regex101.com/r/o01Rfz/3)
 - [Search for an `<a>` element where the `href` attribute is like "odoo.com"](https://regex101.com/r/o01Rfz/5)
 - [Final pass - search for any occurances of `odoo`, and delete them](https://regex101.com/r/FLTxID/11)

I'm not 100% convinced this is the best approach, because it can leave things a bit... weird. For example:
`This website is powered by Odoo!` -> `This website is powered by!`. A more in-depth analysis of Odoo branding on email templates would probably be good.

But at least now it isn't breaking client URLs!

PR #770, opened 2021-09-13
[14.0][IMP] mail_debrand: Keep message body intact while removing branding

### LINKED ISSUE #713: [14.0] mail_debrand: breaks email redirect links in Odoo EE
The regex (`[^(<)(</)]odoo`) in the mail_debrand module seems to be broken.

Firstly - and please correct me if I'm wrong - Python does not support captures like `(xx)` inside groups like `[xx]`. The syntax `[^(<)(</)]` is matching anything that is not `(`, `<`, `/`, or `)`. I believe the intention was to negatively match against `<` or `</`. 

Secondly, this regex _also_ matches any anchor/href containing `odoo`. For example, we have quite a few clients who host Odoo at `https://odoo.company.com`. This regex strips the `/odoo` portion, leaving us with `https:/.company.com`.

Thirdly, the regex is running `.sub()` on the _whole_ match. As above, the match include a single character that is not in `()</`, so it's stripping an extra character preceding `odoo`. Maybe this is the intended behavior?

<hr />

I'd love to make a PR to fix this, but I'm not clear on what the regex is trying to do.

I think maybe `(?<!<)(?<!</)(odoo)(?!\.)` would be a good alternative:
 - Prevents matching if `odoo` is directly preceded by `<` or `</`
 - Prevents matching if `odoo` is directly followed by `.`
 - Does not _match_ the preceding or following part (e.g. `href="odoo" a=1` -> `href="" a=1`, rather than `href= a=1`)

You can test it at on regex101.com [here](https://regex101.com/r/FLTxID/3).

@alexteodor could you clarify a bit on the intent of this regex? Especially, why are we negatively matching `<` and `</`? Also, I'd love to hear your thoughts on the above regex.

joao-p-marques:
Partially fixes https://github.com/OCA/social/issues/713

@Tecnativa
TT31737

ping @pedrobaeza @Yajo @ryanc-me

### DIFF
--- a/mail_debrand/models/mail_render_mixin.py
+++ b/mail_debrand/models/mail_render_mixin.py
@@ -1,6 +1,7 @@
 # Copyright 2019 O4SB - Graeme Gellatly
 # Copyright 2019 Tecnativa - Ernesto Tejeda
 # Copyright 2020 Onestein - Andrea Stirpe
+# Copyright 2021 Tecnativa - João Marques
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 import re
 
@@ -12,7 +13,9 @@ from odoo import api, models
 class MailRenderMixin(models.AbstractModel):
     _inherit = "mail.render.mixin"
 
-    def remove_href_odoo(self, value, remove_parent=True, remove_before=False):
+    def remove_href_odoo(
+        self, value, remove_parent=True, remove_before=False, to_keep=None
+    ):
         if len(value) < 20:
             return value
         # value can be bytes type; ensure we get a proper string
@@ -20,35 +23,41 @@ class MailRenderMixin(models.AbstractModel):
             value = value.decode()
         has_odoo_link = re.search(r"<a\s(.*)odoo\.com", value, flags=re.IGNORECASE)
         if has_odoo_link:
-            tree = etree.HTML(
-                value
-            )  # html with broken links   tree = etree.fromstring(value) just xml
-            odoo_achors = tree.xpath('//a[contains(@href,"odoo.com")]')
-            for elem in odoo_achors:
-                parent = elem.getparent()
-                previous = elem.getprevious()
+            # We don't want to change what was explicitly added in the message body,
+            # so we will only change what is before and after it.
+            if to_keep:
+                to_change = value.split(to_keep)
+            else:
+                to_change = [value]
+                to_keep = ""
+            new_parts = []
+            for part in to_change:
+                tree = html.fromstring(part)
+                if tree is None:
+                    new_parts.append(part)
+                    continue
+                odoo_anchors = tree.xpath('//a[contains(@href,"odoo.com")]')
+                for elem in odoo_anchors:
+                    parent = elem.getparent()
+                    previous = elem.getprevious()
 
-                if remove_before and not remove_parent and previous is not None:
-                    # remove 'using' that is before <a and after </span>
-                    bytes_text = etree.tostring(
-                        previous, pretty_print=True, method="html"
-                    )
-                    only_what_is_in_tags = bytes_text[: bytes_text.rfind(b">") + 1]
-                    data_formatted = html.fromstring(only_what_is_in_tags)
-                    parent.replace(previous, data_formatted)
-                if remove_parent and len(parent.getparent()):
-                    # anchor <a href odoo has a parent powered by that must be removed
-                    parent.getparent().remove(parent)
-                else:
-                    if parent.tag == "td":  # also here can be powerd by
+                    if remove_before and not remove_parent and previous is not None:
+                        # remove 'using' that is before <a and after </span>
+                        previous.tail = ""
+                    if remove_parent and len(parent.getparent()):
+                        # anchor <a href odoo has a parent powered by that must be removed
                         parent.getparent().remove(parent)
                     else:
-                        parent.remove(elem)
-            value = etree.tostring(tree, pretty_print=True, method="html")
-            # etree can return bytes; ensure we get a proper string
-            if type(value) is bytes:
-                value = value.decode()
-        return re.sub("[^(<)(</)]odoo", "", value, flags=re.IGNORECASE)
+                        if parent.tag == "td":  # also here can be powered by
+                            parent.getparent().remove(parent)
+                        else:
+                            parent.remove(elem)
+                part = etree.tostring(
+                    tree, pretty_print=True, method="html", encoding="unicode"
+                )
+                new_parts.append(part)
+            value = to_keep.join(new_parts)
+        return value
 
     @api.model
     def _render_template(
--- a/mail_debrand/tests/test_mail_debrand.py
+++ b/mail_debrand/tests/test_mail_debrand.py
@@ -1,6 +1,7 @@
 # Copyright 2017 Tecnativa - Pedro M. Baeza
 # Copyright 2020 Onestein - Andrea Stirpe
 # Copyright 2021 Sodexis
+# Copyright 2021 Tecnativa - João Marques
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo.tests import common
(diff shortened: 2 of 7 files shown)


mail_debrand/tests/test_mail_debrand.py
```diff
@@ -65,3 +65,19 @@ def test_plaintext_email(self):
         }
         # No exception expected
         MailMessage.create(email_values)
+
+    def test_body_intact(self):
+        """The message body should never be changed"""
+        MailMessage = self.env["mail.mail"]
+        original_body = """<p>And if I send odoo.example.com<br><br>And odoo.com<br><br>And <a target="_blank" rel="noreferrer noopener" href="https://odoo.com">https://odoo.com</a><br><br>And <a target="_blank" rel="noreferrer noopener" href="https://odoo.example.com">https://odoo.example.com</a></p>"""  # noqa: B950
```
pedrobaeza: Can you please split this by lines (and the other one)?

pedrobaeza (approved):

mail_debrand/tests/test_mail_debrand.py
joao-p-marques: Done :+1: Thanks!

mail_debrand/models/mail_render_mixin.py
```diff
@@ -12,43 +12,61 @@
...
+            if to_keep:
+                to_change = value.split(to_keep)
+            else:
+                to_change = [value]
+                to_keep = ""
+            new_parts = []
+            for part in to_change:
+                tree = etree.HTML(
+                    part
+                )  # html with broken links   tree = etree.fromstring(value) just xml
```
yajo: I don't understand this comment.

mail_debrand/models/mail_render_mixin.py
```diff
@@ -12,43 +12,61 @@
...
+                if tree is None:
+                    new_parts.append(part)
+                    continue
+                odoo_achors = tree.xpath('//a[contains(@href,"odoo.com")]')
```
yajo: typo?

```suggestion
                odoo_anchors = tree.xpath('//a[contains(@href,"odoo.com")]')
```

mail_debrand/models/mail_render_mixin.py
```diff
@@ -12,43 +12,61 @@
...
+            # We don't want to change what was explicitly added in the message body,
+            # so we will only change what is before and after it.
+            if to_keep:
+                to_change = value.split(to_keep)
+            else:
+                to_change = [value]
+                to_keep = ""
+            new_parts = []
+            for part in to_change:
+                tree = etree.HTML(
```
yajo: As you are processing parts, possibly this is better suited:

```suggestion
                tree = html.fromstring(
```

mail_debrand/models/mail_render_mixin.py
```diff
@@ -12,43 +12,61 @@
...
-                else:
-                    if parent.tag == "td":  # also here can be powerd by
+                    if remove_before and not remove_parent and previous is not None:
+                        # remove 'using' that is before <a and after </span>
+                        bytes_text = etree.tostring(
+                            previous, pretty_print=True, method="html"
+                        )
+                        only_what_is_in_tags = bytes_text[: bytes_text.rfind(b">") + 1]
+                        data_formatted = html.fromstring(only_what_is_in_tags)
+                        parent.replace(previous, data_formatted)
```
yajo:
```suggestion
                        previous.tail = ""
```

mail_debrand/models/mail_render_mixin.py
```diff
@@ -12,43 +12,61 @@
...
+                        # anchor <a href odoo has a parent powered by that must be removed
                         parent.getparent().remove(parent)
                     else:
-                        parent.remove(elem)
-            value = etree.tostring(tree, pretty_print=True, method="html")
-            # etree can return bytes; ensure we get a proper string
-            if type(value) is bytes:
-                value = value.decode()
-        return re.sub("[^(<)(</)]odoo", "", value, flags=re.IGNORECASE)
+                        if parent.tag == "td":  # also here can be powerd by
```
yajo: I find it weird that you remove the parent even if you got `remove_parent=False`. Does that mean that we can just remove that argument from the method?

```suggestion
                        if parent.tag == "td":  # also here can be powered by
```

mail_debrand/models/mail_render_mixin.py
```diff
@@ -12,43 +12,61 @@
...
+                            parent.getparent().remove(parent)
+                        else:
+                            parent.remove(elem)
+                part = etree.tostring(tree, pretty_print=True, method="html")
+                # etree can return bytes; ensure we get a proper string
+                if type(part) is bytes:
+                    part = part.decode()
+                new_parts.append(
+                    re.sub("[^(<)(</)]odoo", "", part, flags=re.IGNORECASE)
```
yajo: Are there really any `<odoo> whatever </odoo>` elements added to emails? 🤔

mail_debrand/models/mail_render_mixin.py
```diff
@@ -12,43 +12,61 @@
...
-                value = value.decode()
-        return re.sub("[^(<)(</)]odoo", "", value, flags=re.IGNORECASE)
+                        if parent.tag == "td":  # also here can be powerd by
+                            parent.getparent().remove(parent)
+                        else:
+                            parent.remove(elem)
+                part = etree.tostring(tree, pretty_print=True, method="html")
+                # etree can return bytes; ensure we get a proper string
+                if type(part) is bytes:
+                    part = part.decode()
```
yajo:
```suggestion
                part = etree.tostring(tree, pretty_print=True, method="html", encoding="unicode")
```

yajo (changes requested):

mail_debrand/models/mail_render_mixin.py
joao-p-marques: Well, I think this regex is actually badly constructed, as @ryanc-me also suggested and tried to fix in https://github.com/OCA/social/pull/714

Right now, it is actually matching (according to https://regexr.com):
- A negated set of:
    - a `(` character
    - a `<` character
    - a `)` character
    - all the above again :sweat_smile: 
    - a `/` character
- The `odoo` word

So, it obviously needs changing... However, as @ryanc-me seems to have a fix ready in the PR mentioned above, I just wanted to fix another part of the problem, and then maybe we can proceed from what he already did...

joao-p-marques:
@Yajo thanks for the review!

Most of your comments actually refer to the previous code that was not my own, so I can't explain it completely :sweat_smile: ... That's why the goal with this PR was not to change it, but just to add this new checks to avoid extra changes... However, I attended most of them, as they where mostly typos and obvious fixes... 

As for the only one left unattended (about the `<odoo>` tags), as explained, maybe it is better to leave it to another PR, especially as there is one already open.

mail_debrand/models/mail_render_mixin.py
ryanc-me: @joao-p-marques you're exactly right about the regex, there's a bit more info in #713. The PR #714 does "fix" that issue (see this [regex101](https://regex101.com/r/o01Rfz/5)), but I don't think the regex approach is actually correct. It leaves us with `<span>Powered by <a></a></span>`, which is *technically* de-branding, but isn't a nice solution.

I'm not convinced we can remove all branding with a generic regex - maybe instead we need to find all strings like "Odoo" in the  CE/EE sources, and write some specific rules?


Also @Yajo, just for reference, I think the original author was trying not to match the `<odoo></odoo>` tags surrounding all Odoo templates (the top-level element in all XML docs)

joao-p-marques: Thanks for the explanation!

Well, with this PR, as we are making sure to leave the message body intact, we can be a bit more "strict" while reformatting the rest, which already solves a big part of the problem... From my testing, even with this wrong regex, we are being able to actually hide what we want without removing too much, so that takes some urgency out of the other part.

Once this is merged, maybe you can rebase your PR so that we can keep the discussion there and try out your changes.

yajo: > From my testing, even with this wrong regex, we are being able to actually hide what we want

Yes, that's what the code tells me. AFAICS we're already detecting the "odoo" parts properly and removing the garbage. This regex is useless now (it could be before too, but now it certainly is), so please just remove it.

In case there's some corner case not being detected now, then a future PR can fix it and provide a valid test for it.

joao-p-marques: Ok, removed then. Let's work on reviewing and #714 to make sure we don't miss anything and don't introduce any other errors...

joao-p-marques:
Changes done @Yajo

yajo (approved):

joao-p-marques:
The bot is broken :sweat_smile: https://twitter.com/PedroMBaeza/status/1438405500236963841?s=20

PR #773, opened 2021-09-20
[14.0][MIG] mass_mailing_unique

ivantodorovich:


ivantodorovich:
Travis is failing on module `mail_tracking_mass_mailing`,  but I can't see how it would have something to do with `mass_mailing_unique`, probably unrelated?  I'm unable to reproduce it locally o.O

damdam-s (approved):

simahawk:
> Travis is failing on module `mail_tracking_mass_mailing`, but I can't see how it would have something to do with `mass_mailing_unique`, probably unrelated? I'm unable to reproduce it locally o.O

have you tried installing all the modules as travis?

ivantodorovich:
> > Travis is failing on module `mail_tracking_mass_mailing`, but I can't see how it would have something to do with `mass_mailing_unique`, probably unrelated? I'm unable to reproduce it locally o.O
> 
> have you tried installing all the modules as travis?

Got it. It was actually `mass_mailing_partner` that was messing tests.
That module is doing pretty invasive things like this https://github.com/OCA/social/blob/bebbfbd72/mass_mailing_partner/models/mailing_contact.py#L63-L64 

For that reason I think we should test it separately: https://github.com/OCA/social/pull/773/commits/12fa468f0609982cc6ff0b705d308595752311b9

pedrobaeza:
> That module is doing pretty invasive things like this https://github.com/OCA/social/blob/bebbfbd72/mass_mailing_partner/models/mailing_contact.py#L63-L64
> 
> For that reason I think we should test it separately: [12fa468](https://github.com/OCA/social/commit/12fa468f0609982cc6ff0b705d308595752311b9)

I don't see it invasive, as everything is done in newId record, so no other constraints should be applied in such draft mode.

ivantodorovich:
> > That module is doing pretty invasive things like this https://github.com/OCA/social/blob/bebbfbd72/mass_mailing_partner/models/mailing_contact.py#L63-L64
> > For that reason I think we should test it separately: [12fa468](https://github.com/OCA/social/commit/12fa468f0609982cc6ff0b705d308595752311b9)
> 
> I don't see it invasive, as everything is done in newId record, so no other constraints should be applied in such draft mode.

Travis disagrees though :(

With `mass_mailing_unique`, we have `email = fields.Char(copy=False)`, to avoid the IntegrityError if the user tries to copy it.
That makes `mass_mailing_partner` NOT to copy the email on those 2 lines, which is later backfiring in other module's tests. I haven't dug too much on why `mass_mailing_partner` is copying data there, but it doesn't seem right. Seems it's not accounting for fields with `copy=False`.

Apart from that, there are other issues in `mass_mailing_partner` tests, like trying to create mailing lists with duplicate names:

- https://github.com/OCA/social/blob/bebbfbd72a554ca5bfbd7d0f5f1f33f21ec80810/mass_mailing_partner/tests/base.py#L31-L38
- https://github.com/OCA/social/blob/bebbfbd72a554ca5bfbd7d0f5f1f33f21ec80810/mass_mailing_partner/tests/test_mail_mass_mailing_contact.py#L116

```
  File "/odoo/external-src/social/mass_mailing_partner/tests/test_mail_mass_mailing_contact.py", line 116, in test_partners_merge
    list_2 = self.create_mailing_list({"name": "List test 2"})
  File "/odoo/external-src/social/mass_mailing_partner/tests/base.py", line 50, in create_mailing_list
    return m_mailing_list.create(vals)
  File "<decorator-gen-69>", line 2, in create
  File "/odoo/src/odoo/api.py", line 344, in _model_create_multi
    return create(self, [arg])
  File "/odoo/src/odoo/addons/base/models/ir_fields.py", line 534, in create
    recs = super().create(vals_list)
  File "<decorator-gen-13>", line 2, in create
  File "/odoo/src/odoo/api.py", line 345, in _model_create_multi
    return create(self, arg)
  File "/odoo/src/odoo/models.py", line 3864, in create
    records = self._create(data_list)
  File "/odoo/src/odoo/models.py", line 3970, in _create
    cr.execute(query, params)
  File "<decorator-gen-3>", line 2, in execute
  File "/odoo/src/odoo/sql_db.py", line 101, in check
    return f(self, *args, **kwargs)
  File "/odoo/src/odoo/sql_db.py", line 298, in execute
    res = self._obj.execute(query, params)
psycopg2.IntegrityError: duplicate key value violates unique constraint "mailing_list_unique_name"
DETAIL:  Key (name)=(List test 2) already exists.
```

---

btw there's more weird stuff that I don't get, like this:
https://github.com/OCA/social/blob/bebbfbd72a554ca5bfbd7d0f5f1f33f21ec80810/mass_mailing_partner/models/mailing_contact.py#L76-L86

Creating a company from the mailing.contact's `company_name` ? I think that's misconceived btw. `company_name` is supposed to be the contact's employeer/company name AFAICS (same concept than company_name in res.partner https://github.com/odoo/odoo/commit/bf88456943bed2e4ed4a996fb4a48126e325126d)


For these reasons I think it's best to test it separately, no?

pedrobaeza:
Yes, `company_name` thing is indeed wrong, and I don't know when it was introduced. Please create an issue with all this stuff and we'll assign and check for not having to isolate it, or isolate it and we can revert it back when fixed.

ivantodorovich:
Done: https://github.com/OCA/social/issues/776

ivantodorovich:
Okay I ended up removing the `copy=False` from mailing.contact's email, because it otherwise crashes with `mass_mailing_partner`.

Anyway, it's not that important.. User won't be able to duplicate mailing.contacts, but that's not so bad.. it's unlikely anyway.
At least now we can test all modules together 👍🏻

alexis-via (approved):

PR #774, opened 2021-09-20
[14.0][ADD] mass_mailing_company_newsletter

ivantodorovich:
This module lets you manage partner's subscriptions to your company's main mailing list directly from the partner form.
It makes it easier to manage partner mass communications, specially if you only use a single mailing list.

### DIFF
--- a/mass_mailing_company_newsletter/models/res_partner.py
+++ b/mass_mailing_company_newsletter/models/res_partner.py
@@ -1,0 +1,191 @@
+# Copyright 2021 Camptocamp (http://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+from odoo.osv import expression
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    main_mailing_list_id = fields.Many2one(
+        comodel_name="mailing.list",
+        string="Company Newsletter",
+        help="Technical field: The company's Newsletter mailing list.",
+        compute="_compute_main_mailing_list_id",
+        compute_sudo=True,
+    )
+    main_mailing_list_subscription_id = fields.Many2one(
+        comodel_name="mailing.contact.subscription",
+        string="Company Newsletter Subscription",
+        help="Technical field: The company's newsletter subscription for this partner.",
+        compute="_compute_main_mailing_list_subscription_id",
+        compute_sudo=True,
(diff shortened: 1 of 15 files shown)


damdam-s (approved):

mass_mailing_company_newsletter/models/res_partner.py
```diff
@@ -0,0 +1,191 @@
...
+        # But without it, Odoo would print an error log saying:
+        # Non-stored field main_mailing_list_subscription_id cannot be searched.
+        # Even though we're not explicitly searching on it, the following line is:
+        # https://github.com/odoo/odoo/blob/c9b6b030f/odoo/models.py#L5860
+        if operator != "in":  # pragma: no cover
+            return NotImplementedError()
+        # Rely on the fact that search takes a query as value since Odoo 14.0
+        # See https://github.com/odoo/odoo/commit/69869ab68
+        mailing_list = self.env.company.main_mailing_list_id
+        search = self.env["mailing.contact.subscription"]._search(
```
simahawk, line 84:
```suggestion
        ids = self.env["mailing.contact.subscription"]._search(
```

mass_mailing_company_newsletter/views/res_partner.xml
```diff
@@ -0,0 +1,86 @@
...
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+-->
+<odoo>
+
+    <record id="view_partner_form" model="ir.ui.view">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//page[@name='sales_purchases']//group[@name='sale']"
```
simahawk, line 14: wouldn't be better to have a specific tab for this?

simahawk (commented):
LG

mass_mailing_company_newsletter/models/res_partner.py
ivantodorovich: As of https://github.com/odoo/odoo/commit/69869ab68, `_search` returns a `Query` object, not ids.

mass_mailing_company_newsletter/views/res_partner.xml
ivantodorovich: I know the xml looks large, but in fact this is only adding 1 row :) No need for a new page

mass_mailing_company_newsletter/models/res_partner.py
simahawk: then `query`? :smile: 
As you prefer.. not important ;)

yankinmax (approved):

yankinmax:
@simahawk Can you pls take a look?

PR #775, opened 2021-09-21
[12.0][FIX] mail_debrand: Enable debranding in more cases

ypapouin:
Enable debranding even when `notif_layout` is False and also run debranding without translation


ypapouin:
Reminder for @ypapouin : This PR is partially superseded by #736 , the `_debrand_body` could be dropped.
But keep `generate_email` and `send_email` overrides as it is an improvement.

Issue #776, opened 2021-09-22
[14.0] mass_mailing_partner : issues with company_name and incompatibility with mass_mailing_unique


ivantodorovich:
As mentioned in this comment: https://github.com/OCA/social/pull/773#issuecomment-924961156

> With `mass_mailing_unique`, we have `email = fields.Char(copy=False)`, to avoid the IntegrityError if the user tries to copy it.
> That makes `mass_mailing_partner` NOT to copy the email on those 2 lines, which is later backfiring in other module's tests. I haven't dug too much on why `mass_mailing_partner` is copying data there, but it doesn't seem right. Seems it's not accounting for fields with `copy=False`.
> 
> Apart from that, there are other issues in `mass_mailing_partner` tests, like trying to create mailing lists with duplicate names:
> 
> * https://github.com/OCA/social/blob/bebbfbd72a554ca5bfbd7d0f5f1f33f21ec80810/mass_mailing_partner/tests/base.py#L31-L38
> * https://github.com/OCA/social/blob/bebbfbd72a554ca5bfbd7d0f5f1f33f21ec80810/mass_mailing_partner/tests/test_mail_mass_mailing_contact.py#L116
> 
> ```
> Traceback (most recent call last):
>   File "/odoo/external-src/social/mass_mailing_partner/tests/test_mail_mass_mailing_contact.py", line 116, in test_partners_merge
>     list_2 = self.create_mailing_list({"name": "List test 2"})
>   File "/odoo/external-src/social/mass_mailing_partner/tests/base.py", line 50, in create_mailing_list
>     return m_mailing_list.create(vals)
>   File "<decorator-gen-69>", line 2, in create
>   File "/odoo/src/odoo/api.py", line 344, in _model_create_multi
>     return create(self, [arg])
>   File "/odoo/src/odoo/addons/base/models/ir_fields.py", line 534, in create
>     recs = super().create(vals_list)
>   File "<decorator-gen-13>", line 2, in create
>   File "/odoo/src/odoo/api.py", line 345, in _model_create_multi
>     return create(self, arg)
>   File "/odoo/src/odoo/models.py", line 3864, in create
>     records = self._create(data_list)
>   File "/odoo/src/odoo/models.py", line 3970, in _create
>     cr.execute(query, params)
>   File "<decorator-gen-3>", line 2, in execute
>   File "/odoo/src/odoo/sql_db.py", line 101, in check
>     return f(self, *args, **kwargs)
[...]

PR #777, opened 2021-09-22
[12.0][FIX] mass_mailing_partner: wrong company implementation

chienandalu:
See issue https://github.com/OCA/social/issues/776

cc @Tecnativa

please review @ernestotejeda @ivantodorovich @pedrobaeza

### DIFF
--- a/mass_mailing_partner/models/mail_mass_mailing_contact.py
+++ b/mass_mailing_partner/models/mail_mass_mailing_contact.py
@@ -76,16 +76,6 @@ class MailMassMailingContact(models.Model):
             super(MailMassMailingContact, contact).write(new_vals)
         return True
 
-    def _get_company(self):
-        company_id = False
-        if self.company_name:
-            company_id = self.env['res.company'].search(
-                [('name', '=', self.company_name)]).id
-            if not company_id:
-                company_id = self.env['res.company'].create(
-                    {'name': self.company_name}).id
-        return company_id
-
     def _get_categories(self):
         ca_ids = self.tag_ids.ids + self.list_ids.mapped('partner_category.id')
         return [[6, 0, ca_ids]]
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Shouldn't you create instead a parent partner from that `company_name`?

ivantodorovich:
Or simply set the partner's company_name when it's created. According to https://github.com/odoo/odoo/commit/bf88456943bed2e4ed4a996fb4a48126e325126d it should behave as having a parent_id without actually having to create the parent. [A button is even displayed on the partner form to create it](https://github.com/odoo/odoo/blob/542cbe80ef8b71f4f48502fcc73d0c6556e679c2/odoo/addons/base/views/res_partner_views.xml#L151)

chienandalu:
Yes, I'll go for @ivantodorovich take.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -96,7 +86,7 @@ def _prepare_partner(self):
             'email': self.email,
             'country_id': self.country_id.id,
             'title': self.title_id.id,
-            'company_id': self._get_company(),
+            'company_id': self.company_name,
```
pedrobaeza:
```suggestion
            'company_name': self.company_name,
```

pedrobaeza:
About `company_id` field, is mailing contact linked to a company_id? If so, we should respect it.

mass_mailing_partner/models/mail_mass_mailing_contact.py
chienandalu: oops, yes :sweat_smile:

chienandalu:
Well, all the mass_mailing stuff is quite company agnostic

pedrobaeza:
Then I think we should force `company_id=False` in the contact for avoiding problems.

chienandalu:
Ok, done

pedrobaeza (approved):

ivantodorovich (approved):

PR #779, opened 2021-09-22
[14.0][FIX] mail_tracking: Change operator to 'ilike' in partner smartbutton

JorgeQuinteros:
This is the result with the actual attribute **` filter_domain="[('recipient', '=', self)]"`** in the recipient and sender field:
GIF 22-09-2021 23-37-44

So I change the operator to "ilike"(**` filter_domain="[('recipient', 'ilike', self)]"`**) because I think it is the right thing to do when filtering by char fields. 
GIF 22-09-2021 23-40-11

### DIFF
--- a/mail_tracking/views/mail_tracking_email_view.xml
+++ b/mail_tracking/views/mail_tracking_email_view.xml
@@ -102,12 +102,12 @@
             <field
                     name="sender"
                     string="Sender"
-                    filter_domain="[('sender', '=', self)]"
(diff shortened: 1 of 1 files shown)


ivantodorovich (approved):

hbrunn (approved):
technically, the local part of an email address is case sensitive, but as none of the big providers uses them like that, effectively email addresses are case insensitive

PR #784, opened 2021-10-04
[ADD] mail_activity_form

hbrunn:
This module allows you to provide some markup in the ``default_description``
field of ``mail.activity.type``, which makes some portions of the ``note``
field of resulting ``mail.activity`` records editable and others computed.

With this, you can implement form based workflows where you have to fill in
certain forms at certain times, and have to prove later what you filled in
when as mail activities, while still being able to access the data filled
in programmatically afterwards.

Usage
=====

This module adds some attributes that can be used in the
``default_description`` field of ``mail.activity.type``:

data-form-id
    A name used to refer to the content of this node afterwards
data-form-type
    One of 'float', 'int', 'str'
data-form-editable
    Only nodes marked with this attribute may be edited in activities
data-form-compute
    An expression to evaluate, ``object`` can be used to access values of the
    object the activity is attached to, and ``activity`` for the activity
    itself

Note that the above implies that you shouldn't use ``object`` and ``activity``
in the ``data-form-id`` attribute.


ofonseca-pyming (approved):
LGTM

rafaelbn (approved):
LGTM

PR #785, opened 2021-10-05
[14.0][MIG] mass_mailing_list_dynamic: Migration to 14.0

victoralmau:
Migration to 14.0

Please @chienandalu and @Yajo , can you review it?

@Tecnativa TT32133


CarlosRoca13 (approved):

mass_mailing_list_dynamic/__manifest__.py
```diff
@@ -15,6 +15,7 @@
     "depends": ["mass_mailing_partner"],
     "data": [
         # This should go first
+        "security/ir.model.access.csv",
```
pedrobaeza, line 17: You are deforming the comment meaning (which should be by the way `This should go before "mailing_list_view.xml"`). Put it on the proper place.

mass_mailing_list_dynamic/tests/test_dynamic_lists.py
```diff
@@ -2,19 +2,19 @@
...
 from mock import patch
 
 from odoo.exceptions import ValidationError
 from odoo.tests import common
 
 
-@common.at_install(False)
-@common.post_install(True)
 class DynamicListCase(common.SavepointCase):
+    at_install = False
```
pedrobaeza: This doesn't serve since v13, so it can be removed if this is working, as no need of deferring it.

pedrobaeza (changes requested):

mass_mailing_list_dynamic/__manifest__.py
pedrobaeza: You have put it in other bad place and not changing the comment

pedrobaeza (approved):

PR #787, opened 2021-10-11
[IMP] mail_tracking{,_mailgun}: process sent events and fix wrong failure event mappings... and refactor

yajo:
## [FIX] mail_tracking: process sent events 

Up until now, the `sent` event type was never processed, probably because it "made no sense", as an unsent email would never trigger an event.

However, it makes sense to process it because you may have a local relay that transmits mails over to the mail provider. In those circumstances, you should have 2 "sent" events (one from the relay and another one from the provider).

Also marked some useless parts of code for removal.

@Tecnativa TT32365

## [IMP] mail_tracking_mailgun: refactor to support modern webhooks

Before this patch, the module was designed after the [deprecated Mailgun webhooks][3]. However Mailgun had the [events API][2] which was quite different. Modern Mailgun has deprecated those webhooks and instead uses new ones that include the same payload as the events API, so you can reuse code.

However, this was incorrectly reusing the code inversely: trying to process the events API through the same code prepared for the deprecated webhooks.

Besides, both `failed` and `rejected` mailgun events were mapped to `error` state, but that was also wrong because [`mail_tracking` doesn't have an `error` state][1].

So the logic of the whole module is changed, adapting it to process the events API payload, both through controllers (prepared for the new webhooks) and manual updates that directly call the events API.

Also, `rejected` is now translated into `reject`, and `failed` is translated into `hard_bounce` or `soft_bounce` depending on the severity, as specified by [mailgun docs][2]. Also, `bounced` and `dropped` mailgun states are removed because they don't exist, and instead `failed` and `rejected` properly get their metadata.

Of course, to know the severity, now the method to obtain that info must change, it' can't be a simple dict anymore.

Added more parameters because for example modern Mailgun uses different keys for signing payload than for accessing the API. As there are so many parameters, configuration is now possible through `res.config.settings`. Go there to autoregister webhooks too.

Since the new webhooks are completely incompatible with the old supposedly-abstract webhooks controllers (that were never really that abstract), support for old webhooks is removed, and it will be removed in the future from `mail_tracking` directly. There is a migration script that attempts to unregister old webhooks and register new ones automatically.

[1]: https://github.com/OCA/social/blob/f73de421e28a43d018176f61725a3a59665f715d/mail_tracking/models/mail_tracking_event.py#L31-L42
[2]: https://documentation.mailgun.com/en/latest/api-events.html#event-types
[3]: https://documentation.mailgun.com/en/latest/api-webhooks-deprecated.html

### DIFF
--- a/mail_tracking_mailgun/controllers/main.py
+++ b/mail_tracking_mailgun/controllers/main.py
@@ -1,0 +1,76 @@
+# Copyright 2021 Tecnativa - Jairo Llopis
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import hashlib
+import hmac
+import logging
+
+from datetime import datetime, timedelta
+
+from werkzeug.exceptions import NotAcceptable
+
+from odoo import _
+from odoo.exceptions import ValidationError
+from odoo.http import request, route
+
+from ...mail_tracking.controllers import main
+from ...web.controllers.main import ensure_db
+
+_logger = logging.getLogger(__name__)
+
+
+class MailTrackingController(main.MailTrackingController):
+    def _mail_tracking_mailgun_webhook_verify(self, timestamp, token, signature):
+        """Avoid mailgun webhook attacks.
+
+        See https://documentation.mailgun.com/en/latest/user_manual.html#securing-webhooks
+        """  # noqa: E501
+        # Request cannot be old
+        processing_time = datetime.utcnow() - datetime.utcfromtimestamp(int(timestamp))
+        if not timedelta() < processing_time < timedelta(minutes=10):
+            raise ValidationError(_("Request is too old"))
+        # Avoid replay attacks
+        try:
+            processed_tokens = (
+                request.env.registry._mail_tracking_mailgun_processed_tokens
+            )
+        except AttributeError:
+            processed_tokens = (
+                request.env.registry._mail_tracking_mailgun_processed_tokens
+            ) = set()
+        if token in processed_tokens:
+            raise ValidationError(_("Request was already processed"))
+        processed_tokens.add(token)
+        params = request.env["mail.tracking.email"]._mailgun_values()
+        # Assert signature
+        if not params.webhook_signing_key:
+            _logger.warning(
+                "Skipping webhook payload verification. "
+                "Set `mailgun.webhook_signing_key` config parameter to enable"
+            )
+            return
+        hmac_digest = hmac.new(
+            key=params.webhook_signing_key.encode(),
+            msg=("{}{}".format(timestamp, token)).encode(),
+            digestmod=hashlib.sha256,
+        ).hexdigest()
+        if not hmac.compare_digest(str(signature), str(hmac_digest)):
+            raise ValidationError(_("Wrong signature"))
+
+    @route(["/mail/tracking/mailgun/all"], auth="none", type="json", csrf=False)
+    def mail_tracking_mailgun_webhook(self):
+        """Process webhooks from Mailgun."""
+        ensure_db()
+        # Verify and return 406 in case of failure, to avoid retries
+        # See https://documentation.mailgun.com/en/latest/user_manual.html#routes
+        try:
+            self._mail_tracking_mailgun_webhook_verify(
+                **request.jsonrequest["signature"]
+            )
+        except ValidationError as error:
+            raise NotAcceptable from error
+        # Process event
+        request.env["mail.tracking.email"].sudo()._mailgun_event_process(
+            request.jsonrequest["event-data"],
+            self._request_metadata(),
+        )
--- a/mail_tracking_mailgun/models/mail_tracking_email.py
+++ b/mail_tracking_mailgun/models/mail_tracking_email.py
@@ -1,9 +1,11 @@
 # Copyright 2016 Tecnativa - Antonio Espinosa
 # Copyright 2017 Tecnativa - David Vidal
+# Copyright 2021 Tecnativa - Jairo Llopis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-import hashlib
-import hmac
+from collections import namedtuple
+from urllib.parse import urljoin
+
 import requests
 from datetime import datetime
 from odoo import _, api, fields, models
@@ -12,6 +14,22 @@ from odoo.tools import email_split
 
 import logging
 _logger = logging.getLogger(__name__)
+
+MailgunParameters = namedtuple(
+    "MailgunParameters",
+    (
+        "api_key",
+        "api_url",
+        "domain",
+        "validation_key",
+        "webhooks_domain",
+        "webhook_signing_key",
+    ),
+)
+
+
+class EventNotFoundWarning(Warning):
+    pass
 
 
 class MailTrackingEmail(models.Model):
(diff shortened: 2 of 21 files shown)


pedrobaeza:
Who is calling to that method?

mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -32,10 +32,16 @@ def _mailgun_mandatory_fields(self):
         return ('event', 'timestamp', 'token', 'signature',
                 'tracking_email_id', 'odoo_db')
 
-    @property
-    def _mailgun_event_type_mapping(self):
-        return {
-            # Mailgun event type: tracking event type
+    @api.model
+    def _mailgune_event2state(self, event, default="UNKNOWN"):
```
pedrobaeza: `mailgune`?

yajo: Oops, changed, thanks.

yajo:
> Who is calling to that method?

This: https://github.com/OCA/social/blob/f73de421e28a43d018176f61725a3a59665f715d/mail_tracking/models/mail_tracking_email.py#L322

yajo:
Finally, ready to review.

pedrobaeza:
Travis is red.

chienandalu (commented):
You should adapt tests to the new event mapping

yajo:
😓 hard one... ready to review.

yajo:
Travis is green finally.

mail_tracking_mailgun/controllers/main.py
```diff
@@ -0,0 +1,76 @@
...
+        try:
+            processed_tokens = (
+                request.env.registry._mail_tracking_mailgun_processed_tokens
+            )
+        except AttributeError:
+            processed_tokens = (
+                request.env.registry._mail_tracking_mailgun_processed_tokens
+            ) = set()
+        if token in processed_tokens:
+            raise ValidationError(_("Request was already processed"))
```
pedrobaeza, line 42: Should this be an error?

mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -27,43 +46,30 @@ def _country_search(self, country_code):
...
-    @property
-    def _mailgun_event_type_mapping(self):
-        return {
-            # Mailgun event type: tracking event type
+        Args:
+            event: Mailgun event response from API.
+            default: Value to return when not found.
+        """
+        # Mailgun event type: tracking event type
+        equivalents = {
```
pedrobaeza, line 57: Shouldn't this be a "constant" on top of the file? At least is only populated once.

mail_tracking_mailgun/wizards/res_config_settings.py
```diff
@@ -0,0 +1,120 @@
...
+    def get_values(self):
+        """Is Mailgun enabled?"""
+        result = super().get_values()
+        result["mail_tracking_mailgun_enabled"] = bool(
+            self.env["ir.config_parameter"].get_param("mailgun.apikey")
+        )
+        return result
+
+    def mail_tracking_mailgun_unregister_webhooks(self):
+        """Remove existing Mailgun webhooks."""
```
pedrobaeza, line 78: Is this removing also deprecated webhooks?

mail_tracking_mailgun/readme/CONFIGURE.rst
```diff
@@ -1,27 +1,11 @@
...
-- `mailgun.validation_key`: If you want to be able to check mail address
-  validity you must config this parameter with your account Public Validation
-  Key.
+To configure this module, you need to:
+
+#. Go to Mailgun, create an account and validate your sending domain.
+#. Go back to Odoo.
+#. Go to *Settings > General Settings > Discuss > Enable mail tracking with Mailgun*.
+#. Fill all the values. The only one required is the API key.
+#. Optionally click *Unregister Mailgun webhooks* and accept.
```
pedrobaeza, line 7: You shouldn't be required to press these buttons (they should be only for manual actions), but on change of the parameter, you should first before changing data to unregister hooks, and then change them and register hooks.

yajo: Not because:

1. I don't want to override the setting/getting of `ir.config_parameter`, as it is very widely used. And that'd be the only way to get this properly automated.
2. This could unregister webhooks that are not from Odoo.
3. Infrastructure configuration should be predictable.
4. Doing this manually implicitly helps checking that your api key is valid.

mail_tracking_mailgun/wizards/res_config_settings.py
yajo: Yes.

mail_tracking_mailgun/controllers/main.py
yajo: Yes, to avoid replay attacks. See Mailgun docs.

mail_tracking_mailgun/models/mail_tracking_email.py
yajo: The failed status is conditional.

mail_tracking_mailgun/readme/CONFIGURE.rst
pedrobaeza: > 1.  I don't want to override the setting/getting of ir.config_parameter, as it is very widely used. And that'd be the only way to get this properly automated.

That's because you are thinking of the field name automatism on upstream, but using another field name, you can do the set and get you want.

> 2. This could unregister webhooks that are not from Odoo.

So you are saying that clicking manually on this unregister other webhooks not created by Odoo??! That's the problem, and you should know (by the name or something), which ones were created by Odoo and only unregister that ones.

> 3. Infrastructure configuration should be predictable.

Yes, that's why you should automate it. If not, Ansible wouldn't exist.

> 4. Doing this manually implicitly helps checking that your api key is valid.

And if not valid, you receive an error when saving, so that's the desired effect of not storing something invalid.

So I still think you should do it automatically.

yajo: Doing that automatically can suppose unpredictable behavior.

For example, you might not be adding a domain for mailgun and rely instead on web.base.url. web.base.url is not exposed in this dialog, so to change it you have to go to ICP section. If you do so, res.config.settings doesn't know about that change. Later, you go to a res.config.settings because you want to... say... configure B2B mode in your website. When you press save, we'll detect a change and apply the webhook updates. That may be more or less correct depending on who you ask, but in any case it's unpredictable.

In a similar fashion, you could change the mail.catchall.domain in this same dialog without expecting anything to change on mailgun configurations section. Again, a change there could trigger webhook calls, and that'd be unpredictable.

You could also want to register temporarily your sandbox domain webhooks without having to remove your main domain webhooks.

Or you could move from US API to EU API, in which case you need to manually unregister US webhooks first, then change API endpoint, and then register new webhooks. Or alternatively, you could decide to leave both webhooks online (although that would probably require another change in the module).

OTOH if you want to automate it, all you need to do is to call this method from your ansible playbook or equivalent. And if you prefer to configure it manually, the configuration is now self-explained and straightforward, as opposed to how it was before by dealing with ICPs and so forth. And this is still a huge enhancement as compared to what we had.

So, to avoid the cognitive load of having to think in all these scenarios while trying to understand an already quite complex module, it's better to not automate this.

pedrobaeza: I don't see such complication: it's just that you change one of the 5/6 configuration, you unregister before changing them, and register after saving them.

chienandalu (approved):
:+1: (Not strong opinion about webhooks manual/automation though)

mail_tracking_mailgun/readme/CONFIGURE.rst
pedrobaeza: Can you please draft this to see the complications?

yajo: There's no need. Complications are not on code but on concept. Just read above please.

pedrobaeza: But I have read them, and I see no conceptual complications. I did a question though: aren't you able to distinguish the hooks added by this module? If not, you are going to have the same problem doing it automatically or manually.

yajo: I could try to detect that based on the hostname or url endpoint, but things like renaming the base url could trick that detection.

Adding this automatism will mean hours of work. Just to save 2 clicks. Migration code does a best-effort attempt to migrate webhooks automatically. Configuring mailgun will always mean manual interaction, and this doesn't need to be done again later. So given what you ask is neither important nor urgent, if you want I drop a note about it to roadmap and leave it for a future enhancement. I see no blocking points here.

pedrobaeza: The automation doesn't take too much time. It's more time arguing than implementing. You just need to add in res.config.settings the load and save actions. What worries me is that right now, if you manually click on "Unregister hook", it will remove all existing hooks, no matter if yours or from other automatism. And also if you click on "Register hook" several times, what happens?

Do you see what I try to say about the proper hook detection?

yajo: Yes, but there's a warning telling you that could happen. That's not a problem that will affect our deployments, so a different contributor can patch it if it really happens later.

pedrobaeza: Add all of this to ROADMAP, and let's not block it more, but I'm not happy letting this unfinished.

yajo: Thanks, done.

pedrobaeza:
Please forward-port it to v13 and v14:

/ocabot merge major

PR #791, opened 2021-10-26
[14.0] [FIX] mail_notification_custom_subject: drop `subtype_id` kwarg check

naglis:
As `subtype_id` is not in the signature of `message_post`, `subtype_id = kwargs.get('subtype_id', False)` results in `subtype_id`always being `False` (as it will never be in `kwargs`) and the actual value passed via `signature_id` is not used.

This was introduced in migration from 13.0: https://github.com/OCA/social/pull/757 as the signature of `message_post` was updated to reflect the signature in Odoo core.

### DIFF
--- a/mail_notification_custom_subject/models/mail_thread.py
+++ b/mail_notification_custom_subject/models/mail_thread.py
@@ -28,7 +28,6 @@ class MailThread(models.AbstractModel):
         record_name=False,
         **kwargs
     ):
-        subtype_id = kwargs.get("subtype_id", False)
         if not subtype_id and subtype_xmlid:
             subtype_id = self.env["ir.model.data"].xmlid_to_res_id(
                 subtype_xmlid,


pedrobaeza:
I see the argument `subtype_id` in both 13.0:

https://github.com/odoo/odoo/blob/b58b6ae6368bb2d76b7d9f7a14063dd9dc27fc7a/addons/mail/models/mail_thread.py#L1815

and 14.0:

https://github.com/odoo/odoo/blob/475e6b3cb027b3127b8e46fd8c9f90abc35162d7/addons/mail/models/mail_thread.py#L1785

so not sure to understand the problem.

naglis:
@pedrobaeza, yes, however in 13.0 version of `mail_notification_custom_subject` the signature of `message_post` was still not updated to reflect Odoo:
https://github.com/OCA/social/blob/002fdc00675b1b86c41e891a5d1b482dc2a10733/mail_notification_custom_subject/models/mail_thread.py#L12-L25
I did that in the migration to 14.0 (https://github.com/OCA/social/pull/757), but I've missed the case where `subtype_id` can't ever be in `kwargs` as `subtype_id` is now in the signature.
Please let me know if it is still unclear.

pedrobaeza:
But then on 13.0, `subtype_id` was not passed to `super`, so 2 options:

1. This is needed for putting the special subject, and we should replicate it.
2. We should fix it the same way in 13.0 as here.

@CarlosRoca13 do you remember any special reason for popping the value from `kwargs`?

etobella:
@naglis Sorry, I found your issue late. It was already done on #836 :cry: 

This change was necessary, as the value was trying to be found on kwargs but was already present on the method parameters (so will never be on kwargs)

https://github.com/OCA/social/pull/791/files#diff-b0cf6e36c5fa6016fdfa0af9e807628004545dad37e6f94f27b3ee42cb67a807L22

https://github.com/OCA/social/pull/791/files#diff-b0cf6e36c5fa6016fdfa0af9e807628004545dad37e6f94f27b3ee42cb67a807L31

I will close this PR as the change is already being merged right now.

pedrobaeza:
@etobella but see my arguments. Maybe the fix is not correct.

etobella:
The problem is that subtype will never be on kwargs, as you already pass it on a parameter. So, IMO, there are two possible options: 

1- remove the subtype_id from the parameters 
2- take it from the parameters.

When migration was done, original signature was copied and noone noticed that it affected the code (this kind of things happen :disappointed:). You can check it on the migration commit:

https://github.com/OCA/social/pull/757/commits/1c42c0f6ab0fcb4017adc715ec6038a923527f84

pedrobaeza:
So current solution is OK?

etobella:
It should be. At least it seems right for me. The problem was detected on a migration, because subtype_id was lost. With the change all seems right from our side

pedrobaeza:
OK, let's see if there's no problem.

PR #793, opened 2021-11-02
[14.0][MIG] mail_full_expand: Migration to 14.0

ernestotejeda:
Cc @Tecnativa TT32131

Main changes:
- Adapt code to OWL


ernestotejeda:
@Yajo could you review, please?

mail_full_expand/static/src/components/message/mail_full_expand.js
```diff
@@ -0,0 +1,29 @@
...
+        "use strict";
+
+        const components = {
+            Message: require("mail/static/src/components/message/message.js"),
+        };
+        const {patch} = require("web.utils");
+
+        patch(
+            components.Message,
+            "mail_full_expand/static/src/components/message/mail_full_expand.js",
```
yajo: I'm not sure if this would be the best patch name. [It looks like `module_name.key` is recommended](https://github.com/odoo/odoo/blob/bf350c3015f86357e609a6302b48e7c22f2c0b47/addons/web/static/src/js/core/patch_mixin.js#L19) instead.

```suggestion
            "mail_full_expand.main",
```

mail_full_expand/static/src/components/message/mail_full_expand.js
```diff
@@ -0,0 +1,29 @@
...
+            {
+                async _onClickMessageFullExpand() {
+                    const action = await this.rpc({
+                        route: "/web/action/load",
+                        params: {action_id: "mail_full_expand.mail_message_action"},
+                    });
+                    action.res_id = this.message.id;
+                    this.trigger("do-action", {action: action});
+                },
```
yajo: I guess there must be some error around here.

It works while on a record chatter, but it fails when in the discuss app:

https://user-images.githubusercontent.com/973709/140027594-0b7d033f-9e7e-4501-8c7d-810eeac63edf.mp4

yajo (changes requested):
Code looks good, but there's a functional bug.

ernestotejeda:
Thanks @Yajo, changes done (bug fixed).

mail_full_expand/static/src/components/message/mail_full_expand.js
```diff
@@ -0,0 +1,26 @@
...
+        patch(components.Message, "mail_full_expand.main", {
+            async _onClickMessageFullExpand(ev) {
+                ev.stopPropagation();
+                const action = await this.rpc({
+                    route: "/web/action/load",
+                    params: {action_id: "mail_full_expand.mail_message_action"},
+                });
+                action.res_id = this.message.id;
+                this.env.bus.trigger("do-action", {action: action});
+            },
```
Tardo, line 23: I think that here you can use directly...
```suggestion
            _onClickMessageFullExpand(ev) {
                ev.stopPropagation();
                this.env.bus.trigger("do-action", {
                    action: "mail_full_expand.mail_message_action",
                    options: {
                        additional_context: {
                            active_id: this.message.id,
                        },
                    },
                });
            },
```

Tardo (changes requested):

mail_full_expand/static/src/components/message/mail_full_expand.js
ernestotejeda: I have test it locally and it doesn't work, I don't know why :(

Tardo: Ok.. how cumbersome Odoo has made it for these cases :/

Tardo (approved):
LGTM :+1: Code review

yajo (approved):

PR #796, opened 2021-11-10
[14.0][MIG] mail_attach_existing_attachment_account: Migration to 14.0

ernestotejeda:
Cc @Tecnativa TT32120


mail_attach_existing_attachment_account/wizard/account_invoice_send_view.xml
```diff
@@ -12,8 +12,9 @@
             <xpath expr="//field[@name='attachment_ids']" position="after">
                 <field name="model" invisible="1" />
                 <field name="res_id" invisible="1" />
-                <field name="can_attach_attachment" invisible="1" />
-                <div attrs="{'invisible': [('can_attach_attachment', '=', False)]}">
+                <div
+                    attrs="{'invisible': [('model', '!=', 'False'), ('res_id', '!=', False), ('composition_mode', '=', 'mass_mail')]}"
+                >
```
ernestotejeda: I have changed this because can_attach_attachment is not taking the value we expect. When sending an invoice it doesn't take True as it happend in v13. Anyway we get the same result

ernestotejeda:
There is a pre-commit error but it is in another module

ernestotejeda:
@victoralmau could you review this?

chienandalu (approved):

mail_attach_existing_attachment_account/wizard/account_invoice_send_view.xml
victoralmau: In my opinion this change is not correct.
In v13 the addon was not migrated, I have created a new PR https://github.com/OCA/social/pull/797 for v13 that solves this problem.

If it seems correct to you we can mix that first and then re-do this PR taking into account the changes of v13.
Does it seem correct to you?

victoralmau: I have done rebase, this is already solved.

victoralmau (approved):

PR #807, opened 2021-11-24
[FIX] mail_tracking: use recipient_address to search emails

### LINKED ISSUE #803: [14.0] mail_tracking: button 'Tracking emails' in partner is showing a different count that shown when you click on it
The mail_tracking in v13 is searching the emails as follows:

https://github.com/OCA/social/blob/78daea28e5763cb132c77fcc986408680dd4e84f/mail_tracking/models/mail_tracking_email.py#L203-L214

It's using a regex to catch cases when the recipient is `<(emailexample@.example.com)>` and if regex is not matched, it's using the email like comes in the record, but in v14.0 is using the function `email_split` and it's only considering when exactly matching and excluding the emails as `<(emailexample@.example.com)>`

https://github.com/OCA/social/blob/147eda96a56c199602c619b5bac3a9911605e918/mail_tracking/models/mail_tracking_email.py#L204-L210

## Module
mail_tracking

## Describe the bug

The mail_tracking in v13 is searching the emails as follows:

https://github.com/OCA/social/blob/78daea28e5763cb132c77fcc986408680dd4e84f/mail_tracking/models/mail_tracking_email.py#L203-L214

It's using a regex to catch cases when the recipient is `<(emailexample@.example.com)>` and if regex is not matched, it's using the email like comes in the record, but in v14.0 is using the function `email_split` and it's only considering the when exactly matching and excluding the emails as `<(emailexample@.example.com)>`

https://github.com/OCA/social/blob/147eda96a56c199602c619b5bac3a9911605e918/mail_tracking/models/mail_tracking_email.py#L204-L210

## To Reproduce
14.0

1. Go to any partner with mails sent.
2. See the emails shown in button `Tracking emails` 
3. Click on button `Tracking emails` 
4. See that is likely it doesn't show the emails correctly

**Expected behavior**
The expected behavior is user sees the same count of emails in `Tracking emails` that when you click on it.

fernandahf:
Description of the issue/feature this PR addresses:

This commit creates a migration script to fix all emails computed incorrectly with old version of `email_split` 

Also, we are using the field `recipient_address` in the filter of the search view
of `mail.tracking.email` in order to use the same field used in computed
method:

https://github.com/OCA/social/blob/147eda96a56c199602c619b5bac3a9911605e918/mail_tracking/models/res_partner.py#L27

Current behavior before PR:
Inconsistencies between the counter field in partners to show the number of emails tracked and the shown when clicking on itself.

Desired behavior after PR is merged:
Show correctly the same number of emails tracked in the field in partners than the shown when clicking on itself.

Fix https://github.com/OCA/social/issues/803

-- I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

### DIFF
--- a/mail_tracking/views/mail_tracking_email_view.xml
+++ b/mail_tracking/views/mail_tracking_email_view.xml
@@ -105,9 +105,9 @@
                     filter_domain="[('sender', 'ilike', self)]"
                 />
             <field
-                    name="recipient"
-                    string="Recipient"
-                    filter_domain="[('recipient', 'ilike', self)]"
+                    name="recipient_address"
+                    string="Recipient Address"
+                    filter_domain="[('recipient_address', '=', self)]"
                 />
             <field name="name" string="Subject" />
             <field name="time" string="Time" />
--- a/mail_tracking/views/res_partner_view.xml
+++ b/mail_tracking/views/res_partner_view.xml
@@ -12,8 +12,8 @@
         <div name="button_box" position="inside">
             <button
                     name="%(mail_tracking.action_view_mail_tracking_email)d"
-                    context="{'search_default_recipient': email,
-                              'default_recipient': email}"
+                    context="{'search_default_recipient_address': email,
+                              'default_recipient_address': email}"
                     type="action"
                     class="oe_stat_button"
                     icon="fa-envelope-o"


fernandahf:
@moylop260 

Could you review this, please?

moylop260:
Could you check the unittest failing in the CI, please?

Also, Could you tag to maintainer/authors of the module, please?

fernandahf:
@pedrobaeza 

Could review this, please?

mail_tracking/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 {
     "name": "Email tracking",
     "summary": "Email tracking system for all mails sent",
-    "version": "14.0.2.0.0",
+    "version": "14.0.2.0.1",
```
pedrobaeza: Don't change manually the version. The bot will do it.

mail_tracking/migrations/14.0.2.0.1/README.rst
```diff
@@ -0,0 +1 @@
+lo que sea
```
pedrobaeza: This file should go out

mail_tracking/migrations/14.0.2.0.1/post-migration.py
```diff
@@ -0,0 +1,14 @@
+from odoo import SUPERUSER_ID, api
+
+
+def migrate(cr, version):
+    fix_recipient_address_mail_tracking(cr)
+
+
+def fix_recipient_address_mail_tracking(cr):
+    """TODO: add docstring"""
```
pedrobaeza, line 9: Why do you need this?

pedrobaeza (changes requested):
There's a method for doing the email split stuff, and if it's not enough, you should fix such method in Odoo, not to switch again to manual regular expressions that can fail in thousand ways. I don't approve this change.

mail_tracking/migrations/14.0.2.0.1/README.rst
fernandahf: Done

mail_tracking/__manifest__.py
fernandahf: Done

fernandahf:
@pedrobaeza 

In the last version of Odoo v14.0 error with the `email_split` function is not reproduced, I still think the migration to fix `recipient_address` is needed, what do you think?

pedrobaeza:
I don't see the reason to merge this PR, as you are limiting the search to only the address, and the migration script seems a problem on your DB, not a general one.

fernandahf:
@pedrobaeza 

I think it's an error to use a field to compute a counter and use another field to search them?

pedrobaeza:
OK, I get your point. I agree, but you should remove the migration script.

fernandahf:
Done

pedrobaeza (approved):

PR #818, opened 2021-12-16
[14.0][ADD] mail_unique_layout: add to 14.0

kevinkhao:


kevinkhao:
due to the fact that this is a heavily opinionated feature and not that generic, we will move it to ak-odoo-incubator https://github.com/akretion/ak-odoo-incubator/pull/229

PR #820, opened 2021-12-16
[IMP] Travis: Fix execution

etobella:
This PR tries to fix runboat execution


etobella:
Error is related to base_fuzzy_search, so it cannot be fixed here :cry:

PR #832, opened 2022-01-17
[13.0][FIX] mail_debrand: make it work when language of recipients is not the one of the current user

alexeirivera87:
make it work when language of recipients differs from the one used by the user

### DIFF
--- a/mail_debrand/models/mail_template.py
+++ b/mail_debrand/models/mail_template.py
@@ -16,8 +16,13 @@ class MailTemplate(models.Model):
     def _debrand_translated_words(self):
         def _get_translated(word):
             return self.env["ir.translation"]._get_source(
-                "ir.ui.view,arch_db", "model_terms", self.env.lang, word
+                "ir.ui.view,arch_db", "model_terms", lang, word
             )
+
+        lang = self.env.lang
+        template_lang = self._get_template_lang()
+        if template_lang and template_lang != lang:
+            lang = template_lang
 
         odoo_word = _get_translated("Odoo") or _("Odoo")
         powered_by = _get_translated("Powered by") or _("Powered by")
@@ -49,3 +54,15 @@ class MailTemplate(models.Model):
     def render_post_process(self, html):
         html = super().render_post_process(html)
         return self._debrand_body(html)
+
+    def _get_template_lang(self):
+        template_lang = False
+        if self._context.get("default_template_id"):
+            template = self.env["mail.template"].browse(
+                self._context.get("default_template_id")
+            )
+            if template.lang and self._context.get("active_id"):
+                template_lang = template._render_template(
+                    template.lang, template.model, self._context.get("active_id")
+                )
+        return template_lang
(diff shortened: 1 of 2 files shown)


rafaelbn:
Hello @ValentinVinagre can you please check this fix please?

rafaelbn:
Hello @alexeirivera87 , could you please rebase? Thank you!

alexeirivera87:
Hello @rafaelbn 

Let me know if it is OK now ?
regards

mail_debrand/models/mail_template.py
```diff
@@ -49,3 +54,17 @@ def _debrand_body(self, html):
...
+        if self._context.get("default_template_id"):
+            template = self.env["mail.template"].browse(
+                self._context.get("default_template_id")
+            )
+            if template.lang and self._context.get("active_id"):
+                langs = template._render_template(
+                    template.lang, template.model, [self._context.get("active_id")]
+                )
+                for _res_ids, lang in langs.items():
+                    template_lang = lang
```
ValentinVinagre: Why are you doing an override every time on this variable? Do you know what "language" values this variable can take?

alexeirivera87: The field lang, model mail.template, should have something like this: ${object.partner_id.lang}

`template_lang = template._render_template(
                    template.lang, template.model, self._context.get("active_id")
                )`

I applied an adjustment, because it is true that only one valid value will be obtained from that mako expression

rafaelbn:
Don't forget to review [[12.0][FIX] mail_debrand: support all languages](https://github.com/OCA/social/pull/736#)
#736 😄

rafaelbn:
Hello @astirpe ! As you did this PR https://github.com/OCA/social/pull/613 could you please review this PR? With your review and @ValentinVinagre one I think we can verify the BUG this PR fixes. Thank!

mail_debrand/models/mail_template.py
```diff
@@ -49,3 +54,15 @@ def _debrand_body(self, html):
...
         return self._debrand_body(html)
+
+    def _get_template_lang(self):
+        template_lang = False
+        if self._context.get("default_template_id"):
+            template = self.env["mail.template"].browse(
+                self._context.get("default_template_id")
+            )
+            if template.lang and self._context.get("active_id"):
+                template_lang = template._render_template(
```
astirpe, line 65: The `_render_template()` does not return a lang, are you sure this is correct? Can you write a test to cover this part?

astirpe (changes requested):

mail_debrand/models/mail_template.py
alexeirivera87: There are several example of this in the standard Odoo code:

like this one in mail.thread:

`template = self.env['mail.template'].browse(self.env.context['default_template_id'])
            if template and template.lang:
                lang = template._render_template(template.lang, self.env.context['default_model'], self.env.context['default_res_id'])`

or this one in sale.order:

`template = self.env['mail.template'].browse(template_id)
        if template.lang:
            lang = template._render_template(template.lang, 'sale.order', self.ids[0])`

astirpe: Could you write a test to cover this part?

rafaelbn:
Hello, we  have couple of BUGs in this modules. It could be nice to have all the fixed merged. 😄

PR #841, opened 2022-01-26
[15.0][MIG] mail_outbound_static: Migration to 15.0

yankinmax:
Including commits from PR #598


yankinmax:
@rousseldenis @zaoral

yankinmax:
@simahawk can you pls take a look?

simahawk (approved):

simahawk:
@yankinmax pls check the build

mail_outbound_static/tests/test_ir_mail_server.py
```diff
@@ -0,0 +1,369 @@
...
+
+import logging
+import os
+from email import message_from_string
+
+import odoo.tools as tools
+from odoo.exceptions import ValidationError
+from odoo.tests.common import TransactionCase
+
+from odoo.addons.base.tests.common import MockSmtplibCase
```
yankinmax, line 12: @simahawk @rousseldenis 
With this improvement https://github.com/odoo/odoo/commit/1b9dd118cb0f305ed38b6326fa814336cbf16252
I propose to improve tests, as they are failing using https://docs.python.org/3/library/unittest.mock.html#module-unittest.mock

mail_outbound_static/tests/test_ir_mail_server.py
```diff
@@ -0,0 +1,369 @@
...
+                ssl_private_key=smtp_ssl_private_key,
+                smtp_debug=smtp_debug,
+                mail_server_id=mail_server_id,
+            )
+
+        send_from, send_to, message_string = self.IrMailServer._prepare_email_message(
+            message, smtp
+        )
+        self.IrMailServer.send_email(message)
+        return message_string
```
yankinmax, line 111: this is needed, just because, I'm not an expert with email servers and couldn't find any other idea how to get `smtp_seesion` parameter to pass directly to `_prepare_email_message` method of `ir.mail_server`.
Would really appreciate any guidance. Thanks

leemannd (approved):
Seems legit, but It is maybe not necessary anymore if we have a good setup of the system parameters and this commit -> https://github.com/odoo/odoo/commit/044532aed6f796cd93e37556b4a8ce47d2c23a7c

simahawk:
hmm it's slightly different here as the FROM comes from the server record and not from config param.
Also, that commit seems not to be in 15.0 yet 

https://github.com/odoo/odoo/blob/044532aed6f796cd93e37556b4a8ce47d2c23a7c/odoo/addons/base/models/ir_mail_server.py#L417

https://github.com/odoo/odoo/blob/15.0/odoo/addons/base/models/ir_mail_server.py

I'd add a TODO to evaluate overriding `_get_email_from` in future versions.

@gurneyalex what do you think?

PR #844, opened 2022-02-03
[14.0][FIX] mass_mailing_subscription_email: /unsubscribe_from_list link

ivantodorovich:
This fixes the case where `/unsubcribe_from_list` is added to the subscription email notification.

As the subscribe notification is sent through a `mail.template`, and not through a `mailing.mailing` record, odoo won't automagically handle the href replacement.

Moreover, the core controller only works with an actual `mailing.mailing` record, that may not even exist at the time the notification is sent. So, a new controller is added to handle this unsubscribe.

### DIFF
--- a/mass_mailing_subscription_email/controllers/main.py
+++ b/mass_mailing_subscription_email/controllers/main.py
@@ -1,0 +1,27 @@
+# Copyright 2022 Camptocamp SA (https://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import http
+from odoo.exceptions import AccessDenied
+from odoo.http import request
+from odoo.tools import consteq
+
+from odoo.addons.mass_mailing.controllers.main import MassMailController
+
+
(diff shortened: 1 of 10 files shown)


mass_mailing_subscription_email/controllers/main.py
```diff
@@ -0,0 +1,27 @@
...
+    @http.route(
+        "/mail/mailing/contact/<int:res_id>/unsubscribe",
+        type="http",
+        website=True,
+        auth="public",
+    )
+    def mailing_contact_unsubscribe(self, res_id, email=None, token="", **post):
+        subscription = request.env["mailing.contact.subscription"].sudo().browse(res_id)
+        if not subscription.exists():
+            return request.redirect("/web")
```
simahawk, line 23: 404 instead?

simahawk (commented):
LG overall

damdam-s (approved):

mass_mailing_subscription_email/controllers/main.py
ivantodorovich: Trying to mimic core here: https://github.com/odoo/odoo/blob/5c9b42fa1714bc9968f524bdced432b536e43042/addons/mass_mailing/controllers/main.py#L82

PR #848, opened 2022-02-21
[FIX] mail_outbound_static: valid email_from

zaoral:
parseaddr does separate the part of an email address and the name but actually:

1. does not check if the email address is a valid one.
2. does not properly work if the incomming Message['From'] is in a wrong format.

Thanks to this, when we are trying to get the domain of the email_from a really generic exception is raised. "IndexError: list index out of range" only because the email_from is not a valid email address.

To avoid this we check the email_from address is a valid one first.

The test case that brings this change was this one: Message['From'] = "Name, Last Name <valid_email@email.from>. Since we have the name_from with a comma but without a "" surrounding the parseaddr result was ('', 'Name')

Now we try to fix it, If this is not possible he proper raise will be shown to the user telling that the given email from is not a valid one.

--
related to adhoc 46491

### DIFF
--- a/mail_outbound_static/models/ir_mail_server.py
+++ b/mail_outbound_static/models/ir_mail_server.py
@@ -3,9 +3,10 @@
 
 import re
 from email.utils import formataddr, parseaddr
+from email_validator import validate_email, EmailSyntaxError
 
 from odoo import _, api, fields, models, tools
-from odoo.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


rafaelbn:
Hello @jjscarafia @eLBati maybe you can review this FIX 😄 thanks!

rafaelbn:
@zaoral hava you review travis please? Could you make all gets 🍏 ? 😄

PR #804, opened 2021-11-23
[IMP] mail_tracking{,_mailgun}: process sent events and fix wrong failure event mappings... and refactor

yajo:
FW-port https://github.com/OCA/social/pull/787.

@Tecnativa TT32365

PR #850, opened 2022-02-25
[14.0][IMP] mail_tracking_mailgun: refactor to support modern webhooks (forward port)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/social/pull/804.

Before this patch, the module was designed after the [deprecated Mailgun webhooks][3]. However Mailgun had the [events API][2] which was quite different. Modern Mailgun has deprecated those webhooks and instead uses new ones that include the same payload as the events API, so you can reuse code.

However, this was incorrectly reusing the code inversely: trying to process the events API through the same code prepared for the deprecated webhooks.

Besides, both `failed` and `rejected` mailgun events were mapped to `error` state, but that was also wrong because [`mail_tracking` doesn't have an `error` state][1].

So the logic of the whole module is changed, adapting it to process the events API payload, both through controllers (prepared for the new webhooks) and manual updates that directly call the events API.

Also, `rejected` is now translated into `reject`, and `failed` is translated into `hard_bounce` or `soft_bounce` depending on the severity, as specified by [mailgun docs][2]. Also, `bounced` and `dropped` mailgun states are removed because they don't exist, and instead `failed` and `rejected` properly get their metadata.

Of course, to know the severity, now the method to obtain that info must change, it' can't be a simple dict anymore.

Added more parameters because for example modern Mailgun uses different keys for signing payload than for accessing the API. As there are so many parameters, configuration is now possible through `res.config.settings`. Go there to autoregister webhooks too.

Since the new webhooks are completely incompatible with the old supposedly-abstract webhooks controllers (that were never really that abstract), support for old webhooks is removed, and it will be removed in the future from `mail_tracking` directly. There is a migration script that attempts to unregister old webhooks and register new ones automatically.

[1]: https://github.com/OCA/social/blob/f73de421e28a43d018176f61725a3a59665f715d/mail_tracking/models/mail_tracking_event.py#L31-L42
[2]: https://documentation.mailgun.com/en/latest/api-events.html#event-types
[3]: https://documentation.mailgun.com/en/latest/api-webhooks-deprecated.html

### DIFF
--- a/mail_tracking_mailgun/migrations/14.0.2.0.0/post-migration.py
+++ b/mail_tracking_mailgun/migrations/14.0.2.0.0/post-migration.py
@@ -1,0 +1,34 @@
+# Copyright 2021 Tecnativa - Jairo Llopis
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import logging
+
+from openupgradelib import openupgrade
+
+_logger = logging.getLogger(__name__)
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Update webhooks.
+
+    This version dropped support for legacy webhooks and added support for
+    webhook auto registering. Do that process now.
+    """
+    if version != "14.0.1.0.0":
+        return
+    settings = env["res.config.settings"].create({})
+    if not settings.mail_tracking_mailgun_enabled:
+        _logger.warning("Not updating webhooks because mailgun is not configured")
+        return
+    _logger.info("Updating mailgun webhooks")
+    try:
+        settings.mail_tracking_mailgun_unregister_webhooks()
+        settings.mail_tracking_mailgun_register_webhooks()
+    except Exception:
+        # Don't fail the update if you can't register webhooks; it can be a
+        # failing network condition or air-gapped upgrade, and that's OK, you
+        # can just update them later
+        _logger.warning(
+            "Failed to update mailgun webhooks; do that manually", exc_info=True
+        )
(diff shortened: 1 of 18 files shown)


mail_tracking_mailgun/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Update webhooks.
+
+    This version dropped support for legacy webhooks and added support for
+    webhook autoregistering. Do that process now.
+    """
+    settings = env["res.config.settings"].create({})
```
pedrobaeza, line 20: This should be done only if `version == "14.0.1.0.0"`

mail_tracking_mailgun/models/res_partner.py
```diff
@@ -67,7 +66,7 @@ def check_email_validity(self):
             ):
                 raise UserError(
                     _(
-                        "Error %s trying to check mail" % res.status_code
+                        "Error %s trying to " "check mail" % res.status_code
```
pedrobaeza: Not sure why this change

```suggestion
                        "Error %s trying to check mail" % res.status_code
```

mail_tracking_mailgun/migrations/14.0.2.0.0/post-migration.py
MiquelRForgeFlow: That's impossible, because it's a post-migration

pedrobaeza (approved):

pedrobaeza:
We can fast-track this for having it ASAP as merged in v13:

/ocabot merge nobump

Thanks for taking care of the fw-port

PR #859, opened 2022-03-03
[ADD] mail_partner_forwarding: new module

nicomacr:


rafaelbn:
/ocabot rebase

Hello @nicomacr thank you! have you already signed CLA?

nicomacr:
@rafaelbn Yes, but in this case, this development was started by a colleague who no longer works with us.
This was the reason of the fails with the CLA signature.
I already fixed this on this PR.

mail_forwarding/models/res_user.py
```diff
@@ -0,0 +1,14 @@
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    forwarding_user_id = fields.Many2one("res.users", string="Forwarding User")
```
yostashiro: Thought it would allow more flexibility (i.e. forwarding emails to partners with no login) if the field was added to res.partner instead.

jjscarafia: we haven't think on a use case for this, but indeed it would be more flexible.
We could add it on partners but allow configuration from res.users and res.partner views.
It seem ok for you?

jjscarafia: Changes done @yostashiro . Now the functionality can be used on partners

mail_forwarding/models/mail_thread.py
```diff
@@ -0,0 +1,43 @@
+from odoo import models
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    def _notify_compute_recipients(self, message, msg_vals):
+        """ Inherit this method to add in the list of partner to be notify
+        the forwarding_user_id of the any partners in the list """
```
jjscarafia:
```suggestion
        the forwarding_partner_id of any partners in the list """
```

mail_forwarding/models/mail_thread.py
```diff
@@ -0,0 +1,43 @@
+from odoo import models
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    def _notify_compute_recipients(self, message, msg_vals):
+        """ Inherit this method to add in the list of partner to be notify
```
jjscarafia:
```suggestion
        """ Inherit this method to add in the list of partners to be notify
```

mail_forwarding/models/res_partner.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    forwarding_partner_id = fields.Many2one(
+        "res.partner",
+        string="Forwarding Partner",
+        help="Only partner following will be forwarding (not messages"
+        " delivered through channels)",
```
jjscarafia:
```suggestion
        help="Messages will be forwarded only for partners that are followers but no partners being notify because they belong to channel that is following the thread",
```

jjscarafia (changes requested):

mail_forwarding/models/res_user.py
```diff
@@ -0,0 +1,16 @@
...
+    forwarding_partner_id = fields.Many2one(
+        related="partner_id.forwarding_partner_id", readonly=False
+    )
+
+    def __init__(self, pool, cr):
+        super().__init__(pool, cr)
+        self.SELF_WRITEABLE_FIELDS = list(self.SELF_WRITEABLE_FIELDS)
+        self.SELF_WRITEABLE_FIELDS.append("forwarding_partner_id")
+        self.SELF_READABLE_FIELDS = list(self.SELF_READABLE_FIELDS)
+        self.SELF_READABLE_FIELDS.append("forwarding_partner_id")
```
yostashiro: I suppose you can remove this file since res.users inherits res.partner.

nicomacr: I tried it, but still be necessary having it.

mail_forwarding/README.rst
```diff
@@ -0,0 +1,74 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/social-13-0/social-13-0-mail_debrand
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/205/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module modifies the functionality of emails to remove the Odoo branding,
```
rafaelbn: Hello! I don't think this is the correct description 😄

mail_forwarding/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2022 AdHoc SA - Pablo Paez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Mail Fowarding",
+    "summary": "cc user for notifications",
```
rafaelbn: I don't understand this is the scope of the module, is it?

nicomacr:
@rafaelbn I've changed those parts suggested in the readme to better comprehension for the use of the module.
Let me know if missing more changes.

rafaelbn:
Thanks! We have to think in other `name for the module` because this is confusing with other modules like: (`mail_forward`) from previous versions https://github.com/OCA/social/tree/10.0/mail_forward AND the scope of the module!

"This module allows to select a related partner (called "Forwarding Partner") in each partner in order to send all the notifications
 to the Forwarding Partner."

`partner_mail_forwarding` I see this one much better

Maybe `mail_partner_forwarding`

You are not forwarding email, this module set a rule to forward all from a partner to another.

What do you think? 😄

jjscarafia:
Hi @rafaelbn ! Thanks for your feedback.
I like partner_mail_forwarding but usually the first part of the module name is related to the app it's touching (that it would be mail in this case)
Some ideas
1. we could keep as it is "mail_forwarding" due to the standard of using same word as the main app the module touches. But, on module name and description use "Partner Mail Forwarding"
2. Another possibility would be to call it "mail_forwarding_by_partner"
3. If there are other modules where the module don't start with the app, or I'm wrong with that standard, no problem to use "partner_mail_forwarding"

Any ideas?

rafaelbn:
Hello @jjscarafia 😄 

I like `mail_forwarding_by_partner`
I like `mail_partner_forwarding`

So up to you!

nicomacr:
Hello @rafaelbn 

I just changed the module name and tidy up the PR( squash commit, etc).
Take a look and tell me your thoughts.

Thanks

rafaelbn:
For me is OK, are you going to install in production?

rafaelbn:
@jjscarafia I think your review is blocking here 😄

jjscarafia:
Hi, thanks @rafaelbn 
We will install in production and after a week of hard test I will give the review, seems ok?
Thanks!

houssine78:
Hi @jjscarafia did you get hard times on production ? :)
Do you approve it ?

thank you

houssine78 (approved):

jjscarafia (approved):
LGTM

jjscarafia:
@houssine78 
we've this on production for more than a month.
at least it doesn't break anything :p
let's merge it!

houssine78:
@rafaelbn I think this one could be merged now

nicomacr:
@rafaelbn I think this PR can be merged, we already use this on v15 over this [PR](https://github.com/OCA/social/pull/923), and works fine.

PR #863, opened 2022-03-14
[14.0][ADD] mail_via

chafique-delli:
When importing email from incoming gateway and sending back them to follower.
Odoo set in the original sender in the from message.
With email provider like mailjet, the mail is not accepted as we do not have the right to send an email with the original sender.

This module will change the FROM with a "via" address and use as name the name with the email of the sender.


chafique-delli:
ping @florian-dacosta

chafique-delli:
@florian-dacosta , @bguillot , @mourad-ehm

bguillot:
Hi, there is an alternative is base module.
We can close this PR.

PR #866, opened 2022-03-18
[15.0][IMP][mail_show_follower] Customize notification appearance

Shide:
Added settings to customize notification and translations to important parts of the message.

`show_followers_message_sent_to`: The first part of the message
`show_followers_partner_format`: Partner format for emails. You can chooose which fields to show
`show_followers_message_response_warning`: Additional text if you want to warn for the reply behavior

Includes a preview of the message on settings (only visible if you are in debug mode)
**The lang of the message is on the same as the recipient.** Some customization settings are translatable (enable another language to test properly)

@Yajo @rafaelbn @HaraldPanten @ValentinVinagre @pedrobaeza @flotho @cvinh @tafaRU  please review 😄 

MT-431 @moduon

### DIFF
--- a/mail_show_follower/models/mail_mail.py
+++ b/mail_show_follower/models/mail_mail.py
@@ -1,16 +1,64 @@
-from odoo import models
+from markupsafe import Markup
+
+from odoo import api, models, tools
 
 
 class MailMail(models.Model):
     _inherit = "mail.mail"
 
-    def _send(self, auto_commit=False, raise_exception=False, smtp_session=None):
-        plain_text = (
-            '<div summary="o_mail_notification" style="padding: 0px; '
-            'font-size: 10px;"><b>CC</b>: %s<hr style="background-color:'
-            "rgb(204,204,204);border:medium none;clear:both;display:block;"
-            'font-size:0px;min-height:1px;line-height:0; margin:4px 0 12px 0;"></div>'
+    @api.model
+    def _build_cc_text(self, partners):
+        if not partners:
+            return ""
+
+        def get_ctx_param(ctx_key, default_parm):
+            if ctx_key in self.env.context:
+                return self.env.context[ctx_key]
+            return default_parm
+
+        def remove_p(markup_txt):
+            if markup_txt.startswith("<p>") and markup_txt.endswith("</p>"):
+                return markup_txt[3:-4]
+            return markup_txt
+
+        company = self.env.company
+        partner_format = get_ctx_param(
+            "partner_format", company.show_followers_partner_format
         )
+        msg_sent_to = get_ctx_param(
+            "msg_sent_to", company.show_followers_message_sent_to
+        )
+        msg_warn = get_ctx_param(
+            "msg_warn", company.show_followers_message_response_warning
+        )
+        partner_message = ", ".join(
+            [
+                partner_format
+                % {
+                    # Supported parameters
+                    "partner_name": p.name,
+                    "partner_email": p.email,
+                    "partner_email_domain": tools.email_domain_extract(p.email),
+                }
+                for p in partners
+            ]
+        )
+        full_text = """
+            <div summary='o_mail_notification' style='padding:5px;
+            margin:10px 0px 10px 0px;font-size:13px;border-radius:5px;
+            font-family:Arial;border:1px solid #E0E2E6;background-color:#EBEBEB;'>
+            {msg_sent_to} {partner_message}
+            {rc}{msg_warn}
+            </div>
+        """.format(
+            msg_sent_to=remove_p(msg_sent_to),
+            partner_message=Markup.escape(partner_message),
+            rc=msg_warn.striptags() and "<br/>" or "",
+            msg_warn=msg_warn.striptags() and remove_p(msg_warn) or "",
+        )
+        return full_text
+
+    def _send(self, auto_commit=False, raise_exception=False, smtp_session=None):
         group_portal = self.env.ref("base.group_portal")
         for mail_id in self.ids:
             mail = self.browse(mail_id)
@@ -73,13 +121,21 @@ class MailMail(models.Model):
                             or x.user_ids  # otherwise, email is not sent
                             and "email" in x.user_ids.mapped("notification_type")
                         )
-                        # get names and emails
-                        final_cc = None
-                        mails = ""
-                        for p in partners:
-                            mails += "%s &lt;%s&gt;, " % (p.name, p.email)
-                        # join texts
-                        final_cc = plain_text % (mails[:-2])
+                        # set proper lang for recipients
+                        langs = list(
+                            filter(
+                                bool,
+                                mail.mapped("recipient_ids.lang")
+                                + [
+                                    mail.author_id.lang,
+                                    self.env.company.partner_id.lang,
+                                ],
+                            )
+                        )
+                        # get show follower text
+                        final_cc = mail.with_context(
+                            lang=langs and langs[0]
+                        )._build_cc_text(partners)
                         # it is saved in the body_html field so that it does
                         # not appear in the odoo log
                         mail.body_html = final_cc + mail.body_html
--- a/mail_show_follower/models/res_company.py
+++ b/mail_show_follower/models/res_company.py
@@ -5,5 +5,24 @@ class ResCompany(models.Model):
     _inherit = "res.company"
 
     show_internal_users_cc = fields.Boolean(
-        string="Show Internal Users CC", default=True
+        string="Show Internal Users CC",
+        default=True,
+    )
+    show_followers_message_sent_to = fields.Html(
+        string="Text 'Sent to'",
+        translate=True,
+        default="This message has been sent to",
+    )
+    show_followers_partner_format = fields.Char(
+        string="Partner format",
+        default="%(partner_name)s",
+        help="Supported parameters:\n"
+        "%(partner_name)s = Partner Name\n"
+        "%(partner_email)s = Partner Email\n"
+        "%(partner_email_domain)s = Partner Email Domain",
+    )
+    show_followers_message_response_warning = fields.Html(
+        string="Text 'Replies'",
+        translate=True,
+        default="Notice: Replies to this email will be sent to all recipients",
     )
--- a/mail_show_follower/models/res_config_settings.py
+++ b/mail_show_follower/models/res_config_settings.py
@@ -1,11 +1,51 @@
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class ResConfigSettings(models.TransientModel):
     _inherit = "res.config.settings"
 
     show_internal_users_cc = fields.Boolean(
-        string="Show Internal Users CC",
         related="company_id.show_internal_users_cc",
         readonly=False,
     )
+    show_followers_message_sent_to = fields.Html(
+        related="company_id.show_followers_message_sent_to",
+        readonly=False,
+    )
+    show_followers_partner_format = fields.Char(
+        related="company_id.show_followers_partner_format",
+        readonly=False,
+        help="Supported parameters:\n"
+        "%(partner_name)s = Partner Name\n"
+        "%(partner_email)s = Partner Email\n"
+        "%(partner_email_domain)s = Partner Email Domain",
+    )
+    show_followers_message_response_warning = fields.Html(
+        related="company_id.show_followers_message_response_warning",
+        readonly=False,
+    )
+    show_followers_message_preview = fields.Html(
+        string="Message preview",
+        readonly=True,
+        store=False,
+    )
+
+    @api.onchange(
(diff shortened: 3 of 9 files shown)


rafaelbn (approved):
Amazing! Thank you! 👍 

@cvinh you have here all you need

mail_show_follower/models/res_company.py
```diff
@@ -7,3 +7,37 @@ class ResCompany(models.Model):
...
+        help="Format accepts HTML tags.\n"
+        "Eg.: This <strong>message</strong> has been...",
+    )
+    show_followers_partner_format = fields.Char(
+        string="Partner format",
+        default="{partner.name}",
+        help="Use keyword 'partner' for calling partner fields.\n"
+        "Format according to plain HTML.\n"
+        "Eg.: &lt;{partner.email}&gt;",
+    )
```
yajo: ⚠️  This opens a dangerous attack surface. One could set a parameter such as:

```
{partner.env.cr.execute('DELETE FROM res_partner')}
```

Instead, the string should be processed using %-formatter, and some useful variables should be given to its context (probably `partner_name`, `partner_email` and `partner_email_domain` should be enough).

---

To make it even more usable, secure and useful, another option is to replace this for several booleans that allow displaying a specific part. For example:

```suggestion
    show_followers_name = fields.Boolean(
        default=True,
        help="Display follower names on messages?",
    )
    show_followers_email = fields.Boolean(
        default=False,
        help="Display follower e-mails on messages?",
    )
    show_followers_domain = fields.Boolean(
        default=True,
        help="Display follower e-mail domains on messages?",
    )
```

mail_show_follower/models/res_company.py
```diff
@@ -7,3 +7,37 @@ class ResCompany(models.Model):
     show_internal_users_cc = fields.Boolean(
         string="Show Internal Users CC", default=True
     )
+    show_followers_message_sent_to = fields.Char(
+        string="Text 'Message sent to'",
+        translate=True,
+        default="This message has been sent to",
+        help="Format accepts HTML tags.\n"
+        "Eg.: This <strong>message</strong> has been...",
+    )
```
yajo: Supporting direct HTML is complicated.

It has security considerations, as it could open a way to inject XSS attacks in the message, unless the wrapper happens before HTML sanitizing. Also, the sanitizing process itself could produce unexpected behaviours. Translation for HTML fields is also managed differently, and this one should be declared as `fields.Html` instead.

So, all of this makes me think: do we *really* need that support? Is there a valid use case for it?

Odoo doesn't let you AFAIK configure the appearance of outgoing message headers. At least not too much. And that has almost never been a problem.

Too many things to take into account for something that isn't really so much needed. I think it's better to start with a MVP and just add the simplest possible support for the few things that are really needed.

So, please, could you remove HTML support?

mail_show_follower/models/res_company.py
```diff
@@ -7,3 +7,37 @@ class ResCompany(models.Model):
...
+        "Format according to plain HTML.\n"
+        "Eg.: &lt;{partner.email}&gt;",
+    )
+    show_followers_message_response_warning = fields.Char(
+        string="Text 'Warn for replies'",
+        translate=True,
+        default="Notice: Replies to this email will be sent to all recipients",
+        help="Format accepts HTML tags.\n"
+        "Eg.: <strong>Warning:</strong> This email...",
+    )
```
yajo: I don't think it's so important to be able to customize this message. A boolean flag should be enough: display it or not? If true, then just display a sane warning.

mail_show_follower/models/res_company.py
```diff
@@ -7,3 +7,37 @@ class ResCompany(models.Model):
...
+    show_followers_border_css = fields.Char(
+        string="Border",
+        default="1px solid #E0E2E6",
+        help="Format according css.\n" "Eg.: 1px solid black",
+    )
+    show_followers_font_family = fields.Char(
+        string="Font family",
+        default=lambda s: s.env.company.font,
+    )
```
yajo: IMHO this is also too much. The default template itself should be properly integrated with Odoo's scheme. If Odoo's style is configured somewhere, then the default template should get those colors/borders/fonts from there. If Odoo's style is hardcoded, then the template here should be hardcoded too.

To make it customizable enough, just add a method in the model, to let others override it if needed:

```py
def _get_show_followers_template(self):
    return "<div ... %(content)s ... </div>" # <-- all the html here
```

Another alternative would be to obtain that partial template from an `ir.ui.view`, which would allow overriding, translating and customization normally, just like Odoo itself does.

With any of those alternatives, we can just remove all these options:

```suggestion
```

mail_show_follower/views/res_config_settings.xml
```diff
@@ -10,12 +10,88 @@
...
                     <div class="o_setting_left_pane">
-                        <field name="show_internal_users_cc" />
                     </div>
                     <div class="o_setting_right_pane">
-                        <label for="show_internal_users_cc" />
-                        <div class="text-muted" id="show_internal_users_cc">
-                            Add internal users in cc mails details
+                        <div class="content-group">
+                            <label
+                                for="show_internal_users_cc"
```
yajo, line 17: There's something weird on how this is rendered. 🤔  It doesn't follow the pattern.

https://user-images.githubusercontent.com/973709/159226040-2c68ca07-c099-4ba7-838d-74e2ebe67c07.mp4

As you can see, the main checkbox is positioned at the left of the vertical bar. Then, the extra options appear or disappear below. Maybe you can improve this part a little drawing inspiration from other modules that do similar stuff.

yajo (changes requested):
Thanks! It seems a good job indeed. I feel it's a bit over-engineered on some points, commented below.

Please attend those comments, to reduce the feature set to only what's really needed, so it becomes more maintainable.

Also, if possible, add tests.

Apart from that, the core feature you're adding seems a great addition indeed. Good job, thanks!

mail_show_follower/models/res_company.py
Shide: Done

mail_show_follower/models/res_company.py
Shide: Done

mail_show_follower/models/res_company.py
Shide: Not done because you may want to incise on some words and it's easy to translate.

mail_show_follower/models/res_company.py
Shide: Done removing the css customization

mail_show_follower/views/res_config_settings.xml
Shide: Done moving chekcbox to the right

mail_show_follower/models/mail_mail.py
```diff
@@ -73,13 +119,21 @@ def _send(self, auto_commit=False, raise_exception=False, smtp_session=None):
...
-                        mails = ""
-                        for p in partners:
-                            mails += "%s &lt;%s&gt;, " % (p.name, p.email)
-                        # join texts
-                        final_cc = plain_text % (mails[:-2])
+                        # set proper lang for recipients
+                        langs = list(
+                            filter(
+                                bool,
+                                mail.mapped("recipient_ids.lang")
```
yajo, line 128: I'm pondering about this line and I have some questions. 🤔 

Let's imagine an email is sent to 3 recipients. They speak Spanish, English and Chinese respectively. Which language will be chosen? Will that one be used to render emails to all participants?

Odoo itself renders buttons such as "View Task" in the destination language always. Is it safe to trust that the language we have already in the context at this point will be the correct one? In that case, would we be able to remove all of this?

mail_show_follower/models/mail_mail.py
```diff
@@ -1,16 +1,62 @@
...
+            "msg_warn", company.show_followers_message_response_warning
         )
+        partner_message = ", ".join(
+            [
+                partner_format
+                % {
+                    # Supported parameters
+                    "partner_name": p.name,
+                    "partner_email": p.email,
+                    "partner_email_domain": p.email.split("@")[-1],
```
yajo: You can use [`email_domain_extract`](https://github.com/odoo/odoo/blob/454aa2587dcda2582ae58650019669ff0e8baa2c/odoo/tools/mail.py#L521-L529)

mail_show_follower/models/res_company.py
```diff
@@ -5,5 +5,25 @@ class ResCompany(models.Model):
...
+        default="This message has been sent to",
+    )
+    show_followers_partner_format = fields.Char(
+        string="Partner format",
+        default="%(partner_name)s",
+        help="Supported parameters:\n"
+        "%(partner_name)s = Partner Name\n"
+        "%(partner_email)s = Partner Email\n"
+        "%(partner_email_domain)s = Partner Email Domain\n"
+        "Format on plain HTML: &lt;(<), &gt;(>)",
```
yajo: It's better to not use plain HTML as format. Instead, do automatic escaping when interpolating the values. IMHO it provides a better UX because asking an average user to write HTML escape sequences is too hard, and opens a venue for unexpected behaviours.

This point is important.

You can just wrap the template with [`markupsafe.escape`](https://markupsafe.palletsprojects.com/en/2.1.x/escaping/#markupsafe.escape) like Odoo does, when using it (see other comment).

Even better, instead of asking the user to write a %(template)s, a nice UX enhancement would be to use a Selection field instead, so you just give the user 3 options:

1. Mitchell Admin
2. Mitchell Admin (admin@example.com)
3. Mitchell Admin (example.com)

The simpler, the better! 😃

mail_show_follower/models/mail_mail.py
```diff
@@ -1,16 +1,62 @@
...
+
+        def remove_p(markup_txt):
+            if markup_txt.startswith("<p>") and markup_txt.endswith("</p>"):
+                return markup_txt[3:-4]
+            return markup_txt
+
+        company = self.env.company
+        partner_format = get_ctx_param(
+            "partner_format", company.show_followers_partner_format
+        )
```
yajo, line 27: As explained above, wrap it please:

```suggestion
        partner_format = markupsafe.escape(get_ctx_param(
            "partner_format", company.show_followers_partner_format
        ))
```

mail_show_follower/models/res_config_settings.py
```diff
@@ -9,3 +9,50 @@ class ResConfigSettings(models.TransientModel):
...
+    )
+    show_followers_partner_format = fields.Char(
+        string="Partner format",
+        related="company_id.show_followers_partner_format",
+        readonly=False,
+        help="Supported parameters:\n"
+        "%(partner_name)s = Partner Name\n"
+        "%(partner_email)s = Partner Email\n"
+        "%(partner_email_domain)s = Partner Email Domain\n"
+        "Format on plain HTML: &lt;(<), &gt;(>)",
```
yajo: Same as above, just pointing it out here.

BTW, you don't really need to repeat any of the parameters that already are found in the related field, such as `translate`, `help`, `string`... These are inherited automatically. 😉

mail_show_follower/views/res_config_settings.xml
```diff
@@ -10,12 +10,70 @@
...
+                            <label
+                                for="show_followers_message_preview"
+                                class="text-muted"
+                                groups="base.group_no_one"
+                            />
+                            <field
+                                name="show_followers_message_preview"
+                                style="width:100%;"
+                                widget="html"
+                                groups="base.group_no_one"
```
yajo: I would let users see this field even if they're not in debug mode. It's a very nice visual hint that helps always IMHO. 👍🏼 

```suggestion
```

yajo (commented):
The improvement looks a lot better now!

There's one point where I think it still needs a bit of enhancement, and other points where I have suggestions or doubts.

Thanks for your great job and patience, this is becoming in very good shape. 😊

mail_show_follower/models/mail_mail.py
Shide: AFAICS, mails are generated only once and sent multiple times to each recipient, so the body_html is in the language of the sender (normal notifications).
Maybe in other modules (like mass mailing) renders the body with other options, but I'm testing with normal responses and uses the sender lang.
I'm doing almost the same than Odoo: partner lang > author lang (or self.env.user) > company base lang

mail_show_follower/models/mail_mail.py
Shide: 👍🏻

mail_show_follower/models/res_company.py
Shide: Yes to markup
No to the selection because maybe users want to format emails with brackets, parenthesis, etc.

mail_show_follower/models/res_config_settings.py
Shide: I will remove unnecesary properties, but `help` is not being inherited.

yajo (approved):

PR #868, opened 2022-03-18
[UPD] mail_autosubscribe: add subscription-check function

SilvioC2C:
Little improvement: an handy function that checks whether a partner is subscribed to a specific model.

### DIFF
--- a/mail_autosubscribe/models/res_partner.py
+++ b/mail_autosubscribe/models/res_partner.py
@@ -14,3 +14,15 @@ class ResPartner(models.Model):
         column1="partner_id",
         column2="model_id",
     )
+
+    def _has_autosubscription_to(self, model_name: str) -> bool:
+        """Checks whether partner has auto-subscription to a specific model
+
+        :param str model_name: model technical name
+        :returns: True if partner has auto-subscription on given model
+        """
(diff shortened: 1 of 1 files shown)


jcoux (approved):

ivantodorovich (commented):
Correct me if I'm wrong but this is basically just a replacement for the one-liner: 

```py
res = (model_name in partner.mail_autosubscribe_ids.mapped("model"))
```

If this method were actually used in the module, rather than just being a helper method, I would be in favour because it would serve as a possible hook point for custom modules. But as it is right now I fail to see its value :/ you can just use the one-liner mentioned above

SilvioC2C:
@ivantodorovich I used in a custom module, then I thought that, given its generic use, it could be better to have it defined directly in here. This way, it'll be accessible to everyone who uses this module.

ivantodorovich:
What I mean is this method is not used to compute the actual followers of a document (ie in `_message_get_autosubscribe_followers`). Nor it should because it would compromise the domain-based filtering performance we have right now.

It's adding ~10 untested loc that aren't really used in the module and that can be replaced by a one-liner. IMO it's unnecessary ^^

Issue #520, opened 2020-03-11
[mass_mailing_partner] what should happen, when partner is archived?


rgl1234:
Hello
By testing mass_mailing_partner, it seems that nothing happens, when a subscribed-partner is archived. In my opinion, there should be the following behaviour, when a partner is archived:

1) unsubscribe partner from all mailing-lists
or
2) delete mass-mailing-contact
or
3) other suggestions ?

What do you think about this?
René

pedrobaeza:
Not sure if this automation should be done...

@rafaelbn @Yajo what do you think about this?

yajo:
The archived partner is excluded by default from all searches unless `['archive', '=', False]` or `.with_context(active_test=False)` are explicitly used, so I don't see a reason to do this.

pedrobaeza:
@Yajo that's not what happens with a mass_mailing.contact linked to an archived partner.

rgl1234:
Yes, I f a partner is archived, then mails are still sent to the mailing-list-contact, because the contact is not touched. In my case, I have been asked by the customer to "deactivate" his address in our database, because he has stopped working. I then "unsubscribed" the mass-mailing-contact manually to not  send any mails to him in the future....
Maybe the correct behaviour would be to archive the mass-mailing-contact. But as this functionality is not available by default, I thought to unsubscribe it could be a "work-around"....

yajo:
The user can click in the unsubscribe link in received mass mailings and alter his preferences, right?
I think Odoo has a mail blacklist in recent versions, which would  also serve your purposes if I'm not wrong. Have you considered these options?

pedrobaeza:
@Yajo you are deviating from main purpose. This is not something on the customer side, but on your side. If you deactivate a partner, then all in your system should be consistent. I'm stating to think that yes, linked mass_mailing.contact should be deactivated/removed on that deactivation.

yajo:
I don't think so.

If you have a partner and still have one or several mass mailing contacts with the same address, that's most likely because you want him to be able to manage his subscriptions separately.

The fact you archive a partner means that it's no longer interesting to you; but the fact someone unsubscribes from a mailing list means that it's no longer interesting to him. You have different tools for different purposes. Otherwise just use partners and tags instead of mailing lists.

If you want to do that in your DB, you can also add a automated action that cares for that.

pedrobaeza:
[...]

PR #871, opened 2022-03-22
[13.0] add mass_mailing_contact_active and mass_mailing_partner_contact_active

grindtildeath:
This PR includes two modules:

 - mass_mailing_contact_active: This module adds an active field on mailing contact and subscription.
 - mass_mailing_partner_contact_active: This module automatically triggers archive/unarchive operations on partner
to the mailing list contact and its subscriptions.

Closes: #520

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,18 +36,18 @@ jobs:
       matrix:
         include:
           - container: ghcr.io/oca/oca-ci/py3.6-odoo13.0:latest
-            include: "mail_restrict_follower_selection"
+            include: "mail_restrict_follower_selection,mass_mailing_contact_active,mass_mailing_partner_contact_active"
             makepot: "true"
             name: test with Odoo
           - container: ghcr.io/oca/oca-ci/py3.6-ocb13.0:latest
-            include: "mail_restrict_follower_selection"
+            include: "mail_restrict_follower_selection,mass_mailing_contact_active,mass_mailing_partner_contact_active"
             name: test with OCB
           - container: ghcr.io/oca/oca-ci/py3.6-odoo13.0:latest
-            exclude: "mail_restrict_follower_selection"
+            exclude: "mail_restrict_follower_selection,mass_mailing_contact_active,mass_mailing_partner_contact_active"
             makepot: "true"
             name: test with Odoo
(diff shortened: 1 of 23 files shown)


ajaniszewska-dev (approved):
LG

sebalix (approved):
Travis failing on other addons, doesn't seem related to this PR

grindtildeath:
@yajo @pedrobaeza Can you guys please have a look at the travis error on this PR? :pray: 

Looks like it happens only when `mass_mailing_partner` is installed before `mass_mailing_custom_unsubscribe`, and that is what is happening here with a new module depending on the former.

pedrobaeza:
It's your module which is triggering a new write. The usual technique is:

- Add at the beginning `from odoo.tools import config`.
- On the conflicting code, add a "test condition" for seeing if it's in test mode and it's testing your module: `test_condition = config["test_enable"] and self.env.context.get("test_mass_mailing_contact_active")`
- Don't do your code `if not test_condition`.
- Add in your module tests the context `test_mass_mailing_contact_active` for testing your behavior.

grindtildeath:
@pedrobaeza The `write` functions are not triggered during the tests of `mass_mailing_custom_unsubscribe`, but it seems that the definition of the `active` field on `mailing.contact.subscription` is breaking this test: 
```
AssertionError: The test code "odoo.__DEBUG__.services['web_tour.tour'].run('mass_mailing_custom_unsubscribe_tour_contact')" failed
Tour mass_mailing_custom_unsubscribe_tour_contact failed at step Uncheck list 1 (trigger: li:contains('test list 1') input:checked)
```
Question is, can these JS tests be fixed considering the possibility of having an  `active` field on `mailing.contact.subscription` or should we better test the modules separately :question:

pedrobaeza:
For me it's an error to segregate modules, because they can be installed together in real installations, so the problem may appear there, but I don't want to block this.

grindtildeath:
@pedrobaeza I did spend more than 4 hours trying to understand what's the issue, and it seems to be linked to the order of modules installation as the issue is not reproducible all the time. I unfortunately can't invest any more time, as it looks to be quite random.

pedrobaeza:
OK, let's keep it then this way and hope that the issue is not reproduced in real environments:

/ocabot merge nobump

PR #877, opened 2022-03-29
[13.0][FIX] missing commits of mail_activity_team

albariera:
Without these commits, the team activity widget did not work properly.

PR #896, opened 2022-04-18
[13.0][FIX] mail_activity_team: activity count

chienandalu:
The user couldn't see the proper count of his own activities. It was
allways showing the team activities counter.

cc @Tecnativa TT35885

please review @Tardo @victoralmau  @MiquelRForgeFlow @LoisRForgeFlow 

Also please forward port features across versions so you don't miss them later, like this one https://github.com/OCA/social/pull/877

### DIFF
--- a/mail_activity_team/models/res_users.py
+++ b/mail_activity_team/models/res_users.py
@@ -14,7 +14,7 @@ class ResUsers(models.Model):
 
     @api.model
     def systray_get_activities(self):
-        if not self._context.get("team_activities", False):
+        if not self.env.context.get("team_activities"):
             return super().systray_get_activities()
         query = """SELECT m.id, count(*), act.res_model as model,
                                 CASE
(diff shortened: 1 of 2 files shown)


mail_activity_team/models/res_users.py
```diff
@@ -14,7 +14,7 @@ class ResUsers(models.Model):
 
     @api.model
     def systray_get_activities(self):
-        if not self._context.get("team_activities", False):
+        if not self.env.context.get("team_activities"):
```
MiquelRForgeFlow, line 17: why this change? 🤔

chienandalu: It's been several months from the patch :sweat_smile: but probably there was an issue with cached context

pedrobaeza: You should not use the internal variable, but the external one. It has the same content for now, but that may change in the future.

pedrobaeza (approved):

pedrobaeza:
@MiquelRForgeFlow can you please approve?

MiquelRForgeFlow (approved):

PR #897, opened 2022-04-20
[FIX] [15.0] base_search_mail_content: Avoid CacheMiss errors

Shide:
Avoid **CacheMiss** errors setting a value on `message_content` field.

Odoo needs to have a value on this `message_content` field because **Cache** doesn't found a previous valid value.

MT-441 @moduon


yajo (approved):

rafaelbn (approved):
LGTM 👍

Issue #860, opened 2022-03-04
Backport support for Google Mail OAuth authentication for Odoo <= 11.0


gurneyalex:
Google announced that they will be requiring oauth for mail access starting on may 30, 2022.

Odoo has made changes starting on version 12.0 (https://github.com/odoo/odoo/commit/336c9aab3619abb5bfbeb07f87ae014facdefb0b) to add this support.

We need to backport this to addons for version 11.0 and before.

dreispt:
I feel that the https://github.com/OCA/server-auth repo is more adequate for this.

pedrobaeza:
I don't think so, as that's for Odoo authentication methods, not for mailing.

Remember to put on the README the Odoo backport disclaimer and put the initial commit with the module as is extracted from Odoo attributed to for example `info@odoo.com`.

dreispt:
@pedrobaeza Right, that makes sense.

yajo:
I think OCB is the place. After all, it only exists with the purpose of backporting stuff from Odoo 🤷🏼‍♂️. The fix linked alters the fetchmail module, which is ubiquitous. I don't see any benefit on increasing the complexity of the backport (having to separate some parts of the commit in OCB and some other parts in a separate module in another repo).

gurneyalex:
I tend to agree with @Yajo here: it seems to me that a fix in OCB would be easier, and these modules don't need to evolve over time and be migrated forward.

pedrobaeza:
OK, as you prefer. OCB is not very prepared in CI terms, but seeing the Odoo commit, there's a minimal change to be done on core `fetchmail` module and  `base`, so that one is needed to be on OCB or monkey-patched in the module, so it can be easier to be done on OCB.

simahawk:
:+1: for OCB

gurneyalex:
Do we need an update in mail_environment or maybe a new module in https://github.com/OCA/server-env ?

dreispt:
AFAIU  OCB is for bugfixes, not feature backports.
The rule was to early merge upstream proposed PRs (even if rejected by Odoo)

In the cases where an Odoo module was back or forward ported, that was done as a module in the appropriate repo.
My feeling is that this should be an OCA/social module, extracted from Odoo and backported.

yankinmax:
I opened PR to resolve this issue. Reviews and help is appreciated. Thanks
https://github.com/OCA/social/pull/903

PatosMarco:
[...]

PR #903, opened 2022-04-30
[10.0][ADD] {google,fetchmail}_gmail: OAuth for gmail servers

yankinmax:
This is a backport of Odoo SA commit https://github.com/odoo/odoo/commit/336c9aab3619abb5bfbeb07f87ae014facdefb0b
This PR is opened in order to resolve an issue https://github.com/OCA/social/issues/860
The way `ir.mail_server` creates smtp connection in v10 differ from v12. That require a complete override of the `connect` and `send_email` methods. But, I'm sure my implementation isn't the best one, and therefore not tested for the moment of PR open Help is highly appreciated to override it correctly.


yankinmax:
though we have compatible with python2.7 `premailer==3.6.2` in requirements https://github.com/OCA/social/blob/f32e179fee58afc27c450c9201b66f31c38ef38d/requirements.txt#L1
runbot is installing last version which doesn't support python2.7
```
Collecting premailer
  Downloading premailer-3.10.0-py2.py3-none-any.whl
```

PatosMarco:
what is the correct procedure to configure the oauth client? After entering client_id and secret, I cannot connect to my account

PatosMarco:
the link that opens always has client_id = false in the url

yankinmax:
@PatosMarco Features introduced in this PR weren't completely resolved, as I couldn't test a real Gmail account. If you have client_id and secret it's great, because you can finish work. I'm not sure my suggestions will be good enough without debugging on real credentials. Sorry.
Please give it a try by yourself. If you need I can give you access to update this PR to finish work. THX

PR #905, opened 2022-05-09
[FIX] Added a functionlity for loge not can see with a condition.

AmmarOfficewalaSerpentCS:


AmmarOfficewalaSerpentCS:
@dreispt , @Chandresh-OSI , 
can you Please review this PR ?

mail_allow_portal_internal_note/models/res_users.py
```diff
@@ -20,3 +20,18 @@ class Users(models.Model):
         " for documents related tho other companies"
         " , other than his parent company",
     )
+
+    def portal_can_see_internal_messages(self, res_model, res_id):
```
dreispt, line 24: Convert this to a computed field on `mail.thread`

mail_allow_portal_internal_note/static/src/js/composer.js
```diff
@@ -2,16 +2,32 @@ odoo.define("mail_allow_portal_internal_note.composer", function (require) {
...
+                var is_portal_see = await ajax.jsonRpc(
+                    "/can_portal_see_internal_messages",
+                    "call",
+                    {
+                        res_model: this.options.res_model,
+                        res_id: this.options.res_id,
+                    }
+                );
+                if (!is_portal_see) {
+                    this.$(".o_portal_chatter_log_note_btn").addClass("o_hidden");
```
dreispt, line 27: Use the QWeb template to add the `o_hidden` depending on the message thread computed value.

dreispt (changes requested):

dreispt (changes requested):
@AmmarOfficewalaSerpentCS I believe this implementation can be simplified, if we place the business logic in the right place.

PR #906, opened 2022-05-10
[13.0][FIX] mail_tracking_mass_mailing: Set traces in error according to tracking

grindtildeath:
If an exception is caught by ir.mail_server.send_email in mail_tracking module,
the mail.tracking.email record will appear in error but the related mailing
trace would still appear as sent because mail.mail._postprocess_sent_message is
called without any failure_type in mail.mail._send in the mail module (as Exception
is not raised after being caught in mail_tracking module).

Since _postprocess_sent_message method not only sets the mailing.trace state in
mass_mailing module but can also delete the mail.mail records in mail module,
we need to ensure the mailing.trace is written accordingly to the tracking here,
and avoid having the mass_mailing module set a 'sent' status if we had an exception,
hence the usage of a context key to ignore possible writes.

### DIFF
--- a/mail_tracking_mass_mailing/models/mail_mail.py
+++ b/mail_tracking_mass_mailing/models/mail_mail.py
@@ -2,7 +2,7 @@
 # Copyright 2017 Vicent Cubells - <vicent.cubells@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import api, models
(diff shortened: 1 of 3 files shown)


sebalix (approved):
Seems legit

mmequignon (approved):

pedrobaeza:
Please fw-port it to v14

/ocabot merge patch

PR #909, opened 2022-05-17
[MIG] mail_tracking_mailgun: Migration to 15.0

nicomacr:


mail_tracking_mailgun/tests/test_mailgun.py
```diff
@@ -405,7 +405,7 @@ def test_email_validity(self, mock_request):
         }
         with self.assertRaises(UserError):
             self.partner.check_email_validity()
-        self.assertTrue(self.partner.email_bounced)
+            self.assertTrue(self.partner.email_bounced)
```
alan196: With this change, you don't execute this line, if the previous line raise an UserError the next line is not executed, I think this test is not valid in this way

alan196 (changes requested):

mail_tracking_mailgun/tests/test_mailgun.py
nicomacr: yes i just having testing and i forget to remove it. The reason is because i have problem with this test.
If you could give me some hint to solve this bug with this line of the test welcome. 
Thanks

mail_tracking_mailgun/tests/test_mailgun.py
```diff
@@ -0,0 +1,460 @@
...
+        with self.assertRaises(UserError):
+            self.partner.check_email_validity()
+        # Unable to fully validate
+        mock_request.get.return_value.json.return_value = {
+            "is_valid": True,
+            "mailbox_verification": "unknown",
+        }
+        with self.assertRaises(UserError):
+            self.partner.check_email_validity()
+        self.assertTrue(self.partner.email_bounced)
```
chienandalu, line 408: This assertion needs to be checked in its correct context, otherwise is going to fail.

```suggestion
                    self.assertTrue(self.partner.email_bounced)
        # Unable to fully validate
        mock_request.get.return_value.json.return_value = {
            "is_valid": True,
            "mailbox_verification": "unknown",
        }
        with self.assertRaises(UserError):
            self.partner.check_email_validity()
```

chienandalu (changes requested):
Thanks @nicomacr 

I've been testing it locally and everything seems to work as expected.


Regarding the tests, you need to change this in order to pass them:

mail_tracking_mailgun/tests/test_mailgun.py
chienandalu: Here's the test fix you need

```suggestion
        with self.assertRaises(UserError):
            self.partner.check_email_validity()
        # If we autocheck, the mail will be bounced
        self.partner.with_context(mailgun_auto_check=True).check_email_validity()
        self.assertTrue(self.partner.email_bounced)
        # Unable to fully validate
        mock_request.get.return_value.json.return_value = {
            "is_valid": True,
            "mailbox_verification": "unknown",
        }
        with self.assertRaises(UserError):
            self.partner.check_email_validity()
```

chienandalu (approved):
Thanks!

PR #912, opened 2022-05-26
[IMP] mail_activity_team

etobella:
This PR includes two changes

1 - Adds a default team on activities in order to set it automatically
2 - Improves tests making them more standard

@LoisRForgeFlow @albariera

PR #919, opened 2022-06-13
[IMP] mail_activity_team: Force team on schedule

olgamarcocb:
If you schedule an activity and you don't force the team as False, errors might araise.
For example, on holidays.

Is an error raised by #912


etobella (approved):
Thanks for the patch!

albariera (approved):

PR #920, opened 2022-06-13
[14.0][IMP] mail_activity_team: Forwarding commits from 13.0 to 14.0

olgamarcocb:
Includes:

- #912 
- #919 

@etobella

### DIFF
--- a/mail_activity_team/__manifest__.py
+++ b/mail_activity_team/__manifest__.py
@@ -16,6 +16,7 @@
         "views/assets_backend.xml",
         "security/ir.model.access.csv",
         "security/mail_activity_team_security.xml",
+        "views/mail_activity_type.xml",
         "views/mail_activity_team_views.xml",
         "views/mail_activity_views.xml",
         "views/res_users_views.xml",
--- a/mail_activity_team/models/mail_activity.py
+++ b/mail_activity_team/models/mail_activity.py
@@ -38,7 +38,7 @@ class MailActivity(models.Model):
         if self.team_id and self.user_id in self.team_id.member_ids:
             return res
         self.team_id = self.with_context(
-            default_res_model=self.res_model_id.id
+            default_res_model=self.res_model_id.model
         )._get_default_team_id(user_id=self.user_id.id)
         return res
 
--- a/mail_activity_team/models/mail_activity.py
+++ b/mail_activity_team/models/mail_activity.py
@@ -1,4 +1,4 @@
-# Copyright 2018 ForgeFlow S.L.
+# Copyright 2018 ForgeFlow, S.L.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import SUPERUSER_ID, _, api, fields, models
 from odoo.exceptions import ValidationError
@@ -18,6 +18,8 @@ class MailActivity(models.Model):
                 ["|", ("res_model_ids", "=", False), ("res_model_ids", "in", model.ids)]
             )
         return self.env["mail.activity.team"].search(domain, limit=1)
+
+    user_id = fields.Many2one(required=False)
 
     team_id = fields.Many2one(
         comodel_name="mail.activity.team", default=lambda s: s._get_default_team_id()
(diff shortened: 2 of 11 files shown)


etobella (approved):

albariera (approved):
LGMT :+1:

mail_activity_team/models/mail_activity.py
```diff
@@ -19,6 +19,8 @@ def _get_default_team_id(self, user_id=None):
             )
         return self.env["mail.activity.team"].search(domain, limit=1)
 
+    user_id = fields.Many2one(required=False)
```
pedrobaeza, line 23: Why removing this?

etobella: Yes, that's true, it is my fault. I will review where did this happen

etobella: After checking the code, it was removed on 14 migration, it should never come back.

kailasprasannakumar:
Hi, 
if we tick trigger next activity and force_next activity. there is an error. odoo is checking if the user assigned to next activity is in the current activity team and raises a validation error

etobella:
@MiquelRForgeFlow I think this is the proper way to solve the warnings. WDYT?

sanchonuria (approved):

MiquelRForgeFlow (approved):

mail_activity_team/__manifest__.py
```diff
@@ -11,11 +11,12 @@
     "author": "ForgeFlow, Sodexis, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "installable": True,
-    "depends": ["mail_activity_board"],
+    "depends": ["mail_activity_board", "web_domain_field"],
```
pedrobaeza: This dependency is not something that is in v13, so it's not only forward-ports, and I would like to avoid new dependencies in the module.

pedrobaeza (changes requested):

MiquelRForgeFlow:
I think https://github.com/OCA/social/commit/3f907b02b07fd63efd5275eeec0bf8e406ae61c8 should be cherry-picked.

mail_activity_team/__manifest__.py
etobella: On 14+ it is not recommended to set a dynamic domain by the onchange. the only way I found was this :thinking: 

Maybe if odoo does a small change on SparseField we could do it, Changing this by https://github.com/odoo/odoo/blob/16.0/addons/base_sparse_field/models/fields.py#L89
```python
            return json.dumps(value) if isinstance(value, (dict, list)) else (value or None)
```

Might do the trick

pedrobaeza: You can propose such change, but for this, you can still use onchanges. Being deprecated doesn't mean they are not allowed. But aren't you able to define a fixed domain not depending on content?

etobella:
I tested it (the commit) and it doesn't work properly.

I made the following on 13.0 runboat:

- Create a team with two members (one is admin)
- Create an activity for that team
- If you see the activities on the top, it shows 1.
- If you press activity button on topbar, it shows 1
- If you press team activities, it shows 0. [Should be 1]
- If you press anywhere (closing the activities widget), it shows 0 [Should be 1]

I will try to find another solution for it :disappointed:

mail_activity_team/__manifest__.py
etobella: You want to filter the users If and only If you have a team selected. We could try with a repeated field, who knows, that might work....

etobella: I made something in order to avoid it. It would be interesting if someone can test it :smile:

pedrobaeza:
Can't you use the operator `=?` like in https://github.com/OCA/partner-contact/blob/0b26df30c0f646674411c477ee233f019f3c545d/base_location/models/res_partner.py#L127 ?

etobella:
> Can't you use the operator `=?` like in https://github.com/OCA/partner-contact/blob/0b26df30c0f646674411c477ee233f019f3c545d/base_location/models/res_partner.py#L127 ?

I tested, but it does not work :cry: I made the following test on odoo shell:
```python
>>> self._where_calc([("activity_team_ids", "=?", False)]).get_sql()
('"res_users"', '(("res_users"."active" = %s) AND \n                        NOT EXISTS (\n                            SELECT 1 FROM "mail_activity_team_users_rel" AS "res_users__activity_team_ids"\n                            WHERE "res_users__activity_team_ids"."res_users_id" = "res_users".id\n                        )\n                    )', [True])
>>> self._where_calc([("activity_team_ids", "=?", 1)]).get_sql()
('"res_users"', '(("res_users"."active" = %s) AND \n                        EXISTS (\n                            SELECT 1 FROM "mail_activity_team_users_rel" AS "res_users__activity_team_ids"\n                            WHERE "res_users__activity_team_ids"."res_users_id" = "res_users".id\n                            AND "res_users__activity_team_ids"."mail_activity_team_id" IN %s\n                        )\n                    )', [True, (1,)])
>>> self._where_calc([]).get_sql()
('"res_users"', '("res_users"."active" = %s)', [True])
```

As you can see, in case of False, only the users that are not defined on a team are returned. And this is obviously not right :cry:

pedrobaeza:
You have gone so low-level, Enric. If you do that in the search, the tuple is removed when the right part is `False`. Please try with `search` method.

etobella:
Obviously, before doing this test, I did the changes in order to try it but it didn't work.

The comment tried to show why it was not working. 

When creating the query, odoo adds the relation and that's the problem, it shouldn't create it, but it does :sob:

mail_activity_team/__manifest__.py
pedrobaeza: Please check how I have avoided the use of this module and still get good performance:

https://github.com/OCA/rma/pull/326

pedrobaeza:
@etobella please check my comment about how to avoid the dependency using dynamic domains.

etobella:
The problem with the domain was fixed by using two fields (one is related to the other, but with different domain, this maintains consistency of all work outside of edition)

https://github.com/OCA/social/pull/920/files#diff-0d736b66cbac9f3ba20fd65c4097cdfb3f0c94c0c04606bea2275a017e4e677eR54-R61

Usage of '=?' didn't work. a Many2many field with the possible values does not seem a good idea to me, as there might be a lot of users if no team is selected :thinking: 

Right now, the dependancy is removed

pedrobaeza (approved):

pedrobaeza:
Now it should fw-ported to v15.

PR #924, opened 2022-06-16
[14.0] Add module mail_notification_with_history

TDu:
Followers of a discussion in the chatter of Odoo can be informed by email
of a new message being created.
This module adds in that email the history of the chatter discussion to help
the recipient with more context.

### DIFF
--- a/mail_notification_with_history/models/mail_message.py
+++ b/mail_notification_with_history/models/mail_message.py
@@ -1,0 +1,28 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class Message(models.Model):
+    _inherit = "mail.message"
+
+    def _get_notification_message_history(self):
+        """Get the list of messages to include into an email notification history."""
+        if not self.env[self.model]._mail_notification_include_history:
+            return self.browse()
+        domain = self._get_notification_message_history_domain()
+        messages = self.env["mail.message"].search(domain, order="date desc")
+        return messages - self
+
+    def _get_notification_message_history_domain(self):
+        """Return the domain for email and send message comments."""
+        return [
+            ("model", "=", self.model),
+            ("res_id", "=", self.res_id),
+            "|",
+            "&",
+            ("message_type", "=", "comment"),
+            ("subtype_id", "=", self.env.ref("mail.mt_comment").id),
+            ("message_type", "=", "email"),
+        ]
--- a/mail_notification_with_history/models/mail_thread.py
+++ b/mail_notification_with_history/models/mail_thread.py
@@ -1,0 +1,10 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    _mail_notification_include_history = False
--- a/mail_notification_with_history/__init__.py
+++ b/mail_notification_with_history/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/mail_notification_with_history/__manifest__.py
+++ b/mail_notification_with_history/__manifest__.py
@@ -1,0 +1,17 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{
+    "name": "Mail Notification With History",
+    "summary": """Add the previous chatter discussion into new email notifications.""",
+    "version": "14.0.1.0.0",
+    "category": "Social",
+    "website": "https://github.com/OCA/social",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": ["mail"],
+    "data": ["data/mail_data.xml"],
+    "development_status": "Beta",
+    "maintainers": ["TDu"],
+}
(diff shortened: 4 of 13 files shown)


mail_notification_with_history/models/mail_thread.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    _mail_notification_include_history = False
+
+    @api.model
+    def with_mail_notification_history(self):
+        return self._mail_notification_include_history
```
ivantodorovich:
```suggestion
```

I'd remove this and stick to using the property.
Properties can still be overriden, anyway.

btw, public model methods are exposed through xmlrpc.
Rule of thumb is keep everything private unless you really want this, IMO.

mail_notification_with_history/models/mail_message.py
```diff
@@ -0,0 +1,29 @@
...
+
+class Message(models.Model):
+    _inherit = "mail.message"
+
+    def get_notification_message_history(self):
+        """Get the list of messages to include into an email notification history."""
+        if not self.env[self.model].with_mail_notification_history():
+            return []
+        if not self._with_notification_message_history():
+            return []
```
ivantodorovich:
```suggestion
        if (
            not self.env[self.model].with_mail_notification_history()
            or not self._with_notification_message_history()
        ):
            return self.browse()
```

note that, since you'd normally return a `mail.message` recordset, here we should return an empty recordset for consistency rather than an empty list.

mail_notification_with_history/models/mail_message.py
```diff
@@ -0,0 +1,29 @@
...
+        return messages - self
+
+    def _get_notification_message_history_domain(self):
+        return [
+            ("model", "=", self.model),
+            ("res_id", "=", self.res_id),
+            ("message_type", "in", ("user_notification", "comment", "email")),
+        ]
+
+    def _with_notification_message_history(self):
```
ivantodorovich:
```suggestion
    def _should_include_notification_message_history(self):
```

Just a suggestion. Feel free to ignore it.

IMO `with_*` implies some kind of change in the recordset. Makes me think of `with_company`, `with_user`, `with_prefetch`...

Also probably as a hook point doesn't provide much value because one could simply override `_get_notification_message_history` to achieve the same 😓 so could even remove it

mail_notification_with_history/models/mail_message.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class Message(models.Model):
+    _inherit = "mail.message"
+
+    def get_notification_message_history(self):
```
ivantodorovich:
```suggestion
    def _get_notification_message_history(self):
```
See comment below about public methods

mail_notification_with_history/models/mail_message.py
```diff
@@ -0,0 +1,29 @@
...
+        """Get the list of messages to include into an email notification history."""
+        if not self.env[self.model].with_mail_notification_history():
+            return []
+        if not self._with_notification_message_history():
+            return []
+        domain = self._get_notification_message_history_domain()
+        messages = self.env["mail.message"].search(domain, order="date desc")
```
ivantodorovich, line 15: How about also implementing a limit? It could be important in case there are long discussions
Could be overrideable with a method `_get_notification_message_history_limit` in the `mail.thread` (see comment below about `_get_notification_message_history_domain`)

mail_notification_with_history/models/mail_message.py
```diff
@@ -0,0 +1,29 @@
...
+        return messages - self
+
+    def _get_notification_message_history_domain(self):
```
ivantodorovich, line 18: I'd move this method to `mail.thread`, so that it can be overriden per-model

mail_notification_with_history/models/mail_message.py
ivantodorovich: Although, as per my comment below, I'd go for:

```suggestion
        if (
            not self.env[self.model]._mail_notification_include_history
            or not self._with_notification_message_history()
        ):
            return self.browse()
```

mail_notification_with_history/models/mail_message.py
TDu: changed.

mail_notification_with_history/models/mail_message.py
TDu: Yes for the browse instead of the list.

mail_notification_with_history/models/mail_message.py
TDu: Not sure...

mail_notification_with_history/models/mail_message.py
TDu: Well I removed this one as it is not really needed.

mail_notification_with_history/models/mail_thread.py
TDu: I added that one when I was toying with the mocking in the tests. It is removed

mail_notification_with_history/models/mail_message.py
TDu: About the limit of messages lets see with the use of the module if this becomes a requirement.

ivantodorovich (approved):
Thanks ! LG !

TDu:
@ivantodorovich Thank you for your reviews.
I have slightly change the domain that search for message to include. As to include only the comment that are not internal note. And added a 2nd test.
Also squashed the fixups.

ajaniszewska-dev (approved):

mmequignon (approved):

PR #926, opened 2022-06-27
[12.0] [FIX] mail_tracking: reduce spammy score

eLBati:


sergiocorato (approved):
Tested and it works, thanks!

pedrobaeza (approved):

pedrobaeza:
/ocabot merge minor

Can you please fw-port it to newer versions?

PR #927, opened 2022-06-27
[FIX] mail_tracking: reduce spammy score

eLBati:

### DIFF
--- a/mail_tracking/models/ir_mail_server.py
+++ b/mail_tracking/models/ir_mail_server.py
@@ -13,9 +13,8 @@ class IrMailServer(models.Model):
     def _tracking_headers_add(self, tracking_email_id, headers):
         """Allow other addons to add its own tracking SMTP headers"""
         headers = headers or {}
-        headers['X-Odoo-Database'] = getattr(
-            threading.currentThread(), 'dbname', None),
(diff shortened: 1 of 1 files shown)


mail_tracking/models/ir_mail_server.py
```diff
@@ -49,9 +48,16 @@ def build_email(self, email_from, email_to, subject, body, email_cc=None,
...
+        try:
+            tracking_email_id = int(
+                message.get(
+                    "X-Odoo-MailTracking-ID",
+                    # Deprecated tracking header, kept as fallback
+                    message["X-Odoo-Tracking-ID"],
+                )
+            )
+        except (TypeError, ValueError, KeyError):
+            tracking_email_id = False
```
chienandalu, line 60: `try`/`catch` is more expensive than a simple `if` (at least up to Python 3.11)
```suggestion
        mt_obj = self.env["mail.tracking.email"]
        tracking_email_id = message.get(
            "X-Odoo-MailTracking-ID",
            # Deprecated tracking header, kept as fallback
            message.get("X-Odoo-Tracking-ID", "")
        )
        if not tracking_email_id.isdigit():
            return mt_obj
        return mt_obj.browse(int(tracking_email_id))
```

PR #928, opened 2022-06-28
[14.0][MIG] mail_template_multi_attachment: Migration to 14.0

baimont:
https://github.com/OCA/social/issues/620


mail_template_multi_attachment/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2021 ACSONE SA/NV (<http://acsone.eu>)
```
simahawk: no need for copyright here

mail_template_multi_attachment/models/mail_template.py
```diff
@@ -0,0 +1,60 @@
...
+            return results
+        if isinstance(res_ids, int):
+            multi_mode = False
+            results = {res_ids: results}
+        # Generate attachments (inspired from Odoo); Just add new attachments
+        # into 'attachments' key
+        for res_id, values in results.items():
+            attachments = values.setdefault("attachments", [])
+            for template_report in self.template_report_ids:
+                report_name = self._render_template(
```
simahawk, line 42: not a blocker but would be nice to have a separated smaller method here

mail_template_multi_attachment/views/mail_template.xml
```diff
@@ -0,0 +1,37 @@
...
+<!-- Copyright 2021 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+  <record model="ir.ui.view" id="mail_template_form_view">
+    <field name="name">mail.template.form (in mail_template_multi_attachment)</field>
+    <field name="model">mail.template</field>
+    <field name="inherit_id" ref="mail.email_template_form" />
+    <field name="priority" eval="250" />
+    <field name="arch" type="xml">
+      <field name="report_template" position="replace" />
```
simahawk, line 11: can't we use `move` instead?

simahawk (commented):
LG overall. I'll test on a real instance before final approval.
Anyway, thanks: this is what I was looking for :wink:

mail_template_multi_attachment/views/mail_template.xml
baimont: I tried but I think we can't because we are not in the case
<xpath position="..."><xpath position="move"/></xpath>

baimont:
> LG overall. I'll test on a real instance before final approval. Anyway, thanks: this is what I was looking for wink

Hello @simahawk 

I adapted the pr following your comments. Thanks for your review!

sebalix (approved):

sebalix:
One thing we would like to do in the future is adding a domain on the report line evaluated against the underlying record, to attach or not the report to the e-mail.

rousseldenis (approved):
Code review

richard-willdooit:
@rousseldenis @baimont @sebalix @simahawk @rousseldenis 

Hey guys,

Does this need any extra input for it to be approved?  We are using a fork and would love to just rely on this version.

Richard

cc @jamesw-willdooit

mail_template_multi_attachment/views/mail_template.xml
```diff
@@ -0,0 +1,37 @@
...
+      <field name="report_name" position="replace" />
```
sebalix, line 12: I'm sure there is a way to not replace these two fields :thinking:

PR #930, opened 2022-07-04
 [ADD] mail_activity_fetchmail: Add in 15.0

ikapasi-initos:


ikapasi-initos:
@jans23 FYI

ikapasi-initos:
Hi, @MiquelRForgeFlow  it would be great if you can add reviews here.

dsolanki-initos (commented):
Code review and functional test, LGTM!
Please fix failed checks

pedrobaeza:
I consider a more appropriate name `fetchmail_mail_activity_team_activity`. Why depending on `mail_activity` anyway? I think you can do the same with multiple followers instead of selecting a team and depending on such complex module. In that case, the name would be just `fetchmail_mail_activity`.

dsolanki-initos:
@pedrobaeza, We have our team activity structure so we require a team activity feature. I'll update the module name to fetchmail_mail_activity_team_activity which sounds good.

Thank you for the suggestion

jans23:
@pedrobaeza We performed one review and implemented your change request. Would another review of ours help to get this merged?

pedrobaeza:
I repeat my question:

Why depending on mail_activity anyway? I think you can do the same with multiple followers instead of selecting a team and depending on such complex module.

fkantelberg:
@pedrobaeza 
Mainly 2 reasons:
- activities can be checked when processed which helps to coordinate multiple users without any other communication channel (Have you processed the mail of order xyz?...)
- activity teams are better because the followers might be on vacation. With the team you have a central configuration you can ensure that atleast somebody is available at any time and followers are everywhere. The process can also just be to forward it mail to the next best person in the team (also some follower restriction modules are in use)

I hope this clears the reasoning behind the module

pedrobaeza:
Yes, but that limits the applicability of this module, as it forces to install such complex module that we already know that has been conflicting in the past (specially with permissions), for having the feature. Using standard activities (without team) in a base module, and being expanded to teams with a glue one, would be the best one. Anyway, I don't want to block this, and if the rest consider it OK, then we can merge.

jans23:
@pedrobaeza I think this MR got the required reviews and earlier you approved it's merge. Is there anything else which prevents merging?

pedrobaeza:
Tests are red.

alihashaam:
@pedrobaeza please see that we fixed the tests

pedrobaeza:
Great, thanks. For not blocking more, I'm merging as it's a new module, although technically more reviews are needed.

/ocabot merge nobump

PR #931, opened 2022-07-05
[FIX] mail_autosubscribe: check partners type

JoelZilli:
make sure partner is a tuple


nicomacr (approved):
LGTM

JoelZilli:
We use an if to first see if "partner_id" is a tuple because in this case Odoo tells us that we can only concatenate tuples with tuples. If so, we must cast "follower_ids" as a tuple.
If "partner_ids" does not come as a tuple (i.e. it comes as a list), it is directly concatenated with "follower_ids" which is also a list

JoelZilli:
@yajo could you please review this PR?

yajo (approved):

PR #937, opened 2022-07-13
[14.0][FIX] mail_debrand: fix #884

### LINKED ISSUE #884: [14.0] mail_debrand empty content for sign up
Using mail_debrand, the content of the mail sent at sign up is empty (headers and footer are properly filled) :
```html
    <!-- CONTENT -->
    <tr>
        <td align="center" style="min-width: 590px;">
            <table border="0" cellpadding="0" cellspacing="0" width="590" style="min-width: 590px; background-color: white; padding: 0px 8px 0px 8px; border-collapse:separate;">
                <tr><td valign="top" style="font-size: 13px;">
                    </td></tr>
                <tr><td style="text-align:center;">
                  <hr width="100%" style="background-color:rgb(204,204,204);border:medium none;clear:both;display:block;font-size:0px;min-height:1px;line-height:0; margin: 16px 0px 16px 0px;">
                </td></tr>
            </table>
        </td>
    </tr>
```

Without mail_debrand module installed I get the following content : 
```html
    <!-- CONTENT -->
    <tr>
        <td align="center" style="min-width: 590px;">
            <table border="0" cellpadding="0" cellspacing="0" width="590" style="min-width: 590px; background-color: white; padding: 0px 8px 0px 8px; border-collapse:separate;">
                <tr><td valign="top" style="font-size: 13px;">
                    <div>
                        Cher(e) test,<br/><br/>
                        Vous avez été invité par Admin LE FILAMENT à rejoindre Odoo.
                        <div style="margin: 16px 0px 16px 0px;">
                            <a href="removed_url" style="background-color: #875A7B; padding: 8px 16px 8px 16px; text-decoration: none; color: #fff; border-radius: 5px; font-size:13px;">
                                Accepter l'invitation
                            </a>
                        </div>
                        Votre nom de domaine Odoo est: <b><a href="removed_url">removed_url</a></b><br/>
                        Votre courriel de connection est: <b><a href="removed_url" target="_blank">removed_email</a></b><br/><br/>
                        Jamais entendu parler d'Odoo? C’est un logiciel de gestion tout-en-un apprécié par plus de 3 millions d’utilisateurs. Il améliorera considérablement votre expérience de travail et augmentera votre productivité.
                        <br/><br/>
                        Faites un tour dans la section <a href="https://www.odoo.com/page/tour?utm_source=db&amp;utm_medium=auth" style="color: #875A7B;">Odoo Tour</a> afin de découvrir l'outil.
                        <br/><br/>
                        Enjoy Odoo!<br/>
                        --<br/>L'équipe LE FILAMENT 
                    </div>
                </td></tr>
                <tr><td style="text-align:center;">
                  <hr width="100%" style="background-color:rgb(204,204,204);border:medium none;clear:both;display:block;font-size:0px;min-height:1px;line-height:0; margin: 16px 0px 16px 0px;"/>
                </td></tr>
            </table>
        </td>
    </tr>
```

The issue seems related to this block : https://github.com/OCA/social/blob/02467d14fa4b2ade959ced981f3a0411d46e1d26/mail_debrand/models/mail_render_mixin.py#L51-L52 where the full div is removed because of Odoo tour section.

I am not sure how to address this without breaking the way this module works for other mail templates, any idea would be welcome !

ilyasProgrammer:


francesco-ooops (approved):
functional review ok

elvise (approved):
LGTM!

Shide (commented):
Recreating the template it's not the solution i think

francesco-ooops:
> Recreating the template it's not the solution i think

what would you suggest?

francesco-ooops:
@OCA/social-maintainers what do you say?

rafaelbn:
Maybe an option is to propose a PR to Odoo for fixing the template of the Auth Signup

This PR try to fix https://github.com/OCA/social/issues/884 opend by @remi-filament who propose this workaround https://github.com/OCA/social/issues/884#issuecomment-1088419962

Yenthe666:
Also see https://github.com/OCA/social/issues/940 which has very similar issues.
I'm honestly not convinced by the solution(s) provided as it only fixes this template but would fail over other templates which also generate some "odoo.com" within it. Perhaps there should be some changes made on the `xpath` to always find/get the last match or so to prevent this from happening? The branding content is always at the end either way.

francesco-ooops:
thanks for your feedback @Yenthe666 , would you be able to propose a PR that could be adapted for both issues?

Yenthe666:
> thanks for your feedback @Yenthe666 , would you be able to propose a PR that could be adapted for both issues?

I currently don't have the time on hand I'm afraid so can't look into it much further but wanted to give everyone a heads-up along with a tip.
Just a hint, it should probably be adapted in:
https://github.com/OCA/social/blob/2ea33be8de9e290e6e2504634accaac6d2a1e1a5/mail_debrand/models/mail_render_mixin.py#L24-L50

Yenthe666:
For those of you looking for a solution.. we built a little change on top of the current app which allows us to add regex options to whitelist URL's.
`__manifest__.py`:
```
'data': [
        'data/ir_config_parameter.xml'
    ],
```

`data/ir_config_parameter.xml`:
```
<?xml version="1.0" encoding="UTF-8"?>
<odoo>
    <data noupdate="1">
        <record id="mail_debrand_extra_regex" model="ir.config_parameter">
            <field name="key">mail_debrand.extra_regex_to_skip</field>
            <field name="value">False</field>
        </record>
    </data>
</odoo>
```

`models/mail_render_mixin.py` (def remove_href_odoo):
```
    def remove_href_odoo(
        self, value, remove_parent=True, remove_before=False, to_keep=None
    ):
        if len(value) < 20:
            return value
        # value can be bytes type; ensure we get a proper string
        if type(value) is bytes:
            value = value.decode()
        has_odoo_link = re.search(r"<a\s(.*)odoo\.com", value, flags=re.IGNORECASE)
        extra_regex_to_skip = self.env['ir.config_parameter'].sudo().get_param(
            'mail_debrand.extra_regex_to_skip',
            False
        )
        # value is required field on ir config_parameter, so we have added
        # safety check for "False"
        if has_odoo_link and extra_regex_to_skip != "False" and extra_regex_to_skip:
            # check each regex to be skipped
            for regex in extra_regex_to_skip.split(','):
                if re.search(r'{}'.format(regex), value, flags=re.IGNORECASE):
                    has_odoo_link = False
                    break
        if has_odoo_link:
            # We don't want to change what was explicitly added in the message body,
            # so we will only change what is before and after it.
            if to_keep:
                value = value.replace(to_keep, "<body_msg></body_msg>")
            tree = html.fromstring(value)
            odoo_anchors = tree.xpath('//a[contains(@href,"odoo.com")]')
            for elem in odoo_anchors:
                parent = elem.getparent()
                previous = elem.getprevious()
                if remove_before and not remove_parent and previous is not None:
                    # remove 'using' that is before <a and after </span>
                    previous.tail = ""
                if remove_parent and len(parent.getparent()):
                    # anchor <a href odoo has a parent powered by that must be removed
                    parent.getparent().remove(parent)
                else:
                    if parent.tag == "td":  # also here can be powered by
                        parent.getparent().remove(parent)
                    else:
                        parent.remove(elem)
            value = etree.tostring(
                tree, pretty_print=True, method="html", encoding="unicode"
            )
            if to_keep:
                value = value.replace("<body_msg></body_msg>", to_keep)
        return value
```

You can then modify the "mail_debrand_extra_regex" `ir.config_parameter` to hold something like this:
```
<a\s(.*)dev.odoo\.com
```

or
```
<a\s(.*)yourdomain.odoo\.com
```

ilyasProgrammer:
@Yenthe666 Is it a separate module? Or should i apply changes you offered ?

sswapnesh:
@ilyasProgrammer you can apply it directly

You can also use multiple regex to be skipped with comma separated string 

` <a\s(.*)dev.odoo\.com,<a\s(.*)yourdomain.odoo\.com`

sswapnesh:
@rafaelbn  

> Maybe an option is to propose a PR to Odoo for fixing the template of the Auth Signup
> 
> This PR try to fix #884 opend by @remi-filament who propose this workaround [#884 (comment)](https://github.com/OCA/social/issues/884#issuecomment-1088419962)

Can this (above mentioned) fix adopted in this module ? (By moving ICP to General settings and creating a method for has_odoo_link)

francesco-ooops:
> @rafaelbn
> 
> > Maybe an option is to propose a PR to Odoo for fixing the template of the Auth Signup
> > This PR try to fix #884 opend by @remi-filament who propose this workaround [#884 (comment)](https://github.com/OCA/social/issues/884#issuecomment-1088419962)
> 
> Can this (above mentioned) fix adopted in this module ? (By moving ICP to General settings and creating a method for has_odoo_link)

@rafaelbn what do you say?

rafaelbn:
Hello! I think @sswapnesh point a good way to! By the way, we cannot invest here more time and we uninstalled it from every customer until we will have time to fix it or open OPW to Odoo to fix some template in order to make this easiest...

Sorry we cannot help here more in this moment 😢

francesco-ooops:
> @ilyasProgrammer you can apply it directly
> 
> You can also use multiple regex to be skipped with comma separated string
> 
> ` <a\s(.*)dev.odoo\.com,<a\s(.*)yourdomain.odoo\.com`

@ilyasProgrammer can you do this?

ilyasProgrammer:
@Yenthe666 @sswapnesh code was updated with what Yenthe offered. Now looks like the complete div with Odoo tour and so on is there. Is there something missing ?

sswapnesh:
@ilyasProgrammer  This fix is to by pass custom domain like *.dev.odoo.com 
(This is needed to test on odoo.sh)

You have to keep your solution as well as odoo adds that in <a> tag which is completely removed. (Specific to signup template)

For you solution: Not sure how it can work directly since the main template has noupdate 1

mail_debrand/models/mail_render_mixin.py
```diff
@@ -22,6 +22,19 @@ def remove_href_odoo(
...
+        extra_regex_to_skip = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("mail_debrand.extra_regex_to_skip", False)
+        )
+        # value is required field on ir config_parameter, so we have added
+        # safety check for "False"
+        if has_odoo_link and extra_regex_to_skip != "False" and extra_regex_to_skip:
+            # check each regex to be skipped
+            for regex in extra_regex_to_skip.split(","):
```
tarteo, line 34: Why do you split the regex expression by a comma? You can also just use a `|` right (standard regex operator)?

tarteo, line 38: Why do you split the regex expression by a comma? You can also just use a `|` right (standard regex operator)?

Yenthe666: What does it even matter? 🤷 Also, loads of params/systems work with a split on "," actually.

francesco-ooops:
@ilyasProgrammer can you check why runboat is down?

@Shide @Yenthe666 can you please review? thanks!

mail_debrand/models/mail_render_mixin.py
tarteo: Nothing, just a question what the reason was out of curiosity. It's not blocking.

Yenthe666 (approved):

mail_debrand/__manifest__.py
```diff
@@ -19,7 +19,10 @@
              Odoo Community Association (OCA)""",
     "license": "AGPL-3",
     "installable": True,
-    "depends": ["mail"],
+    "depends": ["mail", "auth_signup"],
```
simahawk: why do you need this new dependency?

mail_debrand/models/mail_render_mixin.py
```diff
@@ -22,6 +22,19 @@ def remove_href_odoo(
         if type(value) is bytes:
             value = value.decode()
         has_odoo_link = re.search(r"<a\s(.*)odoo\.com", value, flags=re.IGNORECASE)
+        extra_regex_to_skip = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("mail_debrand.extra_regex_to_skip", False)
```
simahawk:
```suggestion
            .get_param("mail_debrand.extra_regex_to_skip", 'False')
```

mail_debrand/models/mail_render_mixin.py
```diff
@@ -22,6 +22,19 @@ def remove_href_odoo(
...
+        )
+        # value is required field on ir config_parameter, so we have added
+        # safety check for "False"
+        if has_odoo_link and extra_regex_to_skip != "False" and extra_regex_to_skip:
```
simahawk:
```suggestion
        if has_odoo_link and extra_regex_to_skip and extra_regex_to_skip.strip().lower() != "false":
```

ilyasProgrammer:
updated

simahawk (approved):
@ilyasProgrammer one more thing if you don't mind: please remove the odoo version from the commits -> makes no sense when we fwd/bkp changes. Thanks :pray: 
Other than this we are good to go.

francesco-ooops:
@simahawk merge? :)

PR #941, opened 2022-07-27
[14.0][mail_tracking] [IMP] Add option to keep aliases in mail tracking

cvinh:
Use case :
- setup an alias (ie info@domain.com) to an object
1- send an email to info@domain.com and cc someone else
2- send an email to someone else and cc info@domain.com

In the chatter, we want to include the alias to know if the alias was was email_to or email_cc because the user has not the same priority to answer (if the alias is email_to --> the user must reply, if the alias is email_cc --> the user has no action)

Here are screenshots after this PR (alias is setup with crm.team)

Test_email_to_alias
Test_email_cc_alias

### DIFF
--- a/mail_tracking/models/mail_message.py
+++ b/mail_tracking/models/mail_message.py
@@ -217,6 +217,12 @@ class MailMessage(models.Model):
     @api.model
     def _drop_aliases(self, mail_list):
         aliases = self.env["mail.alias"].get_aliases()
+        if self.env.company.mail_tracking_show_aliases:
+            IrConfigParamObj = self.env["ir.config_parameter"].sudo()
+            aliases = "{}@{}".format(
+                IrConfigParamObj.get_param("mail.catchall.alias"),
+                IrConfigParamObj.get_param("mail.catchall.domain"),
+            )
 
         def _filter_alias(email):
             email_wn = getaddresses([email])[0][1]
--- a/mail_tracking/models/res_company.py
+++ b/mail_tracking/models/res_company.py
@@ -1,0 +1,10 @@
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    mail_tracking_show_aliases = fields.Boolean(
+        string="Show Aliases in Mail Tracking",
+        default=False,
+    )
--- a/mail_tracking/tests/test_mail_tracking.py
+++ b/mail_tracking/tests/test_mail_tracking.py
@@ -157,6 +157,37 @@ class TestMailTracking(TransactionCase):
         self.assertEqual(tracking_email.error_type, "no_recipient")
         self.assertFalse(self.recipient.email_bounced)
 
+    def test_message_post_show_aliases(self):
+        # Create message with show aliases setup
+        self.env.company.mail_tracking_show_aliases = True
+        # Setup catchall domain
+        IrConfigParamObj = self.env["ir.config_parameter"].sudo()
+        IrConfigParamObj.set_param("mail.catchall.domain", "test.com")
+        # pylint: disable=C8107
+        message = self.env["mail.message"].create(
+            {
+                "subject": "Message test",
+                "author_id": self.sender.id,
+                "email_from": self.sender.email,
+                "message_type": "comment",
+                "model": "res.partner",
+                "res_id": self.recipient.id,
+                "partner_ids": [(4, self.recipient.id)],
+                "email_cc": "Dominique Pinon <unnamed@test.com>, customer-invoices@test.com",
+                "body": "<p>This is another test message</p>",
+            }
+        )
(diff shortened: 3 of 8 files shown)


cvinh:
Dear PSC Members, is it possible to have a review here ?
Thanks in advance
@OCA/social-maintainers

mail_tracking/models/mail_message.py
```diff
@@ -215,7 +215,10 @@ def tracking_status(self):
 
     @api.model
     def _drop_aliases(self, mail_list):
-        aliases = self.env["mail.alias"].get_aliases()
+        if not self.env.user.company_id.show_aliases_mail_tracking:
```
chienandalu:
```suggestion
        if not self.env.company.show_aliases_mail_tracking:
```

chienandalu (commented):
Can you add a test case?

mail_tracking/models/mail_message.py
cvinh: done

cvinh:
> Can you add a test case?

Test case added

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -157,6 +157,34 @@ def test_message_post_partner_no_email(self):
...
+                "body": "<p>This is another test message</p>",
+            }
+        )
+        message._moderate_accept()
+        message_dict = message.message_format()[0]
+        foundAlias = False
+        for tracking in message_dict["partner_trackings"]:
+            if tracking["recipient"] == "customer-invoices@test.com":
+                foundAlias = True
+        self.assertTrue(foundAlias)
```
chienandalu:
```suggestion
        message_dict, *_ = message.message_format()
        self.assertTrue(
            any([tracking["recipient"] == "customer-invoices@test.com" for tracking in message_dict])
        )
```

chienandalu (approved):
Just a little formal suggestion

mail_tracking/models/res_company.py
```diff
@@ -0,0 +1,10 @@
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    show_aliases_mail_tracking = fields.Boolean(
```
pedrobaeza: Please use a better nomenclature with precedence to the "section" (module) and avoid plurals:

```suggestion
    mail_tracking_show_alias = fields.Boolean(
```

pedrobaeza:
Can you visually present the use case as I don't full get it? Maybe this shouldn't be an option and be the same behavior for all.

mail_tracking/models/res_company.py
pedrobaeza: I see that is still keeping the plural. Better to not include it for avoiding typos on development.

cvinh: done

mail_tracking/tests/test_mail_tracking.py
cvinh: done

mail_tracking/models/res_company.py
pedrobaeza: Thanks. It seems there's an error in CI. And what about the use case expanded info for judging if we should do it the standard?

cvinh:
> Can you visually present the use case as I don't full get it? Maybe this shouldn't be an option and be the same behavior for all.

@pedrobaeza thank you for you review
You are right, for us, the alias must be shown in all cases. I did another PR here https://github.com/OCA/social/pull/1064
You might choose which PR you prefer.
@Tardo as you are the author of this commit, can you tell us why you removed the aliases in tracking https://github.com/OCA/social/commit/ffb4a5b83da542e07a94c5387c455da585b707db ?

pedrobaeza:
I have already put an explanation of why dropping aliases that motivated such code in the other PR, so maybe we should just do a smarter dropping.

cvinh:
> I have already put an explanation of why dropping aliases that motivated such code in the other PR, so maybe we should just do a smarter dropping.

@pedrobaeza finally I did not modify the get_aliases method because of your advice, I add aliases only in tracking_status method

pedrobaeza:
Then should we close this PR in favor of the other?

cvinh:
> Can you visually present the use case as I don't full get it? Maybe this shouldn't be an option and be the same behavior for all.

I added screenshots in description

cvinh:
> Then should we close this PR in favor of the other?

This PR includes a setup to let the users choose the behavior. The other PR does not allow choice.
We can wait for @chienandalu @Tardo 
(On our side, we vote for the other PR)

chienandalu:
I'd go for the behavior choice

pedrobaeza (approved):
OK, then let's go with this one:

/ocabot merge minor

PR #942, opened 2022-07-27
[14.0][IMP] mail_tracking: remove legacy hooks

StephaneMangin:
Computes used to process tracking information are no longer store in database. Thus, the hook computing default values are not used anymore.

Removing it !


leemannd (approved):

leemannd:
Open question, should it be added a pre-migration script that will delete the columns in db if we migration from a previous version?

pedrobaeza:
No. In fact, OpenUpgrade doesn't remove any column. If you want to do that, use `database_cleanup` module after.

pedrobaeza:
Please follow commit message guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

`[IMP] mail_tracking: remove legacy hooks`

grindtildeath (approved):

pedrobaeza:
I think that the compute can be also optimized using `read_group` instead of individual `search_count`, but that's another story. Thanks for this cleanup:

/ocabot merge patch

I suppose the same cleanup can be done in v15.

PR #943, opened 2022-07-27
[ADD] mail_chatter_thread_colour

etobella:
With this module, users might see easily if a chat comes from a chat or from a record.

We can even define a color for a model :smile: 

Peek 2022-07-27 16-59


DavidJForgeFlow (approved):
image
Functional review :+1: code also LGTM!

etobella:
Added the option to change the color of the font too (good in some cases) @LoisRForgeFlow

LoisRForgeFlow (approved):
Much better now. Functional review LGTM :+1: 

image

PR #944, opened 2022-08-01
[15.0] [MIG]mass_mailing_unique: Migration to 15.0

XanderDeJaegere:


StefanRijnhart:
Thanks! Can you squash the two migration commits, and suppress your `prepare-commit-msg`(?) hook so that we don't get the `[15.0-mig-mass]` prefix in the commit message?

StefanRijnhart:
Ping @XanderDeJaegere, please check my last comment.

PR #946, opened 2022-08-01
[15.0][FIX] mail_activity_board: read access to ir.model for non admin

AaronHForgeFlow:
Give read access to ir.model for non admin users when accessing records that inherit mail.activity mixing. Otherwise those users will get access error.

Steps to reproduce: Just access a manufacturing order with a non admin user. An access error to ir.model model will be thrown.

@ForgeFlow


AaronHForgeFlow:
@etobella  do you think this solution is fine?

etobella (approved):

LoisRForgeFlow (approved):

PR #329, opened 2018-11-26
[ADD][mail_parent_recipient][10.0] Add module to use mail of first partner

Tonow-c2c:
New module **Mail parent recipient**

This modules add the fact to use partner's parent email as recipient.

PR #963, opened 2022-09-13
[10.0][ADD] mail_parent_recipient

TDu:
This is an attempt to resolve the comments on #329

### DIFF
--- a/mail_parent_recipient/models/company.py
+++ b/mail_parent_recipient/models/company.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# Copyright 2022 Camptocamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+
+    _inherit = "res.company"
+
+    use_parent_mail_address = fields.Boolean(
+        string="Use Parent Mail Address",
+        help="When an email is sent, fallback to partner's parent email if no "
+        "email is set on the recipient partner."
+    )
--- a/mail_parent_recipient/models/mail_template.py
+++ b/mail_parent_recipient/models/mail_template.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models, api
+
+
+class MailTemplate(models.Model):
+
+    _inherit = 'mail.template'
+
+    @api.multi
+    def generate_recipients(self, results, res_ids):
+        """Use partner's parent email as recipient.
+
+        Walk up the hierarchy of recipient partners via `parent_id`
+        and pick the 1st one having an email.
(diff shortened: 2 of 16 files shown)


mail_parent_recipient/tests/test_mail_parent_recipient.py
```diff
@@ -0,0 +1,120 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.addons.mail.tests.common import TestMail
+
+
+class TestMailTemplate(TestMail):
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestMailTemplate, cls).setUpClass()
```
simahawk, line 12: tracking_disable=True at least or creating these partners :pray:

mail_parent_recipient/models/mail_template.py
```diff
@@ -0,0 +1,45 @@
...
+        Walk up the hierarchy of recipient partners via `parent_id`
+        and pick the 1st one having an email.
+        """
+        results = super(MailTemplate, self).generate_recipients(
+            results,
+            res_ids
+        )
+        test_condition = (
+            config['test_enable'] and not self.env.context.get(
+                'test_parent_mail_recipient'
```
simahawk: why do we need this flag?

simahawk (changes requested):

mail_parent_recipient/models/mail_template.py
TDu: It is to enable the machinery of this module during test ONLY when the 'test_parent_mail_recipient'  is set.

simahawk: This is not answering the main question: why do we need this flag?
I guess is to not make other tests fail.
IMO this should be configured either at company level (off by default) or at partner level (off by default).
We can start adding a flag at company level 1st + res.config.settings then we'll see.

simahawk:
I would try to move to GH actions on 10.0 branch first.

simahawk:
@TDu please move the update dotfiles to another PR

simahawk:
#972 is getting merged. You can rebase soon :tada:

mail_parent_recipient/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Thomas Nowicki <thomas.nowicki@camptocamp.com>
```
simahawk, line 1: @TDu you can add your name here

mail_parent_recipient/models/company.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+
+    _inherit = "res.company"
+
+    use_parent_mail_address = fields.Boolean(
+        string="Use Parent Mail Address",
+        help=""
```
simahawk:
```suggestion
        help="When an email is sent, fallback to partner's parent email if no email is set on the recipient partner."
```

simahawk:
we'll maintain only from v15

PR #966, opened 2022-09-16
[13.0][FIX] mail_tracking: permissions

chienandalu:
Forwarding

- #759

It'd be nice to add some tests though.

**Rationale:**

Refine which mail tracking record is allowed to see. We can rely on the ACLs of the related records so we minimize inconvenient message subject leaks.

A regular user can read these mail.tracking.email records:

- Those with a linked `mail.message` that the user can read
- Those with a linked `mail.mail` that the user can read
- Those with *no* message/mail link but a linked *partner* that the user can read.
- Those with no linked records.

cc @Tecnativa TT31399

### DIFF
--- a/mail_tracking/controllers/main.py
+++ b/mail_tracking/controllers/main.py
@@ -81,8 +81,10 @@ class MailTrackingController(MailController):
         metadata = self._request_metadata()
         with db_env(db) as env:
             try:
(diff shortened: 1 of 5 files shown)


pedrobaeza (approved):

pedrobaeza:
@chienandalu please fw-port it to upper versions.

PR #972, opened 2022-09-30
[10.0] Use GitHub action

TDu:


requirements.txt
```diff
@@ -1,2 +1,3 @@
-premailer==3.6.2
-sendgrid<v6.0.0
+# generated from manifests external_dependencies
+premailer
```
simahawk: Check this commit, we need a specific version https://github.com/OCA/social/commit/36703f5d4b0ce45c585c6eefdb94279ff6de388e

simahawk: ah, in this case it must be set on the manifest of the module

simahawk: and if you look on pypi you see that https://pypi.org/project/premailer/3.6.2/ is the last one supporting py2.
The problem on runboat seems to be related to this too (it installs 3.10)

TDu:
@rousseldenis @sebalix  This is the one you helped me with during the oca days.
It fixes the build for the branch 10.0. I had to disable a few tests that did not pass. Being a version 10 I did not want to spent more time fixing this.

rousseldenis:
@TDu So, this is working. I think it's acceptable as is. @simahawk

rousseldenis (approved):

rousseldenis:
@TDu Could you just squash some commits if they can be merged with each other?

TDu:
I squashed the commit related to fixing the python dependencies.
@rousseldenis Thank you again for your help on this

PR #977, opened 2022-10-03
[14.0][FIX] mail_quoted_reply: avoid access right errors when reading action.

LoisRForgeFlow:
Otherwise non-admin users cannot user the feature.

@ForgeFlow


etobella (approved):
LGTM

/ocabot merge patch

PR #856, opened 2022-03-01
[MIG][14.0] mail_notify_bounce

SimoneVagile:

PR #981, opened 2022-10-19
[14.0][MIG] mail_notify_bounce

tafaRU:


yajo:
This module seems to be unnecessary in v12+. Are you sure we need this migration?

tafaRU:
@yajo it seems so in the face of https://github.com/OCA/social/pull/856#issuecomment-1057855444.

yajo:
It seems to me that use case could be fixed by just creating a "bounce" mailing list and subscribing the odoo inbound mail address to it, along with any other mail addresses that should be notified of bounces.

I guess that would render this module unnecessary. Of course maybe some people can't or don't want to create that mailing list, or maybe this idea simply doesn't work because of something I'm missing.

PR #982, opened 2022-10-19
[15.0][MIG] mail_notification_custom_subject: Migration to 15.0

victoralmau:
Migration to 15.0

Please @CarlosRoca13 and @chienandalu can you review it?

@Tecnativa TT36470


CarlosRoca13 (approved):

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -0,0 +1,78 @@
...
+        parent_id=False,
+        subtype_xmlid=None,
+        subtype_id=False,
+        partner_ids=None,
+        attachments=None,
+        attachment_ids=None,
+        add_sign=True,
+        record_name=False,
+        **kwargs
+    ):
```
chienandalu, line 29: Simpler:
```suggestion
    def message_post(self, **kwargs):
```

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -0,0 +1,78 @@
...
+            parent_id=parent_id,
+            subtype_xmlid=subtype_xmlid,
+            subtype_id=subtype_id,
+            partner_ids=partner_ids,
+            attachments=attachments,
+            attachment_ids=attachment_ids,
+            add_sign=add_sign,
+            record_name=record_name,
+            **kwargs,
+        )
```
chienandalu, line 78:
```suggestion
        return super().message_post(**kwargs)
```

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -21,15 +21,14 @@ def message_post(
...
         subtype_id=False,
         partner_ids=None,
-        channel_ids=None,
         attachments=None,
         attachment_ids=None,
         add_sign=True,
         record_name=False,
         **kwargs
     ):
         if not subtype_id and subtype_xmlid:
```
chienandalu, line 30:
```suggestion
        subtype_id = kwargs.pop("subtype_id", False)
        subtype_xmlid = kwargs.pop("subtype_xmlid", False)
        if not subtype_id and subtype_xmlid:
```

chienandalu (commented):
Tested functionaly. Some formal comments:

victoralmau:
> Tested functionaly. Some formal comments:

Changes done, but it seems to have side effects on other modules. Any suggestions to solve it?

pedrobaeza:
You can't do that, as not the whole inheritance chain is respecting keyword arguments, so better to stick to the whole form.

victoralmau:
Changes done (last changes undone)

chienandalu:
> You can't do that, as not the whole inheritance chain is respecting keyword arguments, so better to stick to the whole form.

Odoo's choice, not mine :smile: :

- https://github.com/odoo/odoo/blob/15.0/addons/rating/models/mail_thread.py#L11
- https://github.com/odoo/odoo/blob/15.0/addons/hr/models/hr_employee.py#L457
- https://github.com/odoo/odoo/blob/15.0/addons/purchase/models/purchase.py#L412
- ...

pedrobaeza:
Well, that ones probably will raise an error with this same chain... Maybe it's a problem in one OCA module, but that form requires that you always call super with all the keywords.

PR #988, opened 2022-11-02
[FIX] mail_activity_board: use sudo to read ir action window

jonerikceberio:
Self-explanatory, just in case: go to runboat, enter as demo, go to any contact, press the Activities access in the chatter.

### DIFF
--- a/mail_activity_board/models/mail_activity.py
+++ b/mail_activity_board/models/mail_activity.py
@@ -52,7 +52,9 @@ class MailActivity(models.Model):
 
     @api.model
     def action_activities_board(self):
-        action = self.env.ref("mail_activity_board.open_boards_activities").read()[0]
+        action = (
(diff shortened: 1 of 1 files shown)


mail_activity_board/models/mail_activity.py
```diff
@@ -52,7 +52,9 @@ def open_origin(self):
 
     @api.model
     def action_activities_board(self):
-        action = self.env.ref("mail_activity_board.open_boards_activities").read()[0]
+        action = (
+            self.env.ref("mail_activity_board.open_boards_activities").sudo().read()[0]
+        )
```
ivantodorovich, line 57:
```suggestion
        action = self.env["ir.actions.actions"]._for_xml_id(
            "mail_activity_board.open_boards_activities"
        )
```

FTR, this also does the trick and it's slightly better security wise.

ivantodorovich (approved):

PR #991, opened 2022-11-07
[15.0] [ADD] `mail_activity_security`

ivantodorovich:
This module allows you to configure, for each Activity Type, the security rules that will be checked to executes actions on the activity.

The following actions are configurable independently:

* Mark as done
* Edit
* Cancel

For each action, you can configure the following security levels:

* Creator: The user who created the activity
* User: The user who's assigned to the activity
* Anyone: Any user with write access to the activity (Odoo default)

These levels are nested. For example, if you set the security level to user, both the User and the Creator will have access to the action.

Additionally, you can specificy a list of User Groups that will always have access to all the actions of the activities. This is useful to effectively give "activity admin access" to the Management groups.


ivantodorovich:
@OCA/social-maintainers , can I bother you with reviews or "no stale" label? 🙏🏻

NL66278 (approved):
👍 LGTM

mail_activity_security/models/mail_activity.py
```diff
@@ -0,0 +1,84 @@
...
+                raise AccessError(
+                    _("You are not allowed to mark this activity as done.")
+                )
+            if operation == "cancel":
+                raise AccessError(_("You are not allowed to cancel this activity."))
+            raise AccessError(_("You're not allowed to perform this operation."))
+        return True
+
+    def _check_activity_security(self, operation: str) -> bool:
+        self.ensure_one()
```
simahawk, line 28: TBH I don't understand the use of `self.ensure_one` for methods that are not part of a public api and that are used only in our own loops, like in this case. What's the point of checking all the times that you have one record?

simahawk (approved):
LG overall

mail_activity_security/models/mail_activity.py
NL66278: I must say I use `self.ensure_one` a lot myself in these cases. It documents quite clearly that the method should be called from a record set that contains one and only one record. This will help in the proper usage of these methods by other programmers.

simahawk: I understand that but it adds a check that is useless on a method that is not meant to be used as a public api, hence by other developers that do not know what they are doing. Just my $0.02 of course :)

PR #996, opened 2022-11-08
[12.0][UPD] switch to github actions

thomaspaulb:
Switch this repo's 12.0 branch to using Github Actions, thereby making it green again


.copier-answers.yml
```diff
@@ -0,0 +1,23 @@
+# Do NOT update manually; changes here will be overwritten by Copier
+_commit: v1.11.0
+_src_path: git+https://github.com/oca/oca-addons-repo-template
+ci: GitHub
+dependency_installation_mode: PIP
+generate_requirements_txt: false
```
sbidoul: Hi @thomaspaulb I wouldn't disable this. If there is an issue with that it's likely revealing a issue with external dependencies that needs to be fixed.

thomaspaulb: I disabled because it showed me this:

```
-# mail_inline_css
+# generated from manifests external_dependencies
+extract_msg
 premailer
-extract_msg<0.30
```

The addition of `mail_inline_css` and the pinning of `extract_msg` looked custom to me, so I thought it should be kept. Should I instead change it in the appropriate addon?

hbrunn:
looks like you're just one `pre-commit run -a` away from having this green

.copier-answers.yml
sbidoul: > Should I instead change it in the appropriate addon

Yes, absolutely. Also, since this is Odoo 12, we can't put version specifiers in manifest external_dependencies, so we need to override in setup.py (see https://github.com/acsone/setuptools-odoo#controlling-setuptools-odoo-behaviour).

thomaspaulb:
@hbrunn It was a bit more than that but it looks like I made it

.copier-answers.yml
thomaspaulb: Done

sbidoul (approved):

hbrunn (approved):
thanks

PR #999, opened 2022-11-14
[MIG] mail_attach_existing_attachment_account: Migration to 15.0

nicomacr:


rafaelbn:
Hello @nicomacr , I've tested v15 and this module is not needed. I can use mail_attanch_existing_attachment_account in v15 and is fully working with account without this module.

Have you tested? Maybe I'm missing something, but in this moment I think this module could be deprecated.

Let me know! 😄 

Thank you!

PR #1001, opened 2022-11-18
[ADD]  mail_post_defer: deferred message posting with queue

yajo:
This module enhances mail threads by using the mail queue by default.

Without this module, Odoo attempts to notify recipients of your message immediately.
If your mail server is slow or you have many followers, this can mean a lot of time.
Install this module and make Odoo more snappy!

All emails will be kept in the outgoing queue by at least 30 seconds,
giving you some time to re-think what you wrote. During that time,
you can still delete the message and start again.

## Usage

1. Go to the form view of any record that has a mail thread. It can be a partner, for example.
1. Post a message.

The mail is now in the outgoing mail queue. It will be there for at least 30
seconds. It will be really sent the next time the "Mail: Email Queue Manager"
cron job is executed.

While the message has not been yet sent:

1.  Hover over the little envelope. You will see a paper airplane icon,
    indicating it is still outgoing.
1.  Hover over the message and click on the little trash icon to delete it.
    Mails will not be sent.

## Configuration

You need to do nothing. The module is configured appropriately out of the box.

However, please notice that the mail queue processing is made by a cron job.
This is normal Odoo behavior, not specific to this module. However, since you
will start using that queue for every message posted by any user in any thread,
you will probably want to increase the that job's frequency.

To do so:

1. Log in with an administrator user.
1. Activate developer mode.
1. Go to *Settings > Technical > Automation > Scheduled Actions*.
1. Edit the action named "Mail: Email Queue Manager".
1. Lower down the frequency in the field *Execute Every*. Recommended: 1 minute.

@moduon MT-1579 MT-2480


rafaelbn (approved):
Functionality 👍

rafaelbn (commented):
Works and solve the problem but a second UX-review I don't see clearly the UX, because, this would be problematic for user. We should think about that.

rafaelbn (approved):
👍

Shide:
Will be interesting if defer time (30 sec on code) could be parametrized?

pedrobaeza:
The base seems similar to https://github.com/OCA/queue/pull/119

yajo:
> Will be interesting if defer time (30 sec on code) could be parametrized?

Makes sense, although the real time for sending an email will be more about 1:00-1:30 minutes minimum. This defer time is to guarantee that, if you're running the mail queue cron at its minimal cadence (once per minute) and let's say you send a message 3 seconds before it's executed, you always have *at least* 30s of defer.

So, even if you parametrize it, there's no guarantee that messages will be sent in that time.

Also, everybody loves this feature in Gmail and other mail clients, but I have never seen an option to configure the deferring time. At most, I see the option to enable it or not. So, chances are that real world users never need to configure this. That's why I decided to hardcode the value.

IMHO we can keep it as it is and support the param later if we really see it's needed.



> The base seems similar to [OCA/queue#119](https://github.com/OCA/queue/pull/119)

Thanks, I didn't know that module! However IIUC both modules would be complementary. `mail_queue_job` replaces the mail queue cron for a queue job implementation, which is probably more snappy and reliable. This one instead:

- Makes messages sent from the chatter deferred.
- Lets the user undo queued messages.

If you install both, you probably can have the best of both worlds, but I don't see any feature overlap or interdependency.

I liked from that one the init hook. I think I'll add one here, because when you install this module, chances are you want to have the mail queue cron as fast as possible.

yajo:
Install hook added. I also added @Shide's suggestion to roadmap, just to keep track of that idea.

Shide (approved):
Code LGTM
Functionally Reviewed: Direct message to other user doesn't wait 30s.

PR #1009, opened 2022-12-01
[14.0][IMP] mail_activity_team: Change sudo() to with_user()

victoralmau:
Change sudo() to with_user()

Remove warning: `deprecated use of sudo(user), use with_user(user) instead`

@Tecnativa


pedrobaeza (approved):
Straight-forward:

/ocabot merge patch

PR #1010, opened 2022-12-01
[14.0][IMP] mail_activity_done: Change assertEquals to assertEqual

victoralmau:
Change assertEquals to assertEqual

@Tecnativa


pedrobaeza (approved):

pedrobaeza:
They can be put in the same commit, but let's simply fix it:

/ocabot merge patch

PR #1011, opened 2022-12-01
[13.0][FIX] mail_activity_done: patch _action_done in a more reliable way

hbrunn:
I don't think we need the monkey patch at all, so the resulting code can be much simpler it seems.

The monkey patch also doesn't play well with inheritance

### DIFF
--- a/mail_activity_done/models/mail_activity.py
+++ b/mail_activity_done/models/mail_activity.py
@@ -2,6 +2,8 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 from odoo import api, fields, models
 
+delete_sentinel = object()
+
 
 class MailActivity(models.Model):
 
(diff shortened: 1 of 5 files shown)


mail_activity_done/models/mail_activity.py
```diff
@@ -18,6 +20,18 @@ def _compute_state(self):
         for record in self.filtered(lambda activity: activity.done):
             record.state = "done"
 
+    def unlink(self):
```
SkiBY, line 23: Sorry, but why we can't use simply a True flag for it? Or even a context check for something like

    def unlink(self)
        """Don't unlink if we're asked not to"""
        if not self.env.context.get("mail_activity_done_protect"):
            return super().unlink()

    def _action_done(self, feedback=False, attachment_ids=None):
        """Ask super not to delete the activity and set it to done"""
        self.write({"done": True, "active": False, "date_done": fields.Date.today()})
        return super(
            MailActivity, self.with_context(mail_activity_done_protect=True,
        )._action_done(feedback=feedback, attachment_ids=attachment_ids)

hbrunn: everything that can be json-encoded can be injected in the context by a malicious user, so if we want to be sure that stuff happens because we want to, i don't have a better idea than the sentinel

SkiBY (approved):

SkiBY:
[@JakubWiselka ](https://github.com/JakubWiselka) could you take a look?

PR #1016, opened 2022-12-02
[FIX] mail_notification_custom_subject: do not require being admin to send messages

yajo:
After installing the module, any user was required to have admin rights to be able to search across models.

This fix does just that search with sudo and returns back to non-sudo mode immediately after. This way, we don't need to increase permissions for everybody.

Tests now run unprivileged, to assert this problem never happens again.

@moduon MT-1644 MT-1645

### DIFF
--- a/mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
+++ b/mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
@@ -20,16 +20,22 @@ class TestMailNotificationCustomSubject(common.TransactionCase):
                 ]
             )
         )
+        cls.admin = common.new_test_user(cls.env, "boss", "base.group_system")
+
+    def setUp(self):
+        super().setUp()
+        self.uid = common.new_test_user(self.env, "worker")
 
     def test_email_subject_template_overrides(self):
-        self.env["mail.message.custom.subject"].create(
-            {
-                "name": "Test template 1",
-                "model_id": self.env.ref("base.model_res_partner").id,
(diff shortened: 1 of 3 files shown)


yajo:
@joao-p-marques I guess you don't want to maintain this anymore, so I changed it here.

rafaelbn (approved):
👍 

This is urgent, block the outgoing emails

ghost (approved):
LGTM. Tested with a user with no permits, and could send a PDF, no problem at all.

mail_notification_custom_subject/tests/test_mail_notification_custom_subject.py
```diff
@@ -52,14 +58,15 @@ def test_email_subject_template_overrides(self):
...
     def test_email_subject_template_normal(self):
-        self.env["mail.message.custom.subject"].create(
-            {
-                "name": "Test template 1",
-                "model_id": self.env.ref("base.model_res_partner").id,
-                "subtype_ids": [(6, 0, [self.env.ref("mail.mt_comment").id])],
-                "subject_template": "{{object.name or 'n/a'}} and something more",
-            }
-        )
+        with self.with_user("boss"):
```
pedrobaeza, line 61: You can use `@users` decorator to mimic this, and avoid the big diff.

pedrobaeza:
Although Joao is no longer at Odoo world, Tecnativa is assuming the maintenance of the module, but well, if you want to put yourself, it's not a problem while you do migrations and so on.

PR #1018, opened 2022-12-02
[FIX] mail_show_follower: failure when partner has more than 1 user

yajo:
In the highly improbable but actually possible and real world case that a partner has more than one user associated, this change makes the emails able to send.

Upstream fix proposal: https://github.com/odoo/odoo/pull/107093

@moduon MT-1634


yajo:
Fast-tracking.

/ocamod merge patch

yajo:
Fast-tracking.

/ocabot merge patch

PR #1027, opened 2022-12-14
[15.0] [FIX] mail_notification_custom_subject: Use same approach as Odoo when selecting subtype_id

Shide:
If an user sends a internal message expanding the editor, the message sent doesn't have any `subtype_xml_id` nor `subtype_id`.

Using the [same approach as Odoo](https://github.com/odoo/odoo/blob/15.0/addons/mail/models/mail_thread.py#L1830-L1833) does selecting the `subtype_id`, fallbacks on `mail.mt_note` type, so the module will try to use a `mail.message.custom.subject` record.

MT-134 @moduon @rafaelbn @yajo


yajo (approved):

yajo:
/ocabot merge patch

Obvious fix.

PR #1035, opened 2023-01-02
[16.0][MIG] mail_outbound_static: Migration to 16.0

dreispt:


mail_outbound_static/models/ir_mail_server.py
```diff
@@ -0,0 +1,146 @@
...
+            bounce_alias = (
+                self.env["ir.config_parameter"].sudo().get_param("mail.bounce.alias")
+            )
+            if not bounce_alias:
+                # then, bounce handling is disabled and we want
+                # Return-Path = From
+                if "Return-Path" in message:
+                    message.replace_header("Return-Path", email_from)
+                else:
+                    message.add_header("Return-Path", email_from)
```
yostashiro, line 113: Update of smtp_from and message values should be done by extending _prepare_email_message(), or the updated values get overriden in there.  I suppose the same issue exists in 15.0.

yostashiro (changes requested):
@dreispt Thanks for the PR.  Can you please check this commit https://github.com/qrtl/social/commit/196f6b3499f3c55aa6da326cea958c2deb2be632 which I added on top of yours, and cherry-pick this as appropriate?

CasVissers-360ERP (approved):
We have it running for some months now, all good.

adrienpeiffer:
Hi @dreispt 
I have an issue related with the smtp_sesssion which is set here 

https://github.com/odoo/odoo/blob/439bef710b78afe0e5ba1d54ce04a5e5c3906cdc/addons/mail/models/mail_mail.py#L397

The content of the FROM is replaced by the value on the session here :

https://github.com/odoo/odoo/blob/15aa0e5025e35a5093bd70cae0513cb492806dcc/odoo/addons/base/models/ir_mail_server.py#L602 

Do you have this kind of issue ?

yostashiro:
@adrienpeiffer I believe https://github.com/OCA/social/pull/1051 will fix the issue you mention.

mail_outbound_static/models/ir_mail_server.py
```diff
@@ -0,0 +1,146 @@
...
+            domain_whitelist = self._get_domain_whitelist(item.domain_whitelist)
+            if email_domain in domain_whitelist:
+                mail_server_id = item.id
+                break
+        if not mail_server_id:
+            mail_server_id = self.sudo().search([], order="sequence", limit=1).id
+        return mail_server_id
+
+    @api.model
+    def create(self, values):
```
adrienpeiffer: You should use 
```
@api.model_create_multi
def create(self, vals_list):
```

adrienpeiffer (changes requested):

mail_outbound_static/models/ir_mail_server.py
dreispt: Done

mail_outbound_static/models/ir_mail_server.py
dreispt: @yostashiro It is best if you forward port #1051 to 16.0

dreispt:
Fixed and rebased.
Please re-review.

hbrunn (approved):

PR #1037, opened 2023-01-11
[16.0][MIG] mail_restrict_follower_selection: Migration to 16.0

AungKoKoLin1997:
@qrtl
In this migration , there is a issue in adding domain for `partner_ids` in adding new followers. We need to refresh for each document model.
Related issue: https://github.com/odoo/odoo/issues/109601


mail_restrict_follower_selection/models/mail_wizard_invite.py
```diff
@@ -0,0 +1,43 @@
...
+    @api.model
+    def get_view(self, view_id=None, view_type="form", **options):
+        result = super(MailWizardInvite, self).get_view(
+            view_id=view_id, view_type=view_type, **options
+        )
+        arch = etree.fromstring(result["arch"])
+        for field in arch.xpath('//field[@name="partner_ids"]'):
+            domain = self._mail_restrict_follower_selection_get_domain()
+            domain = ast.literal_eval(domain)
+            domain.append(("type", "!=", "private"))
```
yostashiro: Can we get this domain from `self._fields["partner_ids"].domain` (and use expression.AND(), perhaps)?

I think this domain manipulation should be done inside _mail_restrict_follower_selection_get_domain() instead of here.

mail_restrict_follower_selection/models/mail_wizard_invite.py
```diff
@@ -0,0 +1,47 @@
...
+        return domain
+
+    @api.model
+    def get_view(self, view_id=None, view_type="form", **options):
+        result = super(MailWizardInvite, self).get_view(
+            view_id=view_id, view_type=view_type, **options
+        )
+        arch = etree.fromstring(result["arch"])
+        for field in arch.xpath('//field[@name="partner_ids"]'):
+            field.attrib["domain"] = str(
```
yostashiro, line 43: @AungKoKoLin1997 Do we need this str conversion?

AungKoKoLin1997: Yes. We need str conversion because we can't assign list to domain attribute.

mail_restrict_follower_selection/models/mail_thread.py
```diff
@@ -0,0 +1,30 @@
...
+        )
+        if test_condition or self.env.context.get("no_restrict_follower"):
+            return result
+        domain = self.env[
+            "mail.wizard.invite"
+        ]._mail_restrict_follower_selection_get_domain()
+        for key in result:
+            for partner_id, email, lang, reason in result[key]:
+                if partner_id:
+                    partner = self.env["res.partner"].search(
```
hieulucky111: should be `search_count` to improve performance

hieulucky111 (changes requested):

hieulucky111 (approved):
LGTM.

hieulucky111 (approved):
Code review
LGTM

yostashiro:
https://github.com/OCA/social/issues/1070

@AungKoKoLin1997 Let's see how people respond to this, and make adjustments in this PR if there are no strong concerns.

mail_restrict_follower_selection/models/mail_wizard_invite.py
```diff
@@ -0,0 +1,47 @@
...
+                .get_param(parameter_name, default="[]"),
+            )
+        )
+        domain = expression.AND(
+            [safe_eval(parameter_domain), self._fields["partner_ids"].domain]
+        )
+        return domain
+
+    @api.model
+    def get_view(self, view_id=None, view_type="form", **options):
```
michaelslade54, line 37: This get_view method is not called in every instance, so the domain for the partner_ids field is not set correctly.

get_view is only called in the first instance, after that, onchange method is called, even after changing models.

If you add a follower in one model, then go to another model with a different domain set and add a follower again, the get_view function is not called, leading to an incorrect domain set on partner_ids.

Hope this is clear.

michaelslade54:
Not withstanding the issue with the domain for `partner_ids` not reloading when changing models, while a user **not** within the required domain is not added as a follower as intended, they are still sent an email informing them that they have an invitation to follow, which is undesirable.

Can someone please confirm?

I believe the domain needs to be added to `new_partners` here...
https://github.com/odoo/odoo/blob/a6f8d9c512ffb816e18bb43a5efa510707877cf6/addons/mail/wizard/mail_wizard_invite.py#L56

AungKoKoLin1997:
> I believe the domain needs to be added to new_partners here...
> https://github.com/odoo/odoo/blob/a6f8d9c512ffb816e18bb43a5efa510707877cf6/addons/mail/wizard/mail_wizard_invite.py#L56

@michaelslade54 If I am not mistaken, `add_followers` is called after we choose partners in wizard and click button. What I mentioned the issue was when adding domain to partner_ids.

fkantelberg (approved):
Tested and shortly reviews. LGTM

I think that the missing domain refresh on model swap is outside of the scope of this PR.

NL66278 (approved):
:+1: LGTM Could have removed copyright from init files, but not blocking.

PR #1038, opened 2023-01-12
[13.0][IMP] mail_tracking_mailgun: mass mailing manual tracking

chienandalu:
Mass mailing are tracked from mail.trace as the don't store a message in the db. In order to gather the message_id and be able to do manual checks to mailgun, that's the table where we should get the message id.

cc @Tecnativa TT40816

please review @pedrobaeza

### DIFF
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -44,6 +44,7 @@ class MailTrackingEmail(models.Model):
     mail_message_id = fields.Many2one(
         string="Message", comodel_name="mail.message", readonly=True, index=True
     )
+    message_id = fields.Char(compute="_compute_message_id")
     mail_id = fields.Many2one(string="Email", comodel_name="mail.mail", readonly=True)
     partner_id = fields.Many2one(
         string="Partner", comodel_name="res.partner", readonly=True
@@ -114,6 +115,15 @@ class MailTrackingEmail(models.Model):
         default=lambda s: uuid.uuid4().hex,
         groups="base.group_system",
     )
+
+    @api.depends("mail_message_id")
+    def _compute_message_id(self):
+        """This helper field will allow us to map the message_id from either the linked
+        mail.message or a mass.mailing mail.trace.
+        """
+        self.message_id = False
+        for tracking in self.filtered("mail_message_id"):
+            tracking.message_id = tracking.mail_message_id.message_id
 
     @api.model_create_multi
     def create(self, vals_list):
--- a/mail_tracking_mailgun/models/mail_tracking_email.py
+++ b/mail_tracking_mailgun/models/mail_tracking_email.py
@@ -202,12 +202,8 @@ class MailTrackingEmail(models.Model):
         https://documentation.mailgun.com/en/latest/api-events.html
         """
         api_key, api_url, domain, *__ = self._mailgun_values()
-        for tracking in self:
-            if not tracking.mail_message_id:
-                raise UserError(_("There is no tracked message!"))
-            message_id = tracking.mail_message_id.message_id.replace("<", "").replace(
-                ">", ""
-            )
+        for tracking in self.filtered("message_id"):
+            message_id = tracking.message_id.replace("<", "").replace(">", "")
             events = []
             url = urljoin(api_url, "/v3/%s/events" % domain)
             params = {
--- a/mail_tracking_mailgun/views/mail_tracking_email.xml
+++ b/mail_tracking_mailgun/views/mail_tracking_email.xml
@@ -7,10 +7,12 @@
         <field name="inherit_id" ref="mail_tracking.view_mail_tracking_email_form" />
         <field name="arch" type="xml">
             <field name="state" position="before">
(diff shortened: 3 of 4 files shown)


mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -202,12 +202,18 @@ def action_manual_check_mailgun(self):
         https://documentation.mailgun.com/en/latest/api-events.html
         """
         api_key, api_url, domain, *__ = self._mailgun_values()
+        # Soft dependency to check mass mailing trackings which have their message_id
```
pedrobaeza: There's a mail_tracking_mass_mailing module that should do this.

chienandalu: But this method is exclusive of the mailgun one. The alternative is an extra module

mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -202,12 +202,18 @@ def action_manual_check_mailgun(self):
...
-                raise UserError(_("There is no tracked message!"))
-            message_id = tracking.mail_message_id.message_id.replace("<", "").replace(
-                ">", ""
+            message_id = (
+                tracking.mail_message_id.message_id
+                or mm_tracking
+                and tracking.mail_stats_id.message_id
             )
+            if not message_id:
+                raise UserError(_("There is no tracked message!"))
```
pedrobaeza: Don't put this, as there can be several reasons for not having a message_id

pedrobaeza (changes requested):
Put a field `message_id` on the trace, that is computed as mail_message_id.message_id here, and expanded on `mail_tracking_mass_mailing` to be got from `mail_stats_id`, instead of this.

chienandalu:
> Put a field message_id on the trace, that is computed as mail_message_id.message_id here, and expanded on mail_tracking_mass_mailing to be got from mail_stats_id, instead of this.

Done :+1:

mail_tracking_mailgun/models/mail_tracking_email.py
chienandalu: Do we hide the button in such case then?

pedrobaeza: OK

chienandalu: Done

mail_tracking/models/mail_tracking_email.py
```diff
@@ -115,6 +116,14 @@ class MailTrackingEmail(models.Model):
         groups="base.group_system",
     )
 
+    @api.depends("mail_message_id")
+    def _compute_message_id(self):
+        """This helper field will allow us to map the message_id from either the linked
+        mail.message or a mass.mailing mail.trace"""
```
pedrobaeza: Proper PEP257 format.

```suggestion
        mail.message or a mass.mailing mail.trace.
        """
```

mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -203,11 +203,7 @@ def action_manual_check_mailgun(self):
         """
         api_key, api_url, domain, *__ = self._mailgun_values()
         for tracking in self:
```
pedrobaeza: This still can be empty, so do:

```suggestion
        for tracking in self.filtered("message_id"):
```

mail_tracking_mailgun/views/mail_tracking_email.xml
```diff
@@ -7,10 +7,12 @@
         <field name="inherit_id" ref="mail_tracking.view_mail_tracking_email_form" />
         <field name="arch" type="xml">
             <field name="state" position="before">
+                    <field name="message_id" />
```
pedrobaeza: Invisible?

mail_tracking_mass_mailing/models/mail_tracking_email.py
```diff
@@ -21,6 +21,14 @@ def _statistics_link_prepare(self, tracking):
         """Inherit this method to link other object to mailing.trace"""
         return {"mail_tracking_id": tracking.id}
 
+    @api.depends("mail_stats_id")
+    def _compute_message_id(self):
+        """For mass mailings the message id is stored in the mailing.trace record"""
```
pedrobaeza:
```suggestion
        """For the mass mailings, the message id is stored in the mailing.trace record."""
```

mail_tracking_mailgun/views/mail_tracking_email.xml
chienandalu: Sure :+1:

mail_tracking_mailgun/models/mail_tracking_email.py
chienandalu: :+1:

mail_tracking_mass_mailing/models/mail_tracking_email.py
chienandalu: Changed

pedrobaeza (approved):

PR #1040, opened 2023-01-17
[13.0][ADD] mail_activity_permission

hbrunn:
This module allows to manage exceptional access to records via activities.

It is meant for situations where users need access to some record that according to the permissions shouldn't be accessible, but it's also not feasible to change the global permission model.


hbrunn:
test are fixed in https://github.com/OCA/social/pull/1011

hieulucky111 (approved):
Code Review
LGTM

mail_activity_permission/models/mail_activity_mixin.py
```diff
@@ -0,0 +1,58 @@
...
+        comodel_name="mail.activity",
+        compute="_compute_activity_permission_ids",
+        search="_search_activity_permission_ids",
+        groups="base.group_user",
+        context={"active_test": False},
+    )
+
+    @api.depends("activity_ids")
+    def _compute_activity_permission_ids(self):
+        MailActivity = self.env["mail.activity"].with_context(active_test=False)
```
CoraVanga, line 22: I think you should use another variable name, `MailActivity` is duplicated with the class name

CoraVanga (changes requested):

mail_activity_permission/models/mail_activity_mixin.py
hbrunn: thanks, I've renamed the class instead to its canonical name

CoraVanga (approved):
LGTM

mail_activity_permission/models/mail_activity_mixin.py
CoraVanga: Thanks, it is ok for me

PR #1044, opened 2023-01-24
[13.0][IMP] mail_outbound_static: Reply-to the same address as From

SkiBY:
if server has no catchall support, there should be Reply-To message field modified, so the replies returns to the same address, we used for 'From' field So there is special flag added, user can easily turn this option on

### DIFF
--- a/mail_outbound_static/__manifest__.py
+++ b/mail_outbound_static/__manifest__.py
@@ -4,10 +4,13 @@
 {
     "name": "Mail Outbound Static",
     "summary": "Allows you to configure the from header for a mail server.",
-    "version": "13.0.2.0.0",
+    "version": "13.0.2.1.0",
     "category": "Discuss",
     "website": "https://github.com/OCA/social",
-    "author": "brain-tec AG, LasLabs, Adhoc SA, Odoo Community Association (OCA)",
+    "author": "Solvti sp. z o.o., "
+    "brain-tec AG, "
+    "LasLabs, Adhoc SA, "
+    "Odoo Community Association (OCA)",
     "license": "LGPL-3",
     "application": False,
     "installable": True,
--- a/mail_outbound_static/i18n/mail_outbound_static.pot
+++ b/mail_outbound_static/i18n/mail_outbound_static.pot
@@ -52,6 +52,18 @@ msgid "Not a valid Email From"
 msgstr ""
 
 #. module: mail_outbound_static
+#: model:ir.model.fields,field_description:mail_outbound_static.field_ir_mail_server__reply_to_the_same_address
+msgid "Reply To The Same Address"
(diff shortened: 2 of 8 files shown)


SkiBY:
@dreispt and @AungKoKoLin1997 could you look through this, as you are also work with this addon?

thomaspaulb (approved):
This makes sense in some cases.

My current case is one where I'm not using catchall, and I'm using `mail_outbound_static` together with `mail_multicompany` in order to have "CompanyA" mails come from a "CompanyA.com" address and "CompanyB" mails from a "CompanyB.com" address, even though the "From" address might belong to any of the two (in the case of internal users that can access multi companies) or something completely different (in the case of a portal user).

In this case, I want the Reply-To address to be modified also, otherwise even though my From-header has been CompanyB-branded, the reply-to might still be a CompanyA address.

mail_outbound_static/__manifest__.py
```diff
@@ -4,10 +4,13 @@
 {
     "name": "Mail Outbound Static",
     "summary": "Allows you to configure the from header for a mail server.",
-    "version": "13.0.2.0.0",
+    "version": "13.0.3.0.0",
```
thomaspaulb: Shouldn't this rather be a minor update? `13.0.2.1.0`

I think the definition is 

- Requires restart only
- Requires update
- Full refactor, needs human attention to implement

But i'm not 100% sure

thomaspaulb (commented):
This makes sense in some cases.

My current case is one where I'm not using catchall, and I'm using `mail_outbound_static` together with `mail_multicompany` in order to have "CompanyA" mails come from a "CompanyA.com" address and "CompanyB" mails from a "CompanyB.com" address, even though the "From" address might belong to any of the two (in the case of internal users that can access multi companies) or something completely different (in the case of a portal user).

In this case, I want the Reply-To address to be modified also, otherwise even though my From-header has been CompanyB-branded, the reply-to might still be a CompanyA address.

mail_outbound_static/__manifest__.py
SkiBY: Hm.. yes, I was not sure too. I've updated main version, because of implementation of reply logic, not sending. So it is "logical" up. But if you say so, i can make 2.1.0.
So - change?

thomaspaulb: Leave it for now, lets await a third opinion.

hbrunn: it's a minor update in my book too

hbrunn (approved):

mail_outbound_static/__manifest__.py
SkiBY: Corrected

PR #1045, opened 2023-01-24
[WIP] [MIG] microsoft_outlook_single_tenant: migration to 16.0

ajaniszewska-dev:


ajaniszewska-dev:
seems module isn't needed as in v16 we can modify endpoint in ir config parameter: https://github.com/odoo/odoo/commit/afe82227127d0e41521d5dfbaf65d93fcf52dc18

PR #1048, opened 2023-01-26
[14.0][IMP] mail_tracking_mailgun: mass mailing manual tracking

chienandalu:
- FW of #1038 

Mass mailing are tracked from mail.trace as the don't store a message in the db. In order to gather the message_id and be able to do manual checks to mailgun, that's the table where we should get the message id.

cc @Tecnativa TT40816

please review @victoralmau @pedrobaeza


victoralmau (approved):

pedrobaeza (approved):
Now to v15!

/ocabot merge minor

PR #1051, opened 2023-01-30
[15.0][FIX] mail_outbound_static: Move smtp info update logic to _prepare_email_message()

AungKoKoLin1997:
Since 15.0, smtp info is prepared in _prepare_email_message() which is called from send_email(), and the smtp info manipulation should be done in this method instead of in send_email().

@qrtl


SkiBY:
@AungKoKoLin1997 could you add test case for new _prepare_email_message function?

dreispt (approved):

amkarthik (approved):

yostashiro:
@AungKoKoLin1997 Can you please forward-port this to 16.0.

PR #1052, opened 2023-01-30
[13.0][FIX] mail_tracking: failed message compute performance

chienandalu:
The failed messages computation relies on a dotted notation search on the message mail_tracking_ids. This causes a really bad performance in almost any model with a mail.thread. It's a good use case for auto_join.

Customer database example of worst case scenario performance before:
image

Performance after the patch:
image

cc @Tecnativa TT41397

please review @carlosdauden @pedrobaeza


pedrobaeza (approved):

carlosdauden (approved):

pedrobaeza:
Let's merge it for getting the performance fix:

/ocabot merge patch

PR #1054, opened 2023-01-30
[15.0] [IMP] `mail_activity_team`: Allow to set a team on server actions

ivantodorovich:
image

This PR let you force a team (different than the user's default one) when configuring activities through server actions

----
+bonus: Use `setUpClass` instead of `setUp` during tests

### DIFF
--- a/mail_activity_team/models/ir_actions_server.py
+++ b/mail_activity_team/models/ir_actions_server.py
@@ -1,0 +1,20 @@
+# Copyright 2023 Camptocamp SA (https://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class IrActionsServer(models.Model):
+    _inherit = "ir.actions.server"
+
+    activity_team_id = fields.Many2one(
+        "mail.activity.team",
+        string="Activity Team",
+    )
+
+    def _run_action_next_activity(self, eval_context=None):
+        # OVERRIDE to force the activity team on scheduled actions
+        if self.activity_user_type == "specific" and self.activity_team_id:
+            self = self.with_context(force_activity_team=self.activity_team_id)
+        return super()._run_action_next_activity(eval_context=eval_context)
--- a/mail_activity_team/__manifest__.py
+++ b/mail_activity_team/__manifest__.py
@@ -15,6 +15,7 @@
     "data": [
         "security/ir.model.access.csv",
         "security/mail_activity_team_security.xml",
+        "views/ir_actions_server_views.xml",
         "views/mail_activity_type.xml",
         "views/mail_activity_team_views.xml",
         "views/mail_activity_views.xml",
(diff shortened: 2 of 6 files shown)


chienandalu (approved):
LGTM :+1: Thanks!

yankinmax (approved):

mail_activity_team/models/ir_actions_server.py
```diff
@@ -0,0 +1,20 @@
...
+    activity_team_id = fields.Many2one(
+        "mail.activity.team",
+        string="Activity Team",
+    )
+
+    def _run_action_next_activity(self, eval_context=None):
+        # OVERRIDE to force the activity team on scheduled actions
+        if self.activity_user_type == "specific" and self.activity_team_id:
+            self = self.with_context(force_activity_team=self.activity_team_id)
+        return super()._run_action_next_activity(eval_context=eval_context)
```
pedrobaeza, line 20: It's not a good pattern to overwrite `self` variable that can affect downstream code

```suggestion
        _self = self
        if self.activity_user_type == "specific" and self.activity_team_id:
            _self = self.with_context(force_activity_team=self.activity_team_id)
        return super(IrActionsServer, _self)._run_action_next_activity(eval_context=eval_context)
```

ivantodorovich: Hello @pedrobaeza ! 
Thanks for the review :)

TBH I like the simplicity and readability of re-assigning `self` too much, it's a pattern I use every day, so I'm very much interested in the risks you mention. To my knowledge, as `self` is a local variable here, it has virtually the same effect than your suggestion code 🤔 am I missing something?

pedrobaeza: AFAIK, being self an "object" and not a basic variable (integer, float, boolean), the argument is pass by reference, not by value, so I would say that the internal context assignation is affecting the whole object. But TBH, I'm not sure about that either. I have always avoided for being prudent, and more being the internals a black box that can work today, but not tomorrow.

ivantodorovich: AFAIK, `self` is just a local variable here.

You could name it `foo` if you'd like, and the code would work as usual.
There's nothing special about it, just it position as first argument.

```py
class ResClass(models.Model):
    _name = "res.class"

    def action_confirm(foo):
        foo.write({"state": "confirm"})
```

Yes, it points to an object, and it's passed by reference.. but that's also what happens when you assign and pass `_self` instead.

chienandalu: IMO `self` reasignation is limited to the scope of the method, so it makes no difference to assign a different variable or to reassign it as @ivantodorovich does

pedrobaeza: OK, yes, you are right. When you reassign, it doesn't affect to the source object. A different thing would be if you are passing a dictionary and adding a new key in your method, which is not the case (and the reason why Odoo put `frozendict` in their ORM). Sorry for the noise.

ivantodorovich:
It seems we're missing https://github.com/OCA/oca-addons-repo-template/commit/3ccb33664717dbd7fb52c10cc44b7545d4e2827c

I'll prepare another PR for the copier update

ivantodorovich:
Here it is: 
- https://github.com/OCA/social/pull/1055

Issue #1061, opened 2023-02-01
[15.0] microsoft_outlook_single_tenant: Error on tests


pedrobaeza:
This module is provoking an error in the CI:

https://github.com/OCA/social/actions/runs/4062982084/jobs/6994667509#step:8:679

It may be due to the changes Odoo has changed upstream: 8f623307749a395b1399d3e8b128e375a94d97ae. I think this makes the module useless (or it can be kept for easy config).

But in any case, the tests should be fixed.

@santostelmo can you please check?

santostelmo:
Thanks @pedrobaeza good news that odoo added that in standard. Indeed that module became useless after commit 8f623307749a395b1399d3e8b128e375a94d97ae

pedrobaeza:
Can you then fix the tests on existing module or remove them?

santostelmo:
Yes I will

santostelmo:
@pedrobaeza PRs for that issue referenced above.

pedrobaeza:
Thanks!

PR #1062, opened 2023-02-02
[14.0] Deprecate microsoft_outlook_single_tenant

santostelmo:
Odoo added the possibility to customize Microsoft authentication endpoint. Therefore this module is not needed anymore.

Related with https://github.com/OCA/social/issues/1061


pedrobaeza (approved):
Fast-tracking this to solve the error in tests:

/ocabot merge major

PR #1064, opened 2023-02-07
[IMP] We should always have aliases in tracking

cvinh:
When receiving an email the user should know if the alias is the mail recipient or is a Cc

Use case :

setup an alias (ie [info@domain.com](mailto:info@domain.com)) to an object
1- send an email to [info@domain.com](mailto:info@domain.com) and cc someone else
2- send an email to someone else and cc [info@domain.com](mailto:info@domain.com)
In the chatter, we want to include the alias to know if the alias was was email_to or email_cc because the user has not the same priority to answer (if the alias is email_to --> the user must reply, if the alias is email_cc --> the user has no action)

Here are screenshots after this PR (alias is setup with crm.team)
Test_email_to_alias
Test_email_cc_alias

### DIFF
--- a/mail_tracking/models/mail_message.py
+++ b/mail_tracking/models/mail_message.py
@@ -139,8 +139,8 @@ class MailMessage(models.Model):
                 .search([("mail_message_id", "=", message.id)])
             )
             # String to List
-            email_cc_list = self._drop_aliases(email_split(message.email_cc))
-            email_to_list = self._drop_aliases(email_split(message.email_to))
+            email_cc_list = self._drop_catchall(email_split(message.email_cc))
+            email_to_list = self._drop_catchall(email_split(message.email_to))
             # Search related partners recipients
             partners |= partners.search(
                 [("email", "in", email_cc_list + email_to_list)]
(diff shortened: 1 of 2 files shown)


pedrobaeza:
If you do this, then we should remove such aliases from the suggestions. The goal is to avoid duplications. You should also take into account that sometimes the recipient is an alias (for example, you send to the project alias alias@company.com), and this creates a project task. In such task, you shouldn't put alias@company.com in the tracking info.

cvinh:
> If you do this, then we should remove such aliases from the suggestions. The goal is to avoid duplications. You should also take into account that sometimes the recipient is an alias (for example, you send to the project alias [alias@company.com](mailto:alias@company.com)), and this creates a project task. In such task, you shouldn't put [alias@company.com](mailto:alias@company.com) in the tracking info.

Yes you are right, so I changed the code of my previous PR
Here I did not change the suggestions, I show alias in tracking_status only
We must have alias@company.com in the tracking info indeed, this is to know of alias@company.com was in Cc (we don't need to answer) or in To (we must answer)

pedrobaeza:
Can you put screenshots of the results of each of them?

pedrobaeza:
My fear is that with this, the project alias is shown in the task that has been generated from that alias. Is that correct?

cvinh:
> My fear is that with this, the project alias is shown in the task that has been generated from that alias. Is that correct?

Yes indeed we do want that the alias is shown in the chatter (see updated screenshots above)

cvinh:
We close that one in favor to https://github.com/OCA/social/pull/941

PR #1068, opened 2023-02-09
[15.0][FIX] mail_tracking: fix portal tests

### LINKED ISSUE #1067: [15.0] mail_tracking: portal tests fail
## Module
mail_tracking

## Describe the bug
mail_tracking causes TestPortalWizard.test_portal_wizard_public_user and TestPortalWizard.test_portal_wizard_partner to fail

## To Reproduce
**Affected versions**: 15.0

1. init db with portal and mail_tracking
2. run portal tests

jonasbuchholz:
fixes #1067

### DIFF
--- a/mail_tracking/models/ir_mail_server.py
+++ b/mail_tracking/models/ir_mail_server.py
@@ -45,10 +45,10 @@ class IrMailServer(models.Model):
         if tracking_email_id:
             headers = self._tracking_headers_add(tracking_email_id, headers)
         msg = super(IrMailServer, self).build_email(
-            email_from,
-            email_to,
-            subject,
-            body,
+            email_from=email_from,
+            email_to=email_to,
+            subject=subject,
+            body=body,
             email_cc=email_cc,
             email_bcc=email_bcc,
             reply_to=reply_to,


pedrobaeza:
I don't get the relation, and more being such args non keyword ones:

https://github.com/odoo/odoo/blob/5fe6b57bffb8108f9534dd42b5ad3c6114e05e5b/odoo/addons/base/models/ir_mail_server.py#L425

Without exposing the full problem or how "the tests fail", it's difficult to know.

jonasbuchholz:
Hi @pedrobaeza , thank you for your feedback. The tests fail with
```
ERROR: Subtest TestPortalWizard.test_portal_wizard_partner (login='admin')
  File "<decorator-gen-406>", line 2, in test_portal_wizard_partner
  File "/data/build/odoo/odoo/tests/common.py", line 1769, in _users
    func(*args, **kwargs)
  File "/data/build/odoo/addons/portal/tests/test_portal_wizard.py", line 79, in test_portal_wizard_partner
    self.assertSentEmail(self.env.user.partner_id, [self.partner])
  File "/data/build/odoo/addons/mail/tests/common.py", line 483, in assertSentEmail
    sent_mail = next(
  File "/data/build/odoo/addons/mail/tests/common.py", line 485, in <genexpr>
    if set(mail['email_to']) == set(expected['email_to']) and mail['email_from'] == expected['email_from']
KeyError: 'email_to'
```
This is due to the fact that in the mock mail gateway, the kwargs are used for assertion: https://github.com/odoo/odoo/blob/1839bf83a14a12eb83a8f721f82ece7238319e43/addons/mail/tests/common.py#L56-L59
By passing the arguments as positional arguments, they will be missing from the dict in self._mails, causing the key error above.

chienandalu (approved):
Thanks @jonasbuchholz Indeed is an unintuitive implementation by Odoo

@pedrobaeza here's how that mock test class expects those parameters as keywords https://github.com/odoo/odoo/blob/adbf23b92e97f6d41874033ae69edaa5a8e1f00c/addons/mail/models/mail_mail.py#L448-L451

pedrobaeza:
OK, thanks for the explanations. @tde-banana-odoo maybe you can fix that in core not passing the first 4 arguments as keyword ones, but normal ones.

Meanwhile, we mimic such behavior here:

/ocabot merge patch

chienandalu:
> OK, thanks for the explanations. @tde-banana-odoo maybe you can fix that in core not passing the first 4 arguments as keyword ones, but normal ones.

Or defining all of them as keywords as they expect them to be

PR #1071, opened 2023-02-17
[16.0][MIG] mail_optional_follower_notification: Migration to 16.0

yankinmax:
ref: RD-622


yankinmax:
@adrienpeiffer @sbidoul @pedrobaeza

damdam-s (approved):

mail_optional_follower_notification/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2016 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Mail optional follower notification",
+    "summary": "Choose to notify followers on mail.compose.message",
+    "author": "ACSONE SA/NV," "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/social",
+    "category": "Social Network",
+    "version": "16.0.1.0.1",
```
simahawk:
```suggestion
    "version": "16.0.1.0.0",
```

mail_optional_follower_notification/models/mail_thread.py
```diff
@@ -0,0 +1,30 @@
...
+from odoo.tools import config
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    def _notify_get_recipients(self, message, msg_vals, **kwargs):
+        """Compute recipients to notify based on subtype and followers. This
+        method returns data structured as expected for ``_notify_recipients``."""
+        test_condition = config["test_enable"] and not self.env.context.get(
```
simahawk: why do we need this test condition? :thinking:

simahawk (changes requested):

mail_optional_follower_notification/models/mail_thread.py
yankinmax: I didn't modify the existing approach.
It seems this was added to fix Travis build while migrating to v14.
https://github.com/OCA/social/commit/05c8aea71a38fcce24659f9d03e9653f383be677

Now, probably we can remove it as tests are green

simahawk: Well, there's no explanation of that change on that commit because it was squashed on the migration one :man_shrugging:

simahawk (approved):

yankinmax:
@pedrobaeza can you pls help to get it merged?

PR #904, opened 2022-05-03
[15.0][MIG] email_template_qweb: Migration to 15.0

fshah-initos:

PR #1072, opened 2023-02-20
[MIG] email_template_qweb: Migration to 15.0

dzungtran89:
- This PR superseds https://github.com/OCA/social/pull/904


ivantodorovich (approved):
Technically LG

email_template_qweb/models/mail_template.py
```diff
@@ -26,7 +26,9 @@ def generate_email(self, res_ids, fields):
         ).items():
             self_with_lang = self.with_context(lang=lang)
             for res_id in res_ids:
-                if self.body_type == "qweb" and (not fields or "body_html" in fields):
+                if self.body_type == "qweb_view" and (
+                    not fields or "body_html" in fields
+                ):
```
hbrunn, line 31: can't you by now override https://github.com/odoo/odoo/blob/15.0/addons/mail/models/mail_render_mixin.py#L565 and just pass the value of `body_type` in `engine`?

hbrunn (changes requested):

email_template_qweb/models/mail_template.py
dzungtran89: That actually renders a mail template, while this tries to take the predefined [views](https://github.com/OCA/social/pull/1072/files#diff-3111cb97ab1476860a01e335e26ed4196c9535d5dae0bce9a9358c4668e28ab3R15) to render the mail's body

hbrunn: I know what the module is doing because I'm it's original author ;-)

But now I see this function insists on views with an [xmlid](https://github.com/odoo/odoo/blob/15.0/addons/mail/models/mail_render_mixin.py#L378) for that case, so it's indeed not usable for our case

hbrunn (approved):

PR #1074, opened 2023-02-22
[15.0][ADD] mail_send_confirmation

AungKoKoLin1997:
This module shows confirmation dialog when send mail to the customers.
The purpose of this module is to avoid a mail is sent to the customer instead of an internal note on the chatter.

@qrtl


AungKoKoLin1997:
pre-commit is failing because of eslint. I am not expert at odoo javascript. If anyone can suggest me better approach for this module, I am always welcome and ready to make adjustment.

yostashiro:
Tested on runboat. While it works on the send button in the chatter, no confirmation is requested when you switch to the message modal, which I think makes this solution incomplete.



I was wondering if it would make the design simpler if we:

- add a 'I understand that the message will be sent to the external follower(s)' checkbox in the template and disable the 'Send' button if the checkbox is not selected, and
- do the same for the message modal

Another approach I thought was to migrate https://github.com/OCA/web/tree/14.0/web_send_message_popup to 15.0 (there is already a PR) and add an option in this module to request for a confirmation before sending the message.

PR #1076, opened 2023-02-23
[ADD] mail_activity_unlink_log

etobella:
This addon enforces the creation of a message when an activity is unlinked.
This change tries to avoid a possible bad practice of deletion of an activity without notifying the deletion.


LoisRForgeFlow (commented):
LGTM but maybe a more descriptive module name would be good. On a quick shoot, I can think on mail_activity_log_unlink or mail_activity_notify_unlink

pedrobaeza:
Yeah, `mail_activity_unlink_log` seems OK.

mail_activity_unlink_log/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2023 CreuBlanca
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Mail Activity Unlink",
```
LoisRForgeFlow: Change the name here

LoisRForgeFlow (approved):
:+1:

sanchonuria (approved):

kluna1998 (approved):
LGTM! 👍🏽

PR #1077, opened 2023-02-23
[FIX] mail_tracking_email: tuple index out of range fix

jok-adhoc:


pedrobaeza:
Can you put steps to reproduce the problem and maybe a regression test?

PR #1084, opened 2023-03-03
[16.0][MIG] mass_mailing_partner

nguyenminhchien:


mass_mailing_partner/models/mailing_contact.py
```diff
@@ -34,47 +34,75 @@ def _check_partner_id_list_ids(self):
...
+                list_ids=vals.get("list_ids", []),
+            )
+            # Don't take the tag_ids from the cache, because it triggers to re-update
+            # the values of creating new Partner by removing the key `category_id`.
+            # Therefore, new partner does not have value for Tags
+            tag_vals = vals.get("tag_ids")
+            if record.partner_id.category_id:
+                # Get Tags from Partner to take into account the tags from Mailing List as well
+                tag_vals = [(6, 0, record.partner_id.category_id.ids)]
+            new_vals.update({"tag_ids": tag_vals})
```
victoralmau: Is this really necessary?
Can you add a test to verify it?
Can you homogenize this code and the write code in a new method?

mass_mailing_partner/models/mailing_contact.py
```diff
@@ -34,47 +34,75 @@ def _check_partner_id_list_ids(self):
                     )
 
     @api.onchange("partner_id")
-    def _onchange_partner_mass_mailing_partner(self):
+    def _onchange_partner_mass_mailing_partner(self, change_tag_bool=True):
```
victoralmau: Is this really necessary?
In any case, I think the right thing to do would be to use context to do this.

victoralmau (changes requested):

mass_mailing_partner/models/mailing_contact.py
nguyenminhchien: Of course it is necessary, i don't add these lines for fun. You can play with this by removing them, then compare the result.
the test has been done at this line: https://github.com/OCA/social/blob/7644f8f70daf10f4ea3f9708af973660b27b106e/mass_mailing_partner/tests/base.py#L66

>  Can you homogenize this code and the write code in a new method?

Good idea, WIP on it

nguyenminhchien: @victoralmau:  Updated

mass_mailing_partner/models/mailing_contact.py
nguyenminhchien: updated

mass_mailing_partner/models/mailing_contact.py
```diff
@@ -0,0 +1,122 @@
...
+        if self.partner_id:
+            self.name = self.partner_id.name
+            self.email = self.partner_id.email
+            self.title_id = self.partner_id.title
+            self.company_name = (
+                self.partner_id.company_id.name or self.partner_id.company_name
+            )
+            self.country_id = self.partner_id.country_id
+            category_ids = self.partner_id.category_id
+            if category_ids and not self._context.get("no_update_tag"):
```
yajo: Would you please avoid [unnecessary double negatives](https://www.recallstack.icu/en/2020/09/01/do-not-use-negative-booleans-please/)?

Simpler to read:

```suggestion
            if category_ids and not self.env.context.get("update_tags", True):
```

mass_mailing_partner/models/mailing_contact.py
```diff
@@ -0,0 +1,122 @@
...
+                self.tag_ids = category_ids
+
+    @api.model
+    def _compute_contact_vals(self, origin_vals):
```
yajo: The method name is misleading because it seems like a compute method. Please change it to something like:

```suggestion
    def _get_contact_vals(self, origin_vals):
```

mass_mailing_partner/models/mailing_contact.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+        # Don't take the tag_ids from the cache, because it triggers to re-update
+        # the values of creating new Partner by removing the key `category_id`.
+        # Therefore, new partner does not have value for Tags
+        tag_vals = "tag_ids" in origin_vals and origin_vals.get("tag_ids") or None
+        if record.partner_id.category_id:
+            # Get Tags from Partner to take into account the tags from Mailing List as well
+            tag_vals = [(6, 0, record.partner_id.category_id.ids)]
+        if tag_vals is not None:
+            new_vals.update({"tag_ids": tag_vals})
```
yajo: KISS:

```suggestion
        new_vals["tag_ids"] = origin_vals.get("tag_ids", [(6, 0, record.partner_id.category_id.ids)])
```

yajo (changes requested):
I share the same concerns as @victoralmau. The code added seems a bit strange. That doesn't mean it's wrong, but there's something strange in the fact you need to do it like that. 🤔 

I think this module would benefit a lot from converting the onchange to computed writable fields. That would probably save a lot of weird code (it already had a bit). It can be done later, of course.

In any case, some suggestions below. Thanks for your your work!

@moduon MT-2479

Camille0907 (approved):
LGTM, but please consider @yajo comments

StephaneMangin:
cc @yajo Is it ok for you now ?

yajo (approved):

pedrobaeza:
I have squashed administrative commit history and fix the problem in vi.po and vi_VN.po files:

/ocabot mere nobump

PR #1093, opened 2023-03-09
[FIX] fix value returned by _action_send_mail

ajaniszewska-dev:


mail_optional_follower_notification/wizard/mail_compose_message.py
```diff
@@ -12,5 +12,7 @@ class MailComposeMessage(models.TransientModel):
     def _action_send_mail(self, auto_commit=False):
         for wizard in self:
             wizard = wizard.with_context(notify_followers=wizard.notify_followers)
-            super(MailComposeMessage, wizard)._action_send_mail(auto_commit=auto_commit)
-        return True
+            res = super(MailComposeMessage, wizard)._action_send_mail(
+                auto_commit=auto_commit
+            )
+        return res
```
grindtildeath: This is going to return only the last result of the call to `super`, IMO we need to do something like this instead:
```
def _action_send_mail(self, auto_commit=False):
    result_mails_su, result_messages = self.env['mail.mail'].sudo(), self.env['mail.message']
    for wizard in self:
        wizard = wizard.with_context(notify_followers=wizard.notify_followers)
        res_mail, res_message = super(MailComposeMessage, wizard)._action_send_mail(
            auto_commit=auto_commit
        )
        result_mails_su += res_mail
        result_messages += res_message
    return result_mails_su, result_messages
```

sebalix (approved):

grindtildeath (approved):

hbrunn (approved):

PR #1102, opened 2023-03-15
[ADD] mass_mailing_list_sql: Integrate SQL queries for partner selection in mailing lists

JoanSForgeFlow:
The module allows users to define an SQL query for selecting partners to be included in the mailing list, providing greater flexibility in targeting recipients with complex criteria. 

By integrating this module, users can create more refined and targeted mailing lists, ultimately improving their email marketing strategies.

@JordiBForgeFlow @ForgeFlow

### DIFF
--- a/mass_mailing_list_sql/__init__.py
+++ b/mass_mailing_list_sql/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import tests
--- a/mass_mailing_list_sql/__manifest__.py
+++ b/mass_mailing_list_sql/__manifest__.py
@@ -1,0 +1,15 @@
+# Copyright (C) 2023 ForgeFlow, S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Mass Mailing List SQL",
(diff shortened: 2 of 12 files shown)


GuillemCForgeFlow (approved):
LGTM! 👍🏽 
Just as a nice to have, perhaps adding a user error when no users are found or when there is no partner with a valid email from all the contacts found, it would be great 👍🏽

JordiBForgeFlow:
@GuillemCForgeFlow Perhaps in this case the customers should just be filtered out. Sometimes the SQL may return 0 partners, and that may be ok. Would be bad to raise an error message. Today the results are 0, but tomorrow they may not be.

mariadforgeflow (approved):
LGTM!

mass_mailing_list_sql/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import models
+from . import tests
```
GuillemCForgeFlow, line 2:
```suggestion
```
Not needed.

GuillemCForgeFlow (changes requested):

PR #1104, opened 2023-03-16
[ADD] mail_template_default Module allows to set default mail template in composer base on rules

JakubWiselka:
Module that allows you to create rules base on which the specific template will be applied to mail composer.

Rules are executed in a specific order based on the amount of information entered into the rule.
The first rule to meet all the conditions is taken
Mail Template Rules are sorted base on 4 fields in the following hierarchy:
Context Flag -> Company -> Filter -> Sequence

### DIFF
--- a/mail_template_default/__manifest__.py
+++ b/mail_template_default/__manifest__.py
@@ -1,0 +1,16 @@
+# Copyright 2023 Solvti sp. z o.o. (https://solvti.pl)
+
+{
+    "name": "Mail Template Default",
+    "summary": "Create rules to apply default mail template in composer",
+    "version": "15.0.1.0.0",
+    "category": "Social Network",
+    "author": "Solvti, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["mail"],
+    "website": "https://github.com/OCA/social",
+    "data": [
+        "security/ir.model.access.csv",
+        "views/mail_template_rule_views.xml",
+    ],
+}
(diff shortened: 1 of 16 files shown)


JakubWiselka:
@KDRdev @MSokol00 @MriV13 @SkiBY Can some of you take a few minutes to test this module and maybe do a CR?

mail_template_default/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2023 Solvti sp. z o.o. (https://solvti.pl)
+
+{
+    "name": "Mail Template Default",
```
SkiBY, line 4: Probably, the name should be changed, to avoid misunderstanding.
This addon adds the rules, not default template only for a wizard. You can simply add Rules at the end?

mail_template_default/wizard/mail_compose_message.py
```diff
@@ -0,0 +1,85 @@
...
+
+        res = super(MailComposer, self).default_get(fields)
+
+        res_model = self._context.get("active_model") or self._context.get(
+            "default_model"
+        )
+        res_id = self._context.get("active_id") or self._context.get("default_res_id")
+        force_mt = self._context.get("force_mail_template")
+        domain = []
+        if not res_model or not res_id or force_mt is False:
```
SkiBY, line 37: If you want to use is, it better to use None, simple not or check if key is in context. So, some more clear check.

mail_template_default/wizard/mail_compose_message.py
```diff
@@ -0,0 +1,85 @@
...
+            domain += [
+                "|",
+                ("company_id", "=", False),
+                ("company_id", "=", rec.company_id.id),
+            ]
+
+        rule_ids = (
+            self.env["mail.template.rule"]
+            .sudo()
+            .search([("model_id.model", "=", rec._name)] + domain)
```
SkiBY, line 61: you can add domain base to line 36 from the start

PR #1105, opened 2023-03-16
[FIX] mail_optional_follower_notification: TypeError

Segui-S73:
TypeError: cannot unpack non-iterable bool object


ioans73 (approved):
Tested locally

hbrunn:
thanks, but that was already proposed in https://github.com/OCA/social/pull/1093

PR #1107, opened 2023-03-17
[14.0][ADD] mail_save_eml New module to save outgoing emails as attachments

JakubWiselka:
New module mail_save_eml which adds possibility to save outgoing email as attachments with .eml format.
The module was created with the need to have a proof or copy of the email that were sent to the customer. In one company we use credit control modules and that happen all the time that customers complains that they did not got any mails from us. With this module, we can send the copy of the email which was sent.

Maybe someone will find this useful


SkiBY (changes requested):
Tests should be added

PR #1111, opened 2023-03-23
[16.0][MIG] email_template_qweb: Migration to 16.0

dzungtran89:


santostelmo (approved):

StephaneMangin (approved):

simahawk:
IMO we should improve the description of the module.
Odoo 16 already renders qweb by default on mail templates https://github.com/odoo/odoo/blob/16.0/addons/mail/models/mail_template.py#L59 so we should focus a little bit on what this module offers more than standard.

Furthermore, if qweb is fully supported (judging from the code, not tested)  I tend to say that we could even consider trashing this module. Again, only if the full set of directives is supported (eg: `t-call`, inheritance, etc).

@dzungtran89 can you please check/test and if needed improve the description? Thanks!

dzungtran89:
Hi @simahawk ,
- About the description, I think the module has already done it well in the [README](https://github.com/OCA/social/blob/5be039cf9878aeb6b9ef3ead9e46f6d5ccb916fc/email_template_qweb/README.rst#qweb-for-email-templates) 
    
- About your concern, luckly the module's author and I had a discussion on another [PR](https://github.com/OCA/social/pull/1072#discussion_r1141853978), you can see the thread for more details

dzungtran89:
Another compelling argument is https://github.com/OCA/social/pull/904#issuecomment-1176196768

gurneyalex (approved):

PR #1117, opened 2023-03-29
[15.0][MIG] mail_preview_audio: Migration to 15.0

chienandalu:
- Standard migration

@Tecnativa TT41546

please review @CarlosRoca13 

ping @etobella


etobella (approved):
Tested on runbot

pedrobaeza:
I think this is worth to be proposed to Odoo upstream to master, don't you think?

chienandalu:
> I think this is worth to be proposed to Odoo upstream to master, don't you think?

I spent some time thinking that it was indeed added in v14 as you can preview some audio files in the core attachments widget (you can with `mp3`, but you can't with `wav`, for example) that are run in the browser's video player.

But I was missing the point of the base module that adds a field widget for binary files that allows that previews in place.

image

So the whole binary field widget should be moved to have that.

pedrobaeza:
I don't quite get you. The idea is to use this preview option in the DMS. If not needed for most audio file (OGG, MP3, OPUS...), then let's get rid off this module. WAV is not so important IMO.

chienandalu:
> I don't quite get you. The idea is to use this preview option in the DMS. If not needed for most audio file (OGG, MP3, OPUS...), then let's get rid off this module. WAV is not so important IMO.

`dms` has `mail_preview_base` as a dependency which adds the mentioned `preview_binary` field widget (nothing to do with core) which is extended by this module to support audio files. (https://github.com/OCA/dms/blob/0142cb1fe6ae0ee3316d493e3d3c981f881f3ef0/dms/views/dms_file.xml#L418)

What core does have already is support for audio files in the attachments widget which this module used to support as well but now delegates to Odoo.

pedrobaeza (approved):
/ocabot merge nobump

PR #1128, opened 2023-04-17
[14.0][FIX] mail_tracking: Index error

Gladys-Haelters:
Current code for mail_tracking_email returns an error as the index is not set correctly, [X3] does not exist/has no value
in the _find_allowed_tracking_ids method
```
        return [
            x[0]
            for x in msg_linked
            if (x[1] in msg_ids)  # We can read the linked message
            or (
                not any({x[1], x[2]}) and x[3] in partner_ids
            )  # No linked msg/mail but we can read the linked partner
            or (not any({x[1], x[2], x[3]}))  # No linked record
        ]
       
```
instead of just fixing the indexes, i opted to rewrite a small part of this code for readability
resulting in fixing the problem and making it more clear what is happening.


mail_tracking/models/mail_tracking_email.py
```diff
@@ -166,15 +166,20 @@ def _find_allowed_tracking_ids(self):
...
-            x[0]
-            for x in msg_linked
-            if (x[1] in msg_ids)  # We can read the linked message
-            or (
-                not any({x[1], x[2]}) and x[3] in partner_ids
-            )  # No linked msg/mail but we can read the linked partner
-            or (not any({x[1], x[2], x[3]}))  # No linked record
-        ]
+        res = []
+
```
pedrobaeza, line 170: Don't put empty lines inside a method (remove all of them)

mail_tracking/models/res_partner.py
```diff
@@ -23,10 +23,6 @@ def _compute_email_score_and_count(self):
         mt_obj = self.env["mail.tracking.email"].sudo()
         for partner in partners_mail:
             partner.email_score = mt_obj.email_score_from_email(partner.email)
-            # We don't want performance issues due to heavy ACLs check for large
```
pedrobaeza, line 26: Why this unrelated change?

pedrobaeza (changes requested):
Thanks for the contribution. Qualifying for co-author for this size of contribution is not applicable.

pedrobaeza:
And please put a proper PR title and commit message.

mail_tracking/__manifest__.py
```diff
@@ -7,10 +7,10 @@
 {
     "name": "Email tracking",
     "summary": "Email tracking system for all mails sent",
-    "version": "14.0.3.1.0",
+    "version": "14.0.3.1.1",
     "category": "Social Network",
     "website": "https://github.com/OCA/social",
-    "author": ("Tecnativa, " "Odoo Community Association (OCA)"),
+    "author": ("Tecnativa", "Odoo Community Association (OCA)", "Accomodata BV"),
```
pedrobaeza, line 13: You mustn't add an author for this little change.

mail_tracking/__manifest__.py
```diff
@@ -7,10 +7,10 @@
 {
     "name": "Email tracking",
     "summary": "Email tracking system for all mails sent",
-    "version": "14.0.3.1.0",
+    "version": "14.0.3.1.1",
```
pedrobaeza, line 10: Don't change manually the version. It will be done by the bot.

mail_tracking/models/mail_tracking_email.py
```diff
@@ -166,15 +166,20 @@ def _find_allowed_tracking_ids(self):
...
+        for tracking_id, message_id, partner_id in msg_linked:
+            # We can read the linked message
+            linked_message = message_id in msg_ids
+            # No linked msg/mail but we can read the linked partner
+            linked_partner = not any({tracking_id, message_id}) and partner_id in partner_ids
```
chienandalu, line 175: Sorry, although the issue is real, this fix isn't right as it doesn't address the issue correctly. You won't ever enter this condition or the next one :wink: 

Please, follow: https://github.com/OCA/social/pull/1148

pedrobaeza:
Closing in favor of https://github.com/OCA/social/pull/1148

PR #1129, opened 2023-04-17
[16.0][MIG] mail_activity_team: Migration to 16.0

rven:


mail_activity_team/models/mail_activity_mixin.py
```diff
@@ -23,13 +23,27 @@ def _search_my_activity_date_deadline(self, operator, operand):
             return super(MailActivityMixin, self)._search_my_activity_date_deadline(
```
bealdav: .

bealdav (commented):
Thanks a lot for your contribution.

image

Module works, but when a team is notified, concerned users are not warn as it happens with individual alert
No indicator displayed

I don't know if it's intentional !?
@JordiBForgeFlow @MiquelRForgeFlow

mail_activity_team/models/mail_activity_mixin.py
rven: @bealdav Can this be reproduced in an earlier version?

rven: When the activities pop-up is "closed", it always shows **My Activities**.
It only temporarily shows the Team Activities when you select it in the pop-up.
It was designed like this so I also left this like it was.

mail_activity_team/models/mail_activity.py
```diff
@@ -0,0 +1,87 @@
...
+                        "The assigned user %(user_name)s is "
+                        "not member of the team %(team_name)s.",
+                        user_name=activity.user_id.name,
+                        team_name=activity.team_id.name,
+                    )
+                )
+
+    @api.onchange("activity_type_id")
+    def _onchange_activity_type_id(self):
+        res = super(MailActivity, self)._onchange_activity_type_id()
```
bealdav:
```suggestion
        res = super()._onchange_activity_type_id()
```

mail_activity_team/models/mail_activity_mixin.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+
+    @api.depends("activity_ids")
+    def _compute_activity_team_user_ids(self):
+        for rec in self:
+            rec.activity_team_user_ids = rec.activity_ids.mapped("team_id.member_ids")
+
+    def _search_my_activity_date_deadline(self, operator, operand):
+        if not self._context.get("team_activities", False):
+            return super(MailActivityMixin, self)._search_my_activity_date_deadline(
```
bealdav:
```suggestion
            return super()._search_my_activity_date_deadline(
```

bealdav (approved):
Thanks a lot, LGTM to me

Needs experts on this module to a more appropriate approval

rconjour (approved):

Bert038:
Can this being merged to v16? The percentage of code check is negligible :)

PR #1137, opened 2023-04-24
[14.0][ADD] mail_thread_create_nolog: display create msg on the fly

sebalix:
This module will block the creation in the database of 'Record created' messages on data models inheriting from `mail.thread`, but will instead generate this message on the fly when Odoo displays the messages of a record.

This allows to reduce the size of the `mail_message` table.


mail_thread_create_nolog/models/mail_message.py
```diff
@@ -0,0 +1,60 @@
...
+            res.append(create_message_values)
+        return res
+
+    def _get_create_message_id(self, domain):
+        # Return the negative version of the greatest message ID
+        # of the current record
+        return -self.search(domain, limit=1, order="id DESC").id
+
+    @api.model
+    def _messages_for_record(self, domain):
```
TDu, line 56: Could you explain what it does ?

TDu (commented):
It needs some unit tests

mail_thread_create_nolog/models/mail_message.py
sebalix: I'll add a docstring yes.
It gets the record from the domain, domain used by Odoo web client to get the messages for that record when it displays the chatter. But the `message_fetch` method is also used to retrieve messages of channels, outside of any record context. In that case, no record = no fake-create-message.

sebalix:
@TDu docstring, comments and tests added

mail_thread_create_nolog/models/mail_thread.py
```diff
@@ -0,0 +1,18 @@
...
+
+from odoo import api, models
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        # Overloaded to not create 'Record created' messages that increase
```
simahawk:
```suggestion
        # Overridden to not create 'Record created' messages that increase
```

mail_thread_create_nolog/models/mail_thread.py
```diff
@@ -0,0 +1,18 @@
...
+        # the size of 'mail_message' table for little value.
+        # Instead a message will be generated on the fly when the Odoo client
+        # will retrieve the list of messages for a given record.
+        return super(MailThread, self.with_context(mail_create_nolog=True)).create(
```
simahawk, line 16: I would make this configurable via UI and maybe by model. Worth to add it to the ROADMAP

mail_thread_create_nolog/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+{
+    "name": "Mail Thread - Fake create log",
+    "summary": "Display a fake (non-stored) create log in the chatter.",
+    "version": "14.0.1.0.0",
+    "category": "Mail",
+    "website": "https://github.com/OCA/social",
```
simahawk, line 8: dev status: alpha ;)

simahawk (approved):
LG at first implementation.
I would spend some more words on the to explain the hack tho.

mail_thread_create_nolog/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+{
+    "name": "Mail Thread - Fake create log",
```
mmequignon, line 4: is it nolog or fake logs?
I would align module / manifest names

mail_thread_create_nolog/models/mail_message.py
```diff
@@ -0,0 +1,65 @@
...
+                    "author_id": (author.id, author.name),
+                }
+            )
+            res.append(create_message_values)
+        return res
+
+    def _get_create_message_id(self, domain):
+        # Return the negative version of the greatest message ID
+        # of the current record
+        return -self.search(domain, limit=1, order="id DESC").id
```
mmequignon, line 53: if search returns an empty recordset, you'll return `-False` which is `0`.
Is it ok?

mail_thread_create_nolog/models/mail_message.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @api.model
+    def _messages_for_record(self, domain):
+        """Return the record from the domain used in `message_fetch`."""
+        model = res_id = None
+        for part in domain:
+            if len(part) == 3 and part[0] == "model" and part[1] == "=":
+                model = part[2]
+            if len(part) == 3 and part[0] == "res_id" and part[1] == "=":
+                res_id = part[2]
```
mmequignon, line 63: you could probably break or return here, once both model / res_id are set.

mail_thread_create_nolog/models/mail_thread.py
sebalix: Yes I was thinking the same...

mail_thread_create_nolog/models/mail_message.py
sebalix: Yes, Odoo client doesn't complain

mail_thread_create_nolog/__manifest__.py
sebalix: From user/admin POV it is a fake log (which replaces the real log in the DB), and to not duplicate information on UI the real log is of course not created through `mail_create_nolog` context key.
But other feedbacks are welcomed, it's not easy to name a module and give it a good name/description.

mail_thread_create_nolog/models/mail_message.py
sebalix: Improved the comment

mail_thread_create_nolog/models/mail_message.py
sebalix: Done

sebalix:
I added a fix to check if the data model of the record has the `_log_access` enabled (so we are sure it has `create_uid`/`create_date` fields), `queue.job` has this option disabled for instance.

simahawk:
@sebalix good to go?

sebalix:
@sebalix good to go, in production since a while now, no issue so far.

PR #1138, opened 2023-04-24
[15.0][FIX] mail_tracking: Index error

Gladys-Haelters:
in multiple cases we get the the following error:
IndexError: tuple index out of range

In the current code for mail_tracking_email the index is not set correctly, 
X3] does not exist/has no value in the _find_allowed_tracking_ids method

        return [
            x[0]
            for x in msg_linked
            if (x[1] in msg_ids)  # We can read the linked message
            or (
                not any({x[1], x[2]}) and x[3] in partner_ids
            )  # No linked msg/mail but we can read the linked partner
            or (not any({x[1], x[2], x[3]}))  # No linked record
        ]
       
instead of just fixing the indexes, i opted to rewrite a small part of this code for readability
resulting in fixing the problem and making it more clear what is happening.

As for the tracking_emails_count. The counter only shows the mail count for the admin, 
If users get rights to view mailtracking their counter will still say 0, even if there are mails behind te smartbutton,
which is in my honest opinion misleading.
so i currently took the counter out because it is not funcion like it should.

It could also  work to just take out :
```
   if not self.env.user.has_group("base.group_system"):
                continue

```
like i did in my PR for the  14.0 branch, but since there is the comment that it is for performance, i opted to leave the counter out in this case.

### DIFF
--- a/mail_tracking/__manifest__.py
+++ b/mail_tracking/__manifest__.py
@@ -7,10 +7,10 @@
 {
     "name": "Email tracking",
     "summary": "Email tracking system for all mails sent",
-    "version": "15.0.2.0.2",
+    "version": "15.0.2.0.3",
     "category": "Social Network",
     "website": "https://github.com/OCA/social",
-    "author": ("Tecnativa, " "Odoo Community Association (OCA)"),
+    "author": ("Tecnativa", "Odoo Community Association (OCA)"),
     "license": "AGPL-3",
     "application": False,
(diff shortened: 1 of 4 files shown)


pedrobaeza (changes requested):
Same comments than in 14.0 patch and you mustn't do several things in the same PR. Removing fields is not correct.

pedrobaeza:
Fixed properly in #1149

PR #1141, opened 2023-05-09
[15.0][FIX] mail_tracking: thread permissions

chienandalu:
If a user tries to read a thread on a record and one message is from a partner on which he has no permissions, there will be an exception as this method tries to fetch info from such partner.

cc @Tecnativa TT43075

please review @sergio-teruel @carlosdauden


sergio-teruel (approved):

carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1148, opened 2023-05-22
[14.0][FIX] mail_tracking: allowed trackings compute

chienandalu:
In the forward port of the permission check, a column was dropped from the tuple, but the following code wasn't adapted to the new tuple length. For clarity sake we also name the tuple unpacking into proper variables.

cc @Tecnativa TT43453

ping @pedrobaeza


pedrobaeza (approved):
Too much time with this problem, so let's fast-track it:

/ocabot merge patch

baba75 (commented):
It works!

PR #1149, opened 2023-05-22
[15.0][FIX] mail_tracking: allowed trackings compute

chienandalu:
fw of https://github.com/OCA/social/pull/1148

In the forward port of the permission check, a column was dropped from the tuple, but the following code wasn't adapted to the new tuple length. For clarity sake we also name the tuple unpacking into proper variables.

cc @Tecnativa TT43453

ping @pedrobaeza


pedrobaeza (approved):
Fast-tracking as well:

/ocabot merge patch

PR #1151, opened 2023-05-29
[15.0][IMP] mail_tracking: New config parameter to disable the tracking image

dhx:
- As the tracker img in the body is currently also used to set the X-Odoo-MailTracking-ID mail header we can remove the tag only after this is done
- Added tests to check for the presence and absence of the tracking image tag depending on the new configuration parameter

### DIFF
--- a/mail_tracking/models/ir_mail_server.py
+++ b/mail_tracking/models/ir_mail_server.py
@@ -23,6 +23,21 @@ class IrMailServer(models.Model):
         match = re.search(r'<img[^>]*data-odoo-tracking-email=["\']([0-9]*)["\']', body)
         return str(match.group(1)) if match and match.group(1) else False
 
+    def _tracking_img_disabled(self, tracking_email_id):
+        # while tracking_email_id is not needed in this implementation, it can
+        # be useful for other addons extending this function to make a more
+        # fine-grained decision
+        return (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("mail_tracking.tracking_img_disabled", False)
+        )
+
+    def _tracking_img_remove(self, body):
+        return re.sub(
+            r'<img[^>]*data-odoo-tracking-email=["\'][0-9]*["\'][^>]*>', "", body
+        )
+
     def build_email(
         self,
         email_from,
@@ -41,9 +56,28 @@ class IrMailServer(models.Model):
         body_alternative=None,
         subtype_alternative="plain",
     ):
+        # Unfortunately we currently have to extract the mail.tracking.email
+        # record id from the tracking image in the body here as the core
+        # mail module does not allow headers to be inserted in the
+        # MailMail._send_prepare_values function.
+        # Things to consider before refactoring this:
+        # - There are third party modules completely replacing the
+        #   MailMail._send function, so even when a future version
+        #   of the core mail module supports adding headers there, we might
+        #   want to wait a little until this feature has trickled through.
+        # - While it would be possible to find the mail.tracking.email
+        #   record via the message_id and the email_to criteria, this
+        #   would rely on the message having no duplicate recipient
+        #   (e.g. different contacts having the same email address) and
+        #   no other module inheriting the _send_prepare_values function
+        #   modifying the email_to parameter.
         tracking_email_id = self._tracking_email_id_body_get(body)
         if tracking_email_id:
             headers = self._tracking_headers_add(tracking_email_id, headers)
+            # Only after the X-Odoo-MailTracking-ID header is set we can remove
+            # the tracking image in case it's to be disabled
+            if self._tracking_img_disabled(tracking_email_id):
+                body = self._tracking_img_remove(body)
         msg = super(IrMailServer, self).build_email(
             email_from=email_from,
             email_to=email_to,
--- a/mail_tracking/README.rst
+++ b/mail_tracking/README.rst
@@ -43,6 +43,15 @@ you need to add ``mail_tracking`` addon to wide load addons list
 (by default, only ``web`` addon), setting ``--load`` option.
 For example, ``--load=web,mail_tracking``
 
+Configuration
+=============
+
+As there can be scenarios where sending a tracking img in the email body is
+not desired, there is a global system parameter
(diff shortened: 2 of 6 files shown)


mail_tracking/readme/USAGE.rst
```diff
@@ -73,3 +73,9 @@ with messages in failed status and that needs an user action.
 * Filter
 
   .. image:: ../static/img/failed_message_filter.png
+
+As there can be scenarios where sending a tracking img in the email body is
```
pedrobaeza: This should go in CONFIGURE.rst.

mail_tracking/models/ir_mail_server.py
```diff
@@ -23,6 +23,21 @@ def _tracking_email_id_body_get(self, body):
...
+        # be useful for other addons extending this function to make a more
+        # fine-grained decision
+        return (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("mail_tracking.tracking_img_disabled", False)
+        )
+
+    def _tracking_img_remove(self, body):
+        return re.sub(
```
pedrobaeza, line 37: Please don't do it this way, having to parse the mail. Simply don't add the tracking where it belongs.

pedrobaeza (changes requested):

mail_tracking/models/ir_mail_server.py
dhx: @pedrobaeza thanks for the feedback I'd also prefer to do it differently, but as mentioned in the commit message unfortunately:
... the tracker img in the body is currently also used to set the X-Odoo-MailTracking-ID mail header we can remove the tag only after this is done

Here the header is added with the tracking_email_id parameter extracted from the image tag: https://github.com/OCA/social/blob/15.0/mail_tracking/models/ir_mail_server.py#L44

mail_tracking/readme/USAGE.rst
dhx: Thanks, moved it!

mail_tracking/models/ir_mail_server.py
pedrobaeza: Uhm, then it's Odoo who adds the tracking header, not this module, so I'm not sure we should include in this module the deactivation of it. It should be a separate module `mail_no_tracking_image` or similar. I know it will affect this one, but you can:

- Change here the way database and Tracking ID is extracted.
- Add a glue module.

dhx: Thanks @pedrobaeza for your feedback, I also wished the current mail_tracking codebase was more straightforward here explaining the reasons for using the img tag to extract the id of the relevant mail.tracking.email record - to me it seems it has to do with the patched _send_prepare_values function not allowing to also return headers (so only the body can be used to store the id when creating the tracking record at this stage):
https://github.com/oca/social/blob/15.0/mail_tracking/models/mail_mail.py#L37

As this mail.tracking.email model is only added by this mail_tracking oca module I'd rule out that this header or the `` tag are added/used by the core odoo modules as they contain the id of this model (I also could not find anything in the current odoo core codebase when searching for the header or the 'data-odoo-tracking-email' attribute used in the `` tag).

pedrobaeza: But seeing the code, you can do other techniques not implying to use the  tag, so you should do a little refactoring for housing this new feature, or it will be patch over patch, and the code will become unmaintainable.

dhx: Instead of adding the `` tag in the `_send_prepare_values` function we could add a `<link>` tag to store the mail.tracking.email id in the body and then later in the inherited `build_email` function remove this `<link>` tag, set the `X-Odoo-MailTracking-ID` header and add the `` (in case the configuration does not prevent it).
What do you think @pedrobaeza ?

pedrobaeza: Isn't the possibility to directly pass that data without having to include them in the mail and later extract them?

dhx: AFAICT that would require a patch in the core mail module like this:
```patch
diff --git a/addons/mail/models/mail_mail.py b/addons/mail/models/mail_mail.py
index 1a3fa6e55c5..63a1a73ec28 100644
--- a/addons/mail/models/mail_mail.py
+++ b/addons/mail/models/mail_mail.py
@@ -458,7 +458,7 @@ class MailMail(models.Model):
                         object_id=mail.res_id and ('%s-%s' % (mail.res_id, mail.model)),
                         subtype='html',
                         subtype_alternative='plain',
-                        headers=headers)
+                        headers=dict(headers, **email.get('headers', {})))
                     processing_pid = email.pop("partner_id", None)
                     try:
                         res = IrMailServer.send_email(
```

Seems there are third party addons completely overriding this function (the cc bcc from webkul I found with a fulltext search here e.g.: https://github.com/planetodooofficial/toabh/blob/main/odoo_email_cc_bcc/models/compose_mail.py#L369 ) so relying on this would result in a decreased compatibility I'm afraid.

Another approach would be to not pass the id of the tracking record at all and later in the build_email function search for the created mail.tracking.email record in the db with the mail.message_id and the email_to recipient, but this is also not bulletproof when there are e.g. multiple partners with the same email address and it relies on there being no other `_send_prepare_values` override modifying the email_to key.

Or @pedrobaeza do you see any other angle to approach this?

pedrobaeza: OK, I see the problem... Then let's go with your current approach as the less bad one, but it would be very interesting to reflect a summary of all of this reflections as a comment on the code for not having to come again to it on the future.

dhx: @pedrobaeza please have a look, any suggestions are welcome

pedrobaeza (approved):

dhx:
Hi @jans23 seen you where asking for this feature, so maybe you want to try this PR and add a review so we can get it merged :)

jonasbuchholz (approved):

jans23:
@dhx That's great news. However we don't have 15.0 up and running yet. I hope we will be there in a few weeks so that I can test it.

PR #1152, opened 2023-05-31
[13.0] mass_mailing_contact_active: Fix default value of active field in mass mailing subscription

grindtildeath:
Add default value in DB for active column.

Model mailing.contact.subscription is persisted using database table mailing_contact_list_rel.
This model is used for following O2m fields:
  - mailing.list.subscription_ids
  - mailing.contact.subscription_list_ids

However it is also used to materialize a M2m relation on following fields:
  - mailing.list.contact_ids
  - mailing.contact.list_ids

If records are added using O2m fields, the ORM will properly get the default value defined for the active field before inserting in the table. However, if records are added using M2m fields, the ORM will only use the foreign keys to insert since it's an associative table.

By defining a TRUE value as default for active field, we ensure the records added using the M2m fields will be active by default.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,18 +36,18 @@ jobs:
       matrix:
         include:
           - container: ghcr.io/oca/oca-ci/py3.6-odoo13.0:latest
-            include: "mail_restrict_follower_selection,mass_mailing_contact_active,mass_mailing_partner_contact_active"
+            include: "mail_restrict_follower_selection"
             makepot: "true"
             name: test with Odoo
           - container: ghcr.io/oca/oca-ci/py3.6-ocb13.0:latest
-            include: "mail_restrict_follower_selection,mass_mailing_contact_active,mass_mailing_partner_contact_active"
+            include: "mail_restrict_follower_selection"
             name: test with OCB
           - container: ghcr.io/oca/oca-ci/py3.6-odoo13.0:latest
-            exclude: "mail_restrict_follower_selection,mass_mailing_contact_active,mass_mailing_partner_contact_active"
+            exclude: "mail_restrict_follower_selection"
             makepot: "true"
             name: test with Odoo
           - container: ghcr.io/oca/oca-ci/py3.6-ocb13.0:latest
-            exclude: "mail_restrict_follower_selection,mass_mailing_contact_active,mass_mailing_partner_contact_active"
+            exclude: "mail_restrict_follower_selection"
             name: test with OCB
     services:
       postgres:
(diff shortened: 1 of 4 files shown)


grindtildeath:
@pedrobaeza @yajo Looks like the reason why this fix is needed was the same reason why I had to segregate the testing on original PR with conflict on mass_mailing_custom_unsubscribe module: https://github.com/OCA/social/pull/871#issuecomment-1332144484
However, it was now breaking a test of mass_mailing_list_dynamic, that I could solve by forcing the propagation of the context key to avoid raising the ValidationError in the constraint.

Can you guys please have a look? :pray:

pedrobaeza (approved):

mass_mailing_contact_active/hooks.py
```diff
@@ -0,0 +1,28 @@
...
+    If records are added using O2m fields, the ORM will properly get the default
+    value defined for the active field before inserting in the table.
+    However, if records are added using M2m fields, the ORM will only use the foreign
+    keys to insert since it's an associative table.
+
+    By defining a TRUE value as default for active field, we ensure the records
+    added using the M2m fields will be active by default.
+    """
+    cr.execute(
+        """ALTER TABLE ONLY mailing_contact_list_rel ALTER COLUMN active SET DEFAULT TRUE;"""
```
yajo: You can move this to [the model's `init()` method](https://github.com/odoo/odoo/blob/9b9859a2f901cf4a434bd766ce86e91ea3cb4db8/odoo/models.py#L2561-L2565).

If you do so, this file can be removed.

mass_mailing_contact_active/migrations/13.0.1.1.0/post-migrate.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+
+def migrate(cr, version):
+    cr.execute(
+        """ALTER TABLE ONLY mailing_contact_list_rel ALTER COLUMN active SET DEFAULT TRUE;"""
```
yajo: If you follow my previous suggestion, this file can be removed too.

yajo (changes requested):

grindtildeath:
@yajo thx for the review, comments attended

TDu (approved):

yajo (approved):
Please squash the commits to merge. The bot can't do that yet https://github.com/OCA/oca-github-bot/issues/43.

grindtildeath:
@pedrobaeza seems the bot was out for lunch the last time you commented

PR #1160, opened 2023-06-08
[ADD] mail_composer_cc_bcc: send email to cc, bcc addresses

hailangvn2023:
The native Odoo Mail Composer doesn't support CC fields out-of-the-box (even
when defined on mail.template records): email_to and email_cc are merged in
a single Recipients field. And mails are sent for each separately, which is
quite confusing for users used to the CC semantic.

This module adds cc and bcc fields to corresponding models then allows
sending only one email to multiple recipients.


Saran440:
Hi @hailangvn2023,

This module is useful. :+1: 

I would like to explore the possibility of changing the fields 'cc' and 'bcc' to text fields.
The reason for this change is that in certain cases,
there is a need to include email addresses in the cc field that are not registered in the system.

It similar with PR https://github.com/OCA/social/pull/986

What do you think?

hailangvn:
> The reason for this change is that in certain cases,
> there is a need to include email addresses in the cc field that are not registered in the system.

Thanks for your comment. Is it possible to provide the specific cases?
Why can't we simply add the email to partner list?

Saran440:
Example
Send invoicing via email to partner but sometime I need cc to accountant of company partner or CEO or my other email.
I think it shouldn't add partner because maybe we mail cc 1 time.

hailangvn:
> Example Send invoicing via email to partner but sometime I need cc to accountant of company partner or CEO or my other email. I think it shouldn't add partner because maybe we mail cc 1 time.

So sorry that this module could not cover those edge cases. As you can see,
this module adds default cc and bcc partners to company settings so that
the partners are added automatically to mail compose form most of the times.

cyrilmanuel (approved):

mail_composer_cc_bcc/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    "website": "https://github.com/OCA/social",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
+    "maintainers": ["hailangvn2023"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "preloadable": True,
+    "depends": [
+        "mail",
+        "sale_management",
```
Saran440:
```suggestion
```

Do this module need depend module `sale_management` ?

Saran440 (approved):
Functional Test :+1: 

Have a little comment

mail_composer_cc_bcc/__manifest__.py
hailangvn: You are right. The PR has been updated.

hailangvn: Thanks for your review.

gurneyalex (approved):

PR #1161, opened 2023-06-08
Migrate mail attach existing attachment account to 15.0

oadiazp-processcontrol:


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0.

On the functional side, is this really needed in this version?

PR #1166, opened 2023-06-15
[14.0] [MIG] mail_force_queue

Kiplangatdan:


Kiplangatdan:
@thomaspaulb Please review

mail_force_queue/models/mail_thread.py
```diff
@@ -0,0 +1,35 @@
...
+        msg_vals=None,
+        model_description=False,
+        mail_auto_delete=True,
+        check_existing=False,
+        force_send=True,
+        send_after_commit=True,
+        **kwargs,
+    ):
+        if self.env.user.company_id.force_mail_queue:
+            force_send = False
```
thomaspaulb, line 23: This can still be overridden by the [context key](https://github.com/OCA/OCB/blob/14.0/addons/mail/models/mail_thread.py#L2253), so, if you want to be sure it's set to False always, better set the context key itself.

mail_force_queue/models/mail_thread.py
```diff
@@ -0,0 +1,35 @@
...
+
+        return super()._notify_record_by_email(
+            message,
+            recipients_data,
+            msg_vals=msg_vals,
+            model_description=model_description,
+            mail_auto_delete=mail_auto_delete,
+            check_existing=check_existing,
+            force_send=force_send,
+            send_after_commit=send_after_commit,
```
thomaspaulb, line 33: If you're not actually touching a named variable in your function, you can remove it in the function definition also, and it will be passed on as part of `**kwargs`

mail_force_queue/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+Force outgoing emails to be queued instead of sent immediately.
+Queued emails are sent by 'Email Queue Manager' cron.
+
+This also solves possible TransactionRollbackError while deleting outgoing email (see `https://github.com/odoo/odoo/issues/22148
+<https://github.com/odoo/odoo/issues/22148>`_.)
```
thomaspaulb, line 5: Maybe change this to:

```
In v10.0, this module solved possible TransactionRollbackError while deleting outgoing email (see `https://github.com/odoo/odoo/issues/22148
<https://github.com/odoo/odoo/issues/22148>`_.)

In v14.0, above problem should have been solved, but we still ran into cases with duplicate emails being sent out: first email on the original transaction, second email on the retried transaction after TransactionRollbackError. In our case, it was caused by concurrent transactions triggering the same write to the `message_main_attachment_id` field of an `account.move` record. We did not get to the bottom of this, but forward-ported this module as a workaround for these and other cases that may still exist.
```

thomaspaulb (changes requested):

thomaspaulb:
@Kiplangatdan ping

PR #1172, opened 2023-06-21
[ADD] mass_mailing_batch

SirTakobi:
CI will be fixed after https://github.com/OCA/oca-addons-repo-template/pull/199


mass_mailing_batch/models/mass_mailing.py
```diff
@@ -0,0 +1,29 @@
...
+
+class MassMailing (models.Model):
+    _inherit = 'mail.mass_mailing'
+
+    def get_remaining_recipients(self):
+        recipients_res_ids = super(MassMailing, self).get_remaining_recipients()
+        batch_size = self.env['ir.config_parameter'] \
+            .sudo() \
+            .get_param(
+                'mass_mailing_batch.size',
```
eLBati, line 15: @SirTakobi I would name this parameter `mass_mailing.batch_size`

eLBati (changes requested):

mass_mailing_batch/models/mass_mailing.py
SirTakobi: Thanks for the review!
I prefer to keep the module name in the parameter key, in order to avoid clashing with other parameter names.
If you really don't like the key right now, maybe I could change it to `mass_mailing_batch.batch_size`, what do you think?

eLBati: OK `mass_mailing_batch.size`

eLBati (approved):

PR #1174, opened 2023-06-21
[15.0][IMP] mail_quoted_reply: set email author as default reply partner

jonasbuchholz:
Thus far, the mail_message.partner_ids field was used for default partner_ids when opening the composer from the reply button. However, for most messages, partner_ids will not be set. When replying to a message, it is most intuitive for users if the default recipient of the message will be the author of the original message.
Instead of using the author_id field, we're using the find_or_create method of res.partner to ensure that a default partner is set. The author_id field will be empty if the email address does not match any existing partners.

### DIFF
--- a/mail_quoted_reply/models/mail_message.py
+++ b/mail_quoted_reply/models/mail_message.py
@@ -29,6 +29,9 @@ class MailMessage(models.Model):
             body=self.body,
         )
 
+    def _default_reply_partner(self):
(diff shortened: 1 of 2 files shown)


dhx (approved):

dhx:
@StefanRijnhart see you made the last commit for this module in the 15.0 branch, think this improves the behaviour of the reply function, what do you think?

PR #1178, opened 2023-06-30
[15.0] Add mail_outgoing_server_by_model

mmequignon:
Allow to set outgoing mail servers per model.

### DIFF
--- a/outgoing_email_by_model/README.rst
+++ b/outgoing_email_by_model/README.rst
@@ -1,0 +1,1 @@
+wait 4 da boat
--- a/outgoing_email_by_model/__init__.py
+++ b/outgoing_email_by_model/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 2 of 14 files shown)


mmequignon:
No idea why tests are failing.
I'll check that.

mail_outgoing_server_by_model/models/ir_model.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+OUTGOING_MAILSERVER_DESCRIPTION = """
+    Allows to force the usage of a given outgoin mail server if this setting is set.
```
TDu:
```suggestion
    Allows to force the usage of a given outgoing mail server if this setting is set.
```

mail_outgoing_server_by_model/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+Allow to select an outgoing mail server per model.
+
+To make it work, the model must inherit `mail.thread`.
```
TDu: It seems to only be active for comment on the mail.thread but not when sending a report, would be good to mention that.

PR #1185, opened 2023-07-19
[15.0][FIX] mail_quoted_reply: avoid rendering reply icon multiple times

jonasbuchholz:
Sometimes, the quoted reply icon is rendered and appended to the message action list multiple times in the chatter. With this fix, the icon is added to the action list only if it is not already contained within the list.

### DIFF
--- a/mail_quoted_reply/static/src/components/mail_message_reply.esm.js
+++ b/mail_quoted_reply/static/src/components/mail_message_reply.esm.js
@@ -30,7 +30,12 @@ patch(
                     );
                 }
                 actionLists.forEach((actionList) => {
-                    actionList.append(MessageQuotedReplyIcon);
+                    if (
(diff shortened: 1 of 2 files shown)


dhx (approved):
LGTM

em230418:
> Sometimes, the quoted reply icon is rendered and appended to the message action list multiple times in the chatter.

Posted how to reproduce this issue here.
https://github.com/OCA/social/issues/1239

Applied patch from this PR. It did not work. That added if statement is always true.

jonasbuchholz:
@em230418 thanks for your feedback, the fix did indeed not work. I corrected the if condition, it should work as intended now.

em230418:
As for now, I am using my own fix on my instance.
https://github.com/OCA/social/pull/1242

etobella (approved):

PR #1194, opened 2023-08-08
[15.0][ADD] mail_thread_subject_ignore

AaronHForgeFlow:
This module allows to avoid fetching mails that are supposed to be automatic replies, this is useful to prevent records like leads, helpdesk tickets, etc to have the chatter full of those automatic replies.

cc @ForgeFlow


DavidJForgeFlow (approved):
LGTM!

mail_thread_subject_ignore/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+
+{
+    "name": "Mail Thread Subject Ignore",
+    "version": "15.0.1.0.0",
+    "license": "LGPL-3",
+    "category": "Mail",
+    "summary": """Ignore Mails with Subject""",
+    "website": "https://github.com/OCA/social",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "depends": ["fetchmail", "test_mail"],
```
MiquelRForgeFlow:
```suggestion
    "depends": ["fetchmail"],
```

AaronHForgeFlow: Change done

mail_thread_subject_ignore/models/mail_thread.py
```diff
@@ -0,0 +1,65 @@
...
+    def is_automated_reply(self, subject):
+        automated_keywords = (
+            self.env["mail.thread.subject.ignore"].search([]).mapped("name")
+        )
+        if not automated_keywords:
+            return False
+        pattern = re.compile(
+            r"|".join(map(re.escape, automated_keywords)), re.IGNORECASE
+        )
+        if subject and pattern.search(subject):
```
hbrunn, line 27: why a substring search and not `match` or hooking the patterns above?

hbrunn (changes requested):
while I see how this makes sense, are you aware of [mail_autogenerated_header](https://github.com/OCA/social/tree/12.0/mail_autogenerated_header)?

Systems that generate automated mails are supposed to generate this header, so with this you can filter out all mails marked as automatically sent. And it makes Odoo behave correctly in this case for other systems too.

AaronHForgeFlow:
Interesting module mail_autogenerated_header. I understand that the header in those automatic replies is usually properly configured so we can filter them. I will try that one.

PR #1196, opened 2023-08-08
[15.0][IMP] mail_activity_team: add multi-company check

DavidJForgeFlow:
When having two users in a activity team from different companies, it will raise an Exception of multicompany when adding the team to a document. This fix ensures that all the users in a team are from the same company (at least one of the companies available).

@ForgeFlow

### DIFF
--- a/mail_activity_team/models/mail_activity_team.py
+++ b/mail_activity_team/models/mail_activity_team.py
@@ -1,7 +1,8 @@
 # Copyright 2018-22 ForgeFlow S.L.
 # Copyright 2021 Sodexis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.odoo.exceptions import ValidationError
 
 
 class MailActivityTeam(models.Model):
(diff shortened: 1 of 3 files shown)


etobella (commented):
I don't like the option to add a required company field on teams. It might get confusing in some multicompany environments

AaronHForgeFlow:
My use case is the following. I am using activity teams "Helpdesk company1" and "Helpdesk company2". All the members of the team are assigned to new incoming tickets. If I assign a member of "Helpdesk company2" to a ticket in company1 the user will get access error, which is not desirable.  The only way to distinguish between the teams is the name of the team, and that leads to hardcode things.

AaronHForgeFlow:
Actually, I can filter members by company so I think this is not necessary at this point. I think we can close this.

mail_activity_team/models/mail_activity_team.py
```diff
@@ -1,7 +1,8 @@
 # Copyright 2018-22 ForgeFlow S.L.
 # Copyright 2021 Sodexis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.odoo.exceptions import ValidationError
```
MiquelRForgeFlow, line 5:
```suggestion
from odoo.exceptions import ValidationError
```

MiquelRForgeFlow (changes requested):

AaronHForgeFlow:
I think this can be closed, not needed for now.

PR #1198, opened 2023-08-10
[15.0][FIX]mail_composer_cc_bcc: avoid sending multiple mails to all recipients when no cc or bcc is set

ChristophAbenthungCibex:
It can happen that the same email will be sent multiple times to the same recipient.

Steps to recreate the error:
- Open a composer (for example from the chatter for a single contact or a single task in project)
- Enter multiple recipients in the to-field
- Leave the cc- and bcc-field empty
- Fill out subject and body
- Send the mail

The mail will be sent to each recipient as many times as there are recipients. For example: if there are 3 recipients in the to-field, each recipient gets the same mail 3 times.

### DIFF
--- a/mail_composer_cc_bcc/models/mail_mail.py
+++ b/mail_composer_cc_bcc/models/mail_mail.py
@@ -34,7 +34,8 @@ class MailMail(models.Model):
         ICP = env["ir.config_parameter"].sudo()
         # Mail composer only sends 1 mail at a time.
         is_out_of_scope = len(self.ids) > 1
-        if is_out_of_scope or not (self.email_cc or self.email_bcc):
+        is_from_composer = self.env.context.get("is_from_composer", False)
+        if is_out_of_scope or not is_from_composer:
             return super()._send(
                 auto_commit=auto_commit,
                 raise_exception=raise_exception,
--- a/mail_composer_cc_bcc/tests/test_mail_cc_bcc.py
+++ b/mail_composer_cc_bcc/tests/test_mail_cc_bcc.py
@@ -178,3 +178,24 @@ Test Template<br></p>""",
         # Only 4 partners notified
         self.assertEqual(len(message.notified_partner_ids), 4)
         self.assertEqual(len(message.notification_ids), 4)
+
+    def test_mail_without_cc_bcc(self):
+        # Test without any partner in cc/bcc -> only 1 mail should be sent
+        self.set_company()
+        form = self.open_mail_composer_form()
+        subject = "Testing without cc/bcc single mail"
+        form.subject = subject
+        composer = form.save()
+        composer.partner_cc_ids = None
+        composer.partner_bcc_ids = None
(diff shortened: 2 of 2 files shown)


nilshamerlinck (approved):

ajaniszewska-dev (approved):

AaronHForgeFlow (approved):

AaronHForgeFlow (changes requested):
After done additional tests it seems the issue is still there. Looking closer to the code changes I see that when the message is sent from the composer the relevant code is not inside the if, so the fix proposed is not fixing the issue.

nilshamerlinck:
Hello @AaronHForgeFlow 

This fix is correct for the "no cc or bcc set" case: it makes sure that if _is_from_composer_ then custom code is used, regardless of cc/bcc values. This is needed because other parts of the code are also customized when _is_from_composer_.

However maybe you found a different case that would also need fixing?

AaronHForgeFlow:
@nilshamerlinck , ok, I will not block the change, it fixes one issue.

There is another issue when using the mail composer and not adding cc nor bcc. Duplicate emails are sent in this case.

AaronHForgeFlow (approved):

nilshamerlinck:
@AaronHForgeFlow is there maybe another module installed in your test db that also overrides `mail.mail:_send()`?

AaronHForgeFlow:
@nilshamerlinck, no, I am not overriding that method in a third module. I will do more testing just to be sure it is this module.

MiquelRForgeFlow (approved):

simahawk:
> @nilshamerlinck, no, I am not overriding that method in a third module. I will do more testing just to be sure it is this module.

@AaronHForgeFlow can we move on?

AaronHForgeFlow:
@simahawk yes, this solves the issue when cc- and bcc-fields are empty using the mail composer.

When not using the mail composer I am having similar issue, but can't say it is this module.

simahawk:
/ocabot merge patch

@ChristophAbenthungCibex pls next time keep the commit msg short and keep the explanation in the description.
Example:
```
[FIX] mail_composer_cc_bcc: fix send without cc/bcc

When CC and BCC are not valued, emails are sent to all recipients.
```
Guidelines https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

ChristophAbenthungCibex:
> /ocabot merge patch
> 
> @ChristophAbenthungCibex pls next time keep the commit msg short and keep the explanation in the description. Example:
> 
> ```
> [FIX] mail_composer_cc_bcc: fix send without cc/bcc
> 
> When CC and BCC are not valued, emails are sent to all recipients.
> ```
> 
> Guidelines https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

@simahawk Thx for the example and letting me know. I read the guidelines and i shortened the commit message as a result. Next time I try to improve that even further.

PR #1201, opened 2023-08-13
[15.0][FIX] mail_composer_cc_bcc

nilshamerlinck:
- One of the main goals of `mail_composer_cc_bcc` is to get the Mail Composer wizard to behave like a normal Mail Client
- Concretely, instead of sending one email per partner as Odoo does by default in [`mail.mail:_send()`](https://github.com/odoo/odoo/blob/15.0/addons/mail/models/mail_mail.py#L394), it only [sends one](https://github.com/oca/social/blob/15.0/mail_composer_cc_bcc/models/mail_mail.py#L71) so that headers To: and Cc: would be shown with all recipients, as end users expect
- But in the case where no Cc/Bcc is defined, `To` header is modified by `mail_composer_cc_bcc` while the original `mail.mail:_send()` is still used
- As a consequence, both behaviors are combined and multiple emails are sent to the same recipients :(
- This PR fixes this bug, by relying on the context-key `is_from_composer` to be consistent with the rest of the code

### DIFF
--- a/mail_composer_cc_bcc/models/mail_mail.py
+++ b/mail_composer_cc_bcc/models/mail_mail.py
@@ -28,18 +28,18 @@ class MailMail(models.Model):
     def _send(  # noqa: max-complexity: 4
         self, auto_commit=False, raise_exception=False, smtp_session=None
     ):
-        env = self.env
-        IrMailServer = env["ir.mail_server"]
-        IrAttachment = env["ir.attachment"]
-        ICP = env["ir.config_parameter"].sudo()
-        # Mail composer only sends 1 mail at a time.
+        is_from_composer = self.env.context.get("is_from_composer", False)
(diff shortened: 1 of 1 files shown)


TDu (approved):
LG, would have been interesting to have some test coverage.

ajaniszewska-dev (approved):
LG

ChristophAbenthungCibex:
Hi @nilshamerlinck  it seems we worked on the same error. I pushed a slightly different solution with test coverage. See #1198 . It should behave the same as your solution.

nilshamerlinck:
Hello @ChristophAbenthungCibex seems so indeed :smile: 
equivalent solution but you included a much welcomed test, so I will close my PR in favor of yours :+1: 
thanks!

PR #1213, opened 2023-09-01
[MIG+IMP] mail_post_defer: migrate to 16.0, allow editing messages

yajo:
Rely on the new `mail.message.schedule` model and follow the rest of the refactor from https://github.com/odoo/odoo/pull/95623. FWIW this cron isn't strictly needed to run every minute, as it runs at the specific time automatically.

Follow refactors from https://github.com/odoo/odoo/pull/95500 and https://github.com/odoo/odoo/pull/79259 for the client.

Improve the module by allowing to edit messages, instead of just deleting them. This was simpler than splitting the edit and delete checks in the client-side.

@moduon MT-3088


rafaelbn (approved):
👍🏼 OM

PR #1191, opened 2023-07-27
[16.0][MIG] mail_tracking: Migration to 16.0

payen000:

PR #1091, opened 2023-03-07
[16.0][MIG] mail_tracking: Migration to 16.0

maq-adhoc:
I restart the mail tracking migration, taking changes from PR https://github.com/OCA/social/pull/1029 and adding my own changes. 
I restarted to be able to do it with the odoo-module-migrate tool and the commits are clean and include PRs #1053 ,#1056 ,#1057.  
I also modified the way we inherit MessagingInitializer so that it does not conflict with other modules, e.g. Voip.

PR #1109, opened 2023-03-22
[15.0][IMP] mail_tracking: Add option to keep aliases in mail tracking (forward port)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/social/pull/941.

PR #1216, opened 2023-09-05
[16.0][MIG] mail_tracking: Migration to 16.0

payen000:
Redo of PR https://github.com/OCA/social/pull/1191; this PR is opened due to inactivity of https://github.com/OCA/social/pull/1091 (from which the `MessagingInitializer._init` fix was taken).


payen000:
Hello @luisg123v, could you review this, please?

chienandalu (changes requested):
Some serious issue when trying to test it functionally:

- Whenever a enter a record I get an error when it tries to load the failed messages tracker:
<details>
TypeError: Cannot read properties of undefined (reading 'refreshMessagefailed')
    at model._onThreadIdOrThreadModelChanged (http://oca-social-16-0-pr1216-13154c4fcbef.runboat.odoo-community.org/web/assets/debug/web.assets_backend.js:176759:25) (/mail_tracking/static/src/js/models/chatter.esm.js:20)
    at model._onThreadIdOrThreadModelChanged (http://oca-social-16-0-pr1216-13154c4fcbef.runboat.odoo-community.org/web/assets/debug/web.assets_backend.js:130691:31) (/mail/static/src/model/model_core.js:371)
    at Listener.onChange (http://oca-social-16-0-pr1216-13154c4fcbef.runboat.odoo-community.org/web/assets/debug/web.assets_backend.js:133311:52) (/mail/static/src/model/model_manager.js:886)
    at ModelManager._executeCreatedRecordsOnChange (http://oca-social-16-0-pr1216-13154c4fcbef.runboat.odoo-community.org/web/assets/debug/web.assets_backend.js:133315:26) (/mail/static/src/model/model_manager.js:890)
    at ModelManager._flushUpdateCycle (http://oca-social-16-0-pr1216-13154c4fcbef.runboat.odoo-community.org/web/assets/debug/web.assets_backend.js:133352:14) (/mail/static/src/model/model_manager.js:927)
    at ModelManager._notifyListenersInUpdateCycle (http://oca-social-16-0-pr1216-13154c4fcbef.runboat.odoo-community.org/web/assets/debug/web.assets_backend.js:133522:18) (/mail/static/src/model/model_manager.js:1097)
    at ModelManager._flushUpdateCycle (http://oca-social-16-0-pr1216-13154c4fcbef.runboat.odoo-community.org/web/assets/debug/web.assets_backend.js:133349:14) (/mail/static/src/model/model_manager.js:924)
    at ModelManager._executeCreatedRecordsComputes (http://oca-social-16-0-pr1216-13154c4fcbef.runboat.odoo-community.org/web/assets/debug/web.assets_backend.js:133269:18) (/mail/static/src/model/model_manager.js:844)
    at ModelManager._flushUpdateCycle (http://oca-social-16-0-pr1216-13154c4fcbef.runboat.odoo-community.org/web/assets/debug/web.assets_backend.js:133348:14) (/mail/static/src/model/model_manager.js:923)
    at ModelManager.insert (http://oca-social-16-0-pr1216-13154c4fcbef.runboat.odoo-community.org/web/assets/debug/web.assets_backend.js:132678:14) (/mail/static/src/model/model_manager.js:253)
</details>

- The mail message filter isn't injected into the filter menu:

image

chienandalu:
Also, if you took changes from the other PR is good to respect the contributor attribution in a separate commit

payen000:
Hello @chienandalu, thank you for testing,

the issues you found should now be solved, could you check again, please?

As for this comment:

> Also, if you took changes from the other PR is good to respect the contributor attribution in a separate commit

I added [a new commit](https://github.com/OCA/social/pull/1216/commits/d1a33597700b0b18dc22c7fbb9e3504589841936) as you suggested, but including Co-authorship and referencing the original PR, please let me know if there is a better/more respectful way of doing this, as I was not able to find a similar case in the guidelines,

thanks again!

chienandalu (changes requested):
Thanks @payen000 :) That can be a way to attribute changes, yes :) (normally I try to cherry-pick and adapt, so I keep the original commit as similar as possible, but is ok)

Functionally I still see some issues:

- Clicking on the tracking status icon doesn't show up the trackings dialog.
- The failed messages review doesn't show up in the record thread either

payen000:
Hello again @chienandalu, thank you for the tips, I appreciate it!

I've updated the code, you should be able to see the failed messages and the tracking action now, could you check again, please?

mail_tracking/controllers/discuss.py
```diff
@@ -5,7 +5,7 @@
 
 
 class MailTrackingDiscussController(DiscussController):
-    @http.route()
+    @http.route(type="json")
```
chienandalu: Why do you need to define this again?

mail_tracking/models/mail_message.py
```diff
@@ -253,14 +253,9 @@ def _prepare_dict_failed_message(self):
             return
         failed_partners = failed_trackings.mapped("partner_id")
         failed_recipients = failed_partners.name_get()
-        if self.author_id:
-            author = self.author_id.name_get()[0]
-        else:
-            author = (-1, _("-Unknown Author-"))
```
chienandalu: Why removing this?

chienandalu (commented):
Thanks for the huge effort @payen000 

I'm in doubt with the failed messages (there are 3 demo records). Should all they show up in the list? Only one does now but they did before in v15? Was it wrong? Checking the compute and search methods and debugging the data in my local shell I'm not completely sure :)

Some more comments below:

payen000:
Hello @chienandalu, thank you for testing again and for your words!

> Only one does now but they did before in v15? Was it wrong?

Now that you mention it, the failed messages that show on Marc Demo's chatter are only the failed messages that involve the logged user, you can see the other 2 failed messages if you log in as Marc Demo, and you will see no failed messages on his chatter if you log in as any other user, whereas in v15 all failed messages were shown regardless of the logger user.

I admit this was semi-accidental (as it is a direct result of filtering the messages by their `isFailed` field), but it _is_ consistent with what can be seen on `Discuss` and it makes sense to me that the failed messages filter yields only records with failed messages that involve the user... but do you think all failed messages should be shown on the chatter regardless of if they involve the logged user? :thinking:

pedrobaeza:
It should work as the activities: you see all the failed messages of the record on the chatter.

chienandalu:
> It should work as the activities: you see all the failed messages of the record on the chatter.

That's what I thought (thus my doubt), but it's not in the design: https://github.com/OCA/social/blob/0162d4ffdda6344c5828a288b15c1ea899b75191/mail_tracking/models/mail_message.py#L51-L53

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -108,10 +113,6 @@ def test_message_post(self):
         self.assertEqual(tracking_email.state, "sent")
         # message_dict read by web interface
         message_dict = message.message_format()[0]
-        self.assertTrue(len(message_dict["partner_ids"]) > 0)
-        # First partner is recipient
-        partner_id = message_dict["partner_ids"][0]
-        self.assertEqual(partner_id, self.recipient.id)
```
chienandalu: Why removing this assertions?

chienandalu (commented):
We're almost there :)

Somo additional comments to the previous unanswered ones:

mail_tracking/models/mail_message.py
payen000: Hello @chienandalu, sorry for the delay!

I deleted these lines because I'm using the same JS model for failed messages and "normal" messages: the `Message` model - as opposed to `v15` in which a new JS model was created for failed messages. This allows us to reuse the same components for the failed messages (or inheriting them and tweaking them a bit, as I did in this PR).

Since the `Message` model already has an `author` field referencing the `Partner` model, and this other `author` is a string, the fields would collide because of a type mismatch and raise an error. The `authorName` field already exists on the `Message` model which basically achieves the same result, so it seems this other `author` field is no longer needed, but what do you think? :thinking:

mail_tracking/controllers/discuss.py
payen000: I defined this again here because of the following traceback when running the tests:

>   Traceback (most recent call last):
>   File "/social/mail_tracking/tests/test_mail_tracking.py", line 714, in test_mail_init_messaging
>     res = controller.mail_init_messaging()
>   File "/odoo/odoo/http.py", line 698, in route_wrapper
>     if routing['type'] == 'http':  # _generate_routing_rules() ensures type is set
> KeyError: 'type'

The error corresponds to [this line ](https://github.com/odoo/odoo/blob/6bb16bf8818042b05ff8d40f852fa8e7d9eba92f/odoo/http.py#L698)... however, I've found that by doing this in the `setUp`:

> for _ in http._generate_routing_rules(["mail", "mail_tracking"], nodb_only=False):
>             pass

The error is no longer triggered, even without the explicit `type="json"` on the controller, I'll update the code shortly to remove it, thanks! :)

mail_tracking/tests/test_mail_tracking.py
payen000: This is because, in `v16`, `message_format` no longer returns `partner_ids`, see `v16`'s `_get_message_format_fields`:

https://github.com/odoo/odoo/blob/6bb16bf8818042b05ff8d40f852fa8e7d9eba92f/addons/mail/models/mail_message.py#L978

vs `v15`'s `_get_message_format_fields`:

https://github.com/odoo/odoo/blob/c59d6e109021c72d32884793d48ec60fecb95168/addons/mail/models/mail_message.py#L979

So the assertions raise errors unrelated to `mail_tracking` :)

mail_tracking/controllers/discuss.py
payen000: Done!

payen000:
@chienandalu I've modified the way the messages show in the chatter so that now all failed messages will be shown in the Failed Messages section as in `v15`, could you check again, please? :)

mail_tracking/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 {
     "name": "Email tracking",
```
luisg123v, line 8: Commit description is dirty, it seems a lot of fixups were squashed without writting a proper description.

mail_tracking/__manifest__.py
```diff
@@ -27,21 +27,10 @@
...
-            "mail_tracking/static/src/js/models/thread.esm.js",
-            "mail_tracking/static/src/js/chatter.esm.js",
-            "mail_tracking/static/src/js/discuss/discuss.esm.js",
-        ],
-        "web.assets_qweb": [
-            "mail_tracking/static/src/xml/mail_tracking.xml",
-            "mail_tracking/static/src/xml/failed_message/common.xml",
-            "mail_tracking/static/src/xml/failed_message/thread.xml",
-            "mail_tracking/static/src/xml/failed_message/discuss.xml",
+            "mail_tracking/static/src/models/*.esm.js",
```
luisg123v: Provide explicit file names, please.

Same below.

mail_tracking/static/src/components/discuss/discuss.xml
```diff
@@ -0,0 +1,186 @@
...
+            t-elif="messageListView.threadViewOwner.thread === messaging.history.thread"
+            position="after"
+        >
+            <t
+                t-elif="messageListView.threadViewOwner.thread === messaging.failedmsg.thread"
+            >
+                <div class="o_MessageList_emptyTitle o-neutral-face-icon">
+                    No failed messages
+                </div>
+                Failed messages will be appeared here.
```
luisg123v: will be appeared -> will appear

mail_tracking/static/src/components/discuss/discuss.xml
```diff
@@ -0,0 +1,186 @@
...
+        </t>
+    </t>
+
+    <t t-inherit="mail.ThreadIcon" t-inherit-mode="extension">
+        <xpath
+            expr="//t[@t-elif='thread.mailbox === messaging.history']"
+            position="after"
+        >
+            <t t-elif="thread.mailbox === messaging.failedmsg">
+                <div class="o_ThreadIcon_mailboxModeration fa fa-fw fa-exclamation" />
```
luisg123v: Combine with outer `t-if`

mail_tracking/static/src/components/discuss/discuss.xml
```diff
@@ -0,0 +1,186 @@
...
+            </t>
+        </xpath>
+    </t>
+
+    <t t-name="mail.tracking.status" owl="1">
+        <t t-if="tracking['isCc']">
+            <span class="mail_tracking_cc">
+                <i class="fa fa-cc" />
```
luisg123v: Missing `role="img" aria-label="..."`

Same below cases

mail_tracking/static/src/components/discuss/discuss.xml
```diff
@@ -0,0 +1,186 @@
...
+                <i class="fa fa-check fa-stack-1x" style="margin-left:-3px" />
+            </span>
+        </t>
+    </t>
+
+    <t t-inherit="mail.Message" t-inherit-mode="extension">
+        <xpath expr="//div[hasclass('o_Message_header')]" position="inside">
+            <!-- Show options only for Discuss failed messages and messages from FailedMessage component -->
+            <t
+                t-if="(messageView.message.isFailed &amp;&amp; messageView.isInDiscuss) || (this.props.isFailedMessage &amp;&amp; messageView.message.isFailedChatterMessage)"
```
luisg123v: Please split line

mail_tracking/static/src/components/discuss/discuss.xml
```diff
@@ -0,0 +1,186 @@
...
+            >
+                <span t-attf-class="o_thread_icons">
+                    <a
+                        href="#"
+                        class="btn btn-link btn-success o_thread_icon text-muted btn-sm o_failed_message_reviewed o_activity_link"
```
luisg123v: `role="button"`.

Same below.

mail_tracking/static/src/components/discuss/discuss.xml
```diff
@@ -0,0 +1,186 @@
...
+                        <i class="fa fa-retweet" />
+                        Retry
+                    </a>
+                </span>
+            </t>
+        </xpath>
+    </t>
+
+    <t
+        t-name="mail.MessageTracking"
```
luisg123v: This doesn't belong to the mail module

mail_tracking/static/src/components/discuss/discuss.xml
```diff
@@ -0,0 +1,186 @@
...
+                        <t t-if="!tracking_first">
+                            -
+                        </t>
+                        <t t-if="tracking['partner_id']">
+                            <a
+                                t-attf-class="o_mail_action_tracking_partner #{tracking['isCc'] ? 'o_mail_cc' : ''}"
+                                t-att-data-partner="tracking['partner_id']"
+                                t-attf-href="#model=res.partner&amp;id={{tracking['partner_id']}}"
+                            >
+                                <t t-esc="tracking['recipient']" />
```
luisg123v: - `t-esc` is deprecated in favor of `t-out` (same in other cases)
- `t-if`, `<a>` and `t-esc` may be combined (same below with `<div>`)

mail_tracking/static/src/components/failed_message/failed_message.esm.js
```diff
@@ -0,0 +1,13 @@
+/** @odoo-module **/
+
+import {Message} from "@mail/components/message/message";
+import {registerMessagingComponent} from "@mail/utils/messaging_component";
+
+export class FailedMessage extends Message {}
+
+Object.assign(FailedMessage, {
```
luisg123v: Shouldn't you use the native method `patch` here?

CC @rolandojduartem

mail_tracking/static/src/components/failed_message/failed_message.xml
```diff
@@ -0,0 +1,51 @@
...
+
+    <t
+        t-name="mail_tracking.FailedMessage"
+        t-inherit="mail.Message"
+        t-inherit-mode="primary"
+    >
+        <!-- Avoid setting original message as isActive upon hovering failed message-->
+        <xpath expr="//div[hasclass('o_Message')]" position="attributes">
+            <attribute
+                t-if="this.props.isFailedMessage"
```
luisg123v: I don't think it's possible to condition attribute inheritance

mail_tracking/static/src/components/failed_message/failed_message.xml
```diff
@@ -0,0 +1,51 @@
...
+                name="t-on-mouseenter"
+            >messageView.onMouseenterFailed</attribute>
+        </xpath>
+        <xpath expr="//p[hasclass('o_mail_tracking')]" position="replace">
```
luisg123v: Avoid replaces

mail_tracking/static/src/components/failed_message/failed_message.xml
```diff
@@ -0,0 +1,51 @@
...
+                            <t t-esc="recipient[1]" />
+                        </a>
+                    </t>
+                </div>
+            </t>
+        </xpath>
+        <t t-if="messageView.personaImStatusIconView" position="replace">
+            <div class="o_Activity_iconContainer bg-danger rounded-circle"><i
+                    class="o_Activity_icon fa fa-exclamation"
+                /></div>
```
luisg123v: Indent inner element so autoformatting works correctly

mail_tracking/static/src/components/failed_message_list/failed_message_list.xml
```diff
@@ -0,0 +1,64 @@
...
+                    t-if="messageListView.isMessageFailedBoxVisible"
+                >
+                    <t
+                        t-foreach="messageListView.messageFailedListViewItems"
+                        t-as="messageListViewItem"
+                        t-key="messageListViewItem.localId"
+                    >
+                        <t t-if="!messageListViewItem.message.isEmpty">
+                            <FailedMessage
+                                t-if="messageListViewItem.messageView"
```
luisg123v: Combine with outer `t-if` and add condition here

mail_tracking/static/src/components/thread_view/thread_view.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-inherit="mail.ThreadView" t-inherit-mode="extension">
+        <xpath expr="//div[hasclass('o_ThreadView')]" position="before">
+            <t t-if="this.props.className == 'o_Chatter_thread'">
```
luisg123v: combine

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -52,6 +57,10 @@ def setUp(self, *args, **kwargs):
                 ),
             },
         )
+        for _ in http._generate_routing_rules(
+            ["mail", "mail_tracking"], nodb_only=False
+        ):
+            pass
```
luisg123v, line 58: Why this?

This looks too complicated

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -27,6 +27,11 @@ def __str__(self):
         return "Test suite"
 
 
+class FakeConnection:
+    def cursor(self):
+        return True
```
luisg123v: Why this? I don't think this is necessary

mail_tracking/views/mail_tracking_email_view.xml
```diff
@@ -22,7 +22,7 @@
                     </group>
                     <group>
                         <group>
-                            <field name="mail_message_id" />
+                            <field name="mail_message_id" string="Message" />
```
luisg123v, line 25: String should be at model level preferably

mail_tracking/static/src/models/messaging_initializer.esm.js
```diff
@@ -5,73 +5,12 @@ import {registerPatch} from "@mail/model/model_core";
 registerPatch({
```
luisg123v, line 5: I think this commit could be squashed with the migration one, usually there's only one migration commit.

luisg123v (changes requested):
There's an extra commit here that is not present in the 15.0 branch: 85e2187098cad175b93387cad88e3bf6a18e92c2.

Take into account we're migrating from 15.0, not from 14.0.

If you consider that commit would be useful here, please add it to 15.0 first.

mail_tracking/static/src/components/failed_message/failed_message.esm.js
rolandojduartem: I think no, it seems the approach is creating a new component using the functionalities of Message. Remember patch is to change the behavior of the inherited component, if you use patch, youl will change the props and template of Message. Now I think Object.assign is not needed, just:

```suggestion
FailedMessage.template ="mail_tracking.FailedMessage";
```

As Odoo does [here](https://github.com/odoo/odoo/blob/6bb16bf8818042b05ff8d40f852fa8e7d9eba92f/addons/account/static/src/components/account_move_form/account_move_form.js#L33).

pedrobaeza:
/ocabot migration mail_tracking

Remove the draft state

mail_tracking/static/src/components/failed_message/failed_message.xml
payen000: @luisg123v the use of replace was intentional here because I'm using `t-inherit-mode="primary"` to generate a new component altogether, so the future inheritances of `mail.Message` will not be affected by this. Adding the `d-none ` class instead would result in the original recipients header to be shown when inspecting the element, which I personally dislike since the invisible elements are still rendered and do not contribute to the purpose of the component.

Knowing this, I believe there is no danger in using replace here, but what do you think?

mail_tracking/__manifest__.py
payen000: Got it, I'll clean the description once I'm done implementing your suggestions, thanks!

mail_tracking/static/src/components/failed_message/failed_message.xml
luisg123v: OK.

BTW, the correct way to ide is using `t-if="False"`, not hiding it using classes.

payen000: I see, thanks!

mail_tracking/views/mail_tracking_email_view.xml
payen000: This was chosen because in `v15`, these two fields have the same label:

https://github.com/OCA/social/blob/09f73b72c9a9f6829213bd500fb2ff7dd0d2c292/mail_tracking/models/mail_tracking_email.py#L46-L49

which triggers a warning at installation. Since this is the only view in the whole module using this field, I chose to add the string here directly to preserve the original `v15` look. Would you rather have the string removed and show "Mail Message" instead of "Message"? I admit I am not a fan of modifying labels directly on views either!

mail_tracking/tests/test_mail_tracking.py
payen000: This was added to avoid triggering the following error when running [this test](https://github.com/OCA/social/blob/c245ec8dfa24764e2a900d749287f6e64b90dc10/mail_tracking/tests/test_mail_tracking.py#L708C9-L708C33):

> Traceback (most recent call last):
> File "/social/mail_tracking/tests/test_mail_tracking.py", line 714, in test_mail_init_messaging
> res = controller.mail_init_messaging()
> File "/odoo/odoo/http.py", line 698, in route_wrapper
> if routing['type'] == 'http': # _generate_routing_rules() ensures type is set
> KeyError: 'type'

The problem is that it is expected that [the related controller](https://github.com/OCA/social/blob/c245ec8dfa24764e2a900d749287f6e64b90dc10/mail_tracking/controllers/discuss.py#L9) has a 'type' defined in its route decorator, and if it doesn't (even if it is inherited), there is no 'type' key in `routing` (see also [related comment above](https://github.com/OCA/social/pull/1216#discussion_r1326171503)).

I followed this implementation:

https://github.com/odoo/odoo/blob/d8567aa8420112bf368f25f1ee93e62fe00a0168/addons/website_sale_loyalty/tests/test_apply_pending_coupon.py#L37-L41

and it works as expected for the purposes of unit testing and does not require rewriting any of the already-implemented tests. What do you think?

mail_tracking/tests/test_mail_tracking.py
payen000: You're right, I've removed this, thanks!

This was added to cover [this line](https://github.com/OCA/social/blob/eae4b5096205f3644187c1dfd40e666444afb89c/mail_tracking/controllers/main.py#L28) without triggering a 'Cursor not closed explicitly' warning that I could not suppress with `mute_logger` for some reason, but it is enough to just patch the `odoo.sql_db.db_connect` method to turn the cursor into an empty Mock object, so that it does nothing and no warning is logged. It is an ugly hack, but I believe it is less ugly that returning a `Booelan`, as I did before.

payen000:
@luisg123v thank you for reviewing, after applying your suggestions the code is looking much more solid! Could you review again, please?

Just some notes:

1 - I deleted the commit you referenced in [this comment](https://github.com/OCA/social/pull/1216#pullrequestreview-1636657730) since [there is a forward port to `v15`](https://github.com/OCA/social/pull/1109) in progress since march, but since it has not been merged, I figured it is best not to include it yet as part of the base migration (I added it because it was requested in the original PR: see [this comment](https://github.com/OCA/social/pull/1091#pullrequestreview-1572950175)), but it is not needed for this migration to work properly.

2 - I've squashed all commits but retained the co-authorship message since there were some ideas that I borrowed from [the original PR](https://github.com/OCA/social/pull/1091).

chienandalu:
> 2 - I've squashed all commits but retained the co-authorship message since there were some ideas that I borrowed from [the original PR](https://github.com/OCA/social/pull/1091).

A simple way to respect the attribution chain can be:

- Copy your final work (the whole mail_tracking folder you've got now) somewhere else in your local computer
- Reset the branch to the header
- Start the migration process (git format patch...)
- Cherry pick the relevant commits from the previous PRs and resolve conflicts if any
- Now take your final work you previously saved in other location and copy it on top overwriting everything
- Commit those changes on a separate commit and that will be your separated contribution

payen000:
Hello @chienandalu, I've updated the code so that the `MessagingInitializer` changes from the first PR are present in their original commit (which also contains a good chunk of the first PR's migration) and added my changes on top as you suggested. Could you check, please?

chienandalu (commented):
Looks like your re-rendering the whole window after each action with the failed messages . Can't you use [owl reactivity](https://github.com/odoo/owl/blob/master/doc/reference/reactivity.md#usestate)?

Peek 26-09-2023 09-13

mail_tracking/readme/CONTRIBUTORS.rst
chienandalu, line 1: Put yourself in the contributors list :smile: :+1:

mail_tracking/controllers/discuss.py
```diff
@@ -0,0 +1,28 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import http
+
+from odoo.addons.mail.controllers.discuss import DiscussController
+
+
+class MailTrackingDiscussController(DiscussController):
+    @http.route()
+    def mail_init_messaging(self):
```
chienandalu: This method has a different signature (https://github.com/odoo/odoo/blob/913c8678cba0741e99610b1174ba6c42ef0c21a6/addons/mail/controllers/discuss.py#L194)
```suggestion
    def mail_init_messaging(self, **kwargs):
```

chienandalu (changes requested):
Superb work! I made a succesful functional tour and a final code review

Just two final tweaks :slightly_smiling_face:

payen000:
Hello @chienandalu,

Thank you for reviewing and testing this far, and of course, for your words of encouragement!

I've added your suggested changes, could you check please?

Also, I converted the 'message ids list' inside `store` into a `Set` for a faster/more reliable filtering of reviewed messages, since the counter of failed messages on any chatter was being reduced by the total number of stored IDs from all chatters - but not anymore, it is now working as expected.

Thanks again!

chienandalu:
Oh, after the last changes:

- I can't see the failed messages in the discuss app
- I only see one of the demo failed messages in the record

payen000:
Sorry @chienandalu, it seems I forgot to reset the runbot after my own tests, that's why most of the records were not visible as failed, after resetting the runbot the failed messages were visible again :)

I've done one last push to re-render the "Retry" and "Set as reviewed" options properly from the Inbox/History/Starred mailboxes, could you check please?

chienandalu (approved):
I think we're ready :smile: :+1: 

Thanks!

payen000:
Thank you @chienandalu!

payen000:
Hello @luisg123v, could you review this, please?

mail_tracking/demo/demo.xml
```diff
@@ -0,0 +1,129 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="True">
```
luisg123v, line 2: Demo data shouldn't be marked as noupdate

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -0,0 +1,706 @@
...
+import base64
+import time
+from unittest.mock import patch
+
+from werkzeug.exceptions import BadRequest
+
+from odoo import http
+from odoo.fields import Command
+from odoo.tests.common import TransactionCase
+from odoo.tools.misc import mute_logger
```
luisg123v:
```suggestion
from odoo.tools import mute_logger
```

(as it was in previous reviewed version)

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -0,0 +1,706 @@
...
+                ("mail_message_id", "=", message.id),
+                ("partner_id", "=", self.recipient.id),
+            ]
+        )
+        # The tracking email must be sent
+        self.assertTrue(tracking_email)
+        self.assertEqual(tracking_email.state, "sent")
+        # message_dict read by web interface
+        message_dict = message.message_format()[0]
+        self.assertTrue(len(message_dict["history_partner_ids"]) > 0)
```
luisg123v: `assertGreater(..., 0)`

But since length can't be negative and what you actually want to check it's set, the following would be enough I think:

`assertTrue(message_dict["history_partner_ids"])`

mail_tracking/security/mail_tracking_email_security.xml
```diff
@@ -0,0 +1,21 @@
...
+
+<record model="ir.rule" id="mail_tracking_email_portal_public_rule">
+    <field
+            name="name"
+        >mail_tracking_email: portal/public: read access on my email trackings</field>
+    <field name="model_id" ref="model_mail_tracking_email" />
+    <field name="domain_force">[('partner_id', '=', user.partner_id.id)]</field>
+    <field
+            name="groups"
+            eval="[(4, ref('base.group_portal')), (4, ref('base.group_public'))]"
```
luisg123v: Use the Command syntax

mail_tracking/models/mail_tracking_email.py
```diff
@@ -0,0 +1,472 @@
...
+        if not self.mail_message_id.exists():  # pragma: no cover
+            return True
+        mail_message = self.mail_message_id
+        partners = mail_message.notified_partner_ids | mail_message.partner_ids
+        if self.partner_id and self.partner_id not in partners:
+            # If mail_message haven't tracking partner, then
+            # add it in order to see his tracking status in chatter
+            if mail_message.subtype_id:
+                mail_message.sudo().write(
+                    {"notified_partner_ids": [(4, self.partner_id.id)]}
```
luisg123v: Use the Command syntax

Same below.

luisg123v (changes requested):

mail_tracking/demo/demo.xml
pedrobaeza: It's the contrary. Demo data should be noupdate. Examples:

https://github.com/odoo/odoo/blob/16.0/addons/sale/data/product_demo.xml#L2
https://github.com/odoo/odoo/blob/16.0/addons/purchase/data/purchase_demo.xml#L3

mail_tracking/tests/test_mail_tracking.py
payen000: This does work, thanks!

payen000:
Hello @luisg123v, I've applied your suggestions, could you check please?

mail_tracking/demo/demo.xml
luisg123v: Demo data is noupdate by default, that's why it shouldn't be marked. Pleasew read guidelines.

pedrobaeza: Can you please point where is that stated that is shouldn't be marked this way and why Odoo does it? Anyway, in this case, I prefer "explicit better than implicit", and this is a totally non-blocker.

luisg123v: <odoo>: for noupdate=0 or demo data (demo data is non-updatable by default)

([Source](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id20))

But I agree its not a blocking point, so either way is OK to me.

pedrobaeza: Thanks for the reference, although is not directly linkable (I have searched the term):

Selección_018

I have never tried such behavior, and all demo data I have seen includes the noupdate part. If it's OK for you, let's keep for clarity (the same I keep `position="inside"` on inherited views for clarity).

luisg123v: > I have never tried such behavior, and all demo data I have seen includes the noupdate

I have a different appreciation regarding that. Almost all demo data I've seen don't include the noupdate attribute.

The reason I consider it shouldn't be included (I wrote that part myself) is that, in my understanding,  noupdate makes no sense under a demo data context. Demo data is never updated because when updating the module, demo files are not even processed, which means a `-i module` is required; which would update also noupdate records anyway.

luisg123v (approved):
LGTM :+1:

PR #1219, opened 2023-09-07
[14.0][FIX] mail_activity_team: Use valid field for user avatar

ypapouin:
Like said in #1218, this fixes the `image_small` field has been renamed to `image_128` since the use of `image.mixin`

Note that this replacement was forgotten during the migration of many modules.
image

### DIFF
--- a/mail_activity_board/views/mail_activity_view.xml
+++ b/mail_activity_board/views/mail_activity_view.xml
@@ -235,7 +235,7 @@
                                     </div>
                                     <div class="oe_kanban_bottom_right">
                                         <img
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):

MiquelRForgeFlow:
@simahawk

simahawk:
To be fwd ported, right?
Also, what about the other modules? Maybe we open issues or at least spread the voice?

MiquelRForgeFlow:
> Maybe we open issues or at least spread the voice?

Good question

PR #1222, opened 2023-09-28
[15.0][FIX] mail_activity_team: Fix tests

victoralmau:
Changes done:
- [x] `mail_activity_done`: Re-do method to make it compatible with other modules and core + Improve tests.
- [x] `mail_activity_team`: Fix tests since https://github.com/odoo/odoo/commit/d64fff459b745d247cc19d58994b9da4bd50292a
Possible improvement in:
- [x] `mail_composer_cc_bcc`: https://github.com/OCA/social/pull/1223

@Tecnativa

### DIFF
--- a/mail_activity_done/models/res_users.py
+++ b/mail_activity_done/models/res_users.py
@@ -1,6 +1,9 @@
 # Copyright 2018-22 ForgeFlow <http://www.forgeflow.com>
+# Copyright 2023 Tecnativa - Víctor Martínez
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
-from odoo import api, fields, models, modules
+from collections import defaultdict
+
+from odoo import api, fields, models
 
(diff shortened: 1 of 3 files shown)


victoralmau:
I have found a curious thing, if you installed only `mail_activity_team`, tests failed, but if you had installed `mail_activity_done` they didn't fail because that module re-defined the method completely.

I think it would be good to create a PR so that odoo builds the query in a separate method to add some extra condition in some method. What do you think @pedrobaeza ?

pedrobaeza:
Yes, we can try to get a hook out of Odoo.

MiquelRForgeFlow:
FYI The fix in `mail_composer_cc_bcc` was merged.

MiquelRForgeFlow (approved):
LGTM

pedrobaeza (approved):
/ocabot merge patch

PR #1226, opened 2023-10-03
[FIX][14.0] mail_debrand tour info on signup email

sergiocorato:
The `Auth Signup: Odoo Connection` template is rendered partially because in this line: https://github.com/OCA/OCB/blob/26316dd56fa290b2a6d6366f325cdc9f4dcac0d6/addons/auth_signup/data/auth_signup_data.xml#L152 there is an `www.odoo.com` and the parent tag is the `div` with the main body of the email, which is then removed.

### DIFF
--- a/mail_debrand/models/mail_render_mixin.py
+++ b/mail_debrand/models/mail_render_mixin.py
@@ -53,8 +53,15 @@ class MailRenderMixin(models.AbstractModel):
                     # remove 'using' that is before <a and after </span>
                     previous.tail = ""
                 if remove_parent and len(parent.getparent()):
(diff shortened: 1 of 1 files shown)


sergiocorato:
@OCA/social-maintainers someone interested?

matteoopenf (approved):

simahawk (commented):
LG. Can you please remove the odoo version from the commit?

sergiocorato:
> LG. Can you please remove the odoo version from the commit?

Thanks.
I renamed the commit, if it was the requested change.

Issue #1090, opened 2023-03-07
[16.0] OCA/social/mail_debrand _> New Portal Signup email body empty / Odoo Periodic Digest html


ser-tec:
## Module
OCA/social/mail_debrand

## Describe the bug
After mail_debrand installation, New Portal Signup email body become empty and Odoo Periodic Digest is not rendered (mail is plain text)

## To Reproduce
**Affected versions**: 16

1. install mail_debrand
2. try to invite a new user and check email or setting->review all template->Settings: New Portal Signup->preview
3. try to send odoo Periodic Digest or preview the template

**Additional context**
ubuntu 20.4
odoo16
mail_debrand 16.0.1.0.0

## Attached examples

_Example New Portal Signup email body empty_

![Immagine](https://user-images.githubusercontent.com/113360845/223399779-38b87f02-ab3f-4aac-b804-63d0301427c3.jpg)

_Example Odoo Periodic Digest html_

![Immagine2](https://user-images.githubusercontent.com/113360845/223399785-ea7226d4-a96a-4a35-a324-43ced6ceb3dc.jpg)

jeffreyouni:
I have the same problem, please help!

Florian-Laport:
This is the same issue as #936. In the mail_debrand module, `_render_template` calls the `remove_href_odoo` function without wrapping its return value in a MarkupSafe object. This function returns a plain string regardless of input become its output comes from `etree.tostring()`.

Priddybroderickson:
Thank you for detailing the issue you're facing with the OCA/social/mail_debrand installation. It seems to be affecting the email body content and rendering in Odoo, particularly the New Portal Signup email and the Odoo Periodic Digest. This bug appears to be present in version 16. Your provided steps to reproduce the issue and additional context will be valuable in [Sterne](https://zarasocial.com/category/deutsche-sterne/) addressing this concern effectively. It's recommended to consider reaching out to the Odoo community or support for further assistance.

mathben:
Check my fix for Odoo 14 #1215

raulovallet:
I have same issue

raulovallet:
> Check my fix for Odoo 14 #1215


I tried this and it have same problem in v15.0 
![Screenshot 2023-11-15 at 2 49 02 PM](https://github.com/OCA/social/assets/23384474/08069cfc-fdc5-4152-b251-3584a8d66d04)
[...]

PR #1227, opened 2023-10-06
[16.0][IMP] mail_debrand: Add remove_href_odoo_special_case

cuongnmtm:
When rendering the template, the remove_href_odoo function will remove the parent of the <a> element which has 'odoo.com' in href.

There is a problem with mail.template "New Portal Signup", the parent element is almost the whole content hence most of the content will be removed.

image

remove_href_odoo_special_case() is added to handle this special case.

### DIFF
--- a/mail_debrand/models/mail_render_mixin.py
+++ b/mail_debrand/models/mail_render_mixin.py
@@ -14,6 +14,26 @@ from odoo import api, models, tools
 class MailRenderMixin(models.AbstractModel):
     _inherit = "mail.render.mixin"
 
+    def remove_href_odoo_special_case(self, elem):
+        parent = elem.getparent()
+        previous_elem = elem.getprevious()
+        # auth_signup.set_password_email
+        if previous_elem is not None and previous_elem.tag == "br":
+            head_msg = "Have a look at the"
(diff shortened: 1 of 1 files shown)


cuongnmtm:
I see the other PR https://github.com/OCA/social/pull/1215 was added to fix the same issue, but the msg is cut "Have a look at the ", I don't really like that behavior.

image

However, my solution is not generic, tell me if you think we should not need to handle a specific case like this then this PR should be closed.

cuongnmtm:
Fix https://github.com/OCA/social/issues/1090

cuongnmtm:
I will close this because it doesn't work with translation...

StefanRijnhart:
Does https://github.com/OCA/social/pull/1272 work for you?

StefanRijnhart:
Oh, I see you are targeting this template: https://github.com/odoo/odoo/blob/597fc00/addons/auth_signup/data/mail_template_data.xml#L146-L151
It is  special case indeed.

StefanRijnhart:
I've extended my fix to cover this case. See the second commit on https://github.com/OCA/social/pull/1272

PR #1228, opened 2023-10-09
[16.0][MIG] mail_activity_reminder

sebalix:
Migration from 13.0 PR https://github.com/OCA/social/pull/1134

PR #1232, opened 2023-10-11
[16.0][IMP] mail_activity_reminder: set message template on activity types

sebalix:
Depends on:
- [x] #1228

Ability to define your own message template on activity types.

Reminders will be sent grouped on these message templates. If no message template is defined, the generic one will be used as usual.

### DIFF
--- a/mail_activity_reminder/models/mail_activity.py
+++ b/mail_activity_reminder/models/mail_activity.py
@@ -111,25 +111,43 @@ class MailActivity(models.Model):
         """
         Group reminders by user and type and send them together
         """
-        MailThread = self.env["mail.thread"]
         utc_now = fields.Datetime.now().replace(tzinfo=UTC)
         for user in self.mapped("user_id"):
             activities = self.filtered(lambda activity: activity.user_id == user)
             tz = timezone(user.sudo().tz or "UTC")
             local_now = utc_now.astimezone(tz)
+            # Send reminders per dedicated mail templates if any
+            mail_templates = activities.activity_type_id.reminder_mail_template_id
+            for template in mail_templates:
+                activities_tmpl = activities.filtered(
+                    lambda o: o.activity_type_id.reminder_mail_template_id == template
+                )
+                self._send_reminder_mail(user, activities_tmpl, template=template)
+                activities -= activities_tmpl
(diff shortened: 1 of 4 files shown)


mail_activity_reminder/models/mail_activity.py
```diff
@@ -111,25 +111,43 @@ def action_remind(self):
...
-                subject=subject,
-                model_description="Activity",
-                notif_layout="mail.mail_notification_light",
-            )
-            activities.update({"last_reminder_local": local_now.replace(tzinfo=None)})
+    def _send_reminder_mail(self, user, activities, template=None):
+        if template is None:
+            template = self.env.ref("mail_activity_reminder.message_activity_assigned")
+        MailThread = self.env["mail.thread"]
+        subject = _("Some activities you are assigned too expire soon.")
```
simahawk, line 141:
```suggestion
        subject = _("Some activities you are assigned to expire soon.")
```

mail_activity_reminder/models/mail_activity.py
```diff
@@ -111,25 +111,43 @@ def action_remind(self):
...
+        body = self.env["ir.qweb"]._render(
+            template.id,
+            dict(activities=activities, model_description="Activities"),
```
simahawk, line 144: no translation for `model_description`?

mail_activity_reminder/models/mail_activity.py
```diff
@@ -111,25 +111,43 @@ def action_remind(self):
...
         Group reminders by user and type and send them together
         """
-        MailThread = self.env["mail.thread"]
         utc_now = fields.Datetime.now().replace(tzinfo=UTC)
         for user in self.mapped("user_id"):
             activities = self.filtered(lambda activity: activity.user_id == user)
             tz = timezone(user.sudo().tz or "UTC")
             local_now = utc_now.astimezone(tz)
+            # Send reminders per dedicated mail templates if any
+            mail_templates = activities.activity_type_id.reminder_mail_template_id
```
simahawk, line 120: wondering: wouldn't be better to set the default template as a default for the field and simplify the logic?

simahawk (changes requested):

Issue #1239, opened 2023-10-17
[15.0] mail_quoted_reply duplicated button icon while moving mouse between messages.


em230418:
## Module
mail_quoted_reply

## Describe the bug
Reply button icon is duplicated while moving mouse between messages.

## To Reproduce
15.0. Did not check others.

1. Make sure chatter has at least 2 messages.
2. Click on text of first message.
3. Move mouse up and down between those messages.

**Expected behavior**
This does not happen

**Additional context**
Reproduced on OCA's runboat.

![Снимок экрана_2023-10-17_13-29-53](https://github.com/OCA/social/assets/52414052/5214a127-d508-4da3-bfe2-c7b20c5fbfd9)

PR #1242, opened 2023-10-19
[FIX] mail_quoted_reply: fix duplicating reply icon

em230418:
Fixes https://github.com/OCA/social/issues/1239

### DIFF
--- a/mail_quoted_reply/readme/CONTRIBUTORS.rst
+++ b/mail_quoted_reply/readme/CONTRIBUTORS.rst
@@ -1,2 +1,3 @@
 * Enric Tobella
 * Lois Rilo <lois.rilo@forgeflow.com>
+* Eugene Molotov <molotov@it-projects.info>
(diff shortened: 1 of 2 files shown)


jonasbuchholz (approved):

jonasbuchholz:
@etobella could you review this fix?

etobella:
It works better now, but I was able to reproduce the error on runboat for the PR

image

etobella:
image


How to reproduce the error:

- Mark a message
- Click on the recipent
- Go Back to the original thread: Now you have two elements

jonasbuchholz:
You are correct, leaving the thread and coming back to it adds additional icons. I tried a different fix in #1185 that doesn't seem to have that issue, but it has a little more overhead because it checks the action list every time. @em230418 what do you think?

etobella:
I prefer the other option, as the bug is shouldn't be there :wink:

PR #1073, opened 2023-02-21
[16.0][MIG] Migration of module mail_tracking_mailgun

alan196:
### Proposed changes

Migration of module mail_tracking_mailgun

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Migration Update

### Checklist

- [x] I have read the CONTRIBUTING doc
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation

### Notes

This PR depends on https://github.com/OCA/social/pull/1029

I have tested functionally with PR https://github.com/OCA/social/pull/1029 and cherry-pick https://github.com/OCA/social/pull/1056

PR #1247, opened 2023-10-31
[16.0][MIG] mail_tracking_mailgun: Migration to 16.0

chienandalu:
Superseds https://github.com/OCA/social/pull/1073/ adding:

- Updated point of migration start, taking the most recent changes
- Fix tests
- Fix warnings
- Improve code
- Get rid of superfluous stuff

cc @Tecnativa TT44207

please review @pedrobaeza 

ping @alan196  @tobru @maq-adhoc


mail_tracking_mailgun/models/ir_config_parameter.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import api, models
+
+
+class IrConfigParameter(models.Model):
+    _inherit = "ir.config_parameter"
+
+    @api.model
+    def _get_mailgun_timeout(self):
+        """Avoid waiting forever in requests"""
+        return self.get_param("mailgun.timeout", 10)
```
pedrobaeza: Why putting this method in this model? And without sudo? I'm not finding too much advantage in it.

mail_tracking_mailgun/models/res_partner.py
```diff
@@ -192,14 +188,6 @@ def _autocheck_partner_email(self):
         for partner in self:
             partner.with_context(mailgun_auto_check=True).check_email_validity()
 
-    @api.model
-    def create(self, vals):
```
pedrobaeza: Why this has been removed?

mail_tracking_mailgun/models/ir_config_parameter.py
chienandalu: Yeah, it isn't ideal. The thing was that having the default timeout spreaded around the code didn't seem very practical in case we wanted to adjust that value in the future.

For the sudo I'm respecting the default access restrictions so none can get that value without permission.

mail_tracking_mailgun/models/res_partner.py
chienandalu: This couldn't be working with the previous code so I decided to just drop it as I assume none is using it

mail_tracking_mailgun/models/ir_config_parameter.py
pedrobaeza: But previous calls have sudo everywhere and now there isn't such sudo, and if you want to save, do it with sudo, but the formula I prefer is to put a global constant MAILGUN_TIMEOUT and import it.

mail_tracking_mailgun/models/res_partner.py
pedrobaeza: Then remove all the references also in documentation.

chienandalu: Done

mail_tracking_mailgun/models/ir_config_parameter.py
chienandalu: Ok, I've done it that way

pedrobaeza (approved):
/ocabot merge nobump

PR #1249, opened 2023-11-02
[15.0][IMP] email_template_qweb Provide standard evaluation context

Rad0van:
While using this module I have taken standard email template as a base (_sale.email_template_edi_sale_) and wanted to create my own slightly changed version. But rendering was crashing on things like **ctx** and **format_amount function** not being available.
So I went deep diving on how these things get present in standard templates and here's my attempt to make it work with this module as well. This is most probably relevant to all other versions as well. Any comments are welcome.


Rad0van:
This is fairly simple. @pedrobaeza @Tardo @chienandalu @victoralmau @etobella @ernestotejeda would you please have a quick look? Thank you.

pedrobaeza (approved):
/ocabot merge patch

PR #1256, opened 2023-11-20
[MIG]mail_debrand: Migration to 17.0

guestisp:
Migrate mail_debrand addon to Odoo 17
Removed parent element as default to better remove the "Powered by" string in mail notifications


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #1165, opened 2023-06-14
[16.0] [MIG] mail_quoted_reply

Borruso:
Migration mail_quoted_reply from 14.0 to 16.0

--
Confermo di aver firmato il CLA https://odoo-community.org/page/cla e di aver letto le linee guida su https://odoo-community.org/page/contributing

Issue #974, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [ ] base_search_mail_content - By @Highcooley - #1154
- [x] base_user_signature - By @Saran440 - #1365
- [ ] fetchmail_thread_default - By @ntsirintanis - #1329
- [x] mail_activity_board - By @augusto-weiss - #1006
- [ ] mail_activity_creator
- [ ] mail_activity_done - By @ignissak - #1059
- [ ] mail_activity_partner
- [x] mail_activity_plan - By @victoralmau - #1335
- [x] mail_activity_reminder - By @sebalix - #1228
- [ ] mail_activity_team - By @rven - #1129
- [ ] mail_attach_existing_attachment
- [x] mail_attach_existing_attachment - By @baimont - #976
- [x] mail_attach_existing_attachment_account - By @sbidoul - #1508
- [x] mail_autogenerated_header - By @LauraCForgeFlow - #1677
- [x] mail_autosubscribe - By @Vandan-OSI - #1115
- [x] mail_composer_cc_bcc - By @sonhd91 - #1224
- [x] mail_debrand - By @SodexisTeam - #1021
- [x] mail_drop_target - By @nguyenminhchien - #1217
- [x] mail_improved_tracking_value - By @ajaniszewska-dev - #1163
- [x] mail_layout_force - By @Abranes - #1333
- [x] mail_notification_custom_subject - By @edlopen - #1454
- [x] mail_optional_autofollow - By @baimont - #975
- [ ] mail_optional_follower_notification
- [x] mail_outbound_static - By @dreispt - #1035
- [x] mail_post_defer - By @yajo - #1213
- [x] ~~mail_preview_base~~ Included in core
- [x] mail_quoted_reply - By @Borruso - #1165
- [ ] mail_restrict_sale_button - By @SMaciasOSI - #1112
- [ ] mail_send_copy - By @legalsylvain - #1271
- [x] mail_show_follower - By @bosd - #1132
- [x] mail_template_multi_attachment - By @atchuthan - #1094
- [x] mail_tracking - By @payen000 - #1216
- [x] mail_tracking_mailgun - By @chienandalu - #1247
- [x] mail_tracking_mass_mailing - By @chienandalu - #1248
- [x] mass_mailing_custom_unsubscribe - By @carolinafernandez-tecnativa - #1315
[...]

PR #1267, opened 2023-12-11
[16.0] [MIG] mail_quoted_reply: Migration to 16.0

mdurepos:
Rewrote JavaScript to match major changes in 15.0->16.0 transition. Prior code relied on methods that no longer exist. New code patches the multiple models in @mail/models to mimic the code present for the "reply to" MessageAction.

This version also fixes a bug in which the quoted reply button appeared on notes as well as emails and sometimes duplicated itself repeatedly after mouseover events.


grindtildeath:
Hello @mdurepos 

Thanks for the effort, but you should follow the [guidelines for migrations](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0) and keep previously existing commits.

By the way, there's already an [open PR](https://github.com/OCA/social/pull/1165) for this module that is linked in the [migration to 16.0 issue](https://github.com/OCA/social/issues/974), so I would suggest we close this one in favour of merging #1165 , and you can still provide fixes afterwards if still needed.

rafaelbn:
Closing as already merged the migration

mdurepos:
> Hello @mdurepos 
> 
> 
> 
> Thanks for the effort, but you should follow the [guidelines for migrations](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0) and keep previously existing commits.
> 
> 
> 
> By the way, there's already an [open PR](https://github.com/OCA/social/pull/1165) for this module that is linked in the [migration to 16.0 issue](https://github.com/OCA/social/issues/974), so I would suggest we close this one in favour of merging #1165 , and you can still provide fixes afterwards if still needed.
> 
> 

Thanks for the pointers and especially the link to the migration guidelines. I will follow up if anything is broken in the recently migrated version.

PR #1273, opened 2024-01-04
[16.0][FIX] mail_composer_cc_bcc: make test work in the new year

StefanRijnhart:
Fixes
```
 2024-01-04 08:40:16,443 320 ERROR odoo odoo.addons.mail_composer_cc_bcc.tests.test_mail_cc_bcc: ERROR: TestMailCcBcc.test_invoice_mail_cc_bcc
Traceback (most recent call last):
  File "/__w/social/social/mail_composer_cc_bcc/tests/test_mail_cc_bcc.py", line 176, in test_invoice_mail_cc_bcc
    form = self.open_invoice_mail_composer_form()
  File "/__w/social/social/mail_composer_cc_bcc/tests/test_mail_cc_bcc.py", line 53, in open_invoice_mail_composer_form
    form = Form(self.env["account.invoice.send"].with_context(**ctx))
  File "/opt/odoo/odoo/tests/common.py", line 1990, in __init__
    self._init_from_defaults(self._model)
  File "/opt/odoo/odoo/tests/common.py", line 2116, in _init_from_defaults
    self._perform_onchange([])
  File "/opt/odoo/odoo/tests/common.py", line 2405, in _perform_onchange
    result = record.onchange(self._onchange_values(), fields, spec)
  File "/opt/odoo/odoo/models.py", line 6438, in onchange
    defaults = self.default_get(missing_names)
  File "/opt/odoo/addons/account/wizard/account_invoice_send.py", line 40, in default_get
    raise UserError(_("You can only send invoices."))
odoo.exceptions.UserError: You can only send invoices.
```
https://github.com/OCA/social/actions/runs/7407270551/job/20153152461#step:8:818

PR #1276, opened 2024-01-04
[16.0][FIX] mass_mailing_unique: Fix test

carolinafernandez-tecnativa:


pedrobaeza:
This one was just a warning, so I don't think it's the real error.

carolinafernandez-tecnativa:
> This one was just a warning, so I don't think it's the real error.

Yes, the other error is apparently solved in https://github.com/OCA/social/pull/1273
Im going to add it on test-requirements.txt and keep checking this other error.

pedrobaeza:
Then no need to do this PR. Just approve the other PR.

PR #1283, opened 2024-01-22
[16.0][[IMP] mail_compose_cc_bcc: split off account specific part

NL66278:
For customers without accounting, need to split this module, as already recognized in the roadmap.


StefanRijnhart (approved):
Thanks for doing the good work of fixing a known issue!

trisdoan (approved):

PR #1287, opened 2024-01-25
[16.0][IMP] mail_debrand: keep the decoration (grey border) where the branding was

StefanRijnhart:
when using `mail.mail_notification_light` as a wrapper

Before:
image

After:
image


pedrobaeza (approved):
I'm not sure of the change in the before/after, but I trust your judgment.

StefanRijnhart:
@pedrobaeza Don't take my word for it, it was a QA tester on my project who noticed this :smiley:

PR #1205, opened 2023-08-21
[16.0][FIX] base_search_mail_content: Allow messsage_content search only for internal users

AungKoKoLin1997:
Related issue: https://github.com/OCA/social/issues/1204

This PR adds a message_content search feature exclusively for internal users, considering constraints in specific model views like project.task. I welcome alternative solutions and suggestions.

@qrtl

PR #1291, opened 2024-01-31
[17.0][MIG] base_search_mail_content: Migration to 17.0

nguyenminhchien:
Ref: BSRD-750


gurneyalex (approved):

gurneyalex (approved):
tested on runboat: works as advertised.

chrisprintfactory:
@nguyenminhchien can it be correct that this release breaks the portal user view of the project module? Installing this module breaks the view with an owl error.

nguyenminhchien:
> @nguyenminhchien can it be correct that this release breaks the portal user view of the project module? Installing this module breaks the view with an owl error.

it works fine for me. would you please record a video of the error you faced? or a screenshot at least, thanks.

yostashiro:
> @nguyenminhchien can it be correct that this release breaks the portal user view of the project module? Installing this module breaks the view with an owl error.

@chrisprintfactory @nguyenminhchien I suppose it's the same issue as what we address in https://github.com/OCA/social/pull/1205. Would be nice if you could review it.

chrisprintfactory:
@nguyenminhchien 

image

Portal users can access projects when shared, that works without any issue on a demo database:
image

The second the module is installed, it breaks the view for portal users, but not internal users:
image

chrisprintfactory:
@yostashiro Tried this fix on the 17.0 release, but it still does not work sadly for portal users.

nguyenminhchien:
@chrisprintfactory from your screenshot, i saw the kanban view. but from my side, it always opens the list view for portal user, and no error at all. did you install other module to switch from list -> kanban? or any configuration some where?
image

chrisprintfactory:
@nguyenminhchien i am using a base install, with some enterprise modules as well (but no custom modules).

only thing installed are helpdesk and project.

@yostashiro coming back, i did this fix with an upgrade, that does not work, but with a reinstall, the fix works as expected. Not the most nice for users, but without the error.

carlos-lopez-tecnativa:
Hi, I don’t understand why the dependency on the `base_search_fuzzy` module was dropped. Could you please clarify, @nguyenminhchien?

I understand that Odoo added some index support, but it still doesn’t support the `% operator`, so the `base_search_fuzzy` module is still necessary. Without it, full-text search doesn't work. Additionally, I created a PR to add this operator in the search view, please check: https://github.com/OCA/social/pull/1627

cc: @pedrobaeza

nguyenminhchien:
> Hi, I don’t understand why the dependency on the `base_search_fuzzy` module was dropped. Could you please clarify, @nguyenminhchien?
> 
> I understand that Odoo added some index support, but it still doesn’t support the `% operator`, so the `base_search_fuzzy` module is still necessary. Without it, full-text search doesn't work. Additionally, I created a PR to add this operator in the search view, please check: #1627
> 
> cc: @pedrobaeza

Hi @carlos-lopez-tecnativa ,
image

In previous version, this module needed the `trgm.index` model created by `base_search_fuzzy`. But, `triagram` has already been supported by Odoo base via attribute index=trigram: for example, `subject = fields.Char(index="trigram")`.

After this change, the module works fine as expected (from its author): the tests were green.

pedrobaeza:
But the problem is that the features are not the same. If you for example have a message "Hello world", with the proper operator '%', you can search by "world hello" or "hel wor", or whatever. Now you can't with trigram using like/ilike operator. The problem is that in 16, the implementation of `base_search_mail_content` was faulty. See #1627

PR #1293, opened 2024-02-06
[15.0][IMP] mail_composer_cc_bcc: filter out partner with no email address and set "" for partner with no name

sonhd91:
- Normally a wizard opens on Odoo client side to prompt user to define an email address for partners that don't have one
- But it's possible to bypass it, so we explicitly filter out partners with no email address
- We also return "" instead of "False" for partner with no name, and odoo's `tools.formataddr` will just use the partner's email address

### DIFF
--- a/mail_composer_cc_bcc/models/mail_mail.py
+++ b/mail_composer_cc_bcc/models/mail_mail.py
@@ -14,9 +14,7 @@ _logger = logging.getLogger(__name__)
 
 
 def format_emails(partners):
-    emails = [
-        tools.formataddr((p.name or "False", p.email or "False")) for p in partners
-    ]
+    emails = [tools.formataddr((p.name, p.email)) for p in partners if p.email]
     return ", ".join(emails)
 
 


mail_composer_cc_bcc/models/mail_mail.py
```diff
@@ -14,9 +14,7 @@
 
 
 def format_emails(partners):
-    emails = [
-        tools.formataddr((p.name or "False", p.email or "False")) for p in partners
-    ]
+    emails = [tools.formataddr((p.name or "", p.email)) for p in partners if p.email]
```
hbrunn:
```suggestion
    emails = [tools.formataddr((p.name, p.email)) for p in partners if p.email]
```

hbrunn (changes requested):

mail_composer_cc_bcc/models/mail_mail.py
sonhd91: Updated, please checked again

hbrunn (approved):

leemannd:
Hello, can this PR be reopened and merged? Unless that I have opened a new one to be able to do a followup if requested.

leemannd (commented):
Approved

leemannd (approved):

hbrunn:
/ocabot merge minor

no idea what's going on with runboat, and I'm overriding @hailangvn2023 as this handle seems inactive. The module should get a new maintainer then or be orphaned

PR #1304, opened 2024-02-09
[15.0][ADD] mail_activity_plan: New module

victoralmau:
New module

@Tecnativa TT47279


mail_activity_plan/models/mail_activity_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+
+from odoo import _, api, fields, models
+
+from odoo.addons.base.models import ir_ui_view
+
+
+class MailActivityMixin(models.AbstractModel):
+    _inherit = "mail.activity.mixin"
+
+    has_activity_plans = fields.Boolean(
```
pedrobaeza: Avoid plurals as possible:

```suggestion
    has_activity_planning = fields.Boolean(
```

mail_activity_plan/models/mail_activity_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+        compute="_compute_has_activity_plans",
+    )
+
+    def _compute_has_activity_plans(self):
+        plan_model = self.env["mail.activity.plan"]
```
pedrobaeza: Can you cover this in tests?

mail_activity_plan/models/mail_activity_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+        for item in self:
+            plans = plan_model._get_plans_from_model(item._name)
+            item.has_activity_plans = True if len(plans) else False
```
pedrobaeza: Avoid spaghetti code:

```suggestion
            item.has_activity_plans = bool(plans)
```

mail_activity_plan/models/mail_activity_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+        return etree.Element(
+            "field", attrib={"name": "has_activity_plans", "invisible": "1"}
+        )
+
+    @api.model
+    def fields_view_get(
+        self, view_id=None, view_type="form", toolbar=False, submenu=False
+    ):
+        """Add (inside <header>) the has_activity_plans field and button to launch
+        activity plans."""
```
pedrobaeza: And remember that a multi-line docstring put the ending 3 quotes in a new line.

```suggestion
        activity plans, creating the element if not exists.
        """
```

mail_activity_plan/readme/CONFIGURE.rst
```diff
@@ -0,0 +1 @@
+Go to Settings > Technical > Discuss > Activity Plans and create the ones you need linked to a model (Contact for example).
```
pedrobaeza: Break documentation lines also at 88 cols:

```suggestion
Go to *Settings > Technical > Discuss > Activity Plans* and create the ones you need 
linked to a model (Contact for example).
```

mail_activity_plan/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows to define activity plans linked to a model and to apply them.
```
pedrobaeza:
```suggestion
This module allows to define activity plans linked to any document type that allows to
plan activities, and later launch them from the documents.
```

pedrobaeza (commented):
Tested on runboat and working as expected. Two refinements:

- If a expression is used, and there's no value in the record, please show a `UserError` saying so, instead of trying to create the activity record and showing the ugly error of field not set.
- Please allow to launch the plan for several records in the list view. In this case, you have to investigate how to add the action in the list (maybe you can investigate on how the "Archive" / "Unarchive" options are put dynamically, or the "Share" button).

mail_activity_plan/security/security.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="rule_multi_company_mail_activity_plan" model="ir.rule">
+        <field name="name">Mail Activity Plan multi-company</field>
+        <field name="model_id" ref="model_mail_activity_plan" />
+        <field name="global" eval="True" />
+        <field name="domain_force">[('company_id', 'in', [False] + company_ids)]</field>
+    </record>
+</odoo>
```
pedrobaeza, line 15: Missing security rule for lines (using parent's company) or related stored field for performance purposes.

victoralmau:
Changes done.

An error is now displayed if the expression does not return a value. 
From the tree view, selecting several records, it is possible to call the action 'Launch Activity Plan'.

mail_activity_plan/static/src/js/mail_activity_plan.js
```diff
@@ -0,0 +1,32 @@
...
+            const state = this.model.get(this.handle);
+            const resIds = await this.getSelectedIdsWithDomain();
+            var self = this;
+            this._rpc({
+                model: state.model,
+                method: "action_wizard_mail_activity_plan",
+                args: [resIds],
+            }).then(function (action) {
+                self.do_action(action);
+            });
```
chienandalu: nitpicking:
```suggestion
            this._rpc({
                model: state.model,
                method: "action_wizard_mail_activity_plan",
                args: [resIds],
            }).then((action) => {
                this.do_action(action);
            });
```

chienandalu (approved):
Tested :+1:

chienandalu (approved):

mail_activity_plan/static/src/js/mail_activity_plan.js
```diff
@@ -0,0 +1,31 @@
+odoo.define("mail_activity_plan.mail_activity_plan", function (require) {
+    "use strict";
+
+    const ListController = require("web.ListController");
+    var core = require("web.core");
+    var _t = core._t;
+
+    ListController.include({
+        async _actionWizardMailActivityPlan() {
```
pedrobaeza, line 10: Can you avoid to show the entry if no existing plans?

chienandalu: You can do it in the `_getActionMenuItems` function with an rpc call to a simple search.

victoralmau: You are right, `_getActionMenuItems()` could be used to make an rpc call, but it is very important to clarify that this method is called **every time** a record is selected / deselected from the list. I will try to use another method (`willStart()` for example) that is only called once.

chienandalu: Ok, that'll be better :+1:

victoralmau: Done.

chienandalu (approved):
Working great!

pedrobaeza:
Just one last change: please put the "Launch activity plan" at the end of all existing buttons, not at the beginning. I'm also just wondering if we should use the same action that the list view instead and forget all the stuff about the header.

victoralmau:
> Just one last change: please put the "Launch activity plan" at the end of all existing buttons, not at the beginning.

It is more difficult than it seems, since the extra actions are added automatically and you would have to set it as a component from https://github.com/odoo/odoo/blob/15.0/addons/web/static/src/legacy/js/components/action_menus.js#L73, which would imply inheriting from _base_ to set a method that can be called in all models, etc. Perhaps it could be added in ROADMAP as a pending issue.

> I'm also just wondering if we should use the same action that the list view instead and forget all the stuff about the header.

I agree, I'm going to change it to "simplify" the logic.

pedrobaeza:
I was talking about the header, no the Action entry, so if you switch to the action in the form, there's nothing more to do

victoralmau:
Changes done. Reduced code. Now there is the action in the tree and form views.

pedrobaeza (approved):
Excellent module we have got

/ocabot merge nobump

PR #1259, opened 2023-11-21
[17.0][MIG] mail_outbound_static

SodexisTeam:

PR #1307, opened 2024-02-15
[17.0][REM] deb package mistakenly added to repo

nilshamerlinck:
Added in https://github.com/OCA/social/pull/1259, not needed


pedrobaeza:
Uhm, this is going to take a lot of MB of download... @dreispt you have to take care about that deb file when merging, as it occupies a lot. I'm in favor of rewriting history for totally removing it. If not, it will take 30 MB of diff adding it, and now other 30 MB of negative diff removing it. What do you think?

nilshamerlinck:
Hi @pedrobaeza good idea
- I just pushed [branch `17.0-rem_deb_package_by_rewriting_history` on my fork](https://github.com/nilshamerlinck/social/tree/17.0-rem_deb_package_by_rewriting_history)
- it's the result of using filter-repo on branch 17.0: `git filter-repo --path pandoc-3.1.9-1-amd64.deb --invert-paths --refs 17.0`
- the idea would be to force push it to replace current branch 17.0
- note that as a consequence, open PRs on branch 17.0 will have to be rebased and some people might be impacted in their deployments, depending on how they handle their repos

pedrobaeza:
Yes, that's the idea. Thanks for doing the work. The problem right now is that the branch is protected against forced pushes. @dreispt or any other board member should do unprotect it. @etobella maybe you can do it?

dreispt:
@pedrobaeza I just disabled the "Require a pull request before merging" for this repo

pedrobaeza:
Thanks, @dreispt but I'm afraid I'm still not able. You should remove the protection of forced pushing.

etobella:
I might be able to do it, but that might be a problem for PRs done after it or weblate :fearful:

pedrobaeza:
Yeah, but carrying in this important repo 60 MB of extra diff for each build you need it is too much. I will rebase Weblate after and take care about notifying people.

etobella:
Ok! Let's do it

pedrobaeza:
Do you do it? I'm not able to force push due to the protection branch.

etobella:
I did it myself

etobella:
I think this can be closed now

PR #1310, opened 2024-02-19
[15.0][BKP][FIX] mail_composer_cc_bcc: send RFQ by email

trisdoan:
Backport of https://github.com/OCA/social/pull/1493


hbrunn (approved):

TTPBruce:
This appears to work other than in one corner case that may not apply to most instances.

When a user is following tracked field changes, the cc user still gets added to that e-mail. This may not be normal/possible on most instances, but this happens to us when a user is following "All" chatter activity.

I had to update _notify_by_email_add_values to check for mail_message_id, browse for the record, and check if message type is notification to avoid adding CC and BCC from there. Not sure if this is the best approach since it will require a browse on each message, but it does resolve the noted issue.

john-herholz-dt:
What is the state here? I have noticed the sending for duplicate emails when using BCC for the 17.0 version of the module.
I wonder if it is related or if it is actually a bug or a feature.

trisdoan:
> What is the state here? I have noticed the sending for duplicate emails when using BCC for the 17.0 version of the module. I wonder if it is related or if it is actually a bug or a feature.

Hi, I am working on the fix, will be ready soon

hbrunn:
@trisdoan please fix conflicts

trisdoan:
Hello @TTPBruce, @hbrunn, how does it look to you?

hbrunn (approved):

PR #1314, opened 2024-02-26
[MIG] mail_template_substitute: Migration to 17.0

SodexisTeam:


hbrunn (approved):

ahaly-compass:
Note that this module doesn't work for account.move.send anymore, because account.move.send doesn't inherit from {'mail.compose.message':'composer_id'} anymore. Maybe this issue could be found at other places too.

atchuthan (approved):
code QA + Functional QA :+1: 
It works with Send & Print at Invoice form view too.

@ahaly-compass please confirm

novamilky (approved):

mail_template_substitute/models/account_move.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2024 Sodexis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def action_invoice_sent(self):
```
hbrunn: this should happen in https://github.com/OCA/OCB/blob/17.0/addons/account/models/account_move.py#L4047 already

hbrunn (changes requested):
I'm not so sure if we should add a dependency on account to this module, or add support for account in another module. What do reviewers think?

mail_template_substitute/models/account_move.py
atchuthan: > this should happen in https://github.com/OCA/OCB/blob/17.0/addons/account/models/account_move.py#L4047 already

_get_mail_template() method will need to provide the External ID of the substitute template, but templates can be created via User Interface too, so external ID will not be available for those records (Do we need to create External Identifier record manually for those created via UI?). @hbrunn

atchuthan:
> I'm not so sure if we should add a dependency on account to this module, or add support for account in another module. What do reviewers think?

I think it would be good to have a separate module(mail_template_substitute_account_move) as it would avoid adding "account" as dependency.  Thus, we create new module with auto_install=True and depends contains mail_template_substitue and account.  Should we add this new module in this repository or other OCA/account related repository?

mail_template_substitute/models/account_move.py
hbrunn: no, but don't override `action_invoice_sent`, override `action_send_and_print`

hbrunn:
`mail_template_substitute_account_move` in oca/account-financial-tools sounds good to me

mail_template_substitute/models/account_move.py
atchuthan: okay, I understood @hbrunn

hbrunn (approved):

sakthi-shan:
> `mail_template_substitute_account_move` in oca/account-financial-tools sounds good to me

New Module created in account-financial-tools
https://github.com/OCA/account-financial-tools/pull/1881
Please review

atchuthan:
@OCA/social-maintainers can we merge this PR?

PR #1317, opened 2024-02-27
17.0 migr mail show follower

thomaspot:
a) Migrated module mail_show_follower to Odoo v17
b) Added 2 extra options in settings to be able to choose font and background color


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #1318, opened 2024-02-28
[16.0][FIX] mail_autosusbscribe: Removing external dependencies and adding context to tests

pilarvargas-tecnativa:
The external dependency odoo_test_helper must be defined in test-requirements.txt, not in the manifest.
Also, a context has been added to the tests to avoid executing the code of the "_message_get_default_recipients" method added by this module in a test environment as long as it is not the module itself that is being tested.

cc TT48102 @Tecnativa


victoralmau (approved):

chienandalu (approved):

pedrobaeza:
This is still red...

pilarvargas-tecnativa:
> This is still red...

I'm checking the error

pilarvargas-tecnativa:
ping @pedrobaeza

mail_autosubscribe/models/models.py
```diff
@@ -24,7 +25,10 @@ def _message_get_default_recipients(self):
         # Overload to include auto follow document partners in the composer
         # Note: This only works if the template is configured with 'Default recipients'
         res = super()._message_get_default_recipients()
-        if self.env.context.get("no_autosubscribe_followers"):
+        test_condition = config["test_enable"] and not self.env.context.get(
+            "no_autosubscribe_followers"
```
pedrobaeza: This is not correct, as the context should be something like `test_mail_autosubscribe`, and pass such context on the test of the module. This way, if we are testing any other module, this code won't apply, but testing the own module, it will apply.

pedrobaeza:
And squash commits into one.

mail_autosubscribe/tests/test_mail_autosubscribe.py
```diff
@@ -97,7 +97,7 @@ def test_mail_template_default_recipients(self):
         self.mail_template.use_default_to = True
         self.mail_template.send_mail(self.order.id)
         message = self.order.message_ids[0]
-        self.assertEqual(message.partner_ids, self.partner_2 | self.partner_3)
+        self.assertEqual(message.partner_ids, self.partner_2)
```
pedrobaeza: The fact of changing the test indicates that something is not OK.

pilarvargas-tecnativa: You are right, I was confusing the context.

mail_autosubscribe/tests/test_mail_autosubscribe.py
```diff
@@ -74,28 +74,36 @@ def test_message_subscribe_disabled(self):
 
     def test_mail_template(self):
         """Test autosubscribe when partner is set in the mail.template partners_to"""
-        self.mail_template.send_mail(self.order.id)
+        self.mail_template.with_context(test_mail_autosubscribe=True).send_mail(
```
pedrobaeza: For not having to put this on each test, just do on `setUpClass`:

```
@classmethod
def setUpClass(cls):
    super().setUpClass()
    cls.env = cls.env.with_context(test_mail_autosubscribe=True)
```

pilarvargas-tecnativa: thanks!

pedrobaeza (approved):
/ocabot merge patch

PR #1319, opened 2024-02-29
[16.0][ADD] mail_template_restrict_by_default

hbrunn:
By default, all Odoo backend users are members of the group
``Mail Template Editor``, which allows editing dynamic templates. This
in turn allows to run code, which allows to escalate privileges.

Checking ``Restrict mail templates edition and QWEB placeholders usage``
in the general settings fixes this, but as that's a manual and therefore
error prone action, this module automates this. It also removes all
users except root from the edit group.


StefanRijnhart (approved):
Thanks!

mail_template_restrict_by_default/hooks.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2024 Hunki Enterprises BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
+from odoo import SUPERUSER_ID, api
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    env["ir.config_parameter"].set_param("mail.restrict.template.rendering", True)
+    env.ref("mail.group_mail_template_editor").write(
```
mart-e, line 9: this shouldn't be needed as long as the admin is in `group_system`

hbrunn: it's meant to remove other users. Didn't debug it yet, but when checking the flag on the [current runbot](https://59019060-17-0-all.runbot211.odoo.com/web?debug=1&reload=true#action=95&model=res.config.settings&view_type=form&cids=1&menu_id=70), it breaks the implied_group relation, but keeps the demo user in the [group](https://59019060-17-0-all.runbot211.odoo.com/web?debug=1#id=14&cids=1&menu_id=59&action=68&model=res.groups&view_type=form)

hbrunn: ah, that's because email marketing/user inherits this too. not sure what to do about that, let's see what others think

StefanRijnhart: I don't see where you are removing the group as an implied group from base.group_user (as added here by default: https://github.com/odoo/odoo/blob/16.0/addons/mail/data/mail_groups.xml#L13-L16)

StefanRijnhart: Ugh it's in https://github.com/odoo/odoo/blob/16.0/addons/mail/models/ir_config_parameter.py#L25-L35

StefanRijnhart: I would expect a module by this name to apply the restriction globally, so removing the group as an implied group everywhere apart from the system group.

hbrunn: this would also take care of https://github.com/odoo/odoo/blob/16.0/addons/mass_mailing/data/res_users_data.xml#L5 which confused me, as the mass mailing user group has the template editor group implied, so when this module is installed, all your users created afterwards are in the mass mailing user group, and thus in the template editor group.

So we want `env['res.groups'].search([]).write({'implied_ids': [(3, ref('group_mail_template_editor')`, but this won't 'protect' users who install mass mailing afterwards. Hmm, difficult

StefanRijnhart: Override res.group's write method and intercept these writes when you encounter the `install_module` context key?

mart-e: Note that the "default user has too many rights" was handled recently in odoo/odoo#150391 (odoo/odoo#149424 for master) if you are not in a "trust everyone by default" mode.

StefanRijnhart: Yes I meant to thank you for that :smile: but that does not solve the implied groups of base.group_user I think.

mart-e: I would say that what you want is:
- with only mail installed: `set_param("mail.restrict.template.rendering", True)`
- with mass_mailing also installed: `default_user.groups_id -= group_mass_mailing_user` (as mass-mailing => qweb edition)

If mass-mailing is already installed, should you remove existing users from mass-mailing? That's debatable but maybe we can assume yes if you are the kind of user that wants this module.

So in this case, the easiest would be to create an auto-install module when both `mass_mailing` and `mail_template_restrict_by_default` is installed maybe?

mail_template_restrict_by_default/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1 @@
+- Holger Brunn \<mail@hunki-enterprises.com> (https://hunki-enterprises.com)
```
hbrunn, line 1: @mart-e @StefanRijnhart thank you now already for your input. any complaints about (names ordered alphabetically by surname)

```suggestion
- Holger Brunn \<mail@hunki-enterprises.com> (https://hunki-enterprises.com)
- Stefan Rijnhart \<stefan@opener.amsterdam> (https://opener.amsterdam)
- Martin Trigaux \<no.mail@github.profile> (https://odoo.com)
```

mart-e: @hbrunn thanks for the credit, you can set my `mat` odoo email address

StefanRijnhart: Not sure I deserve any credits but otherwise this is fine.

mail_template_restrict_by_default/hooks.py
hbrunn: @StefanRijnhart about suppressing writes that set the implied group: I think better not, the first thing I will do after adding it to my ${customer}_base module, is adding the template editor group as implied group to the customers 'Trusted powerusers' group who are the kind of people who may edit mail templates.

It is a lot of fluff for the little the modules actually do, but I think we'll be better off with only removing things explicitly.

StefanRijnhart (approved):
Approving the refactoring since my first approval

mart-e (approved):
Personnaly, I would remove the `write` but I understand your point

mart-e:
Hi passing to also inform the good news that, in 18.0, the employees are no longer template editor by default (odoo/odoo#183062).
We were able to do so because of odoo/odoo#164671 where we allow simple expressions (`object.name`) in templates without being mail editor.
(but this PR is still a good idea for 16.0)

PR #1331, opened 2024-03-18
[16.0][IMP] mail_tracking: display internal notifications status

vincent-hatakeyama:
Use internal notifications status when mails are not generated.

### DIFF
--- a/mail_tracking/models/mail_message.py
+++ b/mail_tracking/models/mail_message.py
@@ -100,6 +100,21 @@ class MailMessage(models.Model):
             "soft-bounced": "error",
         }
 
+    @api.model
+    def _tracking_mail_notification_get_status(self, notification):
+        """Map mail.notification states to be used in chatter"""
+        return (
+            "opened"
+            if notification.is_read
+            else {
+                "ready": "waiting",
+                "sent": "delivered",
+                "bounce": "error",
+                "exception": "error",
+                "canceled": "error",
+            }.get(notification.notification_status, "unknown")
+        )
+
     def _partner_tracking_status_get(self, tracking_email):
         """Determine tracking status"""
         tracking_status_map = self._tracking_status_map_get()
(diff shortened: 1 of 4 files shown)


mail_tracking/models/mail_message.py
```diff
@@ -199,6 +213,28 @@ def tracking_status(self):
...
+                # Search internal mail.notifications (for users using it)
+                # Note that by default, read notifications older than 180 days are
+                # deleted.
+                notification = self.env["mail.notification"].search(
+                    [
+                        ("notification_type", "=", "inbox"),
+                        ("mail_message_id", "=", message.id),
+                        ("res_partner_id", "=", partner.id),
+                    ]
+                )
```
hbrunn, line 224: why not `message.notification_ids.filtered(...)`?

mail_tracking/models/mail_message.py
```diff
@@ -99,6 +99,20 @@ def _tracking_status_map_get(self):
             "soft-bounced": "error",
         }
 
+    def _notification_status(self, notification):
```
hbrunn: please avoid such generic names on models that a lot of addons touch. Possibly prepend with `_tracking`?

hbrunn (dismissed):
please rebase

mail_tracking/models/mail_message.py
vincent-hatakeyama: Left over from previous version.

Fixed in the new version.

mail_tracking/models/mail_message.py
vincent-hatakeyama: I’ve also added the `@api.model`. It might have been better to put it into a mail_notification.py but it was simpler to see the expected values in the same file.

hbrunn (approved):

pedrobaeza (approved):
/ocabot merge patch

Please fw-port it to upper versions.

PR #1353, opened 2024-04-19
[14.0][FIX] mail_activity_team: Manually keep `user_id` in sync with `team_user_id`

ypapouin:
Ensure that the built-in `user_id` field value is in sync with our field `team_user_id` before triggering `create`, otherwise the framework will try to resolve the inverse value by itself and will call `write` before create has ended. That leads to multiple side effects like sending two assignment notifications.

### DIFF
--- a/mail_activity_team/models/mail_activity.py
+++ b/mail_activity_team/models/mail_activity.py
@@ -27,6 +27,24 @@ class MailActivity(models.Model):
         comodel_name="mail.activity.team", default=lambda s: s._get_default_team_id()
     )
 
+    @api.model_create_multi
+    def create(self, vals_list):
+        vals_list = [
+            # ensure that the built-in user_id field value is in sync with our field
+            # team_user_id before triggering create
+            (
+                dict(vals, user_id=vals["team_user_id"])
+                if (
+                    "user_id" in vals
+                    and "team_user_id" in vals
+                    and vals["user_id"] != vals["team_user_id"]
+                )
+                else vals
+            )
+            for vals in vals_list
+        ]
+        return super(MailActivity, self).create(vals_list)
+
     @api.onchange("user_id")
     def _onchange_user_id(self):
         if not self.user_id or (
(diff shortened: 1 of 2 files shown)


mail_activity_team/models/mail_activity.py
```diff
@@ -27,6 +27,19 @@ def _get_default_team_id(self, user_id=None):
...
+        for vals in vals_list:
+            # ensure that the built-in user_id field value is in sync with our field
+            # team_user_id before triggering create
+            if (
+                "user_id" in vals
+                and "team_user_id" in vals
+                and vals["user_id"] != vals["team_user_id"]
+            ):
+                vals["user_id"] = vals["team_user_id"]
+        return super(MailActivity, self).create(vals_list)
```
hbrunn, line 46:
```suggestion
        return super().create(vals_list)
```

mail_activity_team/models/mail_activity.py
```diff
@@ -27,6 +27,19 @@ def _get_default_team_id(self, user_id=None):
...
+    def create(self, vals_list):
+        for vals in vals_list:
+            # ensure that the built-in user_id field value is in sync with our field
+            # team_user_id before triggering create
+            if (
+                "user_id" in vals
+                and "team_user_id" in vals
+                and vals["user_id"] != vals["team_user_id"]
+            ):
+                vals["user_id"] = vals["team_user_id"]
```
hbrunn:
```suggestion
        vals_list = [
            # ensure that the built-in user_id field value is in sync with our field
            # team_user_id before triggering create
            dict(vals, user_id=vals["team_user_id"]) if (
                "user_id" in vals
                and "team_user_id" in vals
                and vals["user_id"] != vals["team_user_id"]
            ) else vals
            for vals in vals_list
        ]
```

(never change your input, make a copy if you're going to change it)

hbrunn (changes requested):

mail_activity_team/models/mail_activity.py
ypapouin: Thank you for the tip

hbrunn (approved):

PR #1354, opened 2024-04-19
[17.0][MIG] mail_quoted_reply: Migration to 17.0

trisdoan:
### Major changes in V17

- Split `_onchange_template_id` + `default_get` + `get_record_data` usage into editable
computed stored fields in [here](https://github.com/odoo/odoo/pull/107356)
- `_evaluate_res_ids` is introduced to parse composer res_ids in [here](https://github.com/odoo/odoo/pull/99482/commits/a5adbc19ada1717e72396060544f8fb2d41d0270#diff-898a3d5c08a567c1a7b82c026a213d796e850e376cbdb1c9e7936409f440d37a). As consequence, 'default_res_id' is deprecated, should use 'default_res_ids'.

**Components:**
- There is a huge refactor in efforts to use OWL in [here](https://github.com/odoo/odoo/commit/7710c3331ebd22f8396870bd0731f8c1152d9c41)


mail_quoted_reply/models/mail_compose_message.py
```diff
@@ -6,18 +6,22 @@
...
-        context = self._context
-        if "is_quoted_reply" in context.keys() and context["is_quoted_reply"]:
-            self.body += Markup(context["quote_body"])
-        return
+    @api.depends("composition_mode", "model", "res_domain", "res_ids", "template_id")
+    def _compute_body(self):
+        res = super()._compute_body()
+        for composer in self:
+            context = composer._context
+            if "is_quoted_reply" in context.keys() and context["is_quoted_reply"]:
```
hbrunn:
```suggestion
            if context.get("is_quoted_reply"):
```

mail_quoted_reply/models/mail_compose_message.py
```diff
@@ -6,18 +6,22 @@
...
 
-    @api.model
-    def get_record_data(self, values):
-        result = super().get_record_data(values)
-        subj = self._context.get("default_subject", False)
-        if subj:
-            result["subject"] = tools.ustr(subj)
-        return result
+    @api.depends(
+        "composition_mode", "model", "parent_id", "res_domain", "res_ids", "template_id"
```
hbrunn: why do you remove `record_name` here?

mail_quoted_reply/models/mail_message.py
```diff
@@ -25,7 +25,7 @@ def _prep_quoted_reply_body(self):
             {body}
             </blockquote>
             </div>
-        """.format(
+        """.format(  # noqa: E501
```
hbrunn: isn't the pre-commit config wrong if we need this?

hbrunn (changes requested):

trisdoan:
Hello @hbrunn, I updated as you suggest :pray:

mail_quoted_reply/models/mail_message.py
trisdoan: Hello, I managed to avoid E501 without adding that tag. Thanks

hbrunn (approved):

Abranes (approved):
Functional test

mail_quoted_reply/models/mail_compose_message.py
```diff
@@ -6,18 +6,28 @@
...
+        "parent_id",
+        "record_name",
+        "res_domain",
+        "res_ids",
+        "template_id",
+    )
+    def _compute_subject(self):
+        res = super()._compute_subject()
+        for composer in self:
+            subj = composer._context.get("default_subject", False)
```
mmequignon, line 32: Same here

mail_quoted_reply/models/mail_compose_message.py
```diff
@@ -6,18 +6,28 @@
...
-        context = self._context
-        if "is_quoted_reply" in context.keys() and context["is_quoted_reply"]:
-            self.body += Markup(context["quote_body"])
-        return
+    @api.depends("composition_mode", "model", "res_domain", "res_ids", "template_id")
+    def _compute_body(self):
+        res = super()._compute_body()
+        for composer in self:
+            context = composer._context
+            if context.get("is_quoted_reply"):
```
mmequignon, line 15: since it appears to depend on a context key, aren't we supposed to add this key in `@api.depends_context()` ?

mmequignon (commented):
Very bad in JS, so only Python review

trisdoan:
Hello @mmequignon, I updated as you suggest, thanks

mail_quoted_reply/models/mail_compose_message.py
```diff
@@ -0,0 +1,35 @@
...
+    _inherit = "mail.compose.message"
+
+    @api.depends("composition_mode", "model", "res_domain", "res_ids", "template_id")
+    @api.depends_context("is_quoted_reply")
+    def _compute_body(self):
+        res = super()._compute_body()
+        for composer in self:
+            context = composer._context
+            if context.get("is_quoted_reply"):
+                composer.body = Markup(context["quote_body"])
```
jaenbox, line 16: It should be noted that when a user replies to a message and wants to load a template from the template_id field, the template must be added to the content that we already have in the body.
You can fix this with this code since the body in the creation of mail.compose.message is False.

`            
if context.get("is_quoted_reply"):
    if composer.body:
        composer.body += Markup(context["quote_body"])
    else:
        composer.body = Markup(context["quote_body"])
`

PR #1355, opened 2024-04-22
[15.0][ADD] outgoing_email_by_model

AungKoKoLin1997:
This PR recreates the module from the closed PR https://github.com/OCA/social/pull/1178. We respect the original work and are ready to step back if the original author decides to continue with it. We're open to their suggestions to improve this module further.

ping @mmequignon 

@qrtl QT2834


hbrunn (changes requested):
note you can always ask to have a PR reopened, and adding your review then makes clear you're serious about it.

I'd suggest to fill the CONFIGURE.rst file, and elaborate in DESCRIPTION.rst that this is about notifications. A short description of a use case would also be good.

kanda999 (approved):
Functional review :  It works as expected in my local env.

hbrunn (approved):

PR #1361, opened 2024-05-06
[17.0][MIG] email_template_qweb

AnaisLForgeFlow:
Standard Migration

@ForgeFlow


email_template_qweb/views/mail_template.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="email_template_form" model="ir.ui.view">
+        <field name="model">mail.template</field>
```
GuillemCForgeFlow, line 5: perhaps would be great to add a name to these views, just as a little improvement. WDYT? 😄

GuillemCForgeFlow (changes requested):
For the rest, LGTM 👍🏿

Rad0van:
While it works fine in some contexts it doesn't work in this scenario:

1. There is a template and a related view created. In my case through data/email_templates.xml - the same way as in here: https://github.com/OCA/social/blob/a4192a637916b9b35c33e537e09e2e90b6edf940/email_template_qweb/demo/mail_template.xml#L1-L10
2. You open email composer from code using something similar to this: https://github.com/odoo/odoo/blob/7564e62114721d6e57c3f4c3d306612d5daaf2e0/addons/sale/models/sale_order.py#L843-L870
3. What you get is an empty email body :-(

The reason being is this:

1. Mail composer computes the body here: https://github.com/odoo/odoo/blob/aff187f814998f4825da25f273eacef724686da3/addons/mail/wizard/mail_compose_message.py#L226-L235
2. Which calls this one: https://github.com/odoo/odoo/blob/aff187f814998f4825da25f273eacef724686da3/addons/mail/wizard/mail_compose_message.py#L1385-L1406

This line is the culprit:
`        template_value = self.template_id[template_fname] if self.template_id else False`

_template_fname_ contains "body_html" However that attribute is not defined in data XML so it returns with no body being rendered.

This usecase used to work in 16.0

email_template_qweb/models/mail_template.py
```diff
@@ -0,0 +1,44 @@
...
+                    custom_context = {
+                        "object": record,
+                        "email_template": self,
+                        "format_datetime": lambda dt,
+                        tz=False,
+                        dt_format=False,
+                        lang_code=False: format_datetime(
+                            self.env, dt, tz, dt_format, lang_code
+                        ),
+                    }
```
Rad0van, line 39:
```suggestion
                    custom_context = self._render_eval_context()
                    custom_context.update(
                        {
                            "object": record,
                            "email_template": self,
                            "format_datetime": lambda dt, tz=False, dt_format=False, lang_code=False: format_datetime(
                                self.env, dt, tz, dt_format, lang_code
                            ),
                        }
                    )
```

This is akin to #1249

Rad0van:
> While it works fine in some contexts it doesn't work in this scenario:
> 
> 1. There is a template and a related view created. In my case through data/email_templates.xml - the same way as in here: https://github.com/OCA/social/blob/a4192a637916b9b35c33e537e09e2e90b6edf940/email_template_qweb/demo/mail_template.xml#L1-L10
> 2. You open email composer from code using something similar to this: https://github.com/odoo/odoo/blob/7564e62114721d6e57c3f4c3d306612d5daaf2e0/addons/sale/models/sale_order.py#L843-L870
> 3. What you get is an empty email body :-(
> 
> The reason being is this:
> 
> 1. Mail composer computes the body here: https://github.com/odoo/odoo/blob/aff187f814998f4825da25f273eacef724686da3/addons/mail/wizard/mail_compose_message.py#L226-L235
> 2. Which calls this one: https://github.com/odoo/odoo/blob/aff187f814998f4825da25f273eacef724686da3/addons/mail/wizard/mail_compose_message.py#L1385-L1406
> 
> This line is the culprit: ` template_value = self.template_id[template_fname] if self.template_id else False`
> 
> _template_fname_ contains "body_html" However that attribute is not defined in data XML so it returns with no body being rendered.
> 
> This usecase used to work in 16.0

I see 2 possible fixes here:

1. override the __compute_body_ in _mail/wizard
/mail_compose_message.py_ - this does not feel exactly right but would do
2. sneak in some dummy body_html in our implementation of __generate_template_ - this does not feel right even more

Any ideas? Maybe @pedrobaeza ?

pedrobaeza:
Check where `self.template_id[template_fname]` is fed and fill it with the QWeb render result.

Rad0van:
> Check where `self.template_id[template_fname]` is fed and fill it with the QWeb render result.

Thanks for quick reply but I am not sure I follow :-(
In my opinion it is not fed anywhere - it is part of the mail.template: https://github.com/odoo/odoo/blob/aff187f814998f4825da25f273eacef724686da3/addons/mail/models/mail_template.py#L63-L65

For common templates it is filled in like this: https://github.com/odoo/odoo/blob/aff187f814998f4825da25f273eacef724686da3/addons/sale/data/mail_template_data.xml#L11-L41

In the __compute_body_ method it asks for _"body_html"_ exactly.

So I fail to see how to feed the result QWeb render to it...

pedrobaeza:
Well, I didn't know as well the internals, but it seems the `__get__` of the model or something is overriden to get that weird behavior returning a specific body when you pass the XML-ID.

simahawk:
This modules is not needed anymore. See https://github.com/OCA/social/issues/1480
Do you see any case where it could still provide value?

simahawk:
> This modules is not needed anymore. See #1480 Do you see any case where it could still provide value?

ops, this is still v17 not v18. Sorry... However, as is not merged yet, we could evaluate if we can get rid of it already in v17.

PR #1366, opened 2024-05-11
[16.0][ADD] mail_hide_chatter_send_message: new module

NL66278:
New module to hide the 'Send message' button on the chatter, to keep discussions internal.

First proposed here: https://github.com/OCA/web/pull/2811


mail_hide_chatter_send_message/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module was written to hide the Send Message button on the chatter/
+
+his is usefull for installations which do not want to send messages
```
yostashiro, line 3:
```suggestion
This module was written to hide the Send Message button on the chatter.

This is useful for installations which do not want to send messages
```

mail_hide_chatter_send_message/static/description/icon.png
yostashiro, line 1: Wasn't sure this icon represented what the module did, but non-blocking.

flotho (approved):
Do what is described on runboat
Code review LGTM
Thanks

HaraldPanten (changes requested):
Hi guys,

I think this module already exists. Why don't you migrate this one? --> https://github.com/OCA/social/tree/14.0/mail_restrict_send_button

@NL66278 @flotho @yostashiro

yostashiro:
> I think this module already exists. Why don't you migrate this one? --> https://github.com/OCA/social/tree/14.0/mail_restrict_send_button

@HaraldPanten The implementation of this module looks quite complicated and I imagine it to be a bit too high-maintenance for a simple need where control by group is not necessary. Just my opinion.

NL66278:
@HaraldPanten I did not know the 14.0 module existed. Note that that module is a lot more complicated, making the presence of the Send button conditional on a group the user belongs to. This can be very useful of-course, but for me it would be a lot of work to migrate the javascript (or even get to understand what al the code does).

HaraldPanten:
> @HaraldPanten I did not know the 14.0 module existed. Note that that module is a lot more complicated, making the presence of the Send button conditional on a group the user belongs to. This can be very useful of-course, but for me it would be a lot of work to migrate the javascript (or even get to understand what al the code does).

Hi @NL66278 I understand that the other module might be more complex, but considering that it makes the same functionallity with permission options I think that both modules shouldn't "coexist". Do you know what I mean?

Maybe JS could be a "headache", but the other module existed first and is more complete. IMHO"mail_restrict_send_button" is the one that should stay in OCA.

@etobella , as a OCA Board member, could you shed some light here?

THX.

etobella:
If it is not a problem, I will answer as PSC of Social. As Board, I would wait for a comment from a PSC :laughing: 

Well, I understand both points of view. From one side, we are duplicating functionalities, but the other module is too complex.

After checking `mail_restrict_send_button` I see to much overengineering for this small functionality. IMO, we should refactor it and then migrate it (or both at the same time). That would be my preferred path, because we would not loose functionality and we will improve the code at the same time.

If it is not refactored, I can understand it is too hard to migrate it as it is.

HaraldPanten:
> If it is not a problem, I will answer as PSC of Social. As Board, I would wait for a comment from a PSC 😆
> 
> Well, I understand both points of view. From one side, we are duplicating functionalities, but the other module is too complex.
> 
> After checking `mail_restrict_send_button` I see to much overengineering for this small functionality. IMO, we should refactor it and then migrate it (or both at the same time). That would be my preferred path, because we would not loose functionality and we will improve the code at the same time.
> 
> If it is not refactored, I can understand it is too hard to migrate it as it is.

Thanks Enric. Then, MIG + IMP is what should be done.

mail_hide_chatter_send_message/readme/DESCRIPTION.rst
NL66278: @yostashiro I corrected the description of the module according to your suggestion.

PR #1305, opened 2024-02-12
[16.0][MIG] mail_gateway + mail_gateway_telegram + mail_gateway_whatsapp: migration to 16.0

etobella:
TODO:

 -  In the chat window with contacts:
  - [ ] Ability to switch the communication medium to one incorporated by this system.
  - [x] Reception of responses in the same window.
  - [x] Support for reactions and emojis.
  - [x] Support for voice messages, images, and stickers.
  - [x] Support edition of messages
 - In the chatter of documents:
  - [x] Ability to send messages through one of these communication mediums.
  - [x] These messages will also be displayed in the contact's chat window with a navigable reference to the document.
  - [ ] Ability for direct navigation from each message to the general chat window with scrolling to the conversation fragment. --> Partially done. Odoo is not using it by default :thinking: 
  - [x] If the WhatsApp/Telegram reply option is used, identifying only the message origin, then the reply will also be placed in the chatter of the original document of the reply.
- Documentation
  - [x] Add documentation for whatsapp configuration
  - [x] Add documentation for telegram configuration

Also it includes web_broker_telegram for testing, because it is easier to configure for testing purposes.

PR #1372, opened 2024-05-24
[IMP] mail_post_defer: defer more mails

yajo:
In the previous version of this module, only emails sent through the chatter were explicitly deferred. However, there are more mails that are auto-generated by Odoo and could benefit a lot from the deferring system. One of such cases is the automatic assignation mail.

To cover that and more cases, now the auto-deferring is done at a lower level. If we get some hint that the mail sending is forced, or that it's already deferred, we skip the machinery. If we get nothing about all that, then we just add our defaults to avoid force-sending and defer by default.

The result is that more Odoo internal systems will use the mail queue, and thus there'll be less blocking.

As an exception, when running `_send_notifications()`, we force-disable the deferring system. That is executed by a cron, or by the user when force-sending a notification. So, only in that case the queue should be skipped by default.

@moduon MT-6204

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -15,7 +15,8 @@ odoo_test_flavor: Both
 odoo_version: 16.0
 org_name: Odoo Community Association (OCA)
 org_slug: OCA
-rebel_module_groups: []
+rebel_module_groups:
+- mail_post_defer
 repo_description: '{''TODO'': ''add repo description.''}'
 repo_name: social
 repo_slug: social
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,8 +36,17 @@ jobs:
       matrix:
         include:
           - container: ghcr.io/oca/oca-ci/py3.10-odoo16.0:latest
+            include: "mail_post_defer"
             name: test with Odoo
           - container: ghcr.io/oca/oca-ci/py3.10-ocb16.0:latest
+            include: "mail_post_defer"
+            name: test with OCB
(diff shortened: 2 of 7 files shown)


etobella:
Is this PR the fix for this?

https://github.com/OCA/social/pull/1305#issuecomment-2119174211

etobella:
This module is deferring notifications too. I think it was not intended, isn't it?

yajo:
Please notice that one assertion done in the test could provide a false positive due to https://github.com/odoo/odoo/pull/166676, but it's not this PR failure. Luckily it would still fail due to other assertions, so we can continue here and hope for the best regarding that PR.

yajo:
This module exists to defer notifications @etobella. It seems to me like it's doing what it should, and I don't think this will fix your problem.

It's just using one native odoo mechanism more than standard Odoo uses it. But if your modules support that system, both should be compatible.

If it's just tests that break, you can mark as rebel or skip some tests conditionally.

etobella:
Then this PR should not affect other modules tests.

In that case, you should make a change in behavior in order to not affect other modules tests. (For example, check that it is not testing or a context variable is passed)

The module breaking the defualt behavior is this one, so if the patch is not enabled, it should be considered as rebel :wink:

etobella:
Thanks @yajo :smile:

yajo:
Yes, I'm marking it rebel here.

pedrobaeza:
We need to solve the CI problem with mail_gateway, so merging:

/ocabot merge minor

PR #1379, opened 2024-05-30
[15.0][ADD] mail_no_user_assign_notification: New module

victoralmau:
New module

Please @pedrobaeza and @carolinafernandez-tecnativa can you review it?

@Tecnativa TT49257


mail_no_user_assign_notification/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+#. Go to Purchase.
```
carolinafernandez-tecnativa: Just a suggestion,  put this as an example, as it is generic for all modules that has user_id.

carolinafernandez-tecnativa:
The only user notification that is not working is when assigned new tasks as it is a many2many field and you can assign multiple users. 

**Configuration:**
image

**Mail received:**
image

I have checked with purchase order and sales order and leads works perfectly!

victoralmau:
Tasks are out of the scope of this module; moreover, it is not just adding the `user_ids` field because the `project` module uses another “_different_” mechanism.

carolinafernandez-tecnativa (approved):

HaraldPanten:
Hi @victoralmau @pedrobaeza I think we did something quite similar (we did it for V16) Maybe you want to have a look and both try to get the best from each module? Maybe we could join efforts --> https://github.com/sygel-technology/sy-social/tree/16.0/mail_block_user_assigned_message

mail_no_user_assign_notification/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,5 @@
+#. Go to Settings > Technical > System Parameters
+#. Edit "mail_no_user_assign_notification.models" key record
+#. Set any mode* (for example: res.partner).
```
chienandalu:
```suggestion
#. Set any model (for example: res.partner).
```

victoralmau:
> Hi @victoralmau @pedrobaeza I think we did something quite similar (we did it for V16) Maybe you want to have a look and both try to get the best from each module? Maybe we could join efforts --> https://github.com/sygel-technology/sy-social/tree/16.0/mail_block_user_assigned_message

I have reviewed what you indicate but I see that (although similar) they are different approaches (apply to the user_id field of the models defined in a system parameter, or indicate for each partner the model on which you want to do it), so I think they should be different modules.

mail_no_user_assign_notification/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module inhibits the auto-assign notification from the models that we want.
```
pedrobaeza:
```suggestion
This module inhibits the assignation notification from the models that we want.
```

mail_no_user_assign_notification/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,5 @@
+#. Go to Settings > Technical > System Parameters
+#. Edit "mail_no_user_assign_notification.models" key record
+#. Set any model* (for example: res.partner).
+
+It is possible to set several models separated by comma (e.g.: model1,model2).
```
pedrobaeza:
```suggestion
NOTE1: It is possible to set several models separated by comma (e.g.: model1,model2).

NOTE2: The model should have the field `user_id` for this to be effective.
```

mail_no_user_assign_notification/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+#. Go to Contacts.
+#. Create a partner and set another user as Salesperson.
+#. User will not receive the automatic assignment email.
+#. The same applies if another partner is modified.
```
pedrobaeza: Remember to not put a specific example, but the generic steps:

```suggestion
#. Go to one of the configured models (like Contacts).
#. Create or edit a record, and assign the field for the user (in contacts, it's called "Salesperson", in purchase orders it's "Purchase representative", etc).
#. The user won't receive the automatic assignment email.
```

mail_no_user_assign_notification/tests/test_mail_no_user_assign_notification.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_partner_create(self):
+        partner = self.env["res.partner"].create(
+            {
+                "name": "Test partner",
+                "user_id": self.user.id,
+            }
+        )
+        self.assertNotIn(
+            self.user.partner_id, partner.message_follower_ids.mapped("partner_id")
```
pedrobaeza: This is a side effect of what we are doing, but you shouldn't test this.

pedrobaeza (changes requested):

pedrobaeza:
Add `ROADMAP.rst` for commenting that this is not valid for project tasks users assignation, as the field is multi-relational. A specific extension would be needed.

victoralmau:
Changes done.

mail_no_user_assign_notification/readme/CONFIGURE.rst
pedrobaeza: Duplicated text

mail_no_user_assign_notification/readme/ROADMAP.rst
```diff
@@ -0,0 +1,2 @@
+This is not valid for project tasks users assignation, as the field is multi-relational.
+A specific extension would be needed.
```
pedrobaeza:
```suggestion
* This is not valid for project tasks users assignation, as the field is multi-relational.
  A specific extension would be needed.
```

mail_no_user_assign_notification/tests/test_mail_no_user_assign_notification.py
```diff
@@ -0,0 +1,38 @@
...
+        )
+        cls.icp.value = "res.partner"
+
+    def test_purchase_order_write(self):
+        partner = self.env["res.partner"].create({"name": "Test partner"})
+        self.assertNotIn(
+            self.user.partner_id, partner.message_follower_ids.mapped("partner_id")
+        )
+        partner.write({"user_id": self.user.id})
+        self.assertNotIn(
```
pedrobaeza: This still testing something different from the purpose of the module.

mail_no_user_assign_notification/models/mail_thread.py
```diff
@@ -0,0 +1,24 @@
...
+    def _message_auto_subscribe_followers(self, updated_values, default_subtype_ids):
+        """Remove user_id from updated_values if this model has been set.
+        This will prevent the user from auto subscrube to records on creation and/or
+        assignment."""
+        if updated_values.get("user_id"):
+            icp = self.sudo().env["ir.config_parameter"]
+            models_to_skip = icp.get_param(
+                "mail_no_user_assign_notification.models", ""
+            )
+            models_to_skip = models_to_skip.split(",")
```
pedrobaeza:
```suggestion
            models_to_skip = [x.strip() for x in models_to_skip.split(",")]
```

mail_no_user_assign_notification/tests/test_mail_no_user_assign_notification.py
victoralmau: Ok, changed the test approach.

pedrobaeza (approved):
/ocabot merge nobump

PR #1380, opened 2024-06-03
[FIX] mail_post_defer: notify from non-thread models and to author

yajo:
In some places of Odoo, there are direct calls to `self.env["mail.thread"].message_notify()` where the target model isn't extending `mail.thread`. These cases were failing, due to the grouping done by `mail.message.schedule`.

In some of those cases, mails were sent with the special `mail_notify_author=True` context, which got lost after the deferring step. In those cases, if the author's message was expected to arrive at the author's inbox, it wouldn't happen anymore.

@moduon MT-6337


yajo:
Fasttracking this important fix.

/ocabot merge patch

PR #1382, opened 2024-06-04
[FIX] mail_post_defer: recover "view $model" button

yajo:
Fixes a regression in https://github.com/OCA/social/pull/1380 that made posted messages lose the "View $model" button, even if the recipient had access.

@moduon MT-6348


yajo (commented):
Fast-tracking important fix.

/ocabot merge patch

yajo (commented):
Fast-tracking important fix.

/ocabot merge patch

PR #855, opened 2022-02-28
[MIG][14.0][mail_footer_notified_partner] Migration to v14 with renaming of module

cvinh:
Superseeds https://github.com/OCA/social/pull/801

PR #1391, opened 2024-06-10
[16.0][FIX][REF] mail_show_follower: internal notes

vvrossem:
This module aims to show the recipients of a message in the email header, but the previous implementation was not suitable for internal notes.

When **logging an internal note** with tagged users and/or partners, those specific recipients should be displayed
as well as the document's followers subscribed to this message subtype.

When **sending an email**, the recipients should be the document's followers.

This is achieved by using the recipient of each notification line


vvrossem:
Tests are failing because of `mail_gateway_telegram`.

mail_show_follower/models/res_partner.py
```diff
@@ -0,0 +1,31 @@
+from odoo import models
+
+
+class RePartner(models.Model):
+    _inherit = "res.partner"
+
+    def _get_excluded_partners_domain(self):
```
yankinmax: maybe add `@api.model`?

mail_show_follower/models/res_partner.py
```diff
@@ -0,0 +1,31 @@
...
+        return [
+            ("active", "in", (True, False)),
+            ("show_in_cc", "=", False),
+            ("groups_id", "not in", self.env.ref("base.group_portal").ids),
+        ]
+
+    def _get_excluded_partners(self):
```
yankinmax: maybe add `@api.model`?

mail_show_follower/models/res_partner.py
```diff
@@ -0,0 +1,31 @@
+from odoo import models
+
+
+class RePartner(models.Model):
```
yankinmax:
```suggestion
class ResPartner(models.Model):
```

yankinmax (approved):
LGTM: code review

mail_show_follower/models/mail_mail.py
```diff
@@ -60,96 +60,36 @@ def remove_p(markup_txt):
...
-                        # get show follower text
-                        final_cc = mail.with_context(
-                            lang=langs and langs[0]
-                        )._build_cc_text(partners)
-                        # it is saved in the body_html field so that it does
-                        # not appear in the odoo log
-                        mail.body_html = final_cc + mail.body_html
+        recipients = self.mapped("recipient_ids")
+
+        for mail in self:
```
yajo: issue: this could be displaying recipients from other mails

```suggestion
        for mail in self:
            recipients = self.recipient_ids
```

mail_show_follower/models/mail_mail.py
```diff
@@ -60,96 +60,36 @@ def remove_p(markup_txt):
...
+            record = self.env[mail.model].browse(mail.res_id)
+            company = (
+                record.company_id
+                if hasattr(record, "company_id")
+                else self.env.user.company_id
+            )
+            show_internal_users = company and company.show_internal_users_cc
+            show_in_cc_recipients = recipients.filtered(
+                lambda r: r._show_in_cc(show_internal_users)
+                and r._has_email_notification()
```
yajo: issue: the fact that the notification is sent by Odoo UI and not by email doesn't mean the recipient shouldn't appear in the list, right?

```suggestion
```

mail_show_follower/models/res_partner.py
```diff
@@ -0,0 +1,33 @@
...
+        self.ensure_one()
+        if not self.user_ids:
+            return True
+        excluded_partners = self._get_excluded_partners()
+        if show_internal_users:
+            return self not in excluded_partners and self.user_ids.mapped("show_in_cc")
+        else:
+            return self not in excluded_partners and self.user_ids.mapped(
+                "groups_id"
+            ) not in self.env.ref("base.group_portal")
```
yajo: issue: this is reversed

```suggestion
            return self not in excluded_partners and self.env.ref("base.group_portal") not in self.user_ids.groups_id
```

mail_show_follower/models/res_partner.py
```diff
@@ -0,0 +1,33 @@
...
+            ("groups_id", "not in", self.env.ref("base.group_portal").ids),
+        ]
+
+    @api.model
+    def _get_excluded_partners(self):
+        domain = self._get_excluded_partners_domain()
+        return self.env["res.users"].search(domain).mapped("partner_id")
+
+    def _has_email_notification(self):
+        return not self.user_ids or "email" in self.user_ids.mapped("notification_type")
```
yajo: Issue: probably an unnecessary method, as per the comment above.

yajo (changes requested):
Some issues detected.

Please also check those test failures...

Please @fcvalgar, @gelojr or @rafaelbn do a functional review.


Thanks!

vvrossem:
@etobella I'm encountering issues with tests failing on `mail_gateway_telegram`.
For example: 
```
  File "/__w/social/social/mail_gateway_telegram/models/mail_gateway_telegram.py", line 71, in _get_channel_vals
    update.message.chat.description or False,
AttributeError: 'Chat' object has no attribute 'description'. Did you mean: 'set_description'?

```
Any idea what could be the issue ?

etobella:
I think they made a change on the library. I will pin a version in order to avoid the issue. I will make it ASAP (tonight or tomorrow morning)

mail_show_follower/models/mail_mail.py
```diff
@@ -60,96 +60,34 @@ def remove_p(markup_txt):
...
+        for mail in self:
+            if not (mail.model and mail.res_id and group_portal):
+                continue
+            recipients = mail.recipient_ids | mail.notified_partner_ids
+            record = self.env[mail.model].browse(mail.res_id)
+            company = (
+                record.company_id
+                if hasattr(record, "company_id")
+                else self.env.user.company_id
+            )
```
yajo: issue: the company comes from the context nowadays

```suggestion
            company = getattr(record, "company_id", self.env.company)
```

mail_show_follower/models/mail_mail.py
```diff
@@ -60,96 +60,34 @@ def remove_p(markup_txt):
...
+            show_in_cc_recipients = recipients.filtered(
+                lambda r: r._show_in_cc(show_internal_users)
+            )
+            if len(show_in_cc_recipients) <= 1:
+                continue
+
+            langs = (
+                mail.recipient_ids.mapped("lang")
+                or mail.author_id.lang
+                or self.env.company.partner_id.lang
```
yajo: issue: better reuse the company from the record if available

```suggestion
                or company.partner_id.lang
```

mail_show_follower/models/res_partner.py
```diff
@@ -0,0 +1,30 @@
...
+        self.ensure_one()
+        if not self.user_ids:
+            return True
+        excluded_partners = self._get_excluded_partners()
+        if show_internal_users:
+            return self not in excluded_partners and self.user_ids.mapped("show_in_cc")
+        else:
+            return self not in excluded_partners and self.env.ref(
+                "base.group_portal"
+            ) not in self.user_ids.mapped("groups_id")
```
yajo: issue(performance): this implementation is not a good idea.

Main method `_show_in_cc()` is called in loop from `MailMail._send()`. For each loop iteration, it will make a search. Making a search involves:
1. Flushing changes to Postgres.
2. Clearing some caches.
3. Asking Postgres for some more data.

We should avoid interaction with DB as much as possible to avoid performance penalties. If we have to contact it, then we should batch it as much as possible.

Here you have an alternative implementation that reuses ORM cache and will be much faster when called in loop. I take the chance to add some comments that will make it more clear what the method does:

```suggestion
    def _filter_shown_in_cc(self, show_internal_users):
        """Get partners that should be displayed as CC on mails."""
        # Never display hidden users
        result = self.filtered_domain(
            [
                "|",
                ("user_ids", "=", False),
                ("user_ids.show_in_cc", "=", True),
            ]
        )
        # Remove internal users from result if needed
        if not show_internal_users:
            internal_users = result.filtered_domain(
                [
                    ("user_ids.active", "=", True),
                    ("user_ids.groups_id", "in",  self.env.ref("base.group_user").ids),
                ]
            )
            result -= internal_users
        return result
```

By using `filtered_domain`, you don't have to ask the ORM again, and you batch operations.

mail_show_follower/models/mail_mail.py
```diff
@@ -60,96 +60,34 @@ def remove_p(markup_txt):
...
+            record = self.env[mail.model].browse(mail.res_id)
+            company = (
+                record.company_id
+                if hasattr(record, "company_id")
+                else self.env.user.company_id
+            )
+            show_internal_users = company and company.show_internal_users_cc
+            show_in_cc_recipients = recipients.filtered(
+                lambda r: r._show_in_cc(show_internal_users)
+            )
```
yajo: Following one suggestion below:

```suggestion
            show_in_cc_recipients = recipients._filter_shown_in_cc()
```

mail_show_follower/models/mail_mail.py
```diff
@@ -60,96 +60,34 @@ def remove_p(markup_txt):
...
-                            lang=langs and langs[0]
-                        )._build_cc_text(partners)
-                        # it is saved in the body_html field so that it does
-                        # not appear in the odoo log
-                        mail.body_html = final_cc + mail.body_html
+
+        for mail in self:
+            if not (mail.model and mail.res_id and group_portal):
+                continue
+            recipients = mail.recipient_ids | mail.notified_partner_ids
```
yajo: suggestion: as said in https://github.com/OCA/social/pull/855#issuecomment-1575081147, one of the failures this module has is:

>  Don't show followers (they are) if the follower use Odoo messaging without emails

I think it would be a good time to fix this problem too, if you want.

The point is that you should include in `recipients` also those that are notified by other systems (not only mail; also those that read notifications in Odoo directly).

If you don't feel like so, this can be handled later, but at least it would be nice if you can add it to `readme/ROADMAP.rst` in case it isn't there yet.

yajo (changes requested):

vvrossem:
Thank you @yajo for your additional review :pray: . I'll handle it asap :)

fcvalgar (approved):
Functional review,

LGTM, thank you @vvrossem 

imageç

image

yajo (approved):
Code review OK.

@fcvalgar please update your functional review. The part you highlight in the Mailhog interface does not belong to this module. Instead, this module adds recipients info in the message body, and I cannot see any of that in the screenshot. Maybe it was not installed when you tried it?

fcvalgar:
> Code review OK.
> 
> @fcvalgar please update your functional review. The part you highlight in the Mailhog interface does not belong to this module. Instead, this module adds recipients info in the message body, and I cannot see any of that in the screenshot. Maybe it was not installed when you tried it?

I think the behaviour is not as expected. The followers of the notes do not appear in the body of the message.

image

image

image

yajo (approved):
I've double-checked the functional test. All seems to be working fine.

As you can see here, if Marc Demo is subscribed to discussions and handles notifications in Odoo, he still appears in the CC list:
image

I create a new user with mail newuser@example.com and subscribe to that record's notes by email. I send an internal note:

image

Now I remove Marc Demo's subscription to notes and send another message:

image

There's no extra header. This was probably the case you triggered, @fcvalgar. IMHO it's OK because there are no more people subscribed, so that header wouldn't add much value.

So this seems like a functional improvement.

/ocabot merge minor

PR #1393, opened 2024-06-12
[15.0][FIX]: Prevent error when the result is empty

hhgabelgaard:
Replaces #1392


StefanRijnhart:
Can you add a test to reproduce the issue?

PR #1402, opened 2024-06-17
16.0 - Migration mail_embed_image

imlopes:


mail_embed_image/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2019 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Mail Embed Image",
+    "version": "16.0.1.0.1",
```
cyrilmanuel: 16.0.1.0.0 (not an update)

mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,92 @@
...
+                    _logger.debug("Fetching image from %s", src)
+                    if response.status_code == 200:
+                        cid = uuid.uuid4().hex
+                        # convert cid to rfc2047 encoding
+                        filename_encoded = "=?utf-8?b?%s?=" % b64encode(
+                            cid.encode("utf-8")
+                        ).decode("utf-8")
+                        image_content = response.content
+                        filepart = MIMEImage(image_content)
+                        filepart.add_header("Content-ID", f"<{cid}>")
```
cyrilmanuel, line 82: ask : shall not be the cid.encode("utf-8") ?

cyrilmanuel (approved):
not an expert but LGTM

gurneyalex (approved):

mail_embed_image/models/ir_mail_server.py
```diff
@@ -83,48 +52,41 @@ def build_email(
...
-                img.set('src', 'cid:%s' % (str(cid)))
-        return root
+        root = fromstring(html_body)
+        images = root.xpath("//img")
+        fileparts = []
+        for img in images:
+            src = img.get("src")
+            if src and not src.startswith("data:") and not src.startswith("base64:"):
+                try:
+                    response = requests.get(src, timeout=10)
```
hbrunn, line 72: the reason the original did the fetching in a somewhat roundabout way is that what you do here allows malicious users to craft emails to themselves and have Odoo fetch arbitrary resources from the internal network. Can be harmless, can be catastrophic depending on what's accessible from there

hbrunn (changes requested):

mail_embed_image/models/ir_mail_server.py
fliot: May be harmless or catastrophic...
Yes, probably.
Is functionally really expected, yes too.
Thanks @imlopes for this code.

imlopes: thanks for the feedback @hbrunn @fliot 
Im really not very comfortable with this kind of module. I've tried to keep as much as I could the original code but lot of changes between the original version and the new one on odoo side.

hbrunn: well as it is I really don't think it should be published.

But what was the problem? Currently you'd call https://github.com/OCA/OCB/blob/16.0/odoo/addons/base/models/ir_http.py#L95 which relieves you from constructing the routing map yourself, and you only need to create a dummy request bound to the public user in a pretty similar way the v10 version does it

StephaneMangin:
superseeded by https://github.com/OCA/social/pull/1499

imlopes:
Thanks a lot @StephaneMangin 🫶

PR #1405, opened 2024-06-21
[14.0][IMP] mail_restrict_follower_selection: allow to use ref in domain

TDu:
With this is change it is now possible to use the `ref` function in the domain set in a parameter, allowing to include xmlids in the domain.

### DIFF
--- a/mail_restrict_follower_selection/models/mail_thread.py
+++ b/mail_restrict_follower_selection/models/mail_thread.py
@@ -2,6 +2,8 @@ from odoo import models
 from odoo.tools import config
 from odoo.tools.safe_eval import safe_eval
 
+from ..utils import _id_get
+
 
 class MailThread(models.AbstractModel):
     _inherit = "mail.thread"
@@ -20,7 +22,9 @@ class MailThread(models.AbstractModel):
         domain = self.env[
             "mail.wizard.invite"
         ]._mail_restrict_follower_selection_get_domain()
-        eval_domain = safe_eval(domain)
+        eval_domain = safe_eval(
+            domain, locals_dict={"ref": lambda str_id: _id_get(self.env, str_id)}
+        )
         for key in result:
             for partner_id, email, reason in result[key]:
                 if partner_id:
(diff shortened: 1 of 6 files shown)


mmequignon (approved):
LGTM

mail_restrict_follower_selection/models/mail_thread.py
```diff
@@ -20,7 +20,7 @@ def _message_add_suggested_recipient(
         domain = self.env[
             "mail.wizard.invite"
         ]._mail_restrict_follower_selection_get_domain()
-        eval_domain = safe_eval(domain)
+        eval_domain = safe_eval(domain, locals_dict={"ref": self.env.ref})
```
StefanRijnhart: It's probably safer to use a version of `ref` that returns the id rather than the browse record, like Odoo does when interpreting XML data: https://github.com/odoo/odoo/blob/14.0/odoo/tools/convert.py#L62. That way there is no vector in which an attacker can call methods on a record from a domain.

TDu: Thanks for the review @StefanRijnhart, good point.
I updated the PR.

mail_restrict_follower_selection/utils.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+
+def _id_get(env, id_str, raise_if_not_found=True):
+    """Have a more secure ref function for use with safe_eval.
+
+    Returning only the ID of the record.
+
+    """
+    return env.ref(id_str, raise_if_not_found).id
```
StefanRijnhart: This is not going to work if not found, even without raising:
```
>>> env.ref("not.exists", False).id
  File "<console>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'id'
```
Your simplest option might be to just drop `raise_if_not_found` from the method signature.

sebalix: I removed this parameter to make things easier, thank you.

Commits squashed as well so the PR is ready to merge.

simahawk (approved):

StefanRijnhart (approved):

PR #1410, opened 2024-07-05
[17.0][MIG] mail_tracking_mailgun: Migration to 17.0 

chienandalu:
Pending:

- [x] <del>Doing practical tests with sandbox API keys</del>

cc @Tecnativa TT49914


mail_tracking_mailgun/models/mail_tracking_email.py
```diff
@@ -78,8 +78,8 @@ def _mailgun_values(self):
         if not api_key:
             raise ValidationError(_("There is no Mailgun API key!"))
         api_url = icp.get_param("mailgun.api_url", "https://api.mailgun.net/v3")
-        catchall_domain = icp.get_param("mail.catchall.domain")
-        domain = icp.get_param("mailgun.domain", catchall_domain)
+        catchall_domain = self.env["mail.alias.domain"].search([], limit=1).name
```
pedrobaeza: Shouldn't this be adapted for all the possible domains?

mail_tracking_mailgun/models/res_partner.py
```diff
@@ -34,9 +34,13 @@ def _email_bounced_set(self, reason, event):
...
             )
-            body = _(
-                "Email has been bounced: %(email)s\nReason: %(reason)s\nEvent: %(event_str)s"
-            ) % ({"email": partner.email, "reason": reason, "event_str": event_str})
+            body = (
+                _(
+                    "Email has been bounced: %(email)s\nReason: "
+                    "%(reason)s\nEvent: %(event_str)s"
+                ),
+                ({"email": partner.email, "reason": reason, "event_str": event_str}),
```
pedrobaeza: This doesn't seem correctly put (with the weird parenthesis), and `_(` accepts already the parameters as new arguments.

mail_tracking_mailgun/models/mail_tracking_email.py
chienandalu: Well, this is done just to get a default value if an specific one isn't set in the settings.

mail_tracking_mailgun/models/res_partner.py
chienandalu: Changed

pedrobaeza (approved):
Thank you!

/ocabot merge nobump

PR #1419, opened 2024-07-15
[17.0][ADD] mail_prevent_bounce_loop

trisdoan:


mail_prevent_bounce_loop/README.rst
```diff
@@ -0,0 +1,110 @@
...
+no_bounce_email as below:
+
+|no_bounce_email|
+
+To never send email notification when receiving auto-reply email from a
+specific email address identified as an auto-reply address, you can tick
+is_automatic_reply_address as below:
+
+|is_automatic_reply|
+
```
ajaniszewska-dev, line 65: in odoo this seems to be not well displayed:

image

ajaniszewska-dev (approved):
functional test for automatic reply ok, minor comment

mail_prevent_bounce_loop/pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["whool"]
+build-backend = "whool.buildapi"
```
hbrunn, line 3: remove

mail_prevent_bounce_loop/models/res_partner.py
```diff
@@ -0,0 +1,16 @@
...
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    no_bounce_email = fields.Boolean(
+        help="When checked, bounced emails will not be forwarded to this address"
+    )
+    is_automatic_reply_address = fields.Boolean(
```
hbrunn, line 13: I find this name a little misleading, better use one that expresses what this *does*, as in `no_chatter_forward` or similar

hbrunn (commented):
I still believe the correct way to handle this is https://github.com/OCA/social/tree/12.0/mail_autogenerated_header and I wonder why this never made it into Odoo core.

ajaniszewska-dev:
> I still believe the correct way to handle this is https://github.com/OCA/social/tree/12.0/mail_autogenerated_header and I wonder why this never made it into Odoo core.

@hbrunn  i am trying here: https://github.com/odoo/odoo/pull/180156

PR #1428, opened 2024-08-07
[FIX] mail_composer_cc_bcc: fix duplicate key value error "unique_mail_message_id_res_partner_id_if_set"

BryanSomko:
Fix duplicate key value violates unique constraint "unique_mail_message_id_res_partner_id_if_set" 
when adding the same user to CC field that is already set as a follower
image


cuongnmtm:
This issue also happens in 17.0

cuongnmtm:
Could you please run pre-commit for your changes?

image


And the automatic tests are failed.

PR #1429, opened 2024-08-07
[15.0][FIX] mail_drop_target: drop multiple files

ArnauCForgeFlow:
I have noticed that when you try to upload multiple files that are not emails, it uploads them multiple times.

image


LoisRForgeFlow (approved):
Functional test, thanks for the fix :+1:

hbrunn (approved):

PR #1340, opened 2024-03-25
[14.0] mail_notification_clean_status_error: remove notifications in error

sebalix:
Delete email notifications in error.

The scheduled action *Notification: Delete Notifications older than 6 Month* provided by Odoo is responsible to delete notifications that have been sent successfully.

However, it doesn't delete the notifications that could not be sent, and their number could keep growing over time, impacting the performance of some queries related to the chatter.

This module extends the scheduled action of Odoo to also delete such notifications.

On a DB with 20M+ of `mail_message` records and 100k notifications in error over 1M+, this allows to speed up the loading of the Discuss app widget (top bar) and the chatter of records from ~35s to ~1s.

Ref. 4352

PR #1430, opened 2024-08-08
[14.0][IMP] mail_tracking: add job to delete old tracking records

henrybackman:
Records in `mail_tracking_email` table grow over time and increase the database size, which slows down backup and restore operations. (For example, in one customer db the `mail_tracking_email` table takes almost 2GB)

Added `autovacuum` to `mail_tracking_email` that removes by default records over 6 months old.
The deletion can be enabled with a configuration variable.
Due to possibly a large number of records to be deleted on first run, set a default limit of 5000 per run.

### DIFF
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -473,3 +473,40 @@ class MailTrackingEmail(models.Model):
         # - return 'NONE' if this request is not for you
         # - return 'ERROR' if any error
         return "NONE"  # pragma: no cover
+
+    def _get_old_mail_tracking_email_domain(self, max_age_days):
+        target_write_date = fields.Datetime.subtract(
+            fields.Datetime.now(), days=max_age_days
+        )
+        return [("write_date", "<", target_write_date)]
+
+    @api.autovacuum
+    def _gc_mail_tracking_email(self, limit=5000):
+        config_max_age_days = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("mail_tracking.mail_tracking_email_max_age_days")
+        )
+        try:
+            max_age_days = int(config_max_age_days)
+        except ValueError:
+            max_age_days = 0
+
+        if not max_age_days > 0:
+            return False
+
+        domain = self._get_old_mail_tracking_email_domain(max_age_days)
+        records_to_delete = self.search(domain, limit=limit).exists()
+        if records_to_delete:
+            _logger.info(
+                "Deleting %s mail.tracking.email records", len(records_to_delete)
+            )
+            self.flush()
+            # Using a direct query to avoid ORM as it causes an issue with
+            # a related field mass_mailing_id in customer DB when deleting
+            # the records. This might be 14.0 specific, so changing to
+            # .unlink() should be tested when forward porting.
+            query = "DELETE FROM mail_tracking_email WHERE id IN %s"
+            args = (tuple(records_to_delete.ids),)
+            self.env.cr.execute(query, args)
+            self.invalidate_cache()
--- a/mail_tracking/models/res_config_settings.py
+++ b/mail_tracking/models/res_config_settings.py
@@ -8,3 +8,9 @@ class ResConfigSettings(models.TransientModel):
         related="company_id.mail_tracking_show_aliases",
         readonly=False,
     )
+    mail_tracking_email_max_age_days = fields.Integer(
+        "Max age in days of mail tracking email records",
+        config_parameter="mail_tracking.mail_tracking_email_max_age_days",
+        help="If set as positive integer enables the deletion of "
+        "old mail tracking records to reduce the database size.",
+    )
--- a/mail_tracking/tests/test_gc_mail_tracking_email.py
+++ b/mail_tracking/tests/test_gc_mail_tracking_email.py
@@ -1,0 +1,87 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields
+from odoo.tests.common import SavepointCase
+
+
+class TestMailTrackingEmailCleanUp(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.settings = cls.env["res.config.settings"].create(
+            {"mail_tracking_email_max_age_days": 365}
+        )
+        cls.settings.set_values()
+        cls.partner = cls.env.ref("base.res_partner_address_28")
+        cls.message = cls.env["mail.message"].create(
+            {
+                "model": "res.partner",
+                "res_id": cls.partner.id,
+                "body": "TEST",
+                "message_type": "email",
+                "subtype_id": cls.env.ref("mail.mt_comment").id,
+                "author_id": cls.partner.id,
+                "date": "2024-03-26",
+            }
+        )
+        cls.recent_mail_tracking_email = cls.env["mail.tracking.email"].create(
+            {"mail_message_id": cls.message.id}
+        )
+        # Can't set the write_date directly as it gets overwritten by the ORM
+        cls.old_mail_tracking_email = cls.env["mail.tracking.email"].create(
+            {"mail_message_id": cls.message.id}
+        )
+        cls.total_count = 2
+        cls.recent_count = 1
+        cls.domain = [
+            ("mail_message_id", "=", cls.message.id),
+        ]
+
+    def _set_write_date(self):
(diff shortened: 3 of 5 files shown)


mail_tracking/models/mail_tracking_email.py
```diff
@@ -473,3 +473,25 @@ def event_process(self, request, post, metadata, event_type=None):
...
+        if not enable_deletion:
+            _logger.info("Mail tracking email deletion job is disabled.")
+            return False
+
+        target_write_date = fields.Datetime.subtract(
+            fields.Datetime.now(), days=max_age_days
+        )
+        domain = [
+            ("write_date", "<", target_write_date),
+        ]
```
sebalix: You should extract the domain computation in a separate method, so it's easy to override to fine-tune the domain (if we want to keep failure values for instance).

mail_tracking/models/res_config_settings.py
```diff
@@ -8,3 +8,10 @@ class ResConfigSettings(models.TransientModel):
         related="company_id.mail_tracking_show_aliases",
         readonly=False,
     )
+    enable_mail_tracking_email_deletion_job = fields.Boolean(
```
sebalix: Not sure if this option is useful if we are using a dedicated cron task: we could simply archive/unarchive it. What do you think?

sebalix: This option could be useful if the deletion was triggered by the `api.autovacuum` feature of Odoo.

pedrobaeza:
Do you have some metrics about performance degradation that leads to this action?

sebalix:
@pedrobaeza this action is here to reduce the DB size (to speed up backup and restore operations mainly), currently we have a DB with the `mail_tracking_value` taking almost 2GB (in the top 10 of biggest table in this DB, 8th position, 1st being `mail_message`).
About performance I checked but there isn't a real impact here (while this similar PR about notifications is reducing a lot the time needed to open the Odoo JS client: https://github.com/OCA/social/pull/1340)

pedrobaeza:
OK, thanks for the information. It's good to add this information initially (and include it in the commit message) for avoiding to guess what is the improvement.

It would be interesting to check if there's a way to optimize the DB schema of this table to reduce this size as well. Maybe there's a related stored or similar fields that increase the size a lot.

mail_tracking/models/mail_tracking_email.py
henrybackman: Done

mail_tracking/models/res_config_settings.py
henrybackman: Switched implementation to use `api.autovacuum` and no new cron jobs required

henrybackman:
Thank you @pedrobaeza for checking. I don't have anything to add to @sebalix comment but I added the relevant details to the description :+1:

sebalix (approved):

henrybackman:
Test added to the `tests/__init__.py` and improved wording in the settings description

mail_tracking/tests/test_gc_mail_tracking_email.py
```diff
@@ -0,0 +1,80 @@
...
+
+class TestMailTrackingEmailCleanUp(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.settings = cls.env["res.config.settings"].create(
+            {"enable_old_mail_tracking_email_deletion": True}
+        )
+        cls.settings.execute()
```
simahawk: Always use `set_values` instead of `execute` as it does too many things that will slow down the setup

simahawk:
@henrybackman when you push again pls report the description in the commit msg.
Note: if you do that before opening a PR you'll get the PR description for free :wink:

mail_tracking/tests/test_gc_mail_tracking_email.py
henrybackman: Changed to `set_values`

mail_tracking/models/mail_tracking_email.py
```diff
@@ -473,3 +473,26 @@ def event_process(self, request, post, metadata, event_type=None):
...
         return "NONE"  # pragma: no cover
+
+    def _get_old_mail_tracking_email_domain(self, max_age_days):
+        target_write_date = fields.Datetime.subtract(
+            fields.Datetime.now(), days=max_age_days
+        )
+        return [("write_date", "<", target_write_date)]
+
+    @api.autovacuum
+    def _gc_mail_tracking_email(self, max_age_days=180, limit=5000):
```
sebalix: `max_age_days` should be configurable through a system parameter as well.
We could replace the current boolean config option by an integer instead, 0 meaning nothing is deleted, or a number of retention days to enable it.

mail_tracking/models/mail_tracking_email.py
```diff
@@ -473,3 +473,26 @@ def event_process(self, request, post, metadata, event_type=None):
...
+        )
+
+        if not enable_deletion:
+            _logger.info("Mail tracking email deletion is disabled.")
+            return False
+
+        domain = self._get_old_mail_tracking_email_domain(max_age_days)
+        records_to_delete = self.search(domain, limit=limit)
+        _logger.debug("Deleting %s mail.tracking.email records", len(records_to_delete))
+        return records_to_delete.unlink()
```
sebalix: After testing this on a customer DB, it triggers the following error (linked to related fields):
```python
  File "/odoo/src/odoo/addons/base/models/ir_autovacuum.py", line 38, in _run_vacuum_cleaner
    func(model)
  File "/odoo/external-src/social/mail_tracking/models/mail_tracking_email.py", line 503, in _gc_mail_tracking_email
    records_to_delete.unlink()
  File "/odoo/external-src/connector/component_event/models/base.py", line 118, in unlink
    result = super(Base, self).unlink()
  File "/odoo/src/odoo/models.py", line 3501, in unlink
    self.flush()
  File "/odoo/src/odoo/models.py", line 5472, in flush
    self.recompute()
  File "/odoo/src/odoo/models.py", line 5931, in recompute
    process(field)
  File "/odoo/src/odoo/models.py", line 5918, in process
    field.recompute(existing)
  File "/odoo/src/odoo/fields.py", line 1178, in recompute
    self.compute_value(record)
  File "/odoo/src/odoo/fields.py", line 1198, in compute_value
    records._compute_field_value(self)
  File "/odoo/src/odoo/models.py", line 4079, in _compute_field_value
    odoo.fields.determine(field.compute, self)
  File "/odoo/src/odoo/fields.py", line 85, in determine
    return needle(records, *args)
  File "/odoo/src/odoo/fields.py", line 574, in _compute_related
    record[self.name] = self._process_related(value[self.related_field.name])
  File "/odoo/src/odoo/models.py", line 5702, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/odoo/src/odoo/fields.py", line 2495, in __get__
    return super().__get__(records, owner)
  File "/odoo/src/odoo/fields.py", line 1025, in __get__
    _("(Record: %s, User: %s)") % (record, env.uid),
odoo.exceptions.MissingError: Record does not exist or has been deleted.
```
I tried different things but only a raw `DELETE` query is working while being simple:
```suggestion
        records_to_delete = self.search(domain, limit=limit).exists()
        if records_to_delete:
            _logger.info("Deleting %s mail.tracking.email records", len(records_to_delete))
            self.flush()
            query = "DELETE FROM mail_tracking_email WHERE id IN %s"
            args = (tuple(records_to_delete.ids),)
            self.env.cr.execute(query, args)
            self.invalidate_cache()
```

sebalix (changes requested):

mail_tracking/models/mail_tracking_email.py
```diff
@@ -473,3 +473,41 @@ def event_process(self, request, post, metadata, event_type=None):
...
+            .sudo()
+            .get_param("mail_tracking.mail_tracking_email_max_age_days")
+        )
+        try:
+            max_age_days = int(config_max_age_days)
+        except ValueError:
+            max_age_days = 0
+
+        if not max_age_days > 0:
+            _logger.info("Mail tracking email deletion is disabled.")
```
sebalix: As the autovacuum is run quite often, I would switch this log to DEBUG:
```suggestion
            _logger.debug("Mail tracking email deletion is disabled.")
```

pedrobaeza: I don't see this as needed really. It will spam the log with no reason.

henrybackman: Good points, removed it

sebalix (approved):

simahawk (approved):

simahawk:
@pedrobaeza good for you?

pedrobaeza (approved):
/ocabot merge minor

PR #1431, opened 2024-08-08
[16.0][FIX] mail_activity_team, fix double send activity notification

ChrisOForgeFlow:
First notification is sent by

https://github.com/odoo/odoo/blob/16.0/addons/mail/models/mail_activity.py#L348

Because new field team_user_id is a related field, when both field are passed to create function, trigger a write to sync fields, because that in validation of change of user is trigger and send as a write

Second notification is send by this one, this is default when record is created

https://github.com/odoo/odoo/blob/16.0/addons/mail/models/mail_activity.py#L302-L303

With change proposed, send same value to both fields if team_user_id is filled

@ForgeFlow

### DIFF
--- a/mail_activity_team/models/mail_activity.py
+++ b/mail_activity_team/models/mail_activity.py
@@ -31,6 +31,15 @@ class MailActivity(models.Model):
         index=True,
     )
 
+    @api.model_create_multi
+    def create(self, values):
+        for value in values:
+            if "team_user_id" in value and value.get(
+                "team_user_id", False
+            ) != value.get("user_id"):
+                value.update({"user_id": value.get("team_user_id", False)})
+        return super().create(values)
+
     @api.onchange("user_id")
     def _onchange_user_id(self):
         if not self.user_id or (


JordiBForgeFlow (approved):

AaronHForgeFlow:
How can the activity be created with team_user_id different from user_id ? Can I do that as a user?

ChrisOForgeFlow:
> How can the activity be created with team_user_id different from user_id ? Can I do that as a user?

It is not done by user, because field is hide, but is filled by other onchanges in same module when change team for example https://github.com/OCA/social/blob/16.0/mail_activity_team/models/mail_activity.py#L45

AaronHForgeFlow (approved):
Ok, thanks for the clarification

AaronHForgeFlow:
Can someone reopen this?

JordiBForgeFlow:
@etobella do you have rights to reopen?

JordiBForgeFlow:
@etobella gracies!

JordiBForgeFlow (approved):

etobella:
Can you rebase?

AaronHForgeFlow:
Rebased! @etobella

etobella:
Tests are failing. Can you check it?

AaronHForgeFlow:
So sad. It seems there is an issue with mass_mailing_custom_unsubscribe & mail_composer_cc_bcc modules. Other PRs in 16.0 are failing because of the same reason. Hope someone fix it :smile:

rousseldenis:
@ChrisOForgeFlow Is it related also to https://github.com/OCA/social/pull/1353 ?

AaronHForgeFlow:
> @ChrisOForgeFlow Is it related also to #1353 ?

It seems to me both resolve the same issue

StefanRijnhart (commented):
Would it be possible to add a reproducing test (maybe patch action_notify and check call_count)?

AlexPForgeFlow (approved):

PR #1434, opened 2024-08-20
[15.0][FIX] mail_activity_done: KeyError user_activites[model.model]

bobslee:
A KeyError occurs for a missing model in the user_activities, when updating the user_activites[model.model] dict.

The fix to initialize the activity data wasn't straightforward.
However just backporting (copied) the code from 16.0 seems to work!

### DIFF
--- a/mail_activity_done/models/res_users.py
+++ b/mail_activity_done/models/res_users.py
@@ -18,7 +18,7 @@ class ResUsers(models.Model):
         for item in res:
             user_activities[item["model"]] = item
         # Redo the method only with the archived records and subtract them.
-        query = """SELECT array_agg(res_id) as res_ids, m.id, count(*),
+        query = """SELECT array_agg(res_id) as res_ids, act.res_model, m.id, count(*),
                     CASE
                         WHEN %(today)s::date - act.date_deadline::date = 0 Then 'today'
                         WHEN %(today)s::date - act.date_deadline::date > 0 Then 'overdue'
@@ -28,7 +28,7 @@ class ResUsers(models.Model):
                 JOIN ir_model AS m ON act.res_model_id = m.id
                 WHERE user_id = %(user_id)s
                 AND act.active = False
-                GROUP BY m.id, states;
+                GROUP BY m.id, act.res_model, states;
                 """
         self.env.cr.execute(
             query,
@@ -42,6 +42,9 @@ class ResUsers(models.Model):
             lambda: {"today": set(), "overdue": set(), "planned": set(), "all": set()}
         )
         for data in activity_data:
+            if not user_activities.get(data["res_model"]):
+                # this especially filters out records due to access rights
+                continue
             records_by_state_by_model[data["id"]][data["states"]] = set(data["res_ids"])
             records_by_state_by_model[data["id"]]["all"] = records_by_state_by_model[
                 data["id"]


bobslee:
The failed tests "with Odoo, OCB" ain't related with the change of this PR.

StefanRijnhart:
That's not a backport from 16, that's a revert of https://github.com/OCA/social/pull/1222 by @victoralmau that never made it into 16!

Would it be possible to add a test that reproduces the issue, so that we can be sure that the problem is not going to resurface after a future refactoring?

StefanRijnhart:
It also raises the question if we can keep the previous improvement and fix the KeyError there.

victoralmau:
Some interesting things to keep in mind:

- I no longer use the `mail_activity_done` module.
- Created https://github.com/odoo/odoo/pull/137070 to add some hooks in odoo and simplify everything, avoiding redoing the `systray_get_activities()` method completely (only to add `act.done = False` to sql).
- We did not FWP to 16.0 from https://github.com/OCA/social/pull/1222 because it was not necessary, i.e. the https://github.com/odoo/odoo/blob/16.0/addons/mail/models/res_users.py#L201 method does not use sql.
- The `mail_activity_done` module in 16.0 is “_incorrect_” because it redoes the `systray_get_activities()` method completely and it is no longer necessary to do anything in that method, the activities you want to “skip” will be archived.
- The `mail_activity_team` module in 16.0 is “_incorrect_” because it redoes the `systray_get_activities()` method completely and it is no longer necessary to do almost nothing in this method, only filtering by the activities of the corresponding team (without doing it by sql).

StefanRijnhart:
Excellent insights @victoralmau, thank you!

bobslee:
@StefanRijnhart , @victoralmau Thanks for your remarks and I agree!
It didn't appeared to me Victor's change was an improvement to reintroduce the `super` call again.
I shall revert and check whether I can add a fix for the original v15 implementation.
Keep you posted!

bobslee:
@StefanRijnhart, @victoralmau I just (force) pushed the change.
Please review.

mail_activity_done/models/res_users.py
```diff
@@ -38,10 +38,28 @@ def systray_get_activities(self):
...
         activity_data = self.env.cr.dictfetchall()
+        model_ids = [a["id"] for a in activity_data]
+        model_names = {
+            n[0]: n[1] for n in self.env["ir.model"].sudo().browse(model_ids).name_get()
+        }
         records_by_state_by_model = defaultdict(
             lambda: {"today": set(), "overdue": set(), "planned": set(), "all": set()}
         )
         for data in activity_data:
+            if not user_activities.get(data["res_model"]):
```
StefanRijnhart, line 45: It seems to me that if this is the case, none of the retrieved records that the activities are linked to are accessible to the current user (as per https://github.com/odoo/odoo/blob/f17428b/addons/mail/models/res_users.py#L178-L182), so you can just `continue` here.

StefanRijnhart (commented):
No test, so you really made me think hard about what might cause the issue :smiling_face_with_tear: but if I understand correctly you might be able to simplify.

mail_activity_done/models/res_users.py
bobslee: @StefanRijnhart Thanks, very good catch!
I reverted and pushed to proper fix now.
Please review and let me know.

bobslee:
> No test, so you really made me think hard about what might cause the issue 🥲 but if I understand correctly you might be able to simplify.

Indeed, I know about the lack of the unit test.
Unfortunately no time (budget) left to write a unittest.

StefanRijnhart (approved):
Thanks for the updates!

PR #1435, opened 2024-08-22
[16.0][FIX]mail_quoted_reply: unintentional tags

ChristophAbenthungCibex:
<...> text of subject and from in the quoted reply block are interpreted as tags and thus not shown in the composer. The unintentional tag may lead to browser problems because the browser sees it as an opening tag without a closing one.

### Steps to reproduce
Reply to any message. The composer will open and from field at the reply will only show the name and not the e-mail address. Inspecting with the developer tools of the browser shows that the e-mail address is interpreted as a tag. The same behavior can be observed for the subject for the reply (in the quoted reply block)


dhx (approved):
LGTM

ChristophAbenthungCibex:
@hbrunn can you please review this?

ChristophAbenthungCibex:
@hbrunn can you please have a look at this? Thank you.

hbrunn (approved):
please note the system only works if you also review PRs by fellow contributors

ChristophAbenthungCibex:
is it still relevant/needed?

dhx:
@ChristophAbenthungCibex we're using this patch for quite some time now to fix this issue, @grindtildeath you've worked on this module, what do you think?

grindtildeath (commented):
@dhx Looks legit but I don't have maintainer rights on this repo and cannot approve a closed PR.

Must ping @OCA/social-maintainers

PR #1437, opened 2024-08-28
[ADD] REPLACE OR REMOVE HREF AND COMPANY NAME

elhamdaoui:
ADD this param config if we want replace "Odoo" by our company name and "href :..odoo.com" by our company website url
**KEY : mail_debrand.action
VALUE : replace**

### DIFF
--- a/mail_debrand/models/mail_mail.py
+++ b/mail_debrand/models/mail_mail.py
@@ -9,7 +9,12 @@ class MailMail(models.AbstractModel):
     _inherit = "mail.mail"
 
     def _prepare_outgoing_body(self):
+        replace = self.env['ir.config_parameter'].sudo().get_param('mail_debrand.action', 'remove') == 'replace'
         body_html = super()._prepare_outgoing_body()
+        if replace:
+            return self.env["mail.render.mixin"].replace_href_odoo(
+                body_html or "", to_keep=self.body
+            )
         return self.env["mail.render.mixin"].remove_href_odoo(
             body_html or "", to_keep=self.body
         )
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I think that should be another module `mail_rebrand` instead of `mail_debrand`.

elhamdaoui:
> I think that should be another module `mail_rebrand` instead of `mail_debrand`.

These changes do not affect the current behavior. The module can serve both roles: debranding and rebranding. However, if you want to replace the terms, you can manually add the config parameter mail_debrand.action=replace. 

If we add a new module alongside **mail_rebrand**, both modules must not be installed at the same time to avoid conflicts. Otherwise, we would need to override the method, which could add complexity. A better approach would be to use the same module and differentiate between debranding and rebranding using a configuration parameter. Instead of mail_debrand.action=replace, we could use mail_debrand.action=debrand or mail_debrand.action=rebrand to clearly specify the desired action.

pedrobaeza:
I was thinking in doing `mail_rebrand` depending of `mail_debrand`. Why separate both? For not adding maintenance burden to the maintainers of this module if they don't want this feature.

Another thing: the commit message must follow the commit message guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message (and please about all in upper).

elhamdaoui:
> I was thinking in doing `mail_rebrand` depending of `mail_debrand`. Why separate both? For not adding maintenance burden to the maintainers of this module if they don't want this feature.
> 
> Another thing: the commit message must follow the commit message guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message (and please about all in upper).

We cannot use the current version of the mail_debrand module with the mail_rebrand module because mail_debrand completely removes "< a >" with href="*.odoo.com" links. To make them compatible, **mail_debrand** needs to be updated so that it does not remove these links when **mail_rebrand** is installed. This modification will enable both modules to work together as intended.

PR #1438, opened 2024-09-11
[ADD] mail_drop_target_mailing_trace

ajaniszewska-dev:
Provide access rights for the model mass_mailing.model_mailing_trace
so that there is no error when user tries to drop email.

This is needed due to odoo commit:
https://github.com/odoo/odoo/commit/95743d51df09ff08bfa30096740d4986cb01539b


mail_drop_target/__manifest__.py
```diff
@@ -8,9 +8,13 @@
     "category": "Discuss",
     "website": "https://github.com/OCA/social",
     "summary": "Attach emails to Odoo by dragging them from your desktop",
-    "depends": ["mail"],
+    "depends": ["mail", "mass_mailing"],
```
grindtildeath: We don't want to force those using `mail_drop_target` to install `mass_mailing`.

I the dependency is needed because what you fix is only happening when `mass_mailing` is installed, then you need to submit a new "glue" module.

ajaniszewska-dev: @grindtildeath  thanks for review, replaced with new module

grindtildeath (commented):
IMO would make sense to reference Odoo commit https://github.com/odoo/odoo/commit/95743d51df09ff08bfa30096740d4986cb01539b here.

mail_drop_target_mailing_trace/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "category": "Marketing",
+    "license": "AGPL-3",
+    "depends": ["mail_drop_target", "mass_mailing"],
+    "data": [
+        "security/groups.xml",
+        "security/ir.model.access.csv",
+    ],
+    "installable": True,
+    "application": True,
+    "auto_install": False,
```
grindtildeath:
```suggestion
    "auto_install": True,
```

ajaniszewska-dev:
> IMO would make sense to reference Odoo commit [odoo/odoo@95743d5](https://github.com/odoo/odoo/commit/95743d51df09ff08bfa30096740d4986cb01539b) here.

@grindtildeath  updated in commit msg/PR description, plus commit mentioned already in module's readme.

grindtildeath (approved):
LGTM, but there's an error on travis.

grindtildeath:
@OCA/social-maintainers Can we please reopen to get this merged? :pray:

mail_drop_target_mailing_trace/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "website": "https://github.com/OCA/social",
+    "category": "Marketing",
+    "license": "AGPL-3",
+    "depends": ["mail_drop_target", "mass_mailing"],
+    "data": [
+        "security/groups.xml",
+        "security/ir.model.access.csv",
+    ],
+    "installable": True,
+    "application": True,
```
hbrunn, line 18:
```suggestion
```

hbrunn (changes requested):
please write more about what problem this solves.

Do users really need full access to the mailing.trace model for drag&drop to work?

PR #1444, opened 2024-09-12
[16.0][FIX] mail_gateway: Pass company in order to ensure that it is executed properly

etobella:


mail_gateway/models/mail_gateway.py
```diff
@@ -113,7 +113,7 @@ def create(self, mvals):
     @tools.ormcache()
     def _get_gateway_map(self, state="integrated", gateway_type=False):
         result = {}
-        for record in self.search(
+        for record in self.sudo().search(
```
pedrobaeza, line 116: Shouldn't you limit here by company in this search?

etobella: No, because this map is intended to be used on the controller entry for the gateway, and the controller does not know the company until we check the gateway

pedrobaeza (approved):
Ok, let's fast-track it then:

/ocabot merge patch

PR #1449, opened 2024-09-30
[16.0][IMP] mail_activity_plan: Remove warning by several names with the same label

victoralmau:
Remove warning by several names with the same label

`WARNING devel odoo.addons.base.models.ir_model: Two fields (model_id, model) of mail.activity.plan() have the same label: Model. [Modules: mail_activity_plan and mail_activity_plan]`

@Tecnativa TT51051


pedrobaeza (approved):
Straight-forward one, so fast-tracking it:

/ocabot merge patch

PR #1455, opened 2024-10-02
[15.0][ADD] mail_forward: new module to forward messages from the chatter

carlos-lopez-tecnativa:
TT51020
@Tecnativa @pedrobaeza @chienandalu @victoralmau could you please review this


chienandalu (commented):
Should we set a FW: prefix in de subject?

image

carlos-lopez-tecnativa:
> Should we set a FW: prefix in de subject?
> 

I added the prefix similar to Gmail. Thanks for the observation.

image

chienandalu (approved):

pedrobaeza (commented):
Tested on runboat and working. Is there any other FA icon that represents better the forwarding (double arrow?)

chienandalu:
> Tested on runboat and working. Is there any other FA icon that represents better the forwarding (double arrow?)

It can be done with `fa-stack`. An example: https://github.com/OCA/social/blob/92640d7d282f251576e76b00e8b77d3e3cf8e598/mail_tracking/static/src/components/discuss/discuss.xml#L78-L100

carlos-lopez-tecnativa:
> > Tested on runboat and working. Is there any other FA icon that represents better the forwarding (double arrow?)
> 
> It can be done with `fa-stack`. An example:
> 

With your example, it looks like this
image

With `fa-forward` it looks like this. What do you think?

image

chienandalu:
I meant stacking your original arrows. Altough the fa-forward could fit

pedrobaeza:
I think stacking the arrow seems better

carlos-lopez-tecnativa:
I updated the icon, and it looks like this. What do you think?
image

pedrobaeza (approved):
/ocabot merge nobump

PR #1460, opened 2024-10-09
[17.0][ADD]mute_notification_user_autosubscribe: Do not send notification to users autosubscribed through user_id field

manuelregidor:
@HaraldPanten @ValentinVinagre 

@luis-ron Could you review, please?

T-6499


pedrobaeza:
Please avoid plurals in module names, and more in the middle of the name. That's food for constant typos.

manuelregidor:
@pedrobaeza done

HaraldPanten:
@manuelregidor Could you adapt the commit and the PR names, considering the correct module name, please? ~~mute_notifications_user_autosubscribe~~ --> mute_notification_user_autosubscribe

manuelregidor:
@HaraldPanten done

HaraldPanten:
Let's wait for @luis-ron 's functional review, then.

luis-ron (approved):
Functional review: LGTM 👍🏻

manuelregidor:
@Tisho99 Could you review, please?

Tisho99 (approved):
LGTM. Technical review

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten:
@pedrobaeza @etobella your point of view here would be nice.

Could you give us a hand?

THX

mute_notification_user_autosubscribe/models/user_autosubscribe_mute.py
```diff
@@ -0,0 +1,52 @@
...
+    _sql_constraints = [
+        (
+            "unique_model_id",
+            "UNIQUE(model_id)",
+            _("Model must be unique in User Autosubscribe Mute instances."),
+        )
+    ]
+
+    @api.model
+    def is_mute_user(self, model, user):
```
pedrobaeza: Why doing this here instead of `res.users` to take advantage of `self` parameter. Also, the method name is a bit awkward. Call it `_user_is_muted`, although if it goes to `res.users`, then it should just be called `_is_muted`. And I have put it as private, as I don't think that method should be publicly exposed.

mute_notification_user_autosubscribe/models/mail_thread.py
```diff
@@ -0,0 +1,41 @@
...
+                subtype = self.env.ref("mute_notification_user_autosubscribe.muted")
+                partner_id = user.partner_id.id
+                final_vals = [p for p in vals if p[0] != partner_id]
+                final_vals += [
+                    (p[0], [subtype.id], p[2]) for p in vals if p[0] == partner_id
+                ]
+        return final_vals
+
+    def _message_auto_subscribe_followers(self, updated_values, default_subtype_ids):
+        vals = super()._message_auto_subscribe_followers(
```
pedrobaeza, line 28: Can you avoid to write things here? If you define a rule, then the rule shouldn't write other support records, but directly apply the rule when sending messages. If not, you may have both data desynchronized.

manuelregidor: I think I've found an alternative way but I need to test it a little bit more. Thank you.

manuelregidor:
@pedrobaeza Changes applied

pedrobaeza (commented):
Well, this is replacement one table (the previous one) by another (the followers one). The question is why can't you determine dynamically in the moment of sending the notification if it should be sent or not.

manuelregidor:
@pedrobaeza We decided to use subtypes to avoid notificactions, so the muting subtype is applied only once, when the user in field user_id is autosubscribed, and after that, notifications will not be sent to this user unless subtypes in the document are changed, with no need of further checks. By doing it this way:
- It is easy to keep a user subscribed to the document while not receiving notifications.
- In case the user needs to receive notifications, the subtypes can be easily changed.
- We can check which contacts are being sent notifications and which are not in each document.

I hope that answers your question.

pedrobaeza:
As said, I don't think this is the way to go, as if you change any mute rule, all the existing documents won't have such configuration, but if you want to go this way, put it on the ROADMAP and I won't block it.

HaraldPanten:
> As said, I don't think this is the way to go, as if you change any mute rule, all the existing documents won't have such configuration, but if you want to go this way, put it on the ROADMAP and I won't block it.

@pedrobaeza 

The use case is:

- We have a Sales team that prepare lots of quotations and sales orders (after confirmation).
- Once these sales orders are processed (materials are delivered, etc.) we create the corresponding invoices.
- When the Accounting team creates the invoice, the Salesperson assigned to the invoice musn't receive notifications when the invoices are sent by email.
- We need to keep the Salesperson as a follower, but they should not receive notifications coming from the invoices, as it is an Accounting issue, not Sales.

Do you know what I mean?

I'm not sure about what you mean. Could you give us more details and which would be a better approach for this situation?

THX

pedrobaeza:
As said, if you change any mute rule, all the existing documents won't have such configuration, but it's just a corner case. You can continue putting that case in the ROADMAP.

manuelregidor:
@pedrobaeza Case included in ROADMAP.

pedrobaeza:
OK, you can go ahead then.

HaraldPanten (approved):

HaraldPanten:
@pedrobaeza Could you merge it, please? I'm not PSC of this repository.

THX!

PR #1461, opened 2024-10-09
[14.0][ADD] mail_chatter_company_tracking

adasatorres:
This add-on includes the company name in the email chatter tracking and filters tracking based on allowed companies.


Abimael1321:
Functionally reviewed. LGTM

Abimael1321 (approved):

Yoana-RB:
LGTM!!

Yoana-RB (approved):

mail_chatter_company_tracking/models/mail_message.py
```diff
@@ -0,0 +1,31 @@
+from odoo import models
+
+
+class Message(models.Model):
+    _inherit = "mail.message"
+
+    def _message_format(self, fnames):
+        vals_list = super()._message_format(fnames)
+        allowed_companies = self.env.context.get("allowed_company_ids", [])
```
christian-ramos-tecnativa: I don't think filtering the msg is a good idea. Mainly because the next case:
I have the next product fields as trackeable:
- Barcode (No company dependant)
- Cost (Company dependant)

Steps:
1. Modify product barcode and cost on company1
2. The msg appear properly
3. image
4. If i change to company2 the msg is not visible, but the barcode is shared, so I should be able to see the change

Also this display doesnt seems right as the barcode is not company dependant:
image

I think you could or just display the company where the change was made at field level (when is company dependant) or move the filtering at line level instead of the hole msg.

Let me know what you think about it.

christian-ramos-tecnativa (changes requested):

mail_chatter_company_tracking/models/mail_message.py
adasatorres: @Christian-RB Hi, I also don’t really think it should be filtered. I’m going to try to make only the company name appear in the company_dependent fields.

mail_chatter_company_tracking/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "summary": """
+       This add-on includes the company name in the email chatter
+       tracking and filters tracking based on allowed companies.
+       """,
+    "author": "Binhex, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/social",
+    "license": "AGPL-3",
+    "category": "social",
+    "version": "14.0.1.0.0",
+    "depends": ["mail", "company_dependent_attribute"],
```
christian-ramos-tecnativa: I understand that this can be useful for understanding how the module works, but technically, it's not necessary. A good approach might be to include it in the README as a helpful optional feature that users can choose to install.
```suggestion
    "depends": ["mail"],
```

mail_chatter_company_tracking/models/mail_message.py
```diff
@@ -0,0 +1,29 @@
...
+                        "id": tracking.id,
+                        "changed_field": tracking.field_desc,
+                        "old_value": tracking.get_old_display_value()[0],
+                        "new_value": tracking.get_new_display_value()[0],
+                        "field_type": tracking.field_type,
+                        "company_name": (
+                            tracking.company_id.name
+                            if tracking.field.company_dependent
+                            else False
+                        ),
```
christian-ramos-tecnativa: I suggest this instead to avoid the dependence:
```suggestion
                        "company_name": (
                            tracking.company_id.name
                            if self.env[tracking.field.model]
                            ._fields[tracking.field.name]
                            .company_dependent
                            else False
                        ),
```

christian-ramos-tecnativa (approved):
I think it looks great that way:
image
Please take a look at my comments regarding the dependencies. I’d like to know your thoughts on this. It’s true that by adding the module, we avoid duplicating the same logic, but without it, we make the module more independent. It's just a point and im not sure wich one should be the best approach. So LGTM

mail_chatter_company_tracking/models/mail_message.py
adasatorres: @Christian-RB Yes, the dependency can be removed, and the line to obtain the company_dependent can be added."

mail_chatter_company_tracking/models/mail_message.py
```diff
@@ -0,0 +1,31 @@
...
+                        "company_name": (
+                            tracking.company_id.name
+                            if self.env[tracking.field.model]
+                            ._fields[tracking.field.name]
+                            .company_dependent
+                            else False
+                        ),
+                    }
+                )
+            vals_list[pos]["tracking_value_ids"] = tracking_value_ids
```
hbrunn: why do you recreate the dictionaries instead of just adding your key? This will destroy the functionality of any other modules that happen to modify other aspects of the result of this function

mail_chatter_company_tracking/static/src/components/message/message.xml
```diff
@@ -0,0 +1,17 @@
...
+  <t t-name="Message" t-inherit="mail.Message" t-inherit-mode="extension" owl="1">
+    <xpath expr="//div[hasclass('o_Message_trackingValueFieldName')]" position="before">
+      <t t-log="value" />
+      <div
+                t-if="value.company_name"
+                class="o_Message_trackingValueFieldCompanyName o_Message_trackingValueItem"
+            >
+        <span style="font-weight:bold;">
+          (
+          <span t-esc="value.company_name" />)
```
hbrunn:
```suggestion
          <!-- prettier-ignore-start -->
          (<span t-esc="value.company_name" />)
          <!-- prettier-ignore-end -->
```

to get rid of the extra space

hbrunn (changes requested):

mail_chatter_company_tracking/models/mail_message.py
adasatorres: Hello, thank you for the correction, I agree; it's only necessary to add the info to the existing dictionary.

hbrunn (approved):

PR #1459, opened 2024-10-08
[17.0][REF] mail_activity_team: use mail.activity.schedule wizard

vvrossem:
# :construction: WORK IN PROGRESS :construction:

PR #1462, opened 2024-10-09
[17.0][IMP] mail_activity_team: add team fields on mail.activity.schedule

imlopes:
When creating an activity, it is not possible to assign it to a team.
A team can be selected on an activity but only through editing it.


imlopes:
Closing to use https://github.com/OCA/social/pull/1459

PR #1463, opened 2024-10-10
[18.0][MIG] mail_notification_with_history: Migration to 18.0

trisdoan:
### Note
- template `message_notification_email` was moved to `data/mail_templates.xml` in https://github.com/odoo/odoo/commit/78c68f4b885c204a1453f80670d46d0ecc53d02c

- Later splitted for better isolate notification layouts in https://github.com/odoo/odoo/commit/a480f1f3cd770a60fca35901d547ed51e56300b7

### This change
- Since t-raw was flagged as deprecated, replace it with t-out
 
### Result
mail_notification_migration


mail_notification_with_history/data/mail_data.xml
```diff
@@ -0,0 +1,31 @@
...
+      <t
+                t-set="message_history"
+                t-value="message._get_notification_message_history()"
+            />
+      <t t-if="message_history">
+        <p class="message_history_title">Discussion history in Odoo:</p>
+        <ul>
+          <t t-foreach="message_history" t-as="msg">
+            <li class="message_history_item">
+            <p><t t-esc="msg.author_id.name or 'unknown'" /> commented on <span
```
henrybackman: `t-esc` is soft deprecated, could we switch to `t-out` here?

henrybackman (approved):
LG with one minor comment

mail_notification_with_history/data/mail_data.xml
simahawk: to be addressed

sebalix (approved):

PR #1464, opened 2024-10-10
[15.0][FIX] mail_show_follower: Use company from the document to render messages correctly

carlos-lopez-tecnativa:
In a multi-company environment, if one company uses the `show_followers_partner_format` by `partner_name` and another company uses `partner_email`, a user with permissions in both companies may experience issues. If a document belongs to Company A, but the current company is set to Company B, the message format is taken from the current company (Company B) instead of the document’s company (Company A). This fix ensures that the message format is rendered based on the document's company.

TT51022 @Tecnativa
@pedrobaeza @chienandalu could you please review it, thanks

### DIFF
--- a/mail_show_follower/models/mail_mail.py
+++ b/mail_show_follower/models/mail_mail.py
@@ -59,97 +59,29 @@ class MailMail(models.Model):
         return full_text
 
     def _send(self, auto_commit=False, raise_exception=False, smtp_session=None):
-        group_portal = self.env.ref("base.group_portal")
-        for mail_id in self.ids:
-            mail = self.browse(mail_id)
-            message_recipients = self.search(
-                [
-                    ("message_id", "=", mail.message_id),
-                ]
-            ).mapped("recipient_ids")
-            # if the email has a model, id and it belongs to the portal group
-            if mail.model and mail.res_id and group_portal:
-                obj = self.env[mail.model].browse(mail.res_id)
-                # those partners are obtained, who do not have a user and
-                # if they do it must be a portal, we exclude internal
-                # users of the system.
-                if hasattr(obj, "message_follower_ids"):
-                    partners_obj = (
-                        obj.message_follower_ids.mapped("partner_id")
-                        | message_recipients
-                    )
-                    # internal partners
-                    user_partner_ids = (
-                        self.env["res.users"]
-                        .search(
-                            [
-                                ("active", "in", (True, False)),
-                                ("show_in_cc", "=", False),
-                            ]
-                        )
-                        .filtered(lambda x: group_portal not in x.groups_id)
-                        .mapped("partner_id")
-                        .ids
-                    )
-                    partners_len = len(
-                        partners_obj.filtered(
-                            lambda x: x.id not in user_partner_ids
-                            and (
-                                not x.user_ids
-                                or group_portal in x.user_ids.mapped("groups_id")
-                            )
-                        )
-                    )
-                    if partners_len > 1:
-                        # get partners
-                        cc_internal = True
-                        # else get company in object
-                        if hasattr(obj, "company_id") and obj.company_id:
-                            cc_internal = obj.company_id.show_internal_users_cc
-                        # get company in user
-                        elif mail.env and mail.env.user and mail.env.user.company_id:
-                            cc_internal = (
-                                self.env.user.company_id.show_internal_users_cc
-                            )
-                        if cc_internal:
-                            partners = partners_obj.filtered(
-                                lambda x: x.id not in user_partner_ids
-                                and (
-                                    not x.user_ids
-                                    or any(x.mapped("user_ids.show_in_cc"))
-                                )
-                            )
-                        else:
-                            partners = partners_obj.filtered(
-                                lambda x: x.id not in user_partner_ids
-                                and (
-                                    not x.user_ids
-                                    or group_portal in x.user_ids.mapped("groups_id")
-                                )
-                            )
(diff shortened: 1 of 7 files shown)


mail_show_follower/models/mail_mail.py
```diff
@@ -104,8 +104,10 @@ def _send(self, auto_commit=False, raise_exception=False, smtp_session=None):
                         # get partners
                         cc_internal = True
                         # else get company in object
+                        company = self.env.company
```
pedrobaeza: I think you can refactor this code to get the put the logic to get the company on one side, and then access to the properties then, so it's clearer. If not, all this `if` are hard to read and prone to errors.

carlos-lopez-tecnativa:
Backport of https://github.com/OCA/social/pull/1391

mail_show_follower/models/mail_mail.py
carlos-lopez-tecnativa: @pedrobaeza I backported the changes from version 16 because it includes a refactor that makes the code more readable. I squashed two commits into a single one and added a commit to fix the multi-company issue. Please let me know if you have any questions.

pedrobaeza (approved):

chienandalu (approved):

mail_show_follower/models/mail_mail.py
```diff
@@ -59,97 +59,31 @@ def remove_p(markup_txt):
...
-                        mail.body_html = final_cc + mail.body_html
+        group_user = self.env.ref("base.group_user")
+
+        for mail in self:
+            if not (mail.model and mail.res_id and group_user):
+                continue
+            # recipients from any Notification Type (i.e. email, inbox, etc.)
+            recipients = mail.notification_ids.res_partner_id
+            record = self.env[mail.model].browse(mail.res_id)
+            company = getattr(record, "company_id", self.env.company)
```
yajo, line 70: suggestion: This is not the recommended way to get a field that you don't know if exists, because that attribute could perfectly be a method (imagine `def company_id():`) and break your logic. Hardly probable, but possible.


The rule of thumb way to make sure it's really a field:

```suggestion
            company = record["company_id"]
```

Of course there can be a `KeyError` which you can circumvent with a try-except block, or a `if "company_id" in record` guard.

mail_show_follower/models/mail_mail.py
```diff
@@ -59,97 +59,31 @@ def remove_p(markup_txt):
...
+            if len(show_in_cc_recipients) <= 1:
+                continue
+
+            langs = (
+                mail.notification_ids.res_partner_id.mapped("lang")
+                or mail.author_id.lang
+                or company.partner_id.lang
+            )
+            final_cc = (
+                mail.with_context(lang=langs[0])
```
yajo: issue: `langs` is of type `list[str] | str | bool` as per the code above. This is only supporting the `list[str]` case.

suggestion: I think the bug you guys discovered deserves at least a new test.

mail_show_follower/models/mail_mail.py
```diff
@@ -59,97 +59,31 @@ def remove_p(markup_txt):
...
+        group_user = self.env.ref("base.group_user")
+
+        for mail in self:
+            if not (mail.model and mail.res_id and group_user):
+                continue
+            # recipients from any Notification Type (i.e. email, inbox, etc.)
+            recipients = mail.notification_ids.res_partner_id
+            record = self.env[mail.model].browse(mail.res_id)
+            company = getattr(record, "company_id", self.env.company)
+            show_internal_users = company and company.show_internal_users_cc
```
yajo, line 71: suggestion: This guard is unnecessary:

```suggestion
            show_internal_users = company.show_internal_users_cc
```

mail_show_follower/models/mail_mail.py
```diff
@@ -59,97 +59,31 @@ def remove_p(markup_txt):
...
-                            )
-                        )
-                        # get show follower text
-                        final_cc = mail.with_context(
-                            lang=langs and langs[0]
-                        )._build_cc_text(partners)
-                        # it is saved in the body_html field so that it does
-                        # not appear in the odoo log
-                        mail.body_html = final_cc + mail.body_html
+        group_user = self.env.ref("base.group_user")
```
yajo, line 62: issue: although the PR description explains a clear and concise problem, the code I see here is a medium refactor of a very important (and yes, spaghethic) piece of code.

For example, you are going away from testing portal group membership into testing backend user membership.

So, the PR goes beyond scope and gets harder to review. Could you split the fix and the refactor in 2 different PRs? Or could you explain a little bit more what's this refactor doing, so it's easier to review?

yajo (changes requested):
Good job, thanks! Some concerns below.

pedrobaeza:
@yajo this is a backport of #1391, with only an extra patch in a separated commit, so all your comments should go to the 16.0 PR when it was done, and it was you who approve it. We are not going to invest more time in that part, as we will finally get to the newest version with the refactoring already done.

The only patch from us applied right now is the company one, but we have changed also the lang part, which indeed can have some side effects (but this is not new). We will fw-port it to 16 if this PR is accepted.

Please accept it as it is to get both versions with the good refactoring, or we will let it die as a PR injected where needed.

yajo (approved):
Thanks! I didn't realize this was a backport and there were 2 commits. Also thanks for that extra fix. Happy to help! ❤️ 

/ocabot merge minor

pedrobaeza:
@carlos-lopez-tecnativa please fw-port the second patch to 16

PR #1471, opened 2024-10-15
[18.0][MIG] mail_inline_css: Migration to 18.0

nguyenminhchien:
- Drop the dependency to `email_template_qweb`:
  - The object `mail.template` from core module `mail` has already supported qweb content via the field `body_html` (https://github.com/odoo/odoo/blob/18.0/addons/mail/models/mail_template.py#L63).
  - So, i adjust this module to drop the dependency to `email_template_qweb`.


TDu (approved):

mail_inline_css/models/mail_template.py
```diff
@@ -0,0 +1,33 @@
...
+
+from odoo import models
+
+try:
+    from premailer import Premailer
+except (OSError, ImportError) as err:  # pragma: no cover
+    import logging
+
+    _logger = logging.getLogger(__name__)
+    _logger.debug(err)
```
ivantodorovich:
```suggestion
from premailer import Premailer
```

since it's already listed as external_dependencies, this try block seems unnecessary

ivantodorovich (approved):

mail_inline_css/models/mail_template.py
nguyenminhchien: Hi, Thanks, i just updated it

ivantodorovich: Thanks!

pedrobaeza (changes requested):

nguyenminhchien:
> This should go to OCA/mail instead.

Hi @pedrobaeza , i opened https://github.com/OCA/mail/pull/6 to move this module to oca/mail.
I closed this PR as a result.

cc @ivantodorovich , @TDu

PR #1476, opened 2024-10-17
[18.0][MIG] mail_tracking: Migration to 18.0

trisdoan:
### Note
- missing commits from 14.0 are ported in https://github.com/OCA/social/pull/1475

#### Backend
- `access_rights_uid` was removed from `_search()` changed in https://github.com/odoo/odoo/pull/171374/commits/21d023e6bc00793ac3b016cfdcafd05c9a81d95d
- `_prepare_outgoing_list` got extra param `mail_server` in https://github.com/odoo/odoo/pull/96008/commits/97ce0844ca0c3d531a7a2e2e47a867d89835b24e
- `_message_format` was converted to use store and `message_format_extras` was removed in https://github.com/odoo/odoo/commit/52df49eabb6e44ea6e0ed30b61e5a555af2e606a. As consequences, `_extras_to_store` is introduced
- `tools.ustr` is considered to be inefficient and flagged as deprecated in https://github.com/odoo/odoo/pull/173811/commits/64a5724e88de6532cf2a04e9243d16b079319ffd. Hence, is replaced by `exception_to_unicode` for the same effect.

#### Frontend
- `mail.messaging` service was removed in https://github.com/odoo/odoo/commit/683258b28107f44fba363e3b4e69f181cac3f80e
- `initMessagingCallback` is removed in https://github.com/odoo/odoo/commit/4f8087f03ad904b99f8f3580557a3ea744a97b76. So new entry point is created for store `onStarted` instead.
- mail.thread service was removed in  https://github.com/odoo/odoo/commit/65e5a7943de094fe538f2b978cb58c028a5694a1

### This changes
- Store helper was introduced in https://github.com/odoo/odoo/pull/150625/commits/c032ccfbaef91da0ee244f0e44e0e4e9d8d8d3fb. As consequences, failed data is now passed through Store.
   - As a side effect, it is also easier to do the search filter_failed(thread) with Store service
  
- As Odoo uses less Jquery, the module runs fine with vanilla. This impacts `mail_tracking/static/src/components/message_tracking/message_tracking.esm.js`

- Took this chance to
   - replace % format with f-string
   - Improve permission constraint
       - Context: This [commit](https://github.com/OCA/social/pull/966/commits/b4ab1600fe0b464937ff3dc06ff7cbf3a29afe3a) porting from 12.0
       - Approach is summarized as below:
           1. When do the searching, 
             - First, use the query to get result
             - Browse those id with the query + Remove any forbidden records
             - Return the accessible records in form of query, to continue the process
           2. When user read the records
             - The low-level method `_check_access` will discard forbidden records + raise exception
       - Added security check test

    - Resolve FIXME, which was added in https://github.com/OCA/social/pull/624/commits/75604439061755da9df64f658a9951a85e196185, then was resolved in https://github.com/odoo/odoo/pull/49999/commits/7fe02c93e410f8e9d11c89f77c651ca4b93810ff

### Result
discuss_sidebar

failed_search_panel

failed_message_review


trisdoan:
~~DRAFT: working on failed test~~

~~> 2024-10-17 05:08:43,628 246 WARNING odoo odoo.addons.mail_tracking.controllers.main: MailTracking email '11' not found~~

mail_tracking/models/mail_message.py
```diff
@@ -29,7 +30,7 @@ class MailMessage(models.Model):
     )
     is_failed_message = fields.Boolean(
         compute="_compute_is_failed_message",
-        search="_search_is_failed_message",
+        store=True,
```
chienandalu: Do we really need to make it stored?

chienandalu (commented):
Huge work @trisdoan !

I should try with a real mail configuration. For the moment:

Installing the module in runboat doesn't load for me the demo failed messages for retry (I saw in them in your screenshot thou...)

image

mail_tracking/models/mail_message.py
trisdoan: Hi, this flag is used to populate the controller: `/mail/failed/messages`

trisdoan: I just tested on runbot and still see demo data, maybe another try?

pedrobaeza: Please don't change how it was, no stored and with search method. Or does any reason for changing it?

trisdoan: Hi @pedrobaeza, yes I agree there is no need to store it in db. But look like the search method is not working. 

I am working on it, will update when it's done

trisdoan:
~~DRAFT: inspect `_search_is_failed_message`~~

pedrobaeza:
What do you think about https://github.com/OCA/social/issues/1458#issuecomment-2427476407 ?

mail_tracking/models/mail_message.py
trisdoan: Hi @pedrobaeza @chienandalu , I just added [fixup! [MIG] mail_tracking: Migration to 18.0](https://github.com/OCA/social/pull/1476/commits/a6bce24c379bb8c8b3be606470b24decfc0fd427)

trisdoan: Hi @pedrobaeza @chienandalu , I just added [fixup! [MIG] mail_tracking: Migration to 18.0](https://github.com/OCA/social/pull/1476/commits/a6bce24c379bb8c8b3be606470b24decfc0fd427) (will squash after we all agree)

- As I inspected, _search_is_failed_message was not working. Without this commit, the condition in final query is always False => return no result
- I added a test to make sure it's covered

ivantodorovich (approved):
Code review. LG. Thanks!

chienandalu (commented):
Looks very good overall. In the testing side, I'm missing reactivity for the failed message actions. An example, where the failed message card should go away (it only does when we reload the window):

Peek 25-10-2024 10-49


Also, can you squash your migration commits?

trisdoan:
> Looks very good overall. In the testing side, I'm missing reactivity for the failed message actions. An example, where the failed message card should go away (it only does when we reload the window):

Hi @chienandalu , yes you're right. It causes confusion indeed.

My approach is to reload the page: `browser.location.reload();`
What do you think?

Look like it gets the same issue in 17.0

pedrobaeza:
Now that OCA/mail has been created (https://github.com/OCA/repo-maintainer-conf/pull/52), can you please move this PR to that repo?

trisdoan:
Superseded by https://github.com/OCA/mail/pull/1

PR #1479, opened 2024-10-17
[18.0][MIG] mail_debrand: Migration to 18.0 

jdebetaz:


pedrobaeza:
What do you think about https://github.com/OCA/social/issues/1458#issuecomment-2427476407 ?

pedrobaeza (changes requested):

PR #1481, opened 2024-10-18
[18.0][MIG] mail_thread_create_nolog: Migration to 18.0

chaule97:
- in odoo 14, [message_fetch](https://github.com/odoo/odoo/blob/f2765d2cab5671a010404c36842bf1b4c4d6350b/addons/mail/models/mail_message.py#L1029) return list of dictionaries of `mail.message`, while in odoo 18 [_message_fetch](https://github.com/odoo/odoo/blob/e4a4806e5c58c599815204a73c78a228ee230613/addons/mail/models/mail_message.py#L1083) returns  `mail.message` model, and the [store](https://github.com/odoo/odoo/blob/e4a4806e5c58c599815204a73c78a228ee230613/addons/mail/controllers/thread.py#L39) will check if messages is still of the `mail.message` model. This makes it difficult for me to override it.
- My ideas is to write method to create message, and then override method `mail_thread_messages` in controller thread to get fake message
- I write `"type": "partner"`  because after [this commit](https://github.com/odoo/odoo/commit/b1a969271acad9210e0fbf75356baa8f4edc4a39#diff-a40efe06b45cbfadbcc29ff716475330e84194f4751a8b64956d417f9d8728bf), it returns default avatar unless type is provided
 image


mail_thread_create_nolog/models/mail_message.py
```diff
@@ -7,44 +7,58 @@
...
                     # so we generate one that doesn't really exist
                     # but based on the current thread so it'll always be
                     # unique (used by the thread_cache client model).
                     "id": self._get_create_message_id(domain),
                     # Author is not preserved above when creating the fake-record
-                    "author_id": (author.id, author.name),
+                    "author": {
+                        "id": author.id,
+                        "name": author.name,
+                        "type": "partner",
```
sebalix, line 55: A comment on this `type: partner` (about the avatar) is welcome :)

mail_thread_create_nolog/tests/test_create_nolog.py
```diff
@@ -1,10 +1,10 @@
 # Copyright 2023 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
-from odoo.tests.common import SavepointCase
+from odoo.tests.common import TransactionCase
 
 
-class TestMailThreadCreateNoLog(SavepointCase):
+class TestMailThreadCreateNoLog(TransactionCase):
```
sebalix: As the module now overrides the std `mail` controller to inject the fake message, should we now test a call on this one instead? Leveraging `odoo.tests.comm.HttpCase`.

What do you think?

sebalix (approved):

sebalix:
Not sure it'll work even if I'm in the maintainers list of the module:

/ocabot merge nobump

PR #1485, opened 2024-10-30
[MIG] mail_optional_follower_notification: Migration to 18.0

carlosdynapps:


mail_optional_follower_notification/wizard/mail_compose_message_view.xml
```diff
@@ -0,0 +1,28 @@
...
+    <record id="email_compose_message_wizard_form" model="ir.ui.view">
+        <field name="model">mail.compose.message</field>
+        <field name="inherit_id" ref="mail.email_compose_message_wizard_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//widget[@name='mail_composer_recipient_list']"
+                position="before"
+            >
+                <field
+                    name="notify_followers"
```
NachoAlesLopez, line 12: This field is missing it's label. Without it, the checkbox doesn't have any text in the mail composer:
Screenshot 2024-10-31 at 09 46 51

NachoAlesLopez (commented):
Functionally, LGTM. The field in the composer is missing its text, but other than that it, appears to work.

Also, mind taking a look at the last two commits? They are missing the module and version in the commit message. Can you rebase and add the correct information?

NachoAlesLopez (approved):

pedrobaeza (changes requested):

PR #1486, opened 2024-11-04
[16.0][FW][FIX] mail_show_follower: Use company from the document to render messages correctly + proper lang handle

carlos-lopez-tecnativa:
Forward-port of  https://github.com/OCA/social/pull/1464
In a multi-company environment, if one company uses the show_followers_partner_format by partner_name and another company uses partner_email, a user with permissions in both companies may experience issues. If a document belongs to Company A, but the current company is set to Company B, the message format is taken from the current company (Company B) instead of the document’s company (Company A). This fix ensures that the message format is rendered based on the document's company.

It also fixes the lang handling for cases with False values.

TT51022
@Tecnativa @pedrobaeza @yajo @chienandalu

### DIFF
--- a/mail_show_follower/models/mail_mail.py
+++ b/mail_show_follower/models/mail_mail.py
@@ -72,14 +72,16 @@ class MailMail(models.Model):
             show_in_cc_recipients = recipients._filter_shown_in_cc(show_internal_users)
             if len(show_in_cc_recipients) <= 1:
                 continue
-
-            langs = (
-                mail.notification_ids.res_partner_id.mapped("lang")
+            lang = (
+                mail.notification_ids.res_partner_id[:1].lang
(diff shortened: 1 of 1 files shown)


mail_show_follower/static/description/index.html
```diff
@@ -8,11 +8,10 @@
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
+:Id: $Id: html4css1.css 8954 2022-01-20 10:10:25Z milde $
```
pedrobaeza: This shouldn't be changed.

pedrobaeza (approved):

yajo (approved):
Thanks!

/ocabot merge patch

PR #1490, opened 2024-11-07
[16.0][IMP] mail_quoted_reply: Sanitize HTML body before quoting

grindtildeath:
As we are adding HTML from external messages into the mail composer HTML widget, we do not control what is in there and it could break the webclient or making it unresponsive depending on its content.

Sanitizing the body of the quoted message might not solve all the issues, but it at least provides a hook for extra processing.


flotho (approved):
easy code review
LGTM

hbrunn (approved):

PR #1491, opened 2024-11-11
[15.0][FIX] mail_show_follower: More models to exclude

pedrobaeza:
Slides and forums are also places where you don't want to show the recipients.

@Tecnativa TT51474

### DIFF
--- a/mail_show_follower/data/config_parameter_data.xml
+++ b/mail_show_follower/data/config_parameter_data.xml
@@ -2,6 +2,8 @@
 <odoo noupdate="1">
     <record id="models_to_exclude_config_parameter" model="ir.config_parameter">
         <field name="key">show_followers.models_to_exclude</field>
-        <field name="value">blog.blog,blog.post</field>
+        <field
+            name="value"
+        >blog.blog,blog.post,slide.slide,slide.channel,forum.forum,forum.post</field>
     </record>
 </odoo>


chienandalu (approved):

victoralmau (approved):

yostashiro:
Can you please add forum posts as well?

pedrobaeza:
@yostashiro done.

yostashiro (approved):
Thank you!

yajo:
Would it make sense to support an allow-list style? This way, the feature would be added only to allowed models.

I think that just for `project.task` would be enough in our use case (and maybe others have similar requirements).

cc @rafaelbn

pedrobaeza:
I don't think so, as that list is way larger and growing/changing with each new module installed or migration.

chienandalu:
What about a model flag? In practice there aren't that many models that the users use to communicate with customers, and the risks of leaking recipients info will show up with every new model...

pedrobaeza:
It's still the same: when a new module containing models with threads is installed, you have to allow it explicitly. That's not what usually users wanting this feature wants to have it.

I still think the exclude model one is the correct here, and without flags, but static model name list, so you don't need to install the modules in your DB to have that list populated, like now.

yajo:
OK, let's keep it like this. If time proves the contrary, it can be added later anyway.

PR #1492, opened 2024-11-12
[16.0][FIX]mail_quoted_reply: reply remove comments

ChristophAbenthungCibex:
Remove html comments of the reply body. They are normally not needed and caused errors in the wizard for our customer. For example: in case of 2 comments are right after each other the user cannot change the content of the composer. Odoo removes comments from all mail related html (No comments possible at mail template body).


hbrunn:
do we still need this with https://github.com/OCA/social/pull/1490?

ChristophAbenthungCibex:
> do we still need this with #1490?

As far as I can tell it is not needed anymore. Thanks for letting me know.

PR #1495, opened 2024-11-14
[MIG] mail_activity_partner: Migration to 18.0

carlosdynapps:


pedrobaeza:
All the mail modules should go now to OCA/mail

PR #1499, opened 2024-11-14
[16.0][MIG] mail_embed_image module

StephaneMangin:
Superseed https://github.com/OCA/social/pull/1402


mail_embed_image/models/ir_mail_server.py
```diff
@@ -0,0 +1,118 @@
...
+        fileparts = []
+        # Limit results to only internal resources to avoid malicious external
+        # image injections
+        for img in root.xpath(
+            ".//img[starts-with(@src, '%s')]"
+            "| .//img[starts-with(@src, '/web/image')]" % (base_url)
+        ):
+            image_path = img.get("src")
+            try:
+                response = requests.get(image_path, timeout=10)
```
hbrunn, line 126: please add a test case where we have an image with a relative url, I believe this will fail here

mail_embed_image/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,26 @@
...
+            <div id="companies_setting" position="inside">
+                <br />
+                <div
+                    class="o_setting_right_pane"
+                    id="mail_templates_setting"
+                    groups="mail.group_mail_template_editor,base.group_system"
+                >
+                    <span class="o_form_label">Email Preprocessing</span>
+                    <div class="text-muted">
+                        Method used to embed images in HTML emails. CIDs attachment does not work with all email clients. Data SRC is more reliable.
```
hbrunn: that actually depends very much on the clients you target, with old outlooks it's the other way around. I'd propose not to make claim about what works better, and rather suggest to test with whatever are the expected mail clients

hbrunn (changes requested):
tests are fixed in https://github.com/OCA/social/pull/1519

mail_embed_image/views/res_config_settings_views.xml
StephaneMangin: Issue regarding Outlook has been tackled elsewhere. WIP
Need more testing regarding it before proposing.

PR #1507, opened 2024-11-21
[16.0][FIX]mail_activity_done: Fix counters and activity_state search

### LINKED ISSUE #1327: mail_activity_done: wrong counters
When using the mail_activity_done module the counters are not correctly computed

The module overrides res.users::systray_get_activities which is used to gather the information displayed in the systray. 

It calls super().systray_get_activities which will compute the counters for all the activities (including the ones in state 'done', which are archived when this module is installed, but removed in a normal implementation) and then tries to compute the number of done activities and to substract that quantity from the counters computed by super(). 

The problem is that the counters are aggregated by models : one counter for crm.lead, one for sale.orders, and the counter is not a counter of activities, but a counter of records of said model which have late activities, activities for today and activities in the future, so bluntly substracting is wrong. 

## To Reproduce
affected versions 16.0 for sure, but probably all others too. 

1. remove all activities assigned to user1 from the database (deletion, not flagging as done)
2. Create a crm.lead with activities assigned to the same user, scheduled 10 days in the past
3. flag one activity as done
4. log in as the user, and check the activities systray

**Expected behavior**

The counter for late activities is 1

**Actual behavior**

The counter for late activities is 0

jesusVMayor:
This PR includes 2 changes:
- Remove systray_get_activities overrride, its not necessary since https://github.com/odoo/odoo/commit/bd4e8aa3cb6b6ce16acd7f1e4bc99f1ce01d18d8
- Override _search_activity_state adding `and mail_activity.done=False` to WHERE clause in the query to allow properly search by this field.

Fixes: https://github.com/OCA/social/issues/1327 CC: @gurneyalex @navarrorico


jesusVMayor:
Fixed tests.

mail_activity_done/models/mail_activity.py
```diff
@@ -92,3 +96,77 @@ def _read_progress_bar(self, domain, group_by, progress_bar):
         """
         domain = expression.AND([domain, [("activity_ids.done", "=", False)]])
         return super()._read_progress_bar(domain, group_by, progress_bar)
+
+    def _search_activity_state(self, operator, value):
+        # Totally override the method.
```
hbrunn: why can't you just take the list of ids you get from super and filter them for being done? That adds very little overhead and you don't break modularity

hbrunn (changes requested):

mail_activity_done/models/mail_activity.py
hbrunn: ...and the inverse for negative operators, your full override doesn't cover that case

jesusVMayor: _search_activity_state returns a list of record ids, if the search is made in res.partner it returns partner_ids, i need to filter the subquery, i don't see a way to make it without making the full query again. 
And we need to filter with done=False allways, even if the search is negative operator, if somebody searchs "activity state is not overdue" or "activity state is today", we need to filter the undone activities in both cases because you want to see records with a particular activity_state based on undone activities.
Am i missing something?

hbrunn: I think I'm convinced about the negative case, "not overdue" can mean "today or planned (and active)", not must mean "anything that's not overdue". We don't do anything with `activity_state` in the mixin, so currently the addon doesn't make a claim about that.

The other part is more complex than I first thought, but I'd rather do
```python
        execute_org = self._cr.execute

        def execute(query, params=None, log_exceptions=True):
            return execute_org(
                query.replace(
                    'WHERE mail_activity.res_model = %(res_model_table)s', 
                    'WHERE mail_activity.res_model = %(res_model_table)s AND '
                    'mail_activity.done = FALSE'
                ),  
                params=params, log_exceptions=log_exceptions
            )   

        self._cr.execute = execute
        try:
            return super()._search_activity_state(operator, value)
        finally:
            self._cr.execute = execute_org
```

Remember that part of the issue you address here is that the previous version did a blunt replacement of the core code. The above is much more likely to keep working with and taking advantage of bugfixes in core, or other overrides of the function.

jesusVMayor: Oh, thanks!, i would never tought that solution

hbrunn (approved):

jesusVMayor:
I have detected a new bug, the _read_progress_bar inherit was not properly working and the count of activities in the kanban view was wrong. We should modify the subquery where clause

hbrunn (changes requested):
looks like you'll have to update the [test](https://github.com/OCA/social/actions/runs/12034768739/job/33552121424?pr=1507#step:8:111) too - I think it's correct to expect a 0 here

jesusVMayor:
Seems a problem with the tests because the function executes the query directly. Testing in shell, if i make the commit after the _action_done, the result is 0. It also happens in the test_activity_state_search.

I changed the tests to move the second activity to demo file, assigned to another partner, and created already done.

hbrunn (approved):

StefanRijnhart (commented):
Thanks! Sounds like your test issues could perhaps have been solved with a `flush_recordset` on the activity.
Do you want to squash your commits in one or maybe two commits? I already picked what remains valid in 18.0 in https://github.com/OCA/mail/pull/31

jesusVMayor:
Done

StefanRijnhart (changes requested):
If I understand correctly, the tests now depend on the demo data. This is against the current grain of making tests less dependent on demo data. The tests are also more reliable if they are for a newly created user. Otherwise, the tests could fail because of demo activities from unrelated modules, or previous activity on a reused database. Can you bring back the test user?

jesusVMayor:
I made some changes in tests:
- Removed demo data in tests, and use flush_recordset to avoid problems.
- Added tz to new user, odoo base _search_activity_state fails if the user has no tz.
- Create test partner with the test user, it causes some security rules problems if is created without it.

navarrorico (approved):
Functional review LGTM :+1:

StefanRijnhart (approved):
Thanks for the updates!

/ocabot merge minor

PR #1508, opened 2024-11-22
[16.0][MIG] mail_attach_existing_attachment_account from 14 to 16

sbidoul:


sbidoul:
/ocabot migration mail_attach_existing_attachment_account

This remains useful for the Send and Print wizard on invoices.

tarteo (approved):
Thanks! Works, tested it on runboat

image

pedrobaeza:
I wonder if applying a similar technique like https://github.com/OCA/web/pull/2804 in the base module, we can avoid this module at all. cc @CarlosRoca13

CarlosRoca13:
> I wonder if applying a similar technique like [OCA/web#2804](https://github.com/OCA/web/pull/2804) in the base module, we can avoid this module at all. cc @CarlosRoca13

@pedrobaeza I think it wouldn't apply in this case... The loading order is used for the assets. In this case, I believe the issue is that the wizard `account.invoice.send` is built from `mail.compose.message`, but it is done using `_inherits`, which requires adapting it in a certain way. 

Although I am not entirely familiar with the module, perhaps @victoralmau can provide some insight on the matter.

pedrobaeza:
OK, thanks. It was just a random thought without digging deeper, as it's a nuisance to require this extension for accounting.

victoralmau (approved):
Functional review OK.

I don't like the need for this module, but indeed https://github.com/OCA/social/pull/1508#issuecomment-2527256570 seems to be necessary.

sbidoul:
Thanks for the reviews and merge.

PR #1516, opened 2024-12-03
[16.0][IMP]mail_quoted_reply: Separate reply body

ChristophAbenthungCibex:
For long replies the message composer gets extremely laggy and slow. This applies to text written in the composer (body) and for opening the composer. 

Steps to reproduce:
1. Go to any chatter
2. Send a message
3. Reply to that message
4. Keep replying to the newly sent message

After a couple of message the composer will open slowly (couple seconds till it opens) and respond slowly to an input.

This pull request separates the reply body into another field. This field is read-only after opening the composer. It can be changed to be writable. However, depending on how long the reply is the time needed to activate it differs. 
This feature can be en-/disabled with the  system parameter "mail_quoted_reply.separate_reply_body".

### DIFF
--- a/mail_quoted_reply/README.rst
+++ b/mail_quoted_reply/README.rst
@@ -31,6 +31,9 @@ Mail Message Reply
 This addon allow to reply on messages from odoo directly.
 It is useful when replying to a message from a customer.
 
+Additionally, it is possible to separate the reply body from the mail body.
+This separation is only for the composer and the sent mail will contain both parts.
+
 **Table of contents**
 
 .. contents::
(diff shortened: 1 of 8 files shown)


hbrunn (changes requested):
code looks fine, but please describe the feature in the readme.

This is of course only fixing a symptom, the underlying issue here is that the html widget doesn't deal well with big/complex content.

ChristophAbenthungCibex:
> code looks fine, but please describe the feature in the readme.
> 
> This is of course only fixing a symptom, the underlying issue here is that the html widget doesn't deal well with big/complex content.

I updated the readme.

Yea it is only fixing a symptom. But as long as the html widget is that slow we need some sort of workaround/fix. Some replies loaded for over a minute before the composer opened and then the typed input was very delayed.

hbrunn (approved):

ChristophAbenthungCibex:
Had to change the compute for "is_separate_body" because the feature was enabled for new messages too. Not just for replies.

ioans73 (approved):
LGTM

hildickethan (approved):

PR #1312, opened 2024-02-21
[17.0][MIG] mail_composer_cc_bcc: Migration to 17.0

trisdoan:
### 1) Summary of changes in 17.0

mail.template
https://github.com/odoo/odoo/blob/17.0/addons/mail/models/mail_template.py

- `generate_recipients` replaced by `_generate_template_recipients`

mail.mail:
https://github.com/odoo/odoo/blob/17.0/addons/mail/models/mail_mail.py

- new method `_prepare_outgoing_list`: https://github.com/odoo/odoo/commit/b1d678b12ac5aac4086771f9ae88d401d919c9fa
- replaces `_send_prepare_values`
- new method `_personalize_outgoing_body` is also introduced

account.move.send: replaces account_invoice_send
- do not inherit mail.compose.message
- `_action_send_mail` replaced by `_send_mail`
- introduce new method `_get_mail_move_values`

### 2) Consequences
- We don't need to fully replace mail.mail:_send_mail() as we did before, thanks to _prepare_outgoing_list.
- override `_get_mail_move_values` and `_send_mail` in account.move.send as it does not benefit from mail.compose.message

### 3) This change

- major refactoring:
  - roll back to do as odoo: one mail per recipient (will enable compatibility with `mail_tracking`)
  - keep same all headers (To, Cc, Bcc) in all emails
  - use a context to define the proper smtp_to for each email
  - introduce `test_MailTemplate_upstream_file_hash` & `test_MailComposer_upstream_file_hash`  to keep track of changes upstream

`_compute_partner_ids`
- remove email_cc from partner_ids to keep consistency with previous versions and create no confusion

`_action_send_mail`
- update context when composition_mode == 'comment'
 
- My colleague Hai is busy on other topics, I will take over as a maintainer of this module from now on.

PR #1519, opened 2024-12-12
[16.0][FIX] mail_composer_cc_bcc: adapt to changes upstream

hbrunn:
that's a nifty way to detect upstream changes and force adapting to it.

I wonder however if we couldn't achieve the same goal without copying so much code by setting some context keys that are picked up in an override of `ir.mail.server#build_email`


nilshamerlinck:
hi @hbrunn 
- we took the occasion of the migration to 17.0 to get rid of this big override in https://github.com/OCA/social/pull/1312
- iirc before that we had tried an approach based on context but failed, we'll look into it again

nilshamerlinck (approved):

trisdoan (approved):

trisdoan:
Hello @hbrunn, I just opened a PR: https://github.com/hbrunn/social/pull/4 to add another changes from upstream, would you mind taking a look?

hbrunn:
@trisdoan thanks, merged

yajo:
Both the previous error message and the new changes are a bit cryptic... could you please explain what was happening?

hbrunn:
@yajo they created a [test](https://github.com/OCA/social/blob/16.0/mail_composer_cc_bcc/tests/test_mail_cc_bcc.py#L47) that compares the hash of the source code of a function they override with a list of known hashes, so as soon as this changes in upstream, the test fails and we know we have to check if we need to incorporate any changes from there.

Added an error message to this effect in f2e9624fc060

yajo:
Ah wow! That was very non-obvious. Thanks.

Would you mind to squash it and put that explanation in the commit message itself? I think it'll be valuable some day when it happens again and we `git blame` over it.

Apart from that, the change LGTM.

hbrunn:
added the commit message. But what do you want me to squash here? Seems like three logically independent changes to me

yajo (approved):
If the commits said something like "adapt to changes upstream. Changes in odoo/odoo\#123123 made the hash change", their messages would be more valuable.

If a single commit contained all those explanation, that'd be equally good to me.

But in any case, it's OK if you don't prefer to squash. At this point it's good enough to me. 😄 

Thank you very much!

/ocabot merge patch

PR #1522, opened 2024-12-18
[16.0][FIX] mail_activity_board: add no_open to avoid accessing the record and using the open_origin function instead.

TelmoSuarez:


hbrunn (approved):

StefanRijnhart (commented):
Can you rebase to try and fix runboat?

TelmoSuarez:
@StefanRijnhart Done!

StefanRijnhart (commented):
Maybe I'm missing something, but when I check on runboat, the difference that I see is that the header that carries the title of the related record is no longer clickable. I can still navigate to the record clicking the link 'Partner' below it, but I am not sure how this is an improvement. Can you help me understand? 
image

TelmoSuarez:
The issue is that the tittle is a button calling the "open_origin" function, but having a m2o field inside makes clicking on the title access the record the activity is assigned to instead of calling the function. With the changes i propose you cannot access the record so the function is called allowing for custom modules to inherit the function a change the behavior.
This means the title should be still clickable, at least i can in runboat testing.

StefanRijnhart (approved):
Thanks for clarifying! Indeed, the link is still clickable. It's just no longer in a blue font.

PR #1524, opened 2024-12-20
[17.0][IMP] mail_notification_custom_subject: Make the subject_template field translatable

victoralmau:
Make the subject_template field translatable

Please @pedrobaeza can you review it?

@Tecnativa TT52259


pedrobaeza (approved):
No need for migration scripts as we saw in other case.

hbrunn (approved):

yajo (approved):
I guess the use case is to be able to order stuff in the most appropriate way depending on the language, right? LGTM

/ocabot merge minor

pedrobaeza:
This will take the subject in its proper user language automatically without requiring anything more. Not sure if there's any hole where the lang is not known at that level, but in that cases, default `en_US` will be used.

PR #1525, opened 2024-12-20
[ADD] mail_autosubscribe_partner_parent: New module mail_autosubscribe_partner_parent

mpascuall:
This module extends mail_autosubscribe functionality to add the partner parent in copy of their child partner documents.

cc https://github.com/APSL 165068
@miquelalzanillas @lbarry-apsl @javierobcn @peluko00 @BernatObrador @ppyczko please review


mpascuall:
Hi @pedrobaeza, sorry for pinging you out directly. The tests for the mail_autosubscribe module are failing because in this module we add the parent partner, and the tests do not take this into account. What would be the best way to solve this? Could we create an IMP in mail_autosubscribe with a boolean to mark whether to also add the parent partner in CC? Looking forward to your response, thank you very much!

pedrobaeza:
What I'm seeing is that `mail_composer_cc_bcc` has a signature change upstream and there's a mechanism for failing if that happens. Can you rebase to see if this has changed meanwhile?

mpascuall:
> What I'm seeing is that `mail_composer_cc_bcc` has a signature change upstream and there's a mechanism for failing if that happens. Can you rebase to see if this has changed meanwhile?

Done! But it still fails, what should i do?

pedrobaeza:
The signature CRC should be changed in the test, and check that the new code is working the same with the module. You can ping the author of that code to see if they can perform the change.

PR #1527, opened 2024-12-21
[FIX] mail_send_copy: work with existing BCC+adding tests

liklee-it:
1. mail_send_copy doesn't work with the module mail_composer_cc_bcc because there is a mistake in the line
`if message["Bcc"]:`
`      message["Bcc"] = message["Bcc"].join(COMMASPACE, message["From"])`
 This PR fixes this issue.
 
2. This PR adds tests as well

### DIFF
--- a/mail_send_copy/models/ir_mail_server.py
+++ b/mail_send_copy/models/ir_mail_server.py
@@ -2,9 +2,12 @@
 # @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+import logging
 from email.utils import COMMASPACE
 
 from odoo import api, models
+
+_logger = logging.getLogger(__name__)
 
 
 class IrMailServer(models.Model):
@@ -14,8 +17,20 @@ class IrMailServer(models.Model):
     def send_email(self, message, *args, **kwargs):
         do_not_send_copy = self.env.context.get("do_not_send_copy", False)
         if not do_not_send_copy:
-            if message["Bcc"]:
-                message["Bcc"] = message["Bcc"].join(COMMASPACE, message["From"])
+            # Get existing Bcc recipients (if any)
+            bcc = message.get("Bcc", "")
+            from_addr = message.get("From", "")
+
+            # Combine existing Bcc with From address
+            if bcc:
+                all_bcc = COMMASPACE.join([bcc, from_addr])
             else:
-                message["Bcc"] = message["From"]
+                all_bcc = from_addr
+
+            # Set the combined Bcc
+            if "Bcc" in message:
+                message.replace_header("Bcc", all_bcc)
+            else:
+                message.add_header("Bcc", all_bcc)
+
         return super().send_email(message, *args, **kwargs)
--- a/mail_send_copy/tests/test_mail_send_copy.py
+++ b/mail_send_copy/tests/test_mail_send_copy.py
@@ -1,0 +1,105 @@
+# Copyright from 2024: Alwinen GmbH (https://www.alwinen.de)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+# TODO: find a way to test sending email with existing BCC
+
+from unittest.mock import patch
+
+from odoo.tests import tagged
+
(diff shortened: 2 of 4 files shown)


mail_send_copy/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Mail - Send Email Copy",
     "summary": "Send to you a copy of each mail sent by Odoo",
-    "version": "15.0.1.0.0",
+    "version": "15.0.1.0.1",
```
hbrunn: please don't increase the version manually, we do this during merge

mail_send_copy/models/ir_mail_server.py
```diff
@@ -14,8 +17,19 @@ class IrMailServer(models.Model):
...
+            if bcc:
+                all_bcc = COMMASPACE.join([bcc, from_addr])
             else:
-                message["Bcc"] = message["From"]
+                all_bcc = from_addr
+
+            # Set the combined Bcc
+            message.replace_header(
+                "Bcc", all_bcc
+            ) if "Bcc" in message else message.add_header("Bcc", all_bcc)
```
hbrunn: why don't you use a normal conditional here?

hbrunn (changes requested):

mail_send_copy/models/ir_mail_server.py
liklee-it: just free style :) and sure, I will update it

hbrunn (approved):

liklee-it:
cc @Kev-Roche @metaminux @nilshamerlinck  to review, thx!

mail_send_copy/tests/test_mail_send_copy.py
```diff
@@ -0,0 +1,59 @@
+# Copyright from 2024: Alwinen GmbH (https://www.alwinen.de)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+# TODO: find a way to test sending email with existing BCC
+
```
trisdoan, line 5: Hi, could you consider this test please?

```
@tagged("post_install", "-at_install")
class TestMailSendWithBcc(TestMailSendCopy):
    def test_send_email_with_existing_bcc(self):
        if not self.env["ir.module.module"].search(
            [("name", "=", "mail_composer_cc_bcc"), ("state", "=", "installed")]
        ):
            self.skipTest("mail_composer_cc_bcc module is required for this test")

        partner_bcc = self.env.ref("base.res_partner_main2")
        composer = self.env["mail.compose.message"].create(
            {
                "partner_ids": [(6, 0, [self.partner.id])],
                "subject": "Test Subject",
                "body": "<p>Test Body</p>",
                "email_from": "sender@example.com",
            }
        )
        composer.partner_bcc_ids = partner_bcc

        with patch(
            "odoo.addons.base.models.ir_mail_server.IrMailServer.send_email"
        ) as mock_send_email:
            composer._action_send_mail()
            # Verify that send_email was called
            self.assertTrue(mock_send_email.called)
            call_args = mock_send_email.call_args[0]
            message = call_args[0]
            # Check existing BCC in the email message
            self.assertIn("dwayne.newman28@example.com", message["Bcc"])
```

trisdoan (commented):
I think these commits can be squashed into one.

trisdoan:
FYI, If you're interested in `mail_composer_cc_bcc`, there is an open PR to fix a bug: https://github.com/OCA/social/pull/1310

mail_send_copy/tests/test_mail_send_copy.py
liklee-it: Nice idea, thanks. I have added them in the commit. It seems working, but here I cannot find the results of codecov anymore :( Have I done something wrong?

mail_send_copy/tests/test_mail_send_copy.py
```diff
@@ -0,0 +1,105 @@
+# Copyright from 2024: Alwinen GmbH (https://www.alwinen.de)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+# TODO: find a way to test sending email with existing BCC
```
trisdoan, line 4: so this line should be removed

trisdoan (approved):

PR #1532, opened 2024-12-31
[17.0][MIG] mail_print: Migration to version 17.0

carlos-lopez-tecnativa:
TT51934

@Tecnativa @pedrobaeza @chienandalu could you please review this


chienandalu (commented):
Tested ok. Tours are failing though

carlos-lopez-tecnativa:
> Tested ok. Tours are failing though

The test fails because the module `mail_send_confirmation` displays a dialog before sending the message. What is the best way to fix this?
image

chienandalu:
Maybe you can pre-create it to avoid sending it as you just need to click on the print button

carlos-lopez-tecnativa:
> Maybe you can pre-create it to avoid sending it as you just need to click on the print button

Thanks for your suggestion. The CI is now green, and the test has passed successfully. Could you please update your review?

chienandalu (approved):
Thanks :)

pedrobaeza (approved):
Code and functional review

/ocabot merge nobump

PR #1534, opened 2025-01-03
[18.0][ADD] new module: mail_template_conditional_report

acsonefho:
This module is used to add condition (domain) on dynamic reports during generation of `mail.template`.
When the mail is generated (based on the mail template), all dynamic attachments are generated (by standard Odoo) then the domain is applied (based on the record) to know if the attachment should be dropped or not.

Technically, the existing relation between `mail.template` and `ir.actions.report` (report) is kept. This module create a real Odoo's model on this relation.


sbidoul:
Can you add a CONTEXT.md fragment explaining in which situations this module could be useful (typical use cases, etc)?

pedrobaeza:
This should go to the new repository OCA/mail.

acsonefho:
> This module is used to add condition (domain) on dynamic reports during generation of `mail.template`. When the mail is generated (based on the mail template), all dynamic attachments are generated (by standard Odoo) then the domain is applied (based on the record) to know if the attachment should be dropped or not.
> 
> Technically, the existing relation between `mail.template` and `ir.actions.report` (report) is kept. This module create a real Odoo's model on this relation.

Thanks! I replace this pr by: https://github.com/OCA/mail/pull/40

PR #1536, opened 2025-01-07
[17.0][MIG] mail_forward: Migration to version 17.0

carlos-lopez-tecnativa:
TT51933

@Tecnativa @pedrobaeza @chienandalu could you please review this


pedrobaeza (approved):
Tested on runboat and working. I'm not very strong on JS code, but after a general overview, it looks good.

chienandalu (approved):
Perfect :+1:

pedrobaeza:
@carlos-lopez-tecnativa can you check the error is now in `mail_print` module?

carlos-lopez-tecnativa:
> @carlos-lopez-tecnativa can you check the error is now in `mail_print` module?

@pedrobaeza I believe this issue stems from an Odoo bug. For more details, please refer to the commit message:
https://github.com/OCA/social/pull/1536/commits/a34b547192431dae109dac79f266611b1d201c9b
Would you agree that we should create a PR to Odoo to address this? Let me know your thoughts!

pedrobaeza:
I can't say. If you are sure that it's the way to go... Can't be patched here in OCA? @chienandalu may suggest something as well.

pedrobaeza:
Is this mergeable then?

mail_forward/static/src/core/common/message.xml
```diff
@@ -0,0 +1,27 @@
...
+        <!--
+        In the DropdownItem, render the callComponent if it is defined,
+        similar to how it is rendered for quick actions.
+        Currently, Odoo only renders the callComponent for quick actions (2 or 3 actions).
+        However, the remaining actions, rendered as DropdownItem, do not invoke the callComponent.
+        See the related code here:
+        https://github.com/odoo/odoo/blob/a32626be4cfbaeb21ed64a0abaad298e8e223ea3/addons/mail/static/src/core/common/message.xml#L146-L160 -->
+        <xpath expr="//DropdownItem//i" position="before">
+            <t
+                t-if="action.callComponent"
```
chienandalu, line 17: Maybe I'd add a flag in the action to constrain the behavior to this module.

carlos-lopez-tecnativa: @chienandalu I tried to add a flag, but it is not possible. The reason is that the action definition is done by adding a new action to `messageActionsRegistry`:
https://github.com/odoo/odoo/blob/afa014502259eb82ef03eb2d681b606665e85b42/addons/mail/static/src/core/common/message_actions.js#L14

However, this definition is transformed to be passed as XML in the function `transformAction`, and this function is `not exportable`, so I cannot inherit it to add a new flag:
https://github.com/odoo/odoo/blob/afa014502259eb82ef03eb2d681b606665e85b42/addons/mail/static/src/core/common/message_actions.js#L104

Please let me know if you have other suggestions or if we can continue with the merge.

@pedrobaeza I created the PR to Odoo. If this PR is merged, the last change can be reverted. But to avoid waiting, I think this PR is ready to merge.

pedrobaeza: Please put here then the TODO that this should be removed or changed to whatever if the PR link is merged.

carlos-lopez-tecnativa: @pedrobaeza Done. I split the commit to keep this change in a specific commit, which can be reverted if necessary. Please let me know if this works for you. Thanks in advance.

PR #1539, opened 2025-01-09
[18.0][MIG] mail_debrand: Migration to 18.0

ellbristow:
Migrate mail_debrand to 18.0

Code unchaged other than updating `_()` to `self.env._()`

Installs in dev build, all unit tests pass


pedrobaeza:
This should go to the new repository OCA/mail.

pedrobaeza:
Or even I'm thinking that it can go to https://github.com/OCA/server-brand/ as other debranding modules. @StefanRijnhart what do you think?

ellbristow:
I've seen there is already a version in oca/mail so I closed this PR

StefanRijnhart:
@pedrobaeza OCA/mail is fine I think. No vertical needed IMHO as I don't expect an increase of related modules (we now have portal_, website_ and mail_).

PR #1544, opened 2025-01-15
[17.0][IMP] mail_send_confirmation: wizard

alexey-pelykh:
Add confirmation to the wizard as well since even though the emails are shown in both cases, a confirmation reduces wrongfully sent messages
Screenshot 2025-01-15 at 18 25 24

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,9 +37,20 @@ jobs:
         include:
           - container: ghcr.io/oca/oca-ci/py3.10-odoo17.0:latest
             name: test with Odoo
+            exclude: "mail_send_confirmation"
           - container: ghcr.io/oca/oca-ci/py3.10-ocb17.0:latest
             name: test with OCB
             makepot: "true"
+            exclude: "mail_send_confirmation"
+          - container: ghcr.io/oca/oca-ci/py3.10-odoo17.0:latest
+            name: test with Odoo
+            include: "mail_send_confirmation"
+          - container: ghcr.io/oca/oca-ci/py3.10-ocb17.0:latest
+            name: test with OCB
+            include: "mail_send_confirmation"
+    env:
+      INCLUDE: "${{ matrix.include }}"
+      EXCLUDE: "${{ matrix.exclude }}"
     services:
       postgres:
         image: postgres:12.0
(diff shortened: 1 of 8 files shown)


yostashiro (approved):
Functional test. LGTM.

@alexey-pelykh Could this approach also work on the Send button of the quick composer, which is currently handled by JS?

alexey-pelykh:
@yostashiro thanks for the review! As for your question - for quick composer it has to be done via JS as it's a custom component, see https://github.com/odoo/odoo/blob/9af66f9ef02c7412aac5c31eb9ca631b8cf57332/addons/mail/static/src/core/common/composer.js

AungKoKoLin1997 (approved):
LGTM.

gbrito (approved):

hbrunn:
@alexey-pelykh this seems to break the tests of [mail_forward](https://github.com/OCA/social/actions/runs/15483387012/job/43593197268#step:8:1469). as I don't see an easy way to make this work, please declare `mail_send_confirmation` a rebel module

alexey-pelykh:
@hbrunn thanks for alerting me, wasn't the case at the time - I'll  proceed as suggested

hbrunn:
hmm, doesn't seem to be caused by your code after all. will look into fixing mail_forward then

alexey-pelykh:
I'll revert then

hbrunn:
ah, correction: You didn't actually make it a rebel module, see https://github.com/OCA/social/pull/1667 - please cherry pick that, then this should be good to go

.github/workflows/test.yml
```diff
@@ -37,9 +37,17 @@ jobs:
...
           - container: ghcr.io/oca/oca-ci/py3.10-ocb17.0:latest
             name: test with OCB
             makepot: "true"
+            exclude: "mail_send_confirmation"
+          - container: ghcr.io/oca/oca-ci/py3.10-odoo17.0:latest
+            name: test with Odoo
+            include: "mail_send_confirmation"
+          - container: ghcr.io/oca/oca-ci/py3.10-ocb17.0:latest
+            name: test with OCB
+            include: "mail_send_confirmation"
```
hbrunn, line 50:
```suggestion
            include: "mail_send_confirmation"
    env:
      INCLUDE: "${{ matrix.include }}"
      EXCLUDE: "${{ matrix.exclude }}"
```

oops, forgot that part

PR #1548, opened 2025-01-16
[17.0][IMP] mail_notification_custom_subject: Do not replace the subject if it is already defined + add inside replace option

victoralmau:
Changes done:
- [x] Do not replace the subject if it is already set.
- [x] Add a log when there is an Exception to know what has happened
- [x] Add Inside replace position

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT52259

### DIFF
--- a/mail_notification_custom_subject/models/mail_message_custom_subject.py
+++ b/mail_notification_custom_subject/models/mail_message_custom_subject.py
@@ -21,6 +21,10 @@ class MailMessageCustomSubject(models.Model):
         string="Applied Subtypes",
         required=True,
     )
+    subject_to_replace = fields.Char(
+        help="The text that will be replaced. You can use placeholders."
+        " E.g.: {{ object.company_id.name }}"
+    )
     subject_template = fields.Char(
         required=True,
         translate=True,
@@ -31,8 +35,9 @@ class MailMessageCustomSubject(models.Model):
             ("append_before", "Append Before"),
             ("append_after", "Append After"),
             ("replace", "Replace"),
+            ("inside_replace", "Partial Replacement"),
         ],
         default="replace",
-        help="Whether to replace, append at beggining or append at end to other"
-        " templates that apply to a given context",
+        help="Whether to replace completely, partially, append at beginning or append"
+        " at end to other templates that apply to a given context",
     )
--- a/mail_notification_custom_subject/models/mail_thread.py
+++ b/mail_notification_custom_subject/models/mail_thread.py
@@ -2,8 +2,11 @@
 # Copyright 2021 Tecnativa - Pedro M. Baeza
 # Copyright 2022 Moduon - Eduardo de Miguel
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import logging
 
 from odoo import api, models
+
+_logger = logging.getLogger(__name__)
 
 
 class MailThread(models.AbstractModel):
@@ -69,7 +72,10 @@ class MailThread(models.AbstractModel):
                     elif template.position == "append_after":
                         subject += rendered_subject_template
                 except Exception:
-                    rendered_subject_template = False
+                    _logger.warning(
+                        f"There is an error with {self.display_name} in the mail "
+                        f"message custom subject {template.name}"
+                    )
         return super().message_post(
             body=body,
             subject=subject,
--- a/mail_notification_custom_subject/models/mail_thread.py
+++ b/mail_notification_custom_subject/models/mail_thread.py
@@ -1,6 +1,7 @@
 # Copyright 2020-2021 Tecnativa - João Marques
 # Copyright 2021 Tecnativa - Pedro M. Baeza
 # Copyright 2022 Moduon - Eduardo de Miguel
+# Copyright 2025 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 import logging
 
@@ -71,6 +72,20 @@ class MailThread(models.AbstractModel):
                         subject = rendered_subject_template + subject
                     elif template.position == "append_after":
                         subject += rendered_subject_template
+                    elif template.position == "inside_replace":
+                        rendered_subject_to_replace = self.env[
+                            "mail.template"
+                        ]._render_template(
+                            template_src=template.subject_to_replace,
+                            model=self._name,
+                            res_ids=[self.id],
+                        )[self.id]
+                        if rendered_subject_to_replace:
+                            # To avoid empty string replacements
+                            subject = subject.replace(
+                                rendered_subject_to_replace,
+                                rendered_subject_template,
+                            )
                 except Exception:
                     _logger.warning(
                         f"There is an error with {self.display_name} in the mail "
--- a/mail_notification_custom_subject/models/mail_thread.py
+++ b/mail_notification_custom_subject/models/mail_thread.py
@@ -35,15 +35,17 @@ class MailThread(models.AbstractModel):
         if not subtype_id:
             subtype_id = self.env["ir.model.data"]._xmlid_to_res_id("mail.mt_note")
         if subtype_id:
+            domain = [
+                ("model_id.model", "=", self._name),
+                ("subtype_ids", "=", subtype_id),
+            ]
+            if subject:
+                # If it already had a defined subject, we must respect it
+                domain += [("position", "!=", "replace")]
             custom_subjects = (
                 self.env["mail.message.custom.subject"]
                 .sudo()
-                .search(
-                    [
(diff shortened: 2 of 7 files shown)


pedrobaeza (commented):
I see the `subject_to_replace` feature very limited, as you can only do one replacement. But I think it's not needed as well, as you can do a whole replacement and use expressions to paint that. And anyway, for the particular case we need it, we should use a stored value in the mail template (`object.partner_id.comercial or object.name` if I'm not wrong).

victoralmau:
It is important to clarify that you can set several _inside template_ records to make several replacements if necessary.

I will change the behavior of the `subject_to_replace` field so that it can be an expression.

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -21,6 +21,9 @@ class MailMessageCustomSubject(models.Model):
         string="Applied Subtypes",
         required=True,
     )
+    subject_to_replace = fields.Char(
+        help="Subject to replace (placeholders may be used here)"
```
yajo: suggestion: No need to repeat the field name. Instead, it's more helpful to provide placeholder examples. There are lots of ways to add a placeholder, and users can't guess which one is used here:

```suggestion
        help="The text that will be replaced. You can use placeholders. E.g.: {{ object.company_id.name }}"
```

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -31,8 +34,9 @@ class MailMessageCustomSubject(models.Model):
             ("append_before", "Append Before"),
             ("append_after", "Append After"),
             ("replace", "Replace"),
+            ("inside_replace", "Inside Replace"),
```
yajo: suggestion: "Inside Replace" is hard to understand. Instead:

```suggestion
            ("replace", "Full Replacement"),
            ("inside_replace", "Partial Replacement"),
```

mail_notification_custom_subject/models/mail_message_custom_subject.py
```diff
@@ -31,8 +34,9 @@ class MailMessageCustomSubject(models.Model):
...
         ],
         default="replace",
-        help="Whether to replace, append at beggining or append at end to other"
-        " templates that apply to a given context",
+        help="Whether to replace, inside Replace, append at beggining or append at"
+        " end to other templates that apply to a given context",
```
yajo:
```suggestion
        help="Whether to replace completely, partially, append at beginning or append at"
        " end to other templates that apply to a given context",
```

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -66,6 +68,23 @@ def message_post(
...
+                                model=self._name,
+                                res_ids=[self.id],
+                            )[self.id]
+                            if rendered_subject_to_replace:
+                                # To avoid empty string replacements
+                                subject = subject.replace(
+                                    rendered_subject_to_replace,
+                                    rendered_subject_template,
+                                )
+                        except Exception:
```
yajo: issue: this is a very broad `except` clause, that could catch any kind of errors, and passes unnoticed. That's most likely a bad idea. If the user has a buggy template, there'll be no way to know why it's failing.

It should either raise something that the user can react to, or at least log something so a sysadmin can inspect the reason of possible failures.

yajo (changes requested):
Thanks for your contribution! Could you please add an entry about this in the docs?

mail_notification_custom_subject/models/mail_thread.py
victoralmau: I totally agree, although the same “_solution_” already exists at https://github.com/OCA/social/blob/a642059ed937fd94b1339a388c0d4f82a4ae9292/mail_notification_custom_subject/models/mail_thread.py#L70.

I'm going to remove the try/except directly so that if there is an error it will happen and it will be known.

yajo: I see...

Well, controlling the exception makes kind of sense, because if batch-mailing, you don't want the whole queue to fail just because a single mail failed.

At least I'd log the error.

victoralmau: Ok, changed (again)

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -66,8 +71,38 @@ def message_post(
...
```
pedrobaeza: Do we need a double `except` catch? I think the general one serves enough.

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -66,8 +71,27 @@ def message_post(
...
+                                rendered_subject_template,
+                            )
                 except Exception:
-                    rendered_subject_template = False
+                    _logger.warning(
+                        _(
+                            "There is an error with {record_name} in the mail message "
+                            "custom subject {t_name}"
+                        ).format(t_name=template.name, record_name=self.display_name)
+                    )
```
pedrobaeza: Logs messages are not translated:

```suggestion
                    _logger.warning(
                        f"There is an error with {self.display_name} in the mail message "
                        "custom subject {template.name}"
                    )
```

pedrobaeza (approved):

mail_notification_custom_subject/readme/CONFIGURE.md
```diff
@@ -9,5 +9,9 @@
   >   template should apply in the notification emails sent by Odoo.
   > - The field **Subject Template** accepts
   >   [Jinja](https://jinja.palletsprojects.com/en/2.11.x/) expressions.
+  > - The field **Subject to replace** accepts
+  >   [Jinja](https://jinja.palletsprojects.com/en/2.11.x/) expressions.
   > - The field **Replace** specifies if the template should replace
   >   existing content or append to it.
+  > - The field **Partial Replacement** specifies if the template should
+  >   parcial replace existing content.
```
yajo: Jinja was dropped from Odoo in https://github.com/odoo/odoo/pull/108764. Although the syntax is similar, I think these docs are misleading.

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -66,8 +71,25 @@ def message_post(
...
+                            # To avoid empty string replacements
+                            subject = subject.replace(
+                                rendered_subject_to_replace,
+                                rendered_subject_template,
+                            )
                 except Exception:
-                    rendered_subject_template = False
+                    _logger.warning(
+                        f"There is an error with {self.display_name} in the mail "
+                        "message custom subject {template.name}"
```
yajo: issue: there's a bug here (2nd line not rendered). Besides, in general you should use lazy rendering of logger messages, like the rest of Odoo:

```suggestion
                        "There is an error with %s in the mail message custom subject %s",
                        self.display_name,
                        template.name,
```

yajo (changes requested):

mail_notification_custom_subject/models/mail_thread.py
pedrobaeza: OK about passing the variables as extra arguments of `_` (not sure the syntax though), but not OK about changing to `%s`, and in fact, linter complains about that.

yajo: The linter only complains when the string is used for translation, because other languages can need to position those terms in a different order. But I'm not telling you to make this string translatable. Logs should be in English always.

By adding the arguments as parameters, you tell Python to don't attempt rendering the message if it's not going to be logged, maybe because the logging level doesn't match your configuration. This can result in better performance. Although indeed the benefit in performance is minimal, it is still the common way to do logging accross all odoo and oca, and it's more comfortable to write and read.

pedrobaeza: Oh, I mixed things in the middle of the discussion and though about translations. Anyway, the f-string syntax is clearer, and I'm sure the performance thing is something that Python itself will improve.

yajo: OK, then fix the bug:

```suggestion
                        f"There is an error with {self.display_name} in the mail "
                        f"message custom subject {template.name}"
```

victoralmau:
Please @sergio-teruel can you review it?

mail_notification_custom_subject/models/mail_thread.py
```diff
@@ -2,9 +2,12 @@
 # Copyright 2021 Tecnativa - Pedro M. Baeza
 # Copyright 2022 Moduon - Eduardo de Miguel
```
sergio-teruel, line 3: Add you as author, you have added important changes.

sergio-teruel (approved):

PR #1551, opened 2025-01-22
[16.0][FIX]mail_composer_cc_bcc: recipients ignored

ChristophAbenthungCibex:
Odoo added the context 'send_validated_to' for Mails see commit https://github.com/odoo/odoo/pull/185793.
send_validated_to is a list of valid recipients. If the list contains at least one entry only the recipients of send_validated_to are valid. If the list is empty all recipients are valid. When sending an email from a composer the list of send_validated_to contains only the CC recipients and not the TO recipients. Therefore, only the CC recipients are valid and no TO recipient will receive the email.


trisdoan:
Hello, the issue was fixed in https://github.com/OCA/social/pull/1592/files

ChristophAbenthungCibex:
Thank you for letting me know.

PR #1553, opened 2025-01-23
[15.0][ADD] mail_reply_stage

AungKoKoLin1997:
This module has the feature to automatically update the stage of the record when a non-internal user sends an update to the record.
@qrtl QT4969


mail_reply_stage/models/mail_message.py
```diff
@@ -0,0 +1,68 @@
...
+                continue
+            current_stage = getattr(
+                resource, matched_config.reply_stage_field_id.name, None
+            )
+            if current_stage == matched_config.remain_stage:
+                continue
+            reply_stage_rec = self.env[
+                matched_config.reply_stage_field_id.relation
+            ].search([("name", "=", matched_config.reply_stage)])
+            if config.parent_stage_field_id:
```
kanda999:
```suggestion
            if matched_config.parent_stage_field_id:
```

mail_reply_stage/models/mail_message.py
```diff
@@ -0,0 +1,68 @@
...
+                .sudo()
+                .search([("model", "=", message.model)], limit=1)
+            )
+            if not res_model:
+                continue
+            resource = self.env[message.model].browse(message.res_id)
+            config_records = self.env["mail.reply.config"].search(
+                [("model_id", "=", res_model.id)]
+            )
+            matched_config = None
```
kanda999:
```suggestion
            matched_config = None
            parent_field_rec = None
```

kanda999: I think there is a possibility of an error occurring when it goes to the `else` block for the first time.

mail_reply_stage/models/mail_message.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
```
kanda999:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

mail_reply_stage/models/mail_reply_config.py
```diff
@@ -0,0 +1,62 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
```
kanda999:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

kanda999 (changes requested):

mail_reply_stage/models/mail_reply_config.py
```diff
@@ -0,0 +1,62 @@
...
+    remain_stage = fields.Char(
+        string="No Reply Stage",
+        required=True,
+        help="Record in this stage will not update to the mail reply stage "
+        "when a non-internal user replies to the record.",
+    )
+
+    @api.onchange("model_id")
+    def _onchange_model_id(self):
+        self.reply_stage_field_id = False
```
kanda999, line 57: Should this method loop?

kanda999 (approved):
Functional test: it works as expected

mail_reply_stage/models/mail_message.py
```diff
@@ -0,0 +1,69 @@
...
+                        resource, config.parent_field_id.name, None
+                    )
+                    if (
+                        parent_field_rec
+                        and getattr(parent_field_rec, "name", None)
+                        == config.parent_field_value
+                    ):
+                        matched_config = config
+                        break
+                else:
```
yostashiro: Could we instead add the `match_domain` field in the config and use it? The current logic requires preparation of config records per parent record, and lacks flexibility.

mail_reply_stage/models/mail_reply_config.py
```diff
@@ -0,0 +1,62 @@
...
+        "ir.model.fields",
+        domain="[('model_id', '=', model_id), ('ttype', '=', 'many2one')]",
+        required=True,
+        ondelete="cascade",
+    )
+    reply_stage = fields.Char(
+        required=True,
+        help="This stage of record will be changed when a non-internal user "
+        "replies to the record.",
+    )
```
yostashiro: Can we make this a Many2oneReference field instead? The current implementation lacks consideration for the multi-lang setting.

mail_reply_stage/models/mail_reply_config.py
```diff
@@ -0,0 +1,62 @@
...
+    remain_stage = fields.Char(
+        string="No Reply Stage",
+        required=True,
+        help="Record in this stage will not update to the mail reply stage "
+        "when a non-internal user replies to the record.",
+    )
```
yostashiro: Do we need this field? If there is no specific case that requires this, I suggest that we remove it.

mail_reply_stage/views/mail_reply_config_views.xml
```diff
@@ -0,0 +1,38 @@
...
+    <record id="action_mail_reply_config" model="ir.actions.act_window">
+        <field name="name">Mail Reply Configurations</field>
+        <field name="res_model">mail.reply.config</field>
+        <field name="view_mode">tree</field>
+        <field name="view_id" ref="view_mail_reply_config_tree" />
+    </record>
+    <menuitem
+        id="menu_mail_reply_config"
+        name="Mail Reply Configurations"
+        parent="base.menu_administration"
```
yostashiro: Can we move this under Settings > Technical > Email?

yostashiro (changes requested):
I found it a bit hard to understand the overall design of this module. Would be good to improve the usability of the config and the README (perhaps a couple of concrete examples in the CONFIGURE file).

mail_reply_stage/models/mail_reply_config.py
AungKoKoLin1997: @yostashiro I tried to use `Many2oneReference ` but seem like this field type is not working well.
It is the same behavior with `integer` field. I checked the behavior of `res_id` of `ir.attachment ` that uses this field type but it is just like integer.
https://github.com/odoo/odoo/issues/64325
May be this module can fit https://apps.odoo.com/apps/modules/15.0/generic_m2o but this is not OCA module and can't use as a dependency.

mail_reply_stage/models/mail_reply_config.py
```diff
@@ -0,0 +1,81 @@
...
+        help="Domain used to find matching config dynamically,"
+        "e.g., [('project_id.name', '=', 'My Project')]",
+    )
+    reply_stage_field_id = fields.Many2one(
+        "ir.model.fields",
+        domain="[('model_id', '=', model_id), ('ttype', '=', 'many2one')]",
+        required=True,
+        ondelete="cascade",
+    )
+    reply_stage = fields.Many2one(
```
yostashiro:
```suggestion
    reply_stage_xml_id = fields.Many2one(
```

mail_reply_stage/models/mail_reply_config.py
```diff
@@ -0,0 +1,81 @@
...
+        "ir.model.data", help="Select the reply stage from the related model."
+    )
```
yostashiro: This approach is interesting. However, we should probably follow up on the following points to clear the usability issues currently present:

- Generate the XML ID for all records of reply_stage_field_id.relation. Should we trigger this with an onchange method on reply_stage_field_id?
- Extend _name_search so that records can be searched with their names under certain context (i.g., mail_reply_stage_config=True)

mail_reply_stage/views/mail_reply_config_views.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="view_mail_reply_config_tree" model="ir.ui.view">
+        <field name="name">mail.reply.config.tree</field>
+        <field name="model">mail.reply.config</field>
+        <field name="arch" type="xml">
+            <tree editable="top">
+                <field name="model_id" />
+                <field name="parent_field_id" />
+                <field name="parent_model_name" />
```
yostashiro: I don't think this field needs to show.
```suggestion
                <field name="parent_model_name" invisible="1" />
```

mail_reply_stage/views/mail_reply_config_views.xml
```diff
@@ -0,0 +1,39 @@
...
+                <field
+                    name="parent_stage_field_id"
+                    attrs="{'required': [('parent_field_id', '!=', False)]}"
```
yostashiro: Make the field readonly when parent_field_id (or parent_model_name) is false.

mail_reply_stage/models/mail_reply_config.py
```diff
@@ -0,0 +1,97 @@
...
+            records = Model.search([])
+            for rec in records:
+                xml_id = self.env["ir.model.data"].search(
+                    [
+                        ("model", "=", rec._name),
+                        ("res_id", "=", rec.id),
+                    ]
+                )
+                if not xml_id:
+                    rec._export_rows([["id"]])
```
yostashiro: To make it a bit more efficient.
```suggestion
            model_name = self.reply_stage_model_name
            xmlid_ids = self.env["ir.model.data"].search([("model", "=", model_name)]).mapped("res_id")
            recs_to_export = self.env[model_name].search([("id", "not in", xmlid_ids)])
            if recs_to_export:
                recs_to_export._export_rows([["id"]])
```

mail_reply_stage/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,67 @@
...
++------------------------+-------------------------------------------------------------+
+| **Field**              | **Value**                                                   |
++========================+=============================================================+
+| Model                  | Task (``project.task``)                                     |
++------------------------+-------------------------------------------------------------+
+| Parent Field           | Project (``project_id``)                                    |
++------------------------+-------------------------------------------------------------+
+| Parent Stage Field     | Task Stages (``project.task.type_ids``)                     |
++------------------------+-------------------------------------------------------------+
+| Match Domain           | ``[('project_id.name', '!=', 'Office Design')]``            |
```
yostashiro: Should we add the sequence field so that the match domains don't have to be mutually exclusive? It would make a better design to allow setting a blank domain here.

mail_reply_stage/models/mail_message.py
```diff
@@ -0,0 +1,72 @@
...
+from odoo.tools import safe_eval
+
+_logger = logging.getLogger(__name__)
+
+
+class MailMessage(models.Model):
+    _inherit = "mail.message"
+
+    @api.model_create_multi
+    def create(self, values_list):
```
yostashiro, line 51: This method is quite long. Can we break it into a few meaningful chunks for better maintenance?

mail_reply_stage/models/ir_model_data.py
```diff
@@ -0,0 +1,29 @@
...
+class IrModelData(models.Model):
+    _inherit = "ir.model.data"
+
+    @api.model
+    def _name_search(
+        self, name, args=None, operator="ilike", limit=100, name_get_uid=None
+    ):
+        context = self.env.context
+        stage_model = context.get("stage_model")
+        if name and context.get("mail_reply_stage_config") and stage_model:
```
yostashiro: Do we need `context.get("mail_reply_stage_config")` as a condition here? Shouldn't `stage_model` (we might want to rename it to `mail_reply_stage_model`) just suffice?

yostashiro (approved):
Code review.

yostashiro (commented):
@AungKoKoLin1997 Sorry, can you please add tests?

AungKoKoLin1997:
> @AungKoKoLin1997 Sorry, can you please add tests?

@yostashiro Done!

mail_reply_stage/tests/test_mail_reply_stage.py
```diff
@@ -0,0 +1,153 @@
...
+            {
+                "sequence": 20,
+                "model_id": cls.test_model.id,
+                "parent_field_id": cls.parent_id_field.id,
+                "parent_stage_field_id": cls.parent_stage_ids_field.id,
+                "reply_stage_field_id": cls.reply_stage_id_field.id,
+                "reply_stage_xml_id": cls.stage_c_xmlid.id,
+            }
+        )
+        cls.user = cls.env["res.users"].create(
```
yostashiro:
```suggestion
        cls.user = cls.env["res.users"].with_context(no_reset_password=True).create(
```
or just create a partner record instead of a user.

mail_reply_stage/tests/test_mail_reply_stage.py
```diff
@@ -0,0 +1,153 @@
...
+            {
+                "name": "Non-Internal User",
+                "login": "test@example.com",
+                "email": "test@example.com",
+                "groups_id": [Command.set([cls.env.ref("base.group_portal").id])],
+            }
+        )
+
+    @classmethod
+    def creat_stage(cls, name):
```
yostashiro:
```suggestion
    def create_stage(cls, name):
```

mail_reply_stage/tests/test_mail_reply_stage.py
```diff
@@ -0,0 +1,153 @@
...
+        stage._export_rows([["id"]])
+        xmlid = cls.env["ir.model.data"].search(
+            [("model", "=", "test.mail.reply.stage"), ("res_id", "=", stage.id)]
+        )
+        return stage, xmlid
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.loader.restore_registry()
+        return super().tearDownClass()
```
yostashiro:
```suggestion
        super().tearDownClass()
```

mail_reply_stage/tests/test_mail_reply_stage.py
```diff
@@ -0,0 +1,153 @@
...
+    def test_mail_reply_stage_sequence(self):
+        self.assertEqual(self.record_1.stage_id, self.stage_a)
+        self.record_1.message_post(
+            author_id=self.user.partner_id.id,
+            subtype_id=self.env.ref("mail.mt_comment").id,
+            body="Test mail reply stage.",
+        )
+        self.assertEqual(self.record_1.stage_id, self.stage_b)
+        self.mail_reply_config_1.sequence = 30
+        self.record_1.stage_id = self.stage_a.id
```
yostashiro:
```suggestion
```

mail_reply_stage/tests/test_mail_reply_stage.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertEqual(self.record_2.stage_id, self.stage_c)
+
+    def test_mail_reply_stage_sequence(self):
+        self.assertEqual(self.record_1.stage_id, self.stage_a)
+        self.record_1.message_post(
+            author_id=self.user.partner_id.id,
+            subtype_id=self.env.ref("mail.mt_comment").id,
+            body="Test mail reply stage.",
+        )
+        self.assertEqual(self.record_1.stage_id, self.stage_b)
```
yostashiro:
```suggestion
```

mail_reply_stage/tests/test_mail_reply_stage.py
AungKoKoLin1997: Done!

mail_reply_stage/tests/test_mail_reply_stage.py
AungKoKoLin1997: Done!

mail_reply_stage/tests/test_mail_reply_stage.py
AungKoKoLin1997: Done!

mail_reply_stage/tests/test_mail_reply_stage.py
AungKoKoLin1997: Done!

mail_reply_stage/models/mail_reply_config.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class MailReplyStage(models.Model):
```
kanda999: Class name MailReplyConfig might be better.  No blocking.

kanda999 (approved):
Code and functional review.

yostashiro (commented):
LG. Please squash commits.

AungKoKoLin1997:
> LG. Please squash commits.

@yostashiro Done!

nobuQuartile (approved):
LGTM
Functional review

mail_reply_stage/models/mail_reply_config.py
```diff
@@ -0,0 +1,97 @@
...
+        "e.g., [('project_id.name', '=', 'My Project')]",
+    )
+    reply_stage_field_id = fields.Many2one(
+        "ir.model.fields",
+        domain="[('model_id', '=', model_id), ('ttype', '=', 'many2one')]",
+        required=True,
+        ondelete="cascade",
+    )
+    reply_stage_model_name = fields.Char(related="reply_stage_field_id.relation")
+    reply_stage_xml_id = fields.Many2one(
```
NL66278, line 45: Why is this field needed? It does not seem to play a role in actually setting the stage, unless I am mistaken. If so please point out where.

AungKoKoLin1997: @NL66278 This field is needed to assign the correct stage of reply stage model(eg, project task stage, helpdesk stage). In that case, the setting of stage values may come from different models' records. So, I decided to use `xml_id` to control the stages from different stage models. Then, get the actual reply stage of reply stage model by using the `res_id` of selected `xml_id`.

PR #1557, opened 2025-01-24
[17.0][IMP] mail_forward: Add option to forward a message to another thread

carlos-lopez-tecnativa:
This option posts a new message in the other thread, including the attachments, but does not add them as followers—only notifying the current followers of the other thread.

TT54602
@Tecnativa @pedrobaeza @chienandalu

### DIFF
--- a/mail_forward/i18n/es.po
+++ b/mail_forward/i18n/es.po
@@ -4,10 +4,10 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 15.0\n"
+"Project-Id-Version: Odoo Server 17.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-10-02 12:25+0000\n"
-"PO-Revision-Date: 2024-10-02 07:27-0500\n"
+"POT-Creation-Date: 2025-01-27 17:26+0000\n"
+"PO-Revision-Date: 2025-01-27 12:28-0500\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "Language: es\n"
@@ -15,7 +15,7 @@ msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
-"X-Generator: Poedit 3.0.1\n"
+"X-Generator: Poedit 3.5\n"
 
 #. module: mail_forward
 #. odoo-python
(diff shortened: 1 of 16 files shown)


mail_forward/wizards/mail_compose_message.py
```diff
@@ -1,11 +1,164 @@
...
+
+    @api.model
+    def _selection_forward_thread(self):
+        # Get all models available to be selected by the user.
+        # Only consider models that support posted messages and are not transient.
+        models = (
+            self.env["ir.model"]
+            .sudo()
+            .search(
+                [("transient", "=", False), ("is_mail_thread", "=", True)],
```
pedrobaeza: Add a new field at model level for only showing those with it checked, as the list is very huge.

pedrobaeza (commented):
In a functional test, I found the word "Recipients" a bit confusing due to 2 things:

- I think it should be "Followers", as it's the usual nomenclature.
- "Followers" / "Recipient" of what? When forwarding to other thread, that's the doubt. Put "of the target document" when redirecting to other thread.

mail_forward/wizards/mail_compose_message.py
carlos-lopez-tecnativa: Done

mail_forward/models/ir_model.py
```diff
@@ -0,0 +1,7 @@
+from odoo import fields, models
+
+
+class IrModel(models.Model):
+    _inherit = "ir.model"
+
+    enable_forward_to_thread = fields.Boolean()
```
pedrobaeza:
```suggestion
    enable_forward_to = fields.Boolean(help="Enable forwarding messages to records of this model.")
```

mail_forward/readme/CONFIGURE.md
```diff
@@ -0,0 +1,5 @@
+To enable a model to forward messages to another thread:
+
+- With debug mode activated, go to `Settings -> Technical -> Database Structure -> Models`.
+- Search for the model you wish to enable. Ensure it is not a `Transient Model` and that `Has Mail Thread` is enabled.
+- Mark the option `Enable Forward to thread` and save.
```
pedrobaeza:
```suggestion
- Search for the model you wish to enable.
- Mark the option `Enable Forward To` and save. Transient models or those without mail thread won't have this check visible.
```

mail_forward/wizards/mail_compose_message.py
```diff
@@ -1,11 +1,164 @@
...
+
+    @api.model
+    def _selection_forward_thread(self):
+        # Get all models available to be selected by the user.
+        # Only consider models that support posted messages and are not transient.
+        models = (
+            self.env["ir.model"]
+            .sudo()
+            .search(
+                [("enable_forward_to_thread", "=", True)],
```
pedrobaeza: I still think you should keep the other conditions. If the flag is forced by other means (export/import, shell), we can't allow a crash at this level.

mail_forward/i18n/es.po
```diff
@@ -34,20 +44,35 @@ msgstr "Fecha"
 #. module: mail_forward
 #: model:ir.model,name:mail_forward.model_mail_thread
 msgid "Email Thread"
-msgstr "Hilo de correos electrónicos"
+msgstr "Hilo de correo electrónico"
```
pedrobaeza: It was correct before.

mail_forward/i18n/es.po
```diff
@@ -91,13 +121,18 @@ msgstr "De"
 #: code:addons/mail_forward/wizards/mail_compose_message.py:0
 #, python-format
 msgid "Fwd:"
-msgstr ""
+msgstr "Re:"
```
pedrobaeza: This translation is not correct. Maybe "RV:".

pedrobaeza (approved):

chienandalu (approved):
Cool feature. Tested :+1:

PR #1558, opened 2025-01-27
 [17.0][IMP] mail_activity_board: update user_id widget

kafai-lam:


mail_activity_board/views/mail_activity_view.xml
```diff
@@ -152,7 +159,7 @@
                 group_delete="false"
                 group_edit="false"
             >
-                <field name="user_id" />
+                <field name="user_id" widget="many2one_avatar_user" />
```
StefanRijnhart: These fields are not rendered, so I don't think it makes sense to set the widget here.

mail_activity_board/views/mail_activity_view.xml
```diff
@@ -38,7 +38,10 @@
                 <field name="related_model_instance" />
             </field>
             <field name="res_name" position="attributes">
-                <attribute name="invisible">1</attribute>
+                <attribute name="column_invisible">1</attribute>
+            </field>
+            <field name="summary" position="after">
+                <field name="user_id" widget="many2one_avatar_user" />
```
StefanRijnhart: I can't see what this does, because opening this calendar view of an activity raises
```
Error: The following error occurred in onWillStart: "value.id is undefined"
```
image

It seems there is an error rendering the widget for reference field `related_model_instance`. Don't you get this error?

mail_activity_board/views/mail_activity_view.xml
kafai-lam: remove changes for this line

mail_activity_board/views/mail_activity_view.xml
kafai-lam: I intended to change field res_name from `invisible` to `column_invisible` for tree view, not calendar view. Just fixed it

StefanRijnhart (approved):

hbrunn (approved):

StefanRijnhart:
@kafai-lam Thanks for your contribution! Would you be interested to review or test the 18.0 migration of this module, in which I included your fix?https://github.com/OCA/mail/pull/46

PR #1559, opened 2025-01-27
[MIG] migr mail show follower: Migration to 18.0

thomaspot:
MIGRATE to Version 18.
Also added a new feature to be able to set a font- and background color and for text message.


pedrobaeza:
This doesn't have the commit history, but anyway, it has to be moved to OCA/mail. It has been done in https://github.com/OCA/mail/pull/34, so closing.

thomaspot:
Do we have an overview of modules which are replaced to the mail repo?

pedrobaeza:
In principle, all starting with mail_..., which means that are modifying `mail` things. `social` is going to be the host for new modules linking to social networks (LinkedIn, X, Facebook, Instagram, etc), similar to the enterprise ones.

https://www.aeodoo.org/blog/licitaciones-abiertas-2/licitacion-para-el-desarrollo-de-aplicacion-de-marketing-social-91

thomaspot:
Ok. Thanx! We added some additional features (Background- and fontcolour). Will make a request there.

pedrobaeza:
Great, I think that can be added in a later `[IMP]` pull request. Please review to check that the migration itself is correct for merging it, and you can perform that extra PR.

PR #1560, opened 2025-01-28
[17.0][MIG] mail_layout_force: Migration to 17.0

sergioM-S73:
Migration module to 17.0


sergioM-S73:
@ivantodorovich @Abranes @hbrunn @yostashiro You can check the migration and try to merge. I appreciate it in advance

mail_layout_force/models/mail_template.py
```diff
@@ -0,0 +1,26 @@
...
+        domain=[("type", "=", "qweb"), ("mode", "=", "primary")],
+        context={"default_type": "qweb"},
+        help="Force a mail layout for this template.",
+    )
+
+    email_layout_xmlid = fields.Char(
+        compute="_compute_email_layout_id", store=True, readonly=False
+    )
+
+    def _compute_email_layout_id(self):
```
ivantodorovich: I think you should do the opposite: instead of having email_layout_xmlid be computed from force_email_layout_id.. compute force_email_layout_id from email_layout_xmlid and define an inverse method to set email_layout_xmlid

This way you guarantee they are always going to be in sync, because you don't need to store `force_email_layout_id` -- it's just an interface to read/set-up `email_layout_xmlid`

ivantodorovich: Also, please write a migration script to properly set `email_layout_xmlid` on records that had `force_email_layout_id`

ivantodorovich (changes requested):
I'm not sure I see the value for this module anymore in 17.0 🤔 
What's the point, being that you can do the same with the core `email_layout_xmlid` field now?
What's your motivation to migrate it?

mail_layout_force/models/mail_template.py
sergioM-S73: @ivantodorovich done!, thanks for the review

mail_layout_force/wizards/mail_compose_message.py
```diff
@@ -14,4 +14,4 @@ def _action_send_mail(self, auto_commit=False):
         for rec in self:
             if rec.template_id.force_email_layout_id:
                 rec.email_layout_xmlid = rec.template_id.force_email_layout_id.xml_id
-        return super(MailComposer, self)._action_send_mail(auto_commit=auto_commit)
+        return super()._action_send_mail(auto_commit=auto_commit)
```
ivantodorovich: We should drop this method entirely, as this is now already handled by core's `_compute_email_layout_xmlid`, right?

ivantodorovich (changes requested):

mail_layout_force/wizards/mail_compose_message.py
sergioM-S73: Yes, the core _compute_email_layout_xmlid method already handles this, so the override was indeed redundant. I've removed the method and the related wizard file.
Thanks!

ivantodorovich (approved):
Thanks for the work done ❤️

hbrunn:
@ivantodorovich 

> I'm not sure I see the value for this module anymore in 17.0 🤔 What's the point, being that you can do the same with the core `email_layout_xmlid` field now? What's your motivation to migrate it?

I asked myself the same question, what are the remaining use cases for this module?

AungKoKoLin1997:
@sergioM-S73 Can you please include this improvement https://github.com/OCA/social/pull/1620 in your PR?

AungKoKoLin1997 (approved):
LGTM

Reyes4711-S73 (approved):
LGTM
@OCA/social-maintainers Please, can you merge this PR?

jusle (approved):
Functional LGTM

ivantodorovich:
@sergioM-S73 can you rebase? it seems CI is failing upon merge

sergioM-S73:
@ivantodorovich I’ve done the rebase and errors have appeared in the tests of another module.

I’ve created this PR to fix it. -> https://github.com/OCA/social/pull/1801

sergioM-S73:
@ivantodorovich ready for merge?

PR #1568, opened 2025-01-30
[17.0][MIG] mail_contact_type

damdam-s:


mail_contact_type/models/res_partner.py
```diff
@@ -46,47 +46,36 @@ def contact_by_types(self, *codes):
...
             partners = self.search(
                 expression.OR(
                     [
                         [("id", "in", partners)],
-                        [("mail_contact_type_ids", "=ilike", name)],
+                        [("mail_contact_type_ids.name", "=ilike", name)],
                     ]
                 ),
                 limit=limit,
-                order=",".join(fields),
```
petrus-v: I'm wondering if we should user order=order here :thinking:

antoniospneto: Check out my alternative in #1617. It preserves the original item order, providing a solution that maintains the sequence while effectively addressing the issue.

petrus-v: just realise there is no order on name search, so the default order define on model should be good enough

petrus-v (approved):

PR #1588, opened 2025-02-12
[17.0][FIX] mail_forward: Avoid error when forwarding a message while readireading the view

carlos-lopez-tecnativa:
`ir.ui.view` requires `base.group_system` permission to read, so sudo is used instead.

TT55074
This error is also present in V15. I can backport this PR to V15 if that’s OK for you.


pedrobaeza (approved):
Straight-forward bugfix

/ocabot merge patch

PR #1493, opened 2024-11-12
[16.0][FIX] mail_composer_cc_bcc: fix duplicate key value error "unique_mail_message_id_res_partner_id_if_set"

trisdoan:
### Note
- Took over this PR: https://github.com/OCA/social/pull/1428
- depends on: https://github.com/OCA/social/pull/1519#pullrequestreview-2504965314
- Took this chance to
    - Fix `https://github.com/OCA/social/issues/1263`
       - it was fixed in `https://github.com/OCA/social/pull/1309`
       - But there is a case, described in `https://github.com/OCA/social/pull/1310#issuecomment-2186838432` , which was not covered.
    - Backport from `17.0`: using `res.partner `instead `product.template` in test. Otherwise, tests cannot run because model product.template is not in environment.

### Result

[mail_composer_cc_bcc_16.webm](https://github.com/user-attachments/assets/94096617-8331-4427-9364-ad5098ec63d9)

PR #1592, opened 2025-02-13
fix email_to_normalized: add bcc, to and cc addresses so they will pr…

### LINKED ISSUE #1591: BCC will not Propperly be Sent
https://github.com/OCA/social/blob/16.0/mail_composer_cc_bcc/models/mail_mail.py#L298

the super function is called before:

Image

so the "email_to_normalized" will not contain the bcc and cc addresses. 

Image

but here they will be removed since they are not in the validated_to 

Image

agent-z28:
Closes #1591


bizzappdev (approved):
Code and functional Testing LGTM

trisdoan (approved):

trisdoan:
Hello @agent-z28, some improvements for 16.0 might be interesting to you https://github.com/OCA/social/pull/1493

PR #1599, opened 2025-02-18
[17.0][ADD] mail_activity_cancel_tracking: New module

victoralmau:
New module

@Tecnativa TT55162


mail_activity_cancel_tracking/models/mail_activity.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def _action_done(self, feedback=False, attachment_ids=None):
+        """Add the context key to avoid sending the canceled activity email when
+        marking the activity as done.
+        """
+        self = self.with_context(skip_mail_message_activity_cancel=True)
+        return super()._action_done(feedback=feedback, attachment_ids=attachment_ids)
+
+    def _action_send_cancel_mail(self):
+        if self.env.context.get("skip_mail_message_activity_cancel"):
```
pedrobaeza: The context check should be in the affected method (`unlink`), not here.

mail_activity_cancel_tracking/models/mail_activity.py
```diff
@@ -0,0 +1,37 @@
...
+    _inherit = "mail.activity"
+
+    def _action_done(self, feedback=False, attachment_ids=None):
+        """Add the context key to avoid sending the canceled activity email when
+        marking the activity as done.
+        """
+        self = self.with_context(skip_mail_message_activity_cancel=True)
+        return super()._action_done(feedback=feedback, attachment_ids=attachment_ids)
+
+    def _action_send_cancel_mail(self):
```
pedrobaeza: And add docstring:

```suggestion
    def _log_cancel(self):
        """Method for logging a message of subtype "Activities" indicating
        that the activities in `self` have been canceled.
        """
```

mail_activity_cancel_tracking/models/mail_activity.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class MailActivity(models.Model):
+    _inherit = "mail.activity"
+
+    def _action_done(self, feedback=False, attachment_ids=None):
+        """Add the context key to avoid sending the canceled activity email when
+        marking the activity as done.
+        """
+        self = self.with_context(skip_mail_message_activity_cancel=True)
```
pedrobaeza:
```suggestion
        self = self.with_context(skip_mail_activity_cancel_log=True)
```

mail_activity_cancel_tracking/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module leaves a record in the chatter when an activity is cancelled.
```
pedrobaeza:
```suggestion
This module leaves a message in the chatter when an activity is cancelled.
```

mail_activity_cancel_tracking/readme/USAGE.md
```diff
@@ -0,0 +1,4 @@
+#. Go to a record that has activities (Contacts for example).
```
pedrobaeza:
```suggestion
#. Go to a record that allows to record activities (Contacts for example).
```

mail_activity_cancel_tracking/readme/USAGE.md
```diff
@@ -0,0 +1,4 @@
...
+#. Create an activity.
+#. Click on the "Cancel" button.
+#. A message will have been added indicating what has been canceled.
```
pedrobaeza:
```suggestion
#. A message is added in the chatter indicating the details of the activity that has been canceled.
```

pedrobaeza (changes requested):
Test in runboat correct. Please check comments inline and include tour test for both an activity recording and cancellation and activity recording and mark as done to check if there's no extra message.

victoralmau:
Changes done and tour test added.

mail_activity_cancel_tracking/static/tests/tours/mail_activity_cancel_tracking.esm.js
```diff
@@ -0,0 +1,31 @@
...
+    test: true,
+    steps: () => [
+        {
+            trigger: ".o-mail-Activity span:contains('Mark Done')",
+        },
+        {
+            trigger: ".o-mail-ActivityMarkAsDone button[aria-label='Done']",
+        },
+        {
+            trigger: ".o-mail-Message:contains('done'):contains('Play Mario Kart')",
```
pedrobaeza, line 15: I think instead doing this in the tour, at the return, you should check the messages in the chatter, and check that there's no cancel message (apart from the done one).

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

pedrobaeza:
Ouch, I have just seen https://github.com/OCA/social/pull/1597...

victoralmau:
> Ouch, I have just seen #1597...

Oops, I hadn't seen it either. What is the best way to avoid having 2 identical modules in 17.0?

pedrobaeza:
This one is already merged, so for now, let's keep it, and we can converge in this one if @etobella agrees

etobella:
We can converge, just notice that the original module on 16.0 was there from 14 and was merged 2 years ago

https://github.com/OCA/social/tree/14.0/mail_activity_unlink_log

It is true that it was merged on 16 a few days ago, so not a problem. Also, your code has more tests and a tour, so it makes sense to keep this one. Can you set the change on openupgrade in order to mark the change of name?

PR #1609, opened 2025-02-27
18.0 mig mass mailing resend

youelEl:


pedrobaeza:
Sorry for always being the bad guy, but we have created OCA/mass-mailing for hosting these modules.

pedrobaeza:
Please move this to OCA/mass-mailing, preserving commit history.

PR #1610, opened 2025-03-03
[17.0][FIX] mail_activity_cancel_tracking: Allow to delete a record (with activities)

victoralmau:
Allow to delete a record (with activities)

Example use case:
- Create a partner
- Create some activity to the partner
- Delete the partner
- The partner has been successfully deleted

Please @pedrobaeza can you review it?

@Tecnativa TT55313

### DIFF
--- a/mail_activity_cancel_tracking/models/mail_activity.py
+++ b/mail_activity_cancel_tracking/models/mail_activity.py
@@ -22,6 +22,9 @@ class MailActivity(models.Model):
             for record_sudo, activity in zip(
                 records_sudo, activity_data["activities"], strict=True
             ):
+                # Use exists() to avoid creating messages linked to deleted records
+                if not record_sudo.exists():
+                    continue
                 record_sudo.message_post_with_source(
                     "mail_activity_cancel_tracking.message_activity_cancel",
                     author_id=self.env.user.partner_id.id,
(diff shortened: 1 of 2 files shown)


mail_activity_cancel_tracking/models/mail_activity.py
```diff
@@ -18,7 +18,12 @@ def _log_cancel(self):
...
         """
         for model, activity_data in self._classify_by_model().items():
-            records_sudo = self.env[model].sudo().browse(activity_data["record_ids"])
+            # Use exists() to avoid creating messages linked to deleted records
+            records_sudo = (
+                self.env[model].sudo().browse(activity_data["record_ids"]).exists()
+            )
+            if not records_sudo:
+                continue
             for record_sudo, activity in zip(
```
pedrobaeza, line 22: I think this will give problems, as the number of elements won't match between `record_sudo` an `activity_data["activities"]`. Instead, you should do the `exists()` check inside the loop and continue if not exists.

mail_activity_cancel_tracking/tests/test_mail_activity_cancel_tracking.py
```diff
@@ -32,3 +33,8 @@ def test_mail_activity_cancel(self):
             "mail_activity_cancel_tracking_cancel",
             login="test-user",
         )
+
+    @mute_logger("odoo.models.unlink")
+    def test_record_unlink(self):
+        self.partner.unlink()
+        self.assertFalse(self.partner.exists())
```
pedrobaeza: This is not adding value. Just put a comment on the line above saying that this shouldn't fail.

pedrobaeza (approved):

victoralmau:
Ping @pilarvargas-tecnativa

pilarvargas-tecnativa (approved):

PR #1613, opened 2025-03-05
[17.0][MIG] mail_gateway: Migration to version 17.0

peluko00:


pedrobaeza:
cc @chienandalu I think there's more work to do for this to work

chienandalu:
> cc @chienandalu I think there's more work to do for this to work

Yes, the big thing is the client stuff, which hasn't been addressed in this PR yet. Our work in progrees for this same migration: https://github.com/OCA/social/pull/1614

PR #1617, opened 2025-03-12
[16.0][MIG] mail_contact_type: Migration to version 16.0

antoniospneto:


felipemotter (approved):

antoniospneto:
cc @petrus-v

mail_contact_type/models/res_partner.py
```diff
@@ -58,32 +58,32 @@ def _get_name(self):
...
+        if self.env.context.get("show_mail_contact_types") and base_count < (
+            limit or 0
+        ):
+            remaining_limit = limit - base_count if limit else False
+            extra_domain = expression.AND(
+                [args or [], [("mail_contact_type_ids", "=ilike", name)]]
+            )
+            extra_partner_ids = self.search(extra_domain, limit=remaining_limit).ids
+            partner_ids = partner_ids + [
+                pid for pid in extra_partner_ids if pid not in partner_ids
```
petrus-v: Doing this is a clever way to reduce computation. However, I have some concerns:  

- The specific order might not always be satisfied. I'm not sure we have a proper test case to ensure we don't break this behavior. For example, the **account** module [overrides the order](https://github.com/OCA/OCB/blob/16.0/addons/account/models/partner.py#L317-L324).  
- When searching by contact type, it will always hide matches of that type if there are more results than the limit in the current criteria. This can be problematic because data can sometimes be messy—users tend to add *"invoice"* in the label name, which is also the name of the tag. As a result, new partners that don’t follow this convention might be hidden, even though users would expect them to appear first in the given order.  

Hope this makes sense.

antoniospneto: @petrus-v Indeed, I was mistaken about this; there was really no need for it.I have looked into this matter more deeply, and now everything seems to be in order.

Note: I made sure to keep the "args" from the super call in the second search to avoid losing the default security domains.

PR #1517, opened 2024-12-05
[16.0][FIX] mail_show_follower: Company identification in _send()

nobuQuartile:
Before this commit, 'Show Internal Users CC' would always be False, regardless of the setting in the company, in case the message came from a record with a company_id field but without a value assignment to it (e.g., a partner record with no company), as company would be an empty record.

This commit fixes the behavior by capturing company from self.env in these cases.

@qrtl QT5021

PR #1621, opened 2025-03-19
[17.0][FIX] mail_show_follower: Company identification in _send()

nobuQuartile:
This is a forward port of the PR below.
https://github.com/OCA/social/pull/1517

Before this commit, 'Show Internal Users CC' would always be False, regardless of the setting in the company, in case the message came from a record with a company_id field but without a value assignment to it (e.g., a partner record with no company), as company would be an empty record.

This commit fixes the behavior by capturing company from self.env in these cases.

@qrtl QT5021


yostashiro (approved):

AungKoKoLin1997 (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1624, opened 2025-03-26
[16.0] [MIG] mail_alias_with_domain

NL66278:
Make it possible to handle incoming mails with aliases that take the smtp domain of the incoming mail into account. So you can have mail addresses like invoice@my_one_company.com and invoice@my_other_company.com and link them to records in the right company.


NL66278:
Test fail on another module. Can be corrected after merge of this PR: https://github.com/OCA/social/pull/1628

mail_alias_with_domain/models/mail_thread.py
```diff
@@ -0,0 +1,49 @@
...
+            custom_values=custom_values,
+        )
+
+    def _find_alias_with_domain(self, message_dict):
+        """Find all aliasses that match."""
+        Alias = self.env["mail.alias"]
+        emails = {email for email in (tools.email_split(message_dict["recipients"]))}
+        alias_names = []
+        for email in emails:
+            clean_email = Alias.get_clean_email(email)
```
hbrunn, line 44: in which cases does email_split return something else than an email address?

mail_alias_with_domain/models/mail_alias.py
```diff
@@ -0,0 +1,101 @@
...
+        for alias in alias_with_domain:
+            alias.alias_domain = alias.alias_name.split("__at__")[1]
+        alias_without_domain = self - alias_with_domain
+        if alias_without_domain:
+            super(Alias, alias_without_domain)._compute_alias_domain()
+        return None
+
+    alias_entry = fields.Char(
+        help="This will be used to enter an email, complete with domain",
+    )
```
hbrunn, line 25: this should be unique

mail_alias_with_domain/models/mail_alias.py
```diff
@@ -0,0 +1,101 @@
...
+from odoo import api, fields, models
+
+
+class Alias(models.Model):
+    _inherit = "mail.alias"
+
+    @api.depends("alias_name")
+    def _compute_alias_domain(self):
+        alias_with_domain = self.filtered(
+            lambda r: r.alias_name and "__at__" in r.alias_name
```
hbrunn, line 14: I don't find it super far fetched to have this in a mail address. did you consider using a character illegal in both localparts and domain names, like '&'?

NL66278: @hbrunn Taken up most of your remarks in new PR here: https://github.com/OCA/social/pull/1797

Only not this one. Using `&` led to other problems, and apparently is actually allowed also, just like `_` and most printable characters. At least according to this site: https://www.ditig.com/characters-allowed-in-email-addresses.  But even if `__at__` is used in some email address - never encountered anything like it -, it will be extremely unlikely that this also will coincide with en email address used as an alias.

PR #1626, opened 2025-04-03
[16.0][IMP] mail_activity_done - make date_done datetime

enriquemartin:


hbrunn (changes requested):
I don't see which problem this solves? And in addition to that, with the datetime type you introduce all the issues coming with timezones

PR #1627, opened 2025-04-07
[16.0][IMP] base_search_mail_content: Use the GIN index correctly

carlos-lopez-tecnativa:
The purpose of the GIN index is for full-text search, but without specifying the operator, Odoo does not use the index and instead searches using LIKE. This commit explicitly adds the % operator.

TT51245

### DIFF
--- a/base_search_mail_content/models/mail_thread.py
+++ b/base_search_mail_content/models/mail_thread.py
@@ -52,7 +52,9 @@ class MailThread(models.AbstractModel):
             doc = etree.XML(res["arch"])
             for node in doc.xpath("/search/field[last()]"):
                 # Add message_content in search view
-                elem = etree.Element("field", {"name": "message_content"})
+                elem = etree.Element(
+                    "field", {"name": "message_content", "operator": "%"}
+                )
                 node.addnext(elem)
                 res["arch"] = etree.tostring(doc)
         return res
--- a/base_search_mail_content/readme/ROADMAP.rst
+++ b/base_search_mail_content/readme/ROADMAP.rst
@@ -2,3 +2,6 @@ This module restricts the message_content search functionality
 to internal users only, addressing the issue faced by project collaborators (portal users)
 as described in https://github.com/OCA/social/issues/1204. Consequently, portal users no
 longer have the ability to search within mail content.
+
+This module's usability is severely limited for languages that don't separate words with spaces (e.g., Chinese, Japanese, Korean, Thai, etc.), 
+because `pg_trgm` needs three-character tokens and a similarity score above the default cutoff
--- a/base_search_mail_content/README.rst
+++ b/base_search_mail_content/README.rst
@@ -77,6 +77,9 @@ to internal users only, addressing the issue faced by project collaborators (por
 as described in https://github.com/OCA/social/issues/1204. Consequently, portal users no
 longer have the ability to search within mail content.
 
+This module's usability is severely limited for languages that don't separate words with spaces (e.g., Chinese, Japanese, Korean, Thai, etc.), 
+because `pg_trgm` needs three-character tokens and a similarity score above the default cutoff
+
 Bug Tracker
 ===========
 
--- a/base_search_mail_content/static/description/index.html
+++ b/base_search_mail_content/static/description/index.html
@@ -8,10 +8,11 @@
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 8954 2022-01-20 10:10:25Z milde $
+:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
 :Copyright: This stylesheet has been placed in the public domain.
 
 Default cascading style sheet for the HTML output of Docutils.
+Despite the name, some widely supported CSS2 features are used.
 
 See https://docutils.sourceforge.io/docs/howto/html-stylesheets.html for how to
 customize this style sheet.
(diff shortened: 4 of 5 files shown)


pedrobaeza (approved):
Again the problem with the method signature in other modules...

carlos-lopez-tecnativa:
ping @pedrobaeza

base_search_mail_content/models/mail_thread.py
```diff
@@ -52,7 +52,9 @@ def get_view(self, view_id=None, view_type="form", **options):
             doc = etree.XML(res["arch"])
             for node in doc.xpath("/search/field[last()]"):
                 # Add message_content in search view
-                elem = etree.Element("field", {"name": "message_content"})
+                elem = etree.Element(
+                    "field", {"name": "message_content", "operator": "%"}
```
yostashiro, line 56: This seems to break the functionality of searches with double-byte characters (tested on the runboat build for this PR).

image
image

yostashiro (changes requested):
Instead, rewriting `_search_message_content()` in the following manner seems to achieve the desired outcome. Can you please check this approach?
```.py
    def _search_message_content(self, operator, value):
        fields = ["record_name", "subject", "body", "email_from", "reply_to"]
        words  = value.split()
        word_domain_list = []
        for word in words:
            field_domain_list = [[(field, operator, word)] for field in fields]
            if operator in expression.NEGATIVE_TERM_OPERATORS:
                word_domain_list.append(expression.AND(field_domain_list))
            else:
                word_domain_list.append(expression.OR(field_domain_list))
        word_domain = expression.AND(word_domain_list)
        model_domain = expression.AND([[("model", "=", self._name)], word_domain])
        messages = self.env["mail.message"].search(model_domain)
        return [("id", "in", messages.mapped("res_id"))]
```

carlos-lopez-tecnativa:
Hi @yostashiro 
Sorry, but this change is out of scope. If I understand you correctly, you expect Odoo to search using each word entered individually. In my opinion, this is not correct.

Please note that this module uses a `GIN` index. My intention is only to enable the search to correctly use that index, which currently doesn't happen because it's using the `LIKE` operator instead of the `%` (full-text search) operator.

Once the module uses the correct operator, the `GIN` index will support `full-text search`. For example, if a message contains "Hello World", you’ll be able to search using "world hell", see the test case for reference.

Note: Please attach a screenshot with the message in English. I couldn't fully understand what it says.

yostashiro:
@carlos-lopez-tecnativa I understand your intention. It's just that `pg_trgm` doesn't seem to work well with CJK characters, so your implementation will make the module unusable for the users in our market.

> For example, if a message contains "Hello World", you’ll be able to search using "world hell", see the test case for reference.

The code I suggest should function the same way, but I guess it's not as performant as using a GIN index.

I think it would be nice to make the search method configurable. Would you consider doing it in this PR? If not, we could create a PR to take care of that.

> Note: Please attach a screenshot with the message in English. I couldn't fully understand what it says.

I don't think you need to understand what it says. I was just showcasing that the search didn't work with Japanese characters. :)

pedrobaeza:
Yeah, `pg_trm` doesn't work with Japanese characters, but having that this module depends on `base_search_fuzzy`, which bases all its feature in trigram indexes, isn't better to just create a new module `mail_search_content_by_word` or similar with just that specific code? Note that there is [base_name_search_improved](https://odoo-community.org/shop/improved-name-search-666?page=2&search=search#attr=939481) that does something similar.

yostashiro:
@pedrobaeza Thanks, we could go as you suggest. However, IMO, the truly optimal design would be to remove the dependency on `base_search_fuzzy` from `base_search_mail_content`, and instead create a separate module such as `mail_search_content_trgm`  which depends on `base_search_fuzzy` and `base_search_mail_content`. If we go with creating `mail_search_content_by_word`, the majority of the code will be a copy of `base_search_mail_content`.

pedrobaeza:
Well, I'm not sure there's so much code to share between both, but in any case, that should be done on a version where the module didn't land yet, not in this one.

yostashiro:
Technically, I still lean toward my suggestion, as it preserves existing behavior, whereas this PR introduces breaking changes for CJK users, which arguably goes against the intent of the stable policy.

That said, I’m fine with proceeding with this PR, considering the limited contribution from the CJK market. How about adding information to the ROADMAP file?

carlos-lopez-tecnativa:
> That said, I’m fine with proceeding with this PR, considering the limited contribution from the CJK market. How about adding information to the ROADMAP file?

I updated the roadmap. Please let me know if it's OK for you. Thanks.

carlos-lopez-tecnativa:
ping @pedrobaeza @yostashiro

pedrobaeza:
We need @yostashiro's approval

base_search_mail_content/readme/ROADMAP.rst
```diff
@@ -2,3 +2,5 @@ This module restricts the message_content search functionality
 to internal users only, addressing the issue faced by project collaborators (portal users)
 as described in https://github.com/OCA/social/issues/1204. Consequently, portal users no
 longer have the ability to search within mail content.
+
+This module does not work correctly when using the GIN index with Japanese characters.
```
yostashiro: Sorry it took me a while to check. From what I understand, the issue is related to how the trigram mechanism works and is not directly caused by the GIN index itself.

```suggestion
This module's usability is severely limited for languages that don't separate words with spaces (e.g., Chinese, Japanese, Korean, Thai, etc.), because `pg_trgm` needs three-character tokens and a similarity score above the default cutoff.
```

yostashiro (approved):
LGTM. Thanks for the update.

PR #1631, opened 2025-04-10
[16.0][IMP] mass_mailing_partner: display list of duplicate e-mails

remi-filament:
When you try to sync a dynamic list and that would create duplicate contacts, you get a validation error.
On big lists, it is not so easy to find out which partners / e-mails are causing this ValidationError.

This PR proposes to list e-mails causing this ValidationError in the message displayed to user, so that partners could be deduplicated or dynamic list domain could be reworked.

### DIFF
--- a/mass_mailing_partner/models/mailing_list.py
+++ b/mass_mailing_partner/models/mailing_list.py
@@ -28,7 +28,20 @@ class MailingList(models.Model):
                 ["partner_id"],
                 ["partner_id"],
             )
-            if len(list(filter(lambda r: r["partner_id_count"] > 1, data))):
+            duplicates = list(filter(lambda r: r["partner_id_count"] > 1, data))
+            if len(duplicates):
+                duplicates_list = "\n".join(
+                    [
+                        f'({dup["partner_id_count"]}) [{dup["partner_id"][0]}] \
+                        {str(dup["partner_id"][1])}'
+                        for dup in duplicates
+                    ]
+                )
                 raise ValidationError(
-                    _("A partner cannot be multiple times " "in the same list")
+                    _(
+                        "A partner cannot be multiple times in the same list\n"
+                        "Duplicate contacts found:\n"
+                        "%s",
+                        duplicates_list,
+                    )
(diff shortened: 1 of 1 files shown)


mass_mailing_partner/models/mailing_list.py
```diff
@@ -28,7 +28,17 @@ def _check_contact_ids_partner_id(self):
...
+                for partner in duplicates:
+                    partner_ids.append(partner["partner_id"][0])
+                partners = self.env["res.partner"].browse(list(set(partner_ids)))
                 raise ValidationError(
-                    _("A partner cannot be multiple times " "in the same list")
+                    _(
+                        "A partner cannot be multiple times in the same list\n"
+                        "Look for the following e-mail(s) :\n"
+                        " - %s",
+                        "\n - ".join(partners.mapped("email")),
```
chienandalu: Wouldn't it be interesting to have the partner info and number of times the partner is duplicated? (id + name + email, partner_id_count)

remi-filament: I am not sure, since you can easily search by email in mailing list contacts.

chienandalu: Different partners could share the same email (not ideal, but it happens) I mean something like this:
```python
            duplicates = list(filter(lambda r: r["partner_id_count"] > 1, data))
            if len(duplicates):
                duplicates_list = "\n".join([
                    f'({dup["partner_id_count"]}) [{dup["partner_id"][0]}] {str(dup["partner_id"][1])}'
                    for dup in duplicates
                ])
                raise ValidationError(
                    _(
                        "A partner cannot be multiple times in the same list\n"
                        "Duplicate contacts found:\n"
                        "%s",
                        duplicates_list
                    )
                )

remi-filament: Done, thank you @chienandalu

chienandalu (approved):
Thanks :)

MiquelRForgeFlow:
:(

Issue #1630, opened 2025-04-10
mail_tracking_mailgun: Enable option to send emails via API


JordiBForgeFlow:
Some cloud providers such as Digital Ocean are blocking the SMTP ports, and advising their customers to deliver emails via API https://docs.digitalocean.com/support/why-is-smtp-blocked

We are willing to implement the API offered by Mailgun to send emails instead of SMTP.

pedrobaeza:
It's a good option, but I wouldn't put it inside `mail_tracking_mailgun`, as it's goal is another and contains a lot of other things (and dependency to `mail_tracking`), which is huge.

I think you can develop a module `mail_mailgun` to do that, and recopy the needed fields. On the new version, we can converge having a base module `mailgun_api`, that both `mail_tracking_mailgun` and `mail_mailgun` depend on.

What do you think?

PR #1634, opened 2025-04-11
[16.0][ADD] mail_mailgun

ArnauCForgeFlow:
This module provides an alternative to SMTP by using Mailgun's API for email delivery.

Related issue: https://github.com/OCA/social/issues/1630


mail_mailgun/models/mail_mail.py
```diff
@@ -0,0 +1,37 @@
...
+                    raise_exception=raise_exception,
+                    smtp_session=False,
+                )
+                _logger.info(
+                    "Sent batch %s emails via mail server ID #%s",
+                    len(batch_ids),
+                    mail_server_id,
+                )
+            else:
+                return super().send(
```
hbrunn, line 35: won't this cause duplicates in case there are actually multiple non-mailgun servers involved?

why do you need this override at all? I think just returning None in ir.mail.server#connect for mailgun servers should suffice?

hbrunn (changes requested):
I think this absolutely needs unittests

PR #1636, opened 2025-04-16
[16.0][FIX] mass_mailing_list_dynamic: prevent put in queue with errors

chienandalu:
A dynamic mailing list syncing could have errors due to constraints or other side effects. But as the syncing is being made by the cron job, the user can't detect and fix (or report) those errors.

So we'll just sync the lists when the user launches the mailing and prior to put it in the mailing queue.

cc @moudon @yajo @fcvalgar MT-9835


yajo (approved):
Makes sense, thanks!

fcvalgar (approved):
LGTM, thank you @chienandalu

chienandalu:
@yajo @rafaelbn please merge

PR #1639, opened 2025-04-17
[17.0][FIX] mail_tracking_mailgun: Markup link

pedrobaeza:
Correct link message on the chatter, using also the provided native method instead of building it manually.

Before:
imagen

After:
imagen

@Tecnativa


carlos-lopez-tecnativa (approved):

pedrobaeza:
As the patch is clear and solves the problem, fast-tracking it:

/ocabot merge patch

PR #1640, opened 2025-04-21
[16.0][FIX] mass_mailing_partner: avoid changing partner unnecessarily

chienandalu:
When a contact is written, this module checks that the partner's email fits the contact one. In a scenario where several partner share the same email address we could end having that partner changed depending on the searching order prevalence.

cc @moduon @yajo @fcvalgar MT-9835

### DIFF
--- a/mass_mailing_partner/models/mailing_contact.py
+++ b/mass_mailing_partner/models/mailing_contact.py
@@ -104,6 +104,8 @@ class MailingContact(models.Model):
         m_partner = self.env["res.partner"]
         # Look for a partner with that email
         email = self.email.strip()
+        if email == self.partner_id.email:
+            return
         partner = m_partner.search([("email", "=ilike", email)], limit=1)
         if partner:
             if partner == self.partner_id:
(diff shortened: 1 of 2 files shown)


fcvalgar (approved):
Functional review,

LGTM, Thank you @chienandalu 


image

image

image

mass_mailing_partner/models/mailing_contact.py
```diff
@@ -106,7 +106,7 @@ def _set_partner(self):
         email = self.email.strip()
         partner = m_partner.search([("email", "=ilike", email)], limit=1)
```
yajo, line 109: suggestion: you can improve performance if you move the new check up here. You'll save some queries:

```suggestion
        if email == self.partner_id.email:
            return
        partner = m_partner.search([("email", "=ilike", email)], limit=1)
```

If so, remember to leave line 109 as it was.

yajo (approved):

mass_mailing_partner/models/mailing_contact.py
chienandalu: Sure, that's better. Done :slightly_smiling_face:

rafaelbn (approved):
/ocabot merge patch

PR #1644, opened 2025-05-10
[16.0][FIX] mail_compose_cc_bcc allow to send cc to follower

DrHohol:
This PR fixes the following error that occurs when a contact who is already a follower of a record (e.g. sale.order) is also added as a CC recipient when sending a message:

Example use case:

1. Create sale order
2. Add contact as follower
3. Click send by email
4. Select the same contact in the CC field
Result before fix:
Error is raised due to a duplicate entry in mail.notification: `The Operation cannot be completed: duplicate key value violates unique constraint "unique_mail_message_id_res_partner_id_if_set"`
Result after fix:
Email is sent successfully, and the notification is created only once.

### DIFF
--- a/mail_composer_cc_bcc/tests/test_mail_cc_bcc.py
+++ b/mail_composer_cc_bcc/tests/test_mail_cc_bcc.py
@@ -23,6 +23,7 @@ VALID_HASHES = [
     "8f26c4084cc7fc300e64d19ccdc944fe",
     "db6cc0d3513a0c85bd716e4cb0a4d09c",
     "458982c6cb3a347b13008f7d8130f633",  # 2025-04-08, odoo commit 3af276804101
+    "2e0c730330ae3bdf00c6bfd956b0527b",
 ]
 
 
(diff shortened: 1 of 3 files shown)


trisdoan (approved):
LGTM! Thanks

mail_composer_cc_bcc/tests/test_mail_cc_bcc.py
```diff
@@ -23,6 +23,7 @@
     "8f26c4084cc7fc300e64d19ccdc944fe",
     "db6cc0d3513a0c85bd716e4cb0a4d09c",
     "458982c6cb3a347b13008f7d8130f633",  # 2025-04-08, odoo commit 3af276804101
+    "2e0c730330ae3bdf00c6bfd956b0527b",
```
NL66278, line 26: Might be helpful to have date and commit hash of this addition as well.

NL66278 (approved):
:+1: One non-blocking request, but looks good to me.

PR #1647, opened 2025-05-14
[16.0][MIG] mail_follower_custom_notification

hbrunn:
In standard Odoo, receiving mail notifications is an all or nothing affair.
This module allows users to decide if they want to receive emails, inbox notifications or both, as well globally as per followed record.

Administrators can also override this settings per message subtype.

Configuration
=============

You can add defaults per message subtype for this settings in Settings /
Technical / Email / Subtypes.

Usage
=====

When followers edit their subscriptions, they will be offered the choice to
override notification settings.

Note subscriptions are only editable for users of the `Technical settings`
group.


hbrunn:
test failure squelched in https://github.com/OCA/social/pull/1648

PR #1649, opened 2025-05-14
[IMP] mail_layout_force: enable in survey emails

kobros-tech:
to force layouts for survey emails.

@wlin-kencove

### DIFF
--- a/mail_layout_force/__manifest__.py
+++ b/mail_layout_force/__manifest__.py
@@ -11,7 +11,7 @@
     "website": "https://github.com/OCA/social",
     "license": "AGPL-3",
     "category": "Marketing",
-    "depends": ["mail"],
+    "depends": ["mail", "survey"],
     "demo": ["demo/mail_layout.xml"],
     "data": [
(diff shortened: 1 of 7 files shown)


mail_layout_force/__manifest__.py
```diff
@@ -11,7 +11,7 @@
     "website": "https://github.com/OCA/social",
     "license": "AGPL-3",
     "category": "Marketing",
-    "depends": ["mail"],
+    "depends": ["mail", "survey"],
```
ivantodorovich, line 14: Please don't add the `survey` dependency to this module.

Instead, create a new glue module (possibly with auto_install: True)
e.g.: `mail_layout_force_survey` / `survey_mail_layout_force`

ivantodorovich (changes requested):

mail_layout_force/__manifest__.py
kobros-tech: no need to merge the PR by the way.

PR #1653, opened 2025-05-18
[16.0][FIX] mass_mailing_unique: constraint message

kobros-tech:


mass_mailing_unique/models/mailing_contact.py
```diff
@@ -12,6 +12,10 @@ class MailingContact(models.Model):
         (
             "unique_email",
             "UNIQUE(email_normalized)",
-            "There's already a contact with this email address",
+            (
+                "There's already an existing 'mailing.contact' record"
```
pedrobaeza: This message should be more user oriented:

```suggestion
                "There's already an existing mailing contact"
```

kobros-tech: I will commit as soon as possible, but maybe later due to some reason.

pedrobaeza (approved):

hbrunn (approved):

Issue #1652, opened 2025-05-18
mass_mailing_partner & mass_mailing_unique


kobros-tech:
If these two modules are installed you will suffer from SQL unique constraint for mailing.contact model whether you have exception message can pop up in res.partner form.

As the message is not descriptive the user will think the exception is inside res.partner model whether it is actually inside mailing.contact model!

Also inside the mailing.contact model if I update the partner I should get its related email updated, but in case the same email is assigned for other mailing.contact record, the exception will not work and you simply find the record updated the partner and its related fields **but the emal field will become empty**.

I fixed both bugs, and will push their fix in 2 PRs in sequence.

My technical advice:

Don't use any _onchange_foo method ever or call it from any where, it should be standalone and used inside its onchange decorator.

If you need create a method to be called inside onchange method and inside other places.

If you need to have a readonly field and use a method to change its value, use a python constraint method but don't use onchange method as the feld will rrefuse changing its value from onchange.

combining onchange + write method is bad and will result it unnecessary loop that we don't like!

https://github.com/OCA/social/pull/1653
https://github.com/OCA/social/pull/1654

kobros-tech:
https://github.com/user-attachments/assets/f2fe470b-0707-4df6-852e-4c264ef5f9ee

kobros-tech:
https://github.com/user-attachments/assets/c1972e7e-9a3d-4aef-8dac-7ea5dba297ec

kobros-tech:
https://github.com/user-attachments/assets/e59d3066-91f3-4602-a6fe-f895cf1a3418

kobros-tech:
@AlexPForgeFlow

I ping you because we discussed a similar issue #https://github.com/OCA/sign/pull/79 for account_avatax_exemption_sign_oca and were not sure.

It's a simiar cause, hoping we change the technique of updating values of readonly fields.

I also welcome your review if you would like.

kobros-tech:
@wlin-kencove
@dnplkndll

@pedrobaeza 
@etobella

kobros-tech:
@pedrobaeza 

This is the goal of this PR #1654

PR #1654, opened 2025-05-18
[16.0][FIX] mass_mailing_partner: control readonly field

kobros-tech:


mass_mailing_partner/models/mailing_contact.py
```diff
@@ -42,16 +42,20 @@ def _check_partner_id_list_ids(self):
...
-    def _onchange_partner_mass_mailing_partner(self):
-        if self.partner_id:
-            self.name = self.partner_id.name
-            self.email = self.partner_id.email
-            self.title_id = self.partner_id.title
-            self.company_name = (
-                self.partner_id.company_id.name or self.partner_id.company_name
-            )
-            self.country_id = self.partner_id.country_id
+    def update_fileds_based_on_partner(self):
```
pedrobaeza:
```suggestion
    def update_fields_based_on_partner(self):
```

mass_mailing_partner/models/mailing_contact.py
```diff
@@ -42,16 +42,20 @@ def _check_partner_id_list_ids(self):
...
+                rec.email = rec.partner_id.email
+                rec.title_id = rec.partner_id.title
+                rec.company_name = (
+                    rec.partner_id.company_id.name or rec.partner_id.company_name
+                )
+                rec.country_id = rec.partner_id.country_id
+
+    @api.constrains("partner_id")
+    def _check_partner_mass_mailing_partner(self):
+        self.update_fileds_based_on_partner()
```
pedrobaeza: This shouldn't be done in a constraint.

pedrobaeza (changes requested):

mass_mailing_partner/models/mailing_contact.py
kobros-tech: I did not understand, what is needed here?

mass_mailing_partner/models/mailing_contact.py
kobros-tech: I wrote about this in an issue, we couldn't also use _onchange outside its parent decorator.

specially with readonly fields.


Could you give me a cause not to use constrains, apart from its official purpose?

I know 1 side effect which is not real life effect, it will appear after saving only.

but we could overcome the side effect by using _onchange (to show instant updates, but the one will really save is the constraint)

mass_mailing_partner/models/mailing_contact.py
pedrobaeza: See the proposal. There's a typo in the method name.

mass_mailing_partner/models/mailing_contact.py
pedrobaeza: The solution is to add `force_save="1"` in the readonly field. That's it. No need of these hacks.

kobros-tech: I didn't know this before 

If I get a link of reference would be great, 👍

pedrobaeza: https://www.odoo.com/es/forum/ayuda-1/readonly-field-in-onchange-method-not-store-value-when-save-148740
https://dev.to/jeevanizm/leveraging-forcesave-in-odoo-for-boolean-fields-dependent-on-onchange-methods-1p7c

mass_mailing_partner/models/mailing_contact.py
```diff
@@ -42,16 +42,20 @@ def _check_partner_id_list_ids(self):
...
+    def update_fields_based_on_partner(self):
+        for rec in self:
+            if rec.partner_id:
+                rec.name = rec.partner_id.name
+                rec.email = rec.partner_id.email
+                rec.title_id = rec.partner_id.title
+                rec.company_name = (
+                    rec.partner_id.company_id.name or rec.partner_id.company_name
+                )
+                rec.country_id = rec.partner_id.country_id
```
pedrobaeza: To reduce one complexity level:

```suggestion
        for rec in self.filtered("partner_id"):
              rec.name = rec.partner_id.name
              rec.email = rec.partner_id.email
              rec.title_id = rec.partner_id.title
              rec.company_name = (
                  rec.partner_id.company_id.name or rec.partner_id.company_name
              )
              rec.country_id = rec.partner_id.country_id
```

mass_mailing_partner/models/mailing_contact.py
```diff
@@ -72,12 +76,6 @@ def create(self, vals_list):
             contact._overwrite_partner(vals, True)
         return result
 
-    def write(self, vals):
```
pedrobaeza, line 75: Why are you removing this?

kobros-tech:
@pedrobaeza 

hi, is it intended to be merged so I can re create the PR?

pedrobaeza:
Ok, please attend comments.

kobros-tech:
sure

kobros-tech:
@pedrobaeza 

please see if I am done with what you request?

mass_mailing_partner/models/mailing_contact.py
pedrobaeza: This question remains unanswered. And I don't get the goal of the overall patch.

kobros-tech:
@pedrobaeza 

I restored the write method.

pedrobaeza:
And can you explain the goal of this PR?

kobros-tech:
@pedrobaeza 

After reviewing the PR, I guess I was intending to use a specic method inside compute method and in other places because I saw errors are following compute method if it is called in other places.

This is what I guess, because I don't remember exactly, forgive my brain.

kobros-tech:
I should have added the error I propose here from the beginning, or maybe I raised an issue I don't remember.

kobros-tech:
Hi @pedrobaeza 

I found the issue for this PR

https://github.com/OCA/social/issues/1652

mass_mailing_partner/views/mailing_contact_view.xml
```diff
@@ -30,31 +30,37 @@
                 <attribute
                     name="attrs"
                 >{'readonly': [('partner_id', '!=', False)]}</attribute>
+                <attribute name="force_save">1</attribute>
             </field>
             <field name="email" position="attributes">
                 <attribute
                     name="attrs"
-                >{'readonly': [('partner_id', '!=', False)]}</attribute>
+                >{'readonly': [('partner_id', '!=', False), ('email', '!=', False)]}</attribute>
```
pedrobaeza, line 38: Then AFAIK, the changes of this file are the only one affected. The rest is non needed refactoring.

kobros-tech:
@pedrobaeza 

I am going to PR what you ask for in a new PR, this one will be closed.

PR #1657, opened 2025-05-26
[16.0][IMP] email_template_qweb: copy view together with template

NL66278:
Also edit the copied template inline to enhance the usability of the module. This is done with visible language selection.

This commits prevents two common errors:
- After copy and editing the qweb template from the copy, the original template is changed (as there was no copied template);
- The user edits the template, which by default shows en_US, but nothing really changes in the customer language, leading to 'weeping and gnashing of teeth'.

### DIFF
--- a/email_template_qweb/__manifest__.py
+++ b/email_template_qweb/__manifest__.py
@@ -1,4 +1,4 @@
-# Copyright 2016 Therp BV <http://therp.nl>
+# Copyright 2016,2025 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "QWeb for email templates",
--- a/email_template_qweb/models/mail_template.py
+++ b/email_template_qweb/models/mail_template.py
@@ -1,6 +1,6 @@
-# Copyright 2016 Therp BV <http://therp.nl>
+# Copyright 2016,2025 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import fields, models, tools
+from odoo import _, api, fields, models, tools
 
 
 class MailTemplate(models.Model):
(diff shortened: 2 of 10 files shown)


email_template_qweb/models/mail_template.py
```diff
@@ -1,19 +1,51 @@
...
+            this.body_view_arch = this.body_view_id.with_context(
+                lang=this.edit_language
+            ).arch
+
+    def _inverse_body_view_arch(self):
+        for this in self:
+            this.body_view_id.with_context(
+                lang=this.edit_language
+            ).arch = this.body_view_arch
+
```
hbrunn: those go below the field definitions, cf https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#37models

email_template_qweb/models/mail_template.py
```diff
@@ -1,19 +1,51 @@
...
-    body_view_arch = fields.Text(related="body_view_id.arch")
+    body_view_arch = fields.Text(
+        compute="_compute_body_view_arch",
+        inverse="_inverse_body_view_arch",
+        readonly=False,
+    )
+    edit_language = fields.Selection(
+        selection="_get_edit_language_selection",
+        default="en_US",
+        help="Set the language to edit template",
```
hbrunn: I'd suggest to make this nonstored and default to the interface language

email_template_qweb/models/mail_template.py
```diff
@@ -50,3 +82,26 @@ def generate_email(self, res_ids, fields):
...
+                # There is a body_view_id that can be copied.
+                body_default = dict(
+                    name=_("%s (copy)", self.body_view_id.name or ""),
+                )
+                new_view = self.body_view_id.copy(default=body_default)
+                default = dict(
+                    default or {},
+                    body_view_id=new_view.id,
+                )
+                default["body_view_id"] = new_view.id
```
hbrunn:
```suggestion
```

email_template_qweb/views/mail_template.xml
```diff
@@ -9,14 +9,21 @@
...
+                    />
                     <field
                         name="body_view_id"
                         attrs="{'required': [('body_type', '=', 'qweb_view')]}"
                     />
                     <field
                         name="body_view_arch"
                         widget="ace"
-                        attrs="{'required': [('body_type', '=', 'qweb_view')], 'invisible': [('body_view_id', '=', False)]}"
+                        context="{'lang': edit_language}"
```
hbrunn, line 23: do you need this given the compute and inverse functions already set it?

hbrunn (changes requested):
I'd have preferred those quite useful but very distinct features to be two commits

email_template_qweb/__manifest__.py
```diff
@@ -1,8 +1,8 @@
-# Copyright 2016 Therp BV <http://therp.nl>
+# Copyright 2016,2025 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "QWeb for email templates",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.1.0",
```
hbrunn: it's usually better to have ocabot do this on merge time

email_template_qweb/models/mail_template.py
NL66278: I have never seen a nonstored selection field. Do you have any example?

PR #1660, opened 2025-05-28
[FIX]: remove duplicate message in mail_tracking

rbtkay:


StephaneMangin (approved):
Quick Win !

alexandregaldeano (commented):
No updates on the translation files?

mail_tracking/models/mail_tracking_email.py
```diff
@@ -43,7 +43,7 @@ class MailTrackingEmail(models.Model):
         string="Date", readonly=True, compute="_compute_date", store=True
     )
     mail_message_id = fields.Many2one(
-        string="Message", comodel_name="mail.message", readonly=True, index=True
+        string="Mail Message", comodel_name="mail.message", readonly=True, index=True
```
petrus-v: just to let you know in such case it would be the default string generated by odoo so we could do

```suggestion
        comodel_name="mail.message", readonly=True, index=True
```

cf: https://github.com/odoo/odoo/blob/14.0/odoo/fields.py#L398-L403
```python
        if not self.string and not self.related:
            # related fields get their string from their parent field
            self.string = (
                name[:-4] if name.endswith('_ids') else
                name[:-3] if name.endswith('_id') else name
            ).replace('_', ' ').title()
```

petrus-v (approved):

mail_tracking/models/mail_tracking_email.py
rbtkay: Nice! I did not catch that. Thank you for the heads up 🙏 
I'm applying this suggestion as I don't think an explicit label is required here.

petrus-v (approved):

StephaneMangin:
You should maybe increment the version of the module to 14.0.3.2.1

StephaneMangin (approved):

damdam-s (approved):

alexandregaldeano (approved):
LGTM 🔥

hbrunn:
> You should maybe increment the version of the module to 14.0.3.2.1

no, our bot increases version numbers automatically, cf https://github.com/OCA/oca-github-bot?tab=readme-ov-file#commands

PR #1669, opened 2025-06-12
[17.0][FIX] mail_activity_team: open team activities from systray

StefanRijnhart:
_To Reproduce_

1. Create an activity for a team of the current user but with no specific user of for another user in the team.
2. Go to the systray activity widget. Switch to tab Team Activities.
3. Click on `1 Future`.

_Actual outcome_

The activity kanban view opens without any records.

_Expected_ behavior_

The activity kanban view opens with the team activity that was created in step 1.

_Technical description of the fix_

The systray widget will open the view with a domain that selects the users's activities. Thanks to a context key that is already set, we can intercept the domain and replace it with a domain on the user's teams' activities.

Backported from https://github.com/OCA/mail/pull/38

### DIFF
--- a/mail_activity_team/models/mail_activity_mixin.py
+++ b/mail_activity_team/models/mail_activity_mixin.py
@@ -98,3 +98,27 @@ class MailActivityMixin(models.AbstractModel):
                 ),
                 False,
             )
+
+    @api.model
+    def web_search_read(
+        self, domain, specification, offset=0, limit=None, order=None, count_limit=None
+    ):
+        # Intercept the queries from the systray widget and apply team search.
+        # This allows the frontend widget to direct the user to their team activities
+        # Late/Today/Future.
+        if self.env.context.get("team_activities"):
+            orig_domain = domain
+            domain = []
+            for clause in orig_domain:
+                if isinstance(clause, list | tuple) and clause[0] == "activity_user_id":
+                    domain.append(("activity_team_user_ids", clause[1], clause[2]))
+                else:
+                    domain.append(clause)
+        return super().web_search_read(
+            domain,
+            specification,
+            offset=offset,
+            limit=limit,
+            order=order,
+            count_limit=count_limit,
+        )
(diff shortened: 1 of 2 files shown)


hbrunn (approved):

piotrhaya:
@StefanRijnhart Why this is in pending for so long ? Its hard to notice that fix is ready and because of that I managed to create own fix in JS (activity_menu.esm.js).
I'll leave here parts of the code to add, maybe it will come in handy someday :)
import { Domain } from "@web/core/domain";

    openActivityGroup(group, filter = "all") {
        if (this.env.services.user.context.team_activities === true) {
            document.body.click(); 
            const context = {
                force_search_count: 1,
            };

            if (filter === "all") {
                context["search_default_activities_overdue"] = 1;
                context["search_default_activities_today"] = 1;
            }
            else if(filter === "overdue"){
                context["search_default_activities_overdue"] = 1;
            }
            else if(filter === "today"){
                context["search_default_activities_today"] = 1;
            }
            else if(filter === "upcoming_all"){
                context["search_default_activities_upcoming_all"] = 1;
            }
            
            let domain = [["activity_team_user_ids", "in", [this.userId]]];
            if (group.domain) {
                domain = Domain.and([domain, group.domain]).toList();
            }
            const views = this.availableViews(group);

            this.action.doAction(
                {
                    context,
                    domain,
                    name: group.name,
                    res_model: group.model,
                    search_view_id: [false],
                    type: "ir.actions.act_window",
                    views,
                },
                {
                    clearBreadcrumbs: true,
                    viewType: group.view_type,
                }
            );
        
        } else {
           return super.openActivityGroup(group, filter);
        }
    },

StefanRijnhart:
@piotrhaya  This PR does not have the required number of approvals yet. It would be great if you could check that my solution is working for you. Then you can approve this PR, and that means we may get this PR merged.

piotrhaya:
@StefanRijnhart I tested this, it works as expected. Unfortunately I am not OCA member I don't have rights to approve it, sorry

StefanRijnhart:
@piotrhaya Thanks for checking! Review is not restricted to OCA members. You can go to the `Files changed` tab, then click Review -> Approved.

piotrhaya (approved):

PR #1547, opened 2025-01-16
[17.0][mig] mail_attach_existing_attachment_account

damdam-s:

PR #1672, opened 2025-06-20
[17.0][MIG] mail_attach_existing_attachment_account: Migration to 17.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/social/pull/1547
TT56840

@Tecnativa @pedrobaeza @victoralmau @alexandregaldeano @petrus-v could you please review this?


pedrobaeza:
Clicking on "Send & print", I don't get the attachment that already exists:

imagen

carlos-lopez-tecnativa:
> Clicking on "Send & print", I don't get the attachment that already exists:
> 
It should be resolved once this PR is merged: https://github.com/odoo/odoo/pull/215420.
The reason is that `ir.attachment` filters records with `res_field=False`, see [this line](https://github.com/odoo/odoo/blob/d86409d93e096126b32cbb35f442db332628375d/odoo/addons/base/models/ir_attachment.py#L532).
However, these attachments have that field filled, so in order to skip that domain, it is necessary to pass the `skip_res_field_check` key in the context.
But the `many2many_checkboxes` widget does not support context until the related PR is merged.
Note: This only occurs with attachments generated by Odoo using mail templates. If you upload an attachment manually, it is retrieved correctly.

pedrobaeza (approved):
Then it's something outside the module itself.

victoralmau (approved):

alexandregaldeano (approved):
LGTM

PR #1676, opened 2025-07-03
[16.0] [ADD] New module mail_company_aware_sender

NL66278:
New module where mail will be send (configurable per company) from company mail domain.


NL66278:
No idea how the project coverage can decrease, while I have 100% coverage of the patch, and there is no way the code can influence existing tests.

mail_company_aware_sender/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2025 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Mail Sender Company Aware",
+    "summary": "Send emails with company specific mail domain",
+    "version": "16.0.1.0.0",
+    "category": "Social Network",
+    "website": "https://github.com/OCA/social",
```
thomaspaulb, line 9: Doesn't this rather belong in `OCA/multi-company`, next to [mail_multicompany](https://github.com/OCA/multi-company/tree/16.0/mail_multicompany) ?

mail_company_aware_sender/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "author": ("Therp BV, " "Odoo Community Association (OCA)"),
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "mail",
+        "mail_outbound_static",  # Need the domain whitelist on outgoing server.
```
thomaspaulb, line 16: I see that you're using a field from this module to determine the domain. For our situation that works, but I'm a bit skeptical about introducing this dependency for anyone in this use case, because:

- While `mail_outbound_static` is one method that's used to select which mail server should be used for sending, it's not the only one
- Another way is using the [from_filter field](https://github.com/OCA/OCB/blob/5375cd1b68a975c63d4322507871d5c9e8a23935/odoo/addons/base/models/ir_mail_server.py#L762) from Odoo core
- Yet another way, I think, is to use the [company_id field from mail_multicompany](https://github.com/OCA/multi-company/blob/16.0/mail_multicompany/models/ir_mail_server.py#L10) to make only certain mail servers accessible when sending from certain companies
- In all such cases we want the from address of the user to match the domain of the company that he's sending from.

mail_company_aware_sender/models/res_company.py
```diff
@@ -0,0 +1,26 @@
...
+from odoo import fields, models
+from odoo.tools import email_domain_extract
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    use_email_domain = fields.Boolean(
+        help="Use domain part of company for sender email address",
+    )
```
thomaspaulb, line 13: Can't we then rather, instead of making this a boolean and inferring the domain from `mail_outbound_static`, make this a Char field that can be filled or not, and when it's filled, it's filled with the domain that the user's email domain is to be replaced with?

mail_company_aware_sender/models/mail_thread.py
```diff
@@ -0,0 +1,30 @@
...
+
+        Check for current company to see whether we should try to override
+        the email_from domain.
+        """
+        override_from = (not email_from) and self.env.company._override_email_domain()
+        author_id, email_from = super()._message_compute_author(
+            author_id, email_from, raise_on_email=raise_on_email
+        )
+        if override_from and author_id:
+            author = self.env["res.partner"].browse(author_id)
```
thomaspaulb, line 22: I don't know if this should always be done globally for all partners that can be an author.

I think I see now why you only want to override if the partner/user's email domain is a known domain in the system - at least that filters out all portal users that don't have that domain.

But maybe there are also cases where the override should for some reason not happen - perhaps good to make an opt_out field on res.partner for that. I had originally thought to make it even more flexible by providing each partner with a company-aware property field that specifies his or her (pre-computed, but modifiable) from-address, because then you can also cater for situations when someone is `god@kingdom.fr` and `allah@sultanate.ae` in different companies. But that probably also introduces new complications.

mail_company_aware_sender/__manifest__.py
NL66278: I think that rather mail_multicompany is out of place. Anyway, as the module depends on mail_outbound_static for the domain_whitelist field, it seems to me better where I put it now.

mail_company_aware_sender/__manifest__.py
NL66278: Yes, but the module is completely agnostic for which outgoing mail server is used to send the mail, Actually the customer for which the module is initially intended has only one outgoing mail server for all companies. As I think will most often be the case.

mail_company_aware_sender/models/res_company.py
NL66278: The domain part of the outgoing email is not inferred from the whitelist, but needs to be there, otherwise Odoo will not use it to send the email and fallback to a default sender address.

Just using the domain part of the company being enabled is more convenient for the user, as he/she will not have to repeat the info that already is in the company address.

mail_company_aware_sender/models/mail_thread.py
NL66278: An opt-out per partner might be a useful extension. I will put that in.

NL66278: @thomaspaulb I put in the partner opt-out.

NL66278:
The failing test is clearly an error in codecov, as the project coverage can not logically decrease if the patch coverage is 100%.

NL66278:
Running in production without any issues for some time now.

PR #1677, opened 2025-07-04
[16.0][MIG] mail_autogenerated_header: Migration to 16.0

LauraCForgeFlow:
Standard migration. Also added an improvement in a separate commit that lets you control whether outgoing emails are allowed to be marked as autogenerated. See the module’s README.rst for more details.


JordiMForgeFlow (approved):
code review 👍🏼

AndreuOForgeFlow (approved):

pedrobaeza (commented):
Marks all the sent emails as auto-generated are too much IMO. The automatic dropping of autogenerated incoming mails though is a great one that should be split.

LauraCForgeFlow:
@pedrobaeza We could split it into 2 modules for 16.0, but this module is already merged in both 17.0 and 18.0, so I think that could be a bit problematic.

To fix it while keeping a single module, maybe we could add a company setting or a system parameter for allowing certain outgoing mails to be set as auto-generated. WDYT? :)

pedrobaeza:
Ouch, I though it was not merged in upper versions. Yeah, the system parameter option seems a good workaround.

PR #1679, opened 2025-07-11
[ADD] mail_thread_attachment_deduplicate

len-foss:
When this module is installed, when processing attachments in a mail thread,
it will check for duplicates and remove them.

This is typically the case when a customer has inline images in their emails,
like social media links, company logos, etc.
If a customer has 10 images in one email, 10 attachments will be created for each message that is received.
After 5 exchanges, that will be 50 attachments, and potentially relevant ones might be lost in the noise.

Unfortunately, it is not possible to know which attachments are actually relevant, so this module
simply checks for exact duplicates (by checksum) and removes them.
As a result, after install of this module, the attachments will stay at 10, and only new content will be added.


CasVissers-360ERP:
@len-foss Tried to a do a functional test but failed: 
https://app.screencastify.com/watch/61JhfOXz7PmqVGzGTto1

len-foss:
@CasVissers-360ERP  I see.
To be clear, this does not introduce any general mechanism to "share attachments" in some form.
The reason being that it would be a real can of worm, introducing questions about what happens if the "master file" gets modified or deleted, access rights, etc.

As such the first version was only filtering incoming messages.

For outgoing message I've added a second override that fixes the flow that you've shown, as it seems to be similarly limited in potential impact.

hbrunn:
I wonder if you'd be better off to not change the storing part (as of above this might interfere with all kinds of things), but the display part? Why not override [mail.thread#_get_mail_thread_data_attachments](https://github.com/OCA/OCB/blob/16.0/addons/mail/models/mail_thread.py#L3430) and do the deduplication there?

You could even consider marking duplicates and moving them in an extra tab or similar on the client side

len-foss:
@hbrunn 
That makes sense if you really want to avoid any risk of bugs. However, these files can quickly add a few megabytes per ticket, which can contribute a significant size in the long term.

It seems to me that it should not interfere with any essential mechanism? If you see any I would indeed change the code to limit its potential impact.
I'll update if after a while in production there is any issue, but so far it's been running well.

hbrunn:
note that attachments in Odoo's standard file store occupy space only once per checksum, so space isn't really an issue here.

I'd expect embedded images to not be shown in the mail body of subsequent messages with your code, as Odoo replaces the cid: source with a link to the corresponding attachment, which you remove. Is that true? For annoying logos this might be a feature, but for the general case I don't think that's desirable, think spacer images or the like that actually contribute to readability

CasVissers-360ERP (approved):
Functional review

len-foss:
@hbrunn I've ported it to 18.0 with extra processing to deduplicate attachments references in the message body, so that it works fully transparently. 

https://github.com/OCA/social/pull/1716

ivs-cetmix:
> note that attachments in Odoo's standard file store occupy space only once per checksum, so space isn't really an issue here.
> 
> I'd expect embedded images to not be shown in the mail body of subsequent messages with your code, as Odoo replaces the cid: source with a link to the corresponding attachment, which you remove. Is that true? For annoying logos this might be a feature, but for the general case I don't think that's desirable, think spacer images or the like that actually contribute to readability

That's the case, however they will create unnecessary entities in the `ir_attachment` table which also consumes space.

PR #1681, opened 2025-07-15
[16.0][ADD] add new module mass_mailing_disable_tracking

huguesdk:
add new module `mass_mailing_disable_tracking` to allow to disable open and link click tracking in mass mailing messages.


mass_mailing_disable_tracking/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+# SPDX-FileCopyrightText: 2025 hugues de keyzer
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
```
petrus-v, line 3: I didn't know about SPDX before ! I suppose it's fine to use it in OCA until CI is green :)

petrus-v (approved):
Code looks good to me !

Awesome, thanks for this contribution that respect our privacy

mass_mailing_disable_tracking/models/__init__.py
huguesdk: there is no licensing check in the ci (or most modules would break). spdx is an open standard and is used by fsfe’s [reuse](https://reuse.software/). generalizing its use would actually allow to add a licensing check in the ci.

huguesdk:
thanks for the review, @petrus-v!

mass_mailing_disable_tracking/models/__init__.py
rousseldenis: I know @sbidoul had exchanges with @carmenbianca some years ago about that and to include them in our guidelines. But I don't know the resulting conclusions.

sbidoul: I remember Carmen mentioning that at some OCA days, but I don't think there was any conclusion and I have not heard about it since. These comments do not harm I guess, but I honestrly don't know what it would take to make that useful on OCA scale.

isturiz:
LGTM.

Jan-Onestein:
Wonderful, good to see that some still care about privacy !!

Can we please have codecov/project updated and have this merged into OCA/social asap ??

huguesdk:
> Wonderful, good to see that some still care about privacy !!
> 
> Can we please have codecov/project updated and have this merged into OCA/social asap ??

glad to hear that i’m not the only one. ☺️ i’m not sure: do i need to do something on my side? add more tests maybe?

NL66278 (approved):
👍 Very welcome contribution!

PR #1678, opened 2025-07-07
[17.0][IMP] mail_autogenerated_header: system parameter to allow autogenerated outgoing mails

LauraCForgeFlow:
Fw-port of improvement included in #1677.

PR #1686, opened 2025-07-21
[16.0][FIX] mail_autogenerated_header: keep original params in mail_server method

LauraCForgeFlow:
According to the [suggestion](https://github.com/OCA/social/pull/1678#pullrequestreview-3032935974) made by the original module's author, it is a good idea to keep the original parameters in the `_send_email_set_autogenerated` method. In this PR we set the parameter list to its original value.


hbrunn (approved):

pedrobaeza (approved):
/ocabot merge patch

Issue #1655, opened 2025-05-20
[17.0][BUG] mail_activity_team: passing team as False erases user


alexey-pelykh:
https://github.com/OCA/social/blame/0c4e7434b16f7d7dd799206add7f9925b0dee66e/mail_activity_team/models/mail_activity.py#L45

should probably be

```python
            if vals.get("team_id"):
```

cc @vvrossem @ilo - since that was your change, what was the grand idea to erase `user_id` if `team_id` is _just_ provided regardless of value?

vvrossem:
https://github.com/OCA/social/pull/1474
I'm afraid I don't remember.

Hopefully @imlopes has a better memory than I 🤞

vvrossem:
> The problem here are the changes between v16 and v17.
> 
> If you check the runbot for v16, I created an activity team and I have no "default" user : 
> 
> ![image](https://github.com/user-attachments/assets/1f27f3e7-43a1-448b-978a-85f97834e60d)
> 
> and 
> 
> ![image](https://github.com/user-attachments/assets/8b172d3d-945f-4722-9383-04616e521691)
> 
> This is because on v16, odoo is not using the wizard and so not calling `activity_schedule` method, which will force an `user_id`. Odoo uses by default the `create` method as usual, and as `mail_activity_team` change the `user_id` field from `required=True` to `required=False`, all is done with success.
> 
> In v17 we don't have the same behavior. After click on the button to schedule/plan an activity, odoo will call `activity_schedule`, which will force the `user_id` : 
> here https://github.com/odoo/odoo/blob/c9959a78b839c38e3fd4bb205d3aa2aeb4c6a31c/addons/mail/wizard/mail_activity_schedule.py#L296-L304 and here https://github.com/odoo/odoo/blob/c9959a78b839c38e3fd4bb205d3aa2aeb4c6a31c/addons/mail/models/mail_activity_mixin.py#L395-L399
> 
> IMHO the best solution was inherit `create` method and check if we are in the `team activity` case.  If is this case, I remove the `user_id`.
> 
> I'll check the error on the OCA tests.
> 
> Maybe you have another idea to implement ?
> I will check with the consultant if keep the `user_id` is a real problem. I've discussed about that yesterday and it seems that is a problem.
> 
> [EDIT]
> 
[...]

PR #1474, opened 2024-10-16
[17.0][REF] mail_activity_team: use mail.activity.schedule wizard

imlopes:
Replace https://github.com/OCA/social/pull/1459

PR #1694, opened 2025-08-01
[17.0][FIX]``mail_activity_team``: do not systematically remove the user from an activity when a team is set

Camille0907:
In [this PR](https://github.com/OCA/social/pull/1474), we updated the module to follow the v17 changes in standard. However it looks like the changes made are too wide: if no user is set, we indeed want to prevent the standard method to set one by default. But if a user is manually added, we want to keep it. Here we add a context key to determine at the creation of the activity if the user should be kept or erased. We also take [this issue](https://github.com/OCA/social/issues/1655) into account.

### DIFF
--- a/mail_activity_team/models/mail_activity.py
+++ b/mail_activity_team/models/mail_activity.py
@@ -35,14 +35,18 @@ class MailActivity(models.Model):
     def create(self, vals_list):
         # Differently from the previous odoo version,
         # the create method is called from (mail.activity.mixin).activity_schedule()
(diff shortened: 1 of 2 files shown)


ivantodorovich (approved):

imlopes (approved):

dreispt (approved):

CRogos:
@dreispt can you merge this?

yankinmax:
Hello @OCA/social-maintainers can you pls reopen this PR and possibly trigger merge?

PR #1638, opened 2025-04-17
[18.0][FIX] mail_thread_create_nolog: move author assignment with _log_access

rlimaeco:

PR #1700, opened 2025-08-05
[18.0][FIX] mail_thread_create_nolog: prevent improper creation of messages

rlimaeco:
Before this commit, when a user added a message or logged a note on any mail thread, the original creation message was being duplicated on each new message.

### DIFF
--- a/mail_thread_create_nolog/models/mail_message.py
+++ b/mail_thread_create_nolog/models/mail_message.py
@@ -22,9 +22,9 @@ class MailMessage(models.AbstractModel):
         # In that case there is no specific record implied, so no need to generate the
         # the message.
         record = self._messages_for_record(domain)
-        author = record.create_uid.partner_id
 
         if record and record._log_access:
+            author = record.create_uid.partner_id
             creation_message = record._creation_message()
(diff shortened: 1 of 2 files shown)


mail_thread_create_nolog/models/mail_message.py
```diff
@@ -17,14 +17,14 @@ def _get_message_fields(self):
         ]
 
     @api.model
-    def _generate_messsage(self, domain):
+    def _generate_message(self, domain):
         # Generate a creation message only if messages are fetched for a record.
         # In that case there is no specific record implied, so no need to generate the
         # the message.
```
twalter-c2c, line 23:
```suggestion
        # In that case there is no specific record implied, so no need to generate
        # the message.
```

twalter-c2c (approved):
LGTM, verified on runboat.

sebalix:
The first commit has already been merged in that PR: https://github.com/OCA/social/pull/1638

grindtildeath (commented):
You're over complicating things here. Please check https://github.com/OCA/social/pull/1718 and I guess we can close that

PR #1706, opened 2025-08-21
[17.0][FIX] mail_gateway: Messages are not being sent using the email messaging thread

CarlosRoca13:
cc @Tecnativa TT57738

Before these changes, the messages sent using the email thread were not being delivered.

error_envio

After making these changes, since the gateway_notifications key is empty, the super is called and the process continues normally.

envio realizado

ping @carlos-lopez-tecnativa @pedrobaeza


pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
Tested locally and working, thanks.

PR #1712, opened 2025-08-25
[17.0][FIX] mail_thread: Duplicated suggestions when email_cc set

CarlosRoca13:
Add the email formated to suggested recipients as odoo does to avoid having repeated suggestions.

See https://github.com/odoo/odoo/blob/bca4a1184d4c1cc7c5e987eb1753886337223f98/addons/mail/models/mail_thread_cc.py#L46C5-L52C26

cc @Tecnativa TT57760

ping @pedrobaeza @carlos-lopez-tecnativa

### DIFF
--- a/mail_tracking/tests/test_mail_tracking.py
+++ b/mail_tracking/tests/test_mail_tracking.py
@@ -223,7 +223,7 @@ class TestMailTracking(TransactionCase):
         # suggested recipients
         recipients = self.recipient._message_get_suggested_recipients()
         suggested_mails = {email[1] for email in recipients[self.recipient.id]}
-        self.assertIn("unnamed@test.com", suggested_mails)
+        self.assertIn('"Dominique Pinon" <unnamed@test.com>', suggested_mails)
         self.assertEqual(len(recipients[self.recipient.id]), 3)
         # Repeated Cc recipients
         message = self.env["mail.message"].create(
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

carlos-lopez-tecnativa:
Please check the CI

CarlosRoca13:
Solved @carlos-lopez-tecnativa

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -311,10 +311,14 @@ def test_email_to(self):
                 "alias_domain_id": alias_domain_id.id,
             }
         )
-        recipients = self.recipient._message_get_suggested_recipients()
+        recipients = self.recipient.with_context(
+            inwdb=True
```
pedrobaeza: Why this context?

mail_tracking/tests/test_mail_tracking.py
```diff
@@ -311,10 +311,14 @@ def test_email_to(self):
...
+        )._message_get_suggested_recipients()
         self.assertEqual(len(recipients[self.recipient.id]), 2)
         suggested_mails = {email[1] for email in recipients[self.recipient.id]}
-        self.assertNotIn("support+unnamed@test.com", suggested_mails)
+        self.assertNotIn(
+            '"Dominique Pinon" <support+unnamed@test.com>', suggested_mails
```
pedrobaeza, line 318: Is this change really needed?

mail_tracking/tests/test_mail_tracking.py
CarlosRoca13: Used for debugging an error but did not remember to delete it. It is now removed

mail_tracking/tests/test_mail_tracking.py
CarlosRoca13: It’s not strictly necessary, because the entry for `support+unnamed@test.com` won’t be there either, but to test it properly, yes.

pedrobaeza:
To be forward-ported/backported.

PR #1727, opened 2025-09-18
[17.0][FIX] mail_gateway: Users without technical access rights can't link a message

CarlosRoca13:
Before this change, when a user tried to link a message to a record, the system threw an error because they did not have enough access rights to see the ir.model records.

error_link

With this change, the error is no longer raised, and the user can link messages only to the models for which they have write access rights.

no_error_link

cc @Tecnativa TT58073

ping @pedrobaeza @victoralmau


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1729, opened 2025-09-22
[16.0][FIX] mail_gateway: Messages are not being sent using the email messaging thread

pedrobaeza:
Backport of #1706

Before these changes, the messages sent using the email thread were not being delivered.

After making these changes, since the gateway_notifications key is empty, the super is called and the process continues normally.

@Tecnativa


pedrobaeza:
Fast-tracking this as already merged in upper versions and very straight-forward for solving also the problem in 16:

/ocabot merge patch

PR #1730, opened 2025-09-23
[17.0][IMP] mail_gateway: add two new threadActionsRegistry

eduezerouali-tecnativa:
cc @Tecnativa TT58083
ping @CarlosRoca13 @pilarvargas-tecnativa 

Clicking on an author’s avatar now opens their contact form, making navigation more intuitive

### DIFF
--- a/mail_gateway/static/src/core/common/thread_actions.esm.js
+++ b/mail_gateway/static/src/core/common/thread_actions.esm.js
@@ -1,0 +1,93 @@
+/* @odoo-module */
+
+import {threadActionsRegistry} from "@mail/core/common/thread_actions";
+import {_t} from "@web/core/l10n/translation";
+import {useComponent} from "@odoo/owl";
+import {useService} from "@web/core/utils/hooks";
+
+threadActionsRegistry
+    .add("open-gw-new-partner", {
+        condition(component) {
+            return (
+                component.thread?.model === "discuss.channel" &&
+                (!component.props.chatWindow || component.props.chatWindow.isOpen) &&
+                component.thread?.guestId
+            );
+        },
+        icon: "fa fa-fw fa-address-book",
+        name: _t("New Partner"),
+        async setup() {
+            const component = useComponent();
+            const orm = useService("orm");
+            let guestId = null;
+            const guest = await orm.silent.searchRead(
+                "mail.guest",
+                [["name", "=", component.thread.name]],
+                ["id"]
+            );
+            guestId = guest[0]?.id;
+            if (guestId) {
+                component.thread.guestId = guestId;
+            }
+        },
+        async open(component) {
+            const guestId = component.thread?.guestId;
+            await component.env.services.action.doAction(
+                {
+                    type: "ir.actions.act_window",
+                    res_model: "mail.guest.manage",
+                    context: {default_guest_id: guestId},
+                    views: [[false, "form"]],
+                    target: "new",
+                },
+                {
+                    onClose: async () => {
+                        component.thread.guestId = null;
+                    },
+                }
+            );
+        },
+        iconLarge: "fa fa-fw fa-lg fa-address-book",
+        sequence: 18,
+    })
+    .add("open-gw-profile", {
+        condition(component) {
+            return (
+                component.thread?.model === "discuss.channel" &&
+                (!component.props.chatWindow || component.props.chatWindow.isOpen) &&
+                !component.thread?.guestId
+            );
+        },
+        icon: "fa fa-fw fa-user-circle-o",
+        name: _t("Open Contact"),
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Which is the problem this is fixing or which is the enhancement? Please specify in the commit message and the PR message.

eduezerouali-tecnativa:
imp-gateway
This is the new functionality, where before it was not clickable.

pedrobaeza:
Don't you already have the feature in core?

Peek 24-09-2025 15-40

eduezerouali-tecnativa:
can't see that feature, anyway this change was made because if there is no partner linked to it, calls `mail.guest.manage` that pop up a wizard so you can related o create a partner. So now after prevous step it is done, it will send you to the contact attach. All of this where are talking about `mail_gateway` functionality.
git-imp

pedrobaeza:
Yeah, I see that the feature is linked to HR.

Anyway, the problem is that this is not related to the `mail_gateway` (but the wizard is, as you can have the situation where there's no contact attached to the channel). I see this as another option inside that dropdown, and it's the same for regular chats than to WhatsAppp chats, or am I wrong?. Thus, it can be a module `mail_chat_partner_link`, and streamlined into a regular form, not clicking on the avatar.

@etobella what do you think?

eduezerouali-tecnativa:
> Yeah, I see that the feature is linked to HR.
> 
> Anyway, the problem is that this is not related to the `mail_gateway` (but the wizard is, as you can have the situation where there's no contact attached to the channel). I see this as another option inside that dropdown, and it's the same for regular chats than to WhatsAppp chats, or am I wrong?. Thus, it can be a module `mail_chat_partner_link`, and streamlined into a regular form, not clicking on the avatar.
> 
> @etobella what do you think?
Would make more to have 2 feature in the dropdown, more intuitive and user-friendly

eduezerouali-tecnativa:
Having this features in the dropdown, make it more intuitive and user-friendly

eduezerouali-tecnativa:
Should I close this PR and create a new one for the changes request, or can I push them here. @pedrobaeza

pedrobaeza:
What you prefer.

mail_gateway/models/discuss_channel.py
```diff
@@ -25,6 +25,7 @@ class MailChannel(models.Model):
         string="Gateway related Token",
         required=False,
     )
+    gateway_partner_id = fields.Many2one("res.partner")
```
pedrobaeza: Why do you need this extra field? And what happen with existing channels previous to this update?

eduezerouali-tecnativa: Changes done, no need to add anything. We can get to the partner in `res.partner.gateway.channel`, so in this case will not affect previous channels.

CarlosRoca13 (approved):

mail_gateway/static/src/core/common/thread_actions.esm.js
```diff
@@ -0,0 +1,86 @@
...
+        },
+        iconLarge: "fa fa-fw fa-lg fa-address-book",
+        sequence: 18,
+    })
+    .add("open-gw-profile", {
+        condition(component) {
+            return (
+                component.thread?.model === "discuss.channel" &&
+                (!component.props.chatWindow || component.props.chatWindow.isOpen) &&
+                component.store.self.type !== "guest"
```
carlos-lopez-tecnativa: I left two comments regarding this action:

1) I see this action as generic for opening a contact, so it’s not related only to the gateway. If it’s not, and it is meant to be specific for the gateway, please add the extra condition, such as `component.thread?.gateway`.
2) I see the option even when the contact doesn’t exist. I think you can use the `guestId` to decide whether to show `Open Contact` or `New Partner`.
image

eduezerouali-tecnativa: In this case `New Partner` is not as generic as `Open Contact`. Do you think it is a good idea to move `Open Contact` to a separated module?  It is not strictly related to gateway, `New Partner` is.

eduezerouali-tecnativa: I will have a look on `guestId` and why it is not working.

pedrobaeza: @eduezerouali-tecnativa did you check this? If so, you should ping the people doing the claiming for them to review again.

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
@eduezerouali-tecnativa migrate it to 18

eduezerouali-tecnativa:
@pedrobaeza already fwp here https://github.com/OCA/social/pull/1731

PR #1731, opened 2025-09-23
[18.0][IMP] mail_gateway: add two new threadActionsRegistry

eduezerouali-tecnativa:
FWP from 17.0: #1730 
cc @Tecnativa TT58083
ping @CarlosRoca13 @pilarvargas-tecnativa

### DIFF
--- a/mail_gateway/wizards/mail_guest_manage.py
+++ b/mail_gateway/wizards/mail_guest_manage.py
@@ -36,6 +36,7 @@ class MailGuestManage(models.TransientModel):
             self.env["discuss.channel.member"].create(
                 self._channel_member_vals(member, partner)
             )
+            member.channel_id.name = partner.complete_name
             member.unlink()
         self.env["mail.message"].search(
             [("author_guest_id", "=", self.guest_id.id)]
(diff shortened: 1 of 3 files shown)


eduezerouali-tecnativa:
ping @pedrobaeza

mail_gateway/wizards/mail_guest_manage.py
```diff
@@ -36,6 +36,9 @@ def _merge_partner(self, partner):
             self.env["discuss.channel.member"].create(
                 self._channel_member_vals(member, partner)
             )
+            self.env["discuss.channel"].browse(member.channel_id.id).write(
```
pedrobaeza: Why not doing?


```suggestion
            member.channel_id.name = partner.name
```

pedrobaeza: And shouldn't it be `complete_name` instead?

eduezerouali-tecnativa: not thought about it but make more sense. This was implemented so when you merge the guest into a partner, the channel names changes to the new partner name instead of keep the default one from the gateway. With complete_name will make it more clear

pedrobaeza (approved):
Tested and finally working in a viable way, so let's merge it for also doing the translations.

/ocabot merge patch

PR #1732, opened 2025-09-25
[17.0][IMP] mail_show_follower: Allow using records in mail.mail._build_cc_text

david-banon-tecnativa:
Allows using records inside mail.mail._build_cc_text, useful if you need to inherit from this method.
@pedrobaeza @juancarlosonate-tecnativa 
TT58129 @Tecnativa

### DIFF
--- a/mail_show_follower/models/mail_mail.py
+++ b/mail_show_follower/models/mail_mail.py
@@ -1,12 +1,11 @@
 from markupsafe import Markup
 
-from odoo import api, models, tools
+from odoo import models, tools
 
 
 class MailMail(models.Model):
     _inherit = "mail.mail"
 
-    @api.model
     def _build_cc_text(self, partners):
         if not partners:
             return ""


mail_show_follower/models/mail_mail.py
```diff
@@ -1,12 +1,11 @@
...
 
-from odoo import api, models, tools
+from odoo import models, tools
 
 
 class MailMail(models.Model):
     _inherit = "mail.mail"
 
-    @api.model
     def _build_cc_text(self, partners):
```
pedrobaeza, line 9:
```suggestion
    def _build_cc_text(self, partners):
        self.ensure_one()
```

pedrobaeza (approved):

mail_show_follower/models/mail_mail.py
david-banon-tecnativa: This method is used on settings without any actual record, so we can't run ensure_one() without also changing that. https://github.com/OCA/social/blob/4e2dda748a618157d3dc5930526843992436e127/mail_show_follower/models/res_config_settings.py#L51

pedrobaeza: OK, got it. It's a bit tricky then.

victoralmau (approved):

pedrobaeza:
Please fw-port it to 18

PR #1736, opened 2025-10-01
[17.0][IMP] mail_attach_existing_attachment: Add more information in the attachment display name

carlos-lopez-tecnativa:
When a message contains multiple attachments, including duplicates or files with the same name, it becomes difficult to distinguish them.

This commit improves the display name of attachments by adding the creator (user) and creation date, making it easier to differentiate records.

TT57808

Before
image

After
image

### DIFF
--- a/mail_attach_existing_attachment/__init__.py
+++ b/mail_attach_existing_attachment/__init__.py
@@ -1,1 +1,2 @@
+from . import models
 from . import wizard
--- a/mail_attach_existing_attachment/models/__init__.py
+++ b/mail_attach_existing_attachment/models/__init__.py
@@ -1,0 +1,1 @@
+from . import ir_attachment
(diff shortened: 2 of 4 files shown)


pedrobaeza (commented):
This is impacting in the display name everything, not only in the scope of the module, so I think it should be limited.

carlos-lopez-tecnativa:
> This is impacting in the display name everything, not only in the scope of the module, so I think it should be limited.

I updated the code to pass the context `display_full_attachment_name`.
This is possible thanks to the merge of PR https://github.com/odoo/odoo/pull/216588/

carlos-lopez-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge patch

PR #1752, opened 2025-11-03
[16.0][FIX]mail_activity_team: correctly count only active activities

jbuitrago-pyming:
There is a discrepancy between the activity count shown in the systray and the actual number of pending activities.

The systray counter was incorrectly including activities that have already been marked as done. This caused confusion as the badge number did not match the number of items in the activity list after clicking on it.

image
image


ofonseca-pyming (approved):
Functional and code reviewed.

CRogos (approved):
Code LGTM

etobella (changes requested):
The problem with this is that the field is added in `mail_activity_done`.

This change will fail if you only have this module installed (tested locally...)

PR #1758, opened 2025-11-18
[17.0][FIX] mail_tracking: Prevent singleton error when trying to resend a failed message

carlos-lopez-tecnativa:
In the `mail` module, a new resend record is always created https://github.com/odoo/odoo/blob/c956c154755e2630da2f464a88f5c3545d7733e3/addons/mail/wizard/mail_resend_message.py#L44 so this search is limited to 1 to prevent the issue.

image

TT59301
@Tecnativa @pedrobaeza @CarlosRoca13 could you please review this?

### DIFF
--- a/mail_tracking/wizards/mail_resend_message.py
+++ b/mail_tracking/wizards/mail_resend_message.py
@@ -29,7 +29,10 @@ class MailResendMessage(models.TransientModel):
                 )
                 if existing_resend_partner:
                     rec["partner_ids"].extend(
-                        [Command.link(existing_resend_partner.id)]
+                        [
+                            Command.link(resend_partner.id)
+                            for resend_partner in existing_resend_partner
+                        ]
                     )
                     continue
                 # Create only resends that mail.notification didn't prepare already


mail_tracking/wizards/mail_resend_message.py
```diff
@@ -25,7 +25,7 @@ def default_get(self, fields):
                     lambda x, tracking=tracking: x.res_partner_id == tracking.partner_id
                 )
                 existing_resend_partner = self.env["mail.resend.partner"].search(
-                    [("notification_id", "=", notification.id)]
+                    [("notification_id", "=", notification.id)], limit=1
```
pedrobaeza: I think you should add all the existing partners, isn't it?

carlos-lopez-tecnativa: @pedrobaeza done

mail_tracking/wizards/mail_resend_message.py
```diff
@@ -29,7 +29,7 @@ def default_get(self, fields):
                 )
                 if existing_resend_partner:
                     rec["partner_ids"].extend(
-                        [Command.link(existing_resend_partner.id)]
+                        [Command.set(existing_resend_partner.ids)]
```
pedrobaeza: This should be link, not set AFAIK, as it's extending the previous dictionary.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #1763, opened 2025-11-19
[18.0][MIG] mail_partial_autodelete

mathieudelva:
standard migration

@Kev-Roche @metaminux


mail_partial_autodelete/tests/test_mail_partial_autodelete.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2021 Akretion (http://www.akretion.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo.addons.base.tests.common import BaseCommon
+
+
+class TestStockMoveForcedLot(BaseCommon):
```
metaminux, line 7: SavePointCase should be replaced by TransactionCase.
BaseCommon should be used only when there is no call to the DB in tests.

```suggestion
from odoo.tests.common import TransactionCase


class TestMailPartialAutodeleteCase(TransactionCase):
```

mail_partial_autodelete/models/mail_mail.py
```diff
@@ -0,0 +1,45 @@
...
+            # exceptions email can be drop
+            sent_records = self.filtered(lambda s: s.state == "sent")
+            # we purge (for security reason) email that are not a notification
+            # maybe we have secret inside
+            sent_records.filtered(lambda s: not s.is_notification).write(
+                {"body_html": ""}
+            )
+            return super(MailMail, self - sent_records).unlink()
+        else:
+            return super().unlink()
```
metaminux, line 45: As stated in the recommendations for migrating to 15.0 :
> If you want to prevent the deletion of some records according a criteria, instead of overriding the unlink method, use a method with the decorator @api.ondelete. Put at_uninstall=True only if the check should be also performed on module uninstallation. Example: https://github.com/odoo/odoo/blob/74e76bfb9bce3c4caf80a69af3359677b6b40cf3/addons/account/models/account_account.py#L554-L557

mail_partial_autodelete/tests/test_mail_partial_autodelete.py
```diff
@@ -0,0 +1,35 @@
...
+    def setUp(self):
+        super().setUp()
+        self.mail = self.env["mail.mail"].create(
+            {
+                "subject": "Test subject",
+                "body_html": "example body",
+                "email_to": "test@example.com",
+                "email_from": self.env.user.email,
+            }
+        )
```
metaminux, line 17: > If there's a test class using setUp to generate test records, move it to setUpClass as it's way more performant since it runs only once.


```suggestion
    @classmethod
    def setUpClass(cls)
        super().setUpClass()
        cls.mail = cls.env["mail.mail"].create(
            {
                "subject": "Test subject",
                "body_html": "example body",
                "email_to": "test@example.com",
                "email_from": cls.env.user.email,
            }
        )
```

metaminux (changes requested):

metaminux:
Hello @mathieudelva 

I added some comments and suggestions on this migration PR some time ago.
Can you take a look and tell me if you need some help of any kind ?

PR #1773, opened 2025-12-06
[17.0][FIX] mail_gateway: Fix charset error for Werkzeug 2.3+

geomer198:
Add hasattr() check for charset attribute before usage with fallback to 'utf-8' for compatibility with Werkzeug 2.3+ where the attribute was removed.

Werkzeug 2.3.0:

https://werkzeug.palletsprojects.com/en/stable/changes/#version-2-3-0

For odoo 17 requirements:
```
Werkzeug==2.0.2 ; python_version <= '3.10' 
Werkzeug==2.2.2 ; python_version > '3.10' and python_version < '3.12'
Werkzeug==3.0.1 ; python_version >= '3.12'  # (Noble) Avoid deprecation warnings
```

### DIFF
--- a/mail_gateway/controllers/gateway.py
+++ b/mail_gateway/controllers/gateway.py
@@ -51,9 +51,12 @@ class GatewayController(Controller):
                     ("Content-Type", "application/json"),
                 ],
             )
-        jsonrequest = json.loads(
-            request.httprequest.get_data().decode(request.httprequest.charset)
+        charset = (
+            hasattr(request.httprequest, "charset")
+            and request.httprequest.charset
+            or "utf-8"
         )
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Isn't a replacement of such attribute for not hardcoding `utf-8`?

geomer198:
> Isn't a replacement of such attribute for not hardcoding `utf-8`?

The new version of the library uses only UTF-8 encoding. This change therefore supports the behavior of both older versions of the Werkzeug library and the newer ones (>= 2.3).

pedrobaeza (approved):

Bearnard21 (approved):
LGTM

pedrobaeza:
You can fw-port it to upper versions.

PR #1784, opened 2025-12-31
[16.0][FIX] email_template_qweb: use language-aware env for rendering QWeb

KKamaa:
fix bug where in email_template_qweb module: the QWeb renderer (ir.qweb) is instantiated once from the original environment (self.env) before the code switches into the per-language context:

- Subject is rendered via the "normal" mail.template rendering path that uses the per-language self_with_lang context -> it follows template.lang.
- Body is rendered by ir.qweb._render(...) using an IrQweb recordset created before lang was applied -> it keeps whatever env.lang was at the moment generate_email() started (e.g. the website request language, current user language, etc.), which explains the "random" English/Dutch mix.

### DIFF
--- a/email_template_qweb/models/mail_template.py
+++ b/email_template_qweb/models/mail_template.py
@@ -1,4 +1,4 @@
-# Copyright 2016 Therp BV <http://therp.nl>
+# Copyright 2016-2026 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import fields, models, tools
 
@@ -17,36 +17,28 @@ class MailTemplate(models.Model):
 
     def generate_email(self, res_ids, fields):
         multi_mode = True
-        IrQweb = self.env["ir.qweb"]
-
         if isinstance(res_ids, int):
             res_ids = [res_ids]
             multi_mode = False
-        result = super(MailTemplate, self).generate_email(res_ids, fields=fields)
-        for lang, (_template, _template_res_ids) in self._classify_per_lang(
+        result = super().generate_email(res_ids, fields=fields)
+        if self.body_type != "qweb_view" or (fields and "body_html" not in fields):
+            return result if multi_mode else result[res_ids[0]]
+        for lang, (_template, template_res_ids) in self._classify_per_lang(
             res_ids
         ).items():
             self_with_lang = self.with_context(lang=lang)
-            for res_id in res_ids:
-                if self.body_type == "qweb_view" and (
-                    not fields or "body_html" in fields
-                ):
-                    for record in self_with_lang.env[self.model].browse(res_id):
-                        body_html = IrQweb._render(
-                            self_with_lang.body_view_id.id,
-                            {"object": record, "email_template": self_with_lang},
-                        )
-                        # Some wizards, like when sending a sales order, need this
-                        # fix to display accents correctly
-                        body_html = tools.ustr(body_html)
-                        result[res_id][
-                            "body_html"
-                        ] = self_with_lang._render_template_postprocess(
-                            {res_id: body_html}
-                        )[
-                            res_id
-                        ]
-                        result[res_id]["body"] = tools.html_sanitize(
-                            result[res_id]["body_html"]
-                        )
+            IrQweb = self_with_lang.env["ir.qweb"]
+            for res_id in template_res_ids:
+                record = self_with_lang.env[self_with_lang.model].browse(res_id)
+                result_res_id = result[res_id]
+                body_html = IrQweb._render(
+                    self_with_lang.body_view_id.id,
+                    {"object": record, "email_template": self_with_lang},
+                )
+                body_html = tools.ustr(body_html)
+                rendered_post_temp = self_with_lang._render_template_postprocess(
+                    {res_id: body_html}
+                )
+                result_res_id["body_html"] = rendered_post_temp[res_id]
+                result_res_id["body"] = tools.html_sanitize(result_res_id["body_html"])
         return result if multi_mode else result[res_ids[0]]


thomaspaulb (approved):
Code approved, just the commit message should change to the OCA standard:

```
[FIX] email_template_qweb: use language-aware env for rendering QWeb

More detailed information... (maybe can leave out in this case, summary is clear)
```

email_template_qweb/models/mail_template.py
```diff
@@ -17,36 +17,39 @@ class MailTemplate(models.Model):
...
+                    body_html = IrQweb._render(
+                        self_with_lang.body_view_id.id,
+                        {"object": record, "email_template": self_with_lang},
+                    )
+                    body_html = tools.ustr(body_html)
+                    result[res_id][
+                        "body_html"
+                    ] = self_with_lang._render_template_postprocess(
+                        {res_id: body_html}
+                    )[
```
NL66278: Bit overkill to spread `[res_id]`  over three lines.

NL66278 (approved):
👍 One minor remark but for the rest LGTM

email_template_qweb/models/mail_template.py
KKamaa: It looks that way due to the pylint but the res_id used is from the for loop.

NL66278: Maybe I was not clear, I just meant that lines 48 - 50 could be on one line, making it all a bit more compact and readable.

thomaspaulb:
@OCA/social-maintainers This is ready for merge

ntsirintanis (approved):

NL66278:
@KKamaa Please remove the Draft: status of this PR, as that might stop people from merging.

email_template_qweb/models/mail_template.py
KKamaa: updated the readability on the lines affected

NL66278 (approved):
👍 Already looked good. Now even better ;-)

ntsirintanis (approved):

email_template_qweb/models/mail_template.py
```diff
@@ -36,20 +36,18 @@ def generate_email(self, res_ids, fields):
                     not fields or "body_html" in fields
                 ):
                     record = self_with_lang.env[self_with_lang.model].browse(res_id)
+                    result_resId = result[res_id]
```
NL66278: result_resId mixes camelCase en snake_style. We should only use snake_style, so please rename to result_res_id.

NL66278:
@KKamaa Please take up my new comment and then merge commits.

email_template_qweb/models/mail_template.py
KKamaa: fixed this

email_template_qweb/models/mail_template.py
```diff
@@ -17,36 +17,37 @@ class MailTemplate(models.Model):
...
         ).items():
             self_with_lang = self.with_context(lang=lang)
-            for res_id in res_ids:
-                if self.body_type == "qweb_view" and (
+
+            # bind QWeb AFTER switching lang context
+            IrQweb = self_with_lang.env["ir.qweb"]
+
+            for res_id in template_res_ids:  # ✅ use the bucketed ids
+                if self_with_lang.body_type == "qweb_view" and (
```
NL66278: I just noticed that this can be tested straight after line 24:
 `result = super().generate_email(res_ids, fields=fields)`

If this is followed by
```
if self.body_type != "qweb_view" or (fields and "body_html" not in fields):
    return result
```
A lot of processing can be avoided if we do not in the end modify the result anyway, and the code will have an indentation level less.

NL66278:
@KKamaa One more remark, check whether you agree, and still need to smash commits.

email_template_qweb/models/mail_template.py
KKamaa: added it, the early exit, if body isn't a qweb_view or body_html is in fields.

email_template_qweb/models/mail_template.py
```diff
@@ -17,36 +17,40 @@ class MailTemplate(models.Model):
...
```
NL66278: So drop this if and de-indent everything that follows.

email_template_qweb/models/mail_template.py
```diff
@@ -17,36 +17,40 @@ class MailTemplate(models.Model):
...
+        if self.body_type != "qweb_view" or (fields and "body_html" not in fields):
+            return result if multi_mode else result[res_ids[0]]
+
+        for lang, (_template, template_res_ids) in self._classify_per_lang(
             res_ids
         ).items():
             self_with_lang = self.with_context(lang=lang)
-            for res_id in res_ids:
-                if self.body_type == "qweb_view" and (
+
```
NL66278: Also do not add blank lines everywhere. Basically, with some exceptions, the body of a method should not contain blank lines. They make it more difficult to see all the logic on one page.

email_template_qweb/models/mail_template.py
KKamaa: made the fix

email_template_qweb/models/mail_template.py
KKamaa: fixed

KKamaa:
@NL66278 there is a conflict coming from your branch https://github.com/Therp/social/tree/16.0-email_template_qweb-copy-view-also

NL66278:
@OCA/social-maintainers Can somebody please merge this?

hbrunn:
merging will only work after https://github.com/OCA/oca-addons-repo-template/pull/333 is merged and applied to this branch

NL66278:
@hbrunn Merging should work now, I made a PR for the repo template upgrade and it was immediately merged by Pedro.

PR #1395, opened 2024-06-13
[15.0][ADD] mail_message_restrict

AungKoKoLin1997:
Superseding of https://github.com/OCA/social/pull/1177

@qrtl QT3509

PR #1786, opened 2026-01-06
[15.0][ADD] mail_message_restrict

AungKoKoLin1997:
Supersedes https://github.com/OCA/social/pull/1395

This module limits the creation of non-internal messages in the chatter to certain "allowed" models.
This means that, upon installation of the module, users will get an error message when they try to create
a message in the chatter, unless the model of the chatter is part of the "allowed" models.

@qrtl QT6295


kanda999 (approved):
Functional review: this module is working in production env.

yostashiro (approved):
Relaying my prior approval from the other PR.

PR #1793, opened 2026-01-14
[17.0][IMP] mail_history_mark_unread: enable marking notifications as unread from search panel

AndreuOForgeFlow:
Enable a button from the Search Messages panel (accessible via the search icon in Discuss), allowing users to mark any searched message as unread directly from the search results.

image

### DIFF
--- a/mail_history_mark_unread/README.rst
+++ b/mail_history_mark_unread/README.rst
@@ -29,9 +29,13 @@ Mail History Mark as Unread
 |badge1| |badge2| |badge3| |badge4| |badge5|
 
 This module introduces a "Mark as Unread" button for messages in the
-History mailbox of the Discuss App. When done the message is restored
-back to the Inbox, just like if it were a new unread message, and
-disappears from the History mailbox.
+History mailbox and Search results of the Discuss App. When done the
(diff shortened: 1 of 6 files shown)


AaronHForgeFlow:
it seems runboat is not available anymore, can you rebase so it is rebuilt?

AndreuOForgeFlow:
> it seems runboat is not available anymore, can you rebase so it is rebuilt?

Done, @AaronHForgeFlow ! :pray:

AaronHForgeFlow (commented):
It works well, just one thing, when I mark it as unread the message disappears from the search, is that intended?

AndreuOForgeFlow:
> It works well, just one thing, when I mark it as unread the message disappears from the search, is that intended?

@AaronHForgeFlow. Yes, that is intended, since when a message is marked as unread, it is moved back to the inbox and therefore disappears from this search  :+1:

AaronHForgeFlow (approved):
Functional LGTM 👍

AndreuOForgeFlow:
Hi @pedrobaeza and @etobella, could we perhaps merge this one?

PR #1794, opened 2026-01-15
[15.0][ADD] ‎mail_last_message_date

nobuQuartile:
We sometimes need to know the last time a record’s chatter was updated (e.g., in automated actions). 
Added a last_chatter_update datetime field to chatter-enabled models and implemented automatic updates on message creation. 

@qrtl QT5890


AungKoKoLin1997 (commented):
@nobuQuartile I’m not comfortable with the current design because other modules will need to depend on it, and the current module would also require an additional `mixin` model.

I believe we should create a generic module that can be reused across multiple use cases.

My idea is to add a new field on `mail.thread` rather than using a `mixin`, and then update that field when a message is created, based on the configuration we set.

We’ll also need to add configuration options in Settings and in the message subtype to control which subtypes and which models should use this feature.

How do you think that?

nobuQuartile:
@AungKoKoLin1997 CC: @yostashiro 

When I discussed it directly with Yoshi last time, we agreed that we could create a new field in the mail.thread, it would end up adding a field — which may or may not be used — to all models that inherit from mail.thread.

As a result, we concluded that it would be better to implement it as a mixin module and add the field only to the models that actually need it.

Do you agree with this approach?

mail_last_chatter_update_mixin/models/mail_last_chatter_update_mixin.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class MailLastChatterUpdateMixin(models.AbstractModel):
+    _name = "mail.last_chatter_update.mixin"
```
yostashiro:
```suggestion
    _name = "mail.last.message.date.mixin"
```

mail_last_chatter_update_mixin/models/mail_last_chatter_update_mixin.py
```diff
@@ -0,0 +1,23 @@
...
+    _description = "Mail Last Chatter Update Mixin"
+    _last_chatter_update_message_types = ()
+
+    last_chatter_update = fields.Datetime(
```
yostashiro:
```suggestion
    last_message_date = fields.Datetime(
```

mail_last_chatter_update_mixin/models/mail_last_chatter_update_mixin.py
```diff
@@ -0,0 +1,23 @@
...
+        help="Datetime when a message was posted on the record.",
+    )
+
+    def _get_last_chatter_update_message_types(self):
+        return self._last_chatter_update_message_types
+
+    def _last_chatter_update_should_write(self, message):
```
yostashiro:
```suggestion
    def _should_track_message(self, message):
```

mail_last_chatter_update_mixin/models/mail_message.py
```diff
@@ -0,0 +1,31 @@
...
+class MailMessage(models.Model):
+    _inherit = "mail.message"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        messages = super().create(vals_list)
+        messages._update_last_chatter_update()
+        return messages
+
+    def _update_last_chatter_update(self):
```
yostashiro: Can we move this method above create().
```suggestion
    def _update_last_message_date(self):
```

mail_last_chatter_update_mixin/tests/test_models.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class TestMailLastChatterUpdate(models.Model):
+    _name = "test.mail.last.chatter.update"
```
yostashiro:
```suggestion
class TestMailLastMessageDate(models.Model):
    _name = "test.mail.last.message.date"
```

mail_last_chatter_update_mixin/models/mail_last_chatter_update_mixin.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class MailLastChatterUpdateMixin(models.AbstractModel):
+    _name = "mail.last_chatter_update.mixin"
+    _description = "Mail Last Chatter Update Mixin"
+    _last_chatter_update_message_types = ()
```
yostashiro: I think we can drop this constant.
```suggestion
```

mail_last_chatter_update_mixin/models/mail_last_chatter_update_mixin.py
```diff
@@ -0,0 +1,23 @@
...
+
+    last_chatter_update = fields.Datetime(
+        help="Datetime when a message was posted on the record.",
+    )
+
+    def _get_last_chatter_update_message_types(self):
+        return self._last_chatter_update_message_types
```
yostashiro:
```suggestion
    def _get_tracked_message_types(self):
        return []
```

mail_last_chatter_update_mixin/models/mail_message.py
```diff
@@ -0,0 +1,31 @@
...
+        for message in self:
+            if not (message.model and message.res_id):
+                continue
+            record = self.env[message.model].browse(message.res_id)
+            if (
+                not record.exists()
```
yostashiro: Do you have an example of how this condition is needed?

mail_last_chatter_update_mixin/models/mail_message.py
```diff
@@ -0,0 +1,31 @@
...
+                or "last_chatter_update" not in record._fields
+                or not record._last_chatter_update_should_write(message)
+            ):
+                continue
+            record.sudo().write(
+                {
+                    "last_chatter_update": message.date or fields.Datetime.now(),
+                }
+            )
```
yostashiro: I suppose no fallback is needed here.
```suggestion
            record.sudo().write({"last_message_date": message.date})
```

yostashiro (changes requested):

yostashiro:
Module name suggestion: mail_last_message_date

mail_last_message_date/models/mail_last_message_date_mixin.py
```diff
@@ -0,0 +1,21 @@
...
+    )
+
+    def _get_tracked_message_types(self):
+        return []
+
+    def _should_track_message(self, message):
+        message_types = self._get_tracked_message_types()
+        if not message_types:
+            return True
+        return bool(message.message_type) and message.message_type in message_types
```
yostashiro:
```suggestion
        return message.message_type in message_types
```

yostashiro (commented):
Just a quick one. I will come back to this PR for more later. Please update the PR title accordingly.

mail_last_message_date/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,18 @@
...
+#. Example:
+
+   .. code-block:: python
+
+      from odoo import models
+
+      class PurchaseOrder(models.Model):
+          _inherit = ["purchase.order", "mail.last.message.date.mixin"]
+
+          def _get_tracked_message_types(self):
```
AungKoKoLin1997, line 15: Since mixin model is returning []. I think there is no reason to call super().

mail_last_message_date/tests/test_models.py
```diff
@@ -0,0 +1,16 @@
...
+class TestMailLastMessage(models.Model):
+    _name = "test.mail.last.message.date"
+    _inherit = ["mail.thread", "mail.last.message.date.mixin"]
+
+    name = fields.Char()
+
+    def _get_tracked_message_types(self):
+        message_types = super()._get_tracked_message_types()
+        message_types.append("email")
+        return message_types
```
AungKoKoLin1997, line 16: Same as above.

mail_last_message_date/tests/test_models.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class TestMailLastMessage(models.Model):
+    _name = "test.mail.last.message.date"
+    _inherit = ["mail.thread", "mail.last.message.date.mixin"]
```
AungKoKoLin1997, line 9: Please also add `description` to avoid CI error when we do the future migration.

mail_last_chatter_update_mixin/models/mail_message.py
nobuQuartile: No. We don't need the condition.
I have deleted it.

mail_last_message_date/readme/CONFIGURE.rst
nobuQuartile: Done

mail_last_message_date/tests/test_models.py
nobuQuartile: Done

mail_last_message_date/tests/test_models.py
nobuQuartile: Done.
Should I add the _description to other models?

nobuQuartile:
@AungKoKoLin1997 
Sorry for the late update to the code.
Could you continue the review?

PR #1780, opened 2025-12-24
[18.0][IMP] mail_gateway_whatsapp: Add support for buttons and variables in WhatsApp templates

eduezerouali-tecnativa:
This is FWP from https://github.com/OCA/social/pull/1723
cc @Tecnativa TT57145

ping @pedrobaeza @carlos-lopez-tecnativa @CarlosRoca13

PR #1796, opened 2026-01-16
[18.0][IMP] mail_gateway_whatsapp: Add support for messages in server actions

pedrobaeza:
Depends on:

- [x] #1780

You may want to send messages through the WhatsApp gateway in your server actions the same you are able to send mails in the linked documents.

This also allows to select this kind of server action on automation rules (base_automation).

@Tecnativa  TT57123


victoralmau (approved):

carlos-lopez-tecnativa (changes requested):
I got an error when using the server action.
Image

I fixed this error in https://github.com/OCA/social/compare/c29e67864d94aae3e0107e7edb6c540b8d62cffe..356d50f14e48c657fe7614b008f214ef2f2e108e but I see that you reverted my change in https://github.com/OCA/social/compare/f2d97c7b9c41f53225e1b49180715d8be66745c5..32abaf45a6bbdb55080786cc43b8e4b06dde3848 Why?

Note that the function `render_body_message` expects the `default_res_id` or `default_res_ids` keys in the context; otherwise, the record is not set correctly.
https://github.com/OCA/social/blob/9612966c7c13e02d1ade1ea8ed844b89ff508e98/mail_gateway_whatsapp/models/mail_whatsapp_template.py#L436-L442

pedrobaeza:
@carlos-lopez-tecnativa sorry the overwritten was a mistake on my side. The context has been restored.

carlos-lopez-tecnativa (approved):

PR #1805, opened 2026-02-02
[16.0][FIX] Rename Deco Addict partner to AcmeCorporation

NL66278:
This follows the same change as in Odoo, after complaints from a real Deco company.

### DIFF
--- a/mail_activity_reminder/tests/test_mail_activity_reminder.py
+++ b/mail_activity_reminder/tests/test_mail_activity_reminder.py
@@ -21,6 +21,7 @@ class TestMailActivityReminder(common.TransactionCase):
                 no_reset_password=True,
             )
         )
+        cls.Partner = cls.env["res.partner"]
         cls.ResUsers = cls.env["res.users"]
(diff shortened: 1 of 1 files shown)


NL66278:
@OCA/social-maintainers Can this be fast tracked as it is just a needed adaption to a change in Odoo and without it all 16.0 PR's will be red.

mail_activity_reminder/tests/test_mail_activity_reminder.py
```diff
@@ -31,8 +31,8 @@ def setUpClass(cls):
         cls.model_res_partner = cls.env["ir.model"].search(
             [("model", "=", "res.partner")], limit=1
         )
-        cls.partner_DecoAddict = cls.env["res.partner"].search(
-            [("name", "ilike", "Deco Addict")], limit=1
+        cls.partner_AcmeCorporation = cls.env["res.partner"].search(
```
etobella: Can you create a partner to avoid this issue in the future?

etobella (approved):
Thanks!

PR #1717, opened 2025-09-02
[17.0][IMP] mail_gateway, mail_gateway_whatsapp: Handle and display errors when sending a message through the gateway

carlos-lopez-tecnativa:
By default, Odoo only filters notifications with errors for the `notification_type` 'email'.  see https://github.com/odoo/odoo/blob/f40701dd894f01ed1a3a6360b1177a0d53552c49/addons/mail/wizard/mail_resend_message.py#L36

This commit adds the ability to display errors to the user when a gateway message fails to be sent for any reason.

TT50004

@Tecnativa @pedrobaeza @CarlosRoca13 @etobella could you please review this?

Before
image

After

image

PR #1806, opened 2026-02-02
[17.0][IMP] mail_gateway_whatsapp: Handle errors when the message is not sent to the API

carlos-lopez-tecnativa:
In the previous PR https://github.com/OCA/social/pull/1717 we handled the notification via webhook when Meta returns an error. However, when the error occurs in the communication with Meta, the message is displayed as delivered. After this commit, the message is displayed as error when applicable.

The failure details are only written when `failure_type = 'unknown'`. See: https://github.com/OCA/social/blob/695efd6fa3401614118c6a52f58453b40a9f0842/mail_gateway/models/mail_message.py#L114 So, in exception cases, we write this field to ensure it is processed correctly.

TT59775

@Tecnativa @pedrobaeza @CarlosRoca13 @etobella could you please review this?


etobella (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1822, opened 2026-03-11
[16.0][FIX] mail_gateway_whatsapp: We should limit the partner, as we might have 2 partners with the same phone

etobella:
Not usual, but might happen


pedrobaeza (approved):
This is the same as https://github.com/OCA/social/pull/1796/changes/70da8f381fe1b52c16a0882ad90f405982d613be

pedrobaeza:
fw-ported to 17 in #1823

PR #1826, opened 2026-03-16
[16.0][UPD] mail_composer_cc_bcc: update hash table

NL66278:
Test was showing errors because upstream _send function changed. The changes themselves where not relevant, because overridden function works on one message at a time, not a list.

### DIFF
--- a/mail_composer_cc_bcc/tests/test_mail_cc_bcc.py
+++ b/mail_composer_cc_bcc/tests/test_mail_cc_bcc.py
@@ -11,7 +11,7 @@ from odoo.tests import Form
 from odoo.addons.mail.models.mail_mail import MailMail as upstream
 from odoo.addons.mail.tests.test_mail_composer import TestMailComposer
 
-# When Odoo upstream function _send in the mall.mail model, that has been fully
+# When Odoo upstream function _send in the mail.mail model, that has been fully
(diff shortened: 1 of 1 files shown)


mail_composer_cc_bcc/tests/test_mail_cc_bcc.py
```diff
@@ -23,8 +23,9 @@
     "46172c91183f2cb50b22a6b3b5e3869b",
     "8f26c4084cc7fc300e64d19ccdc944fe",
     "db6cc0d3513a0c85bd716e4cb0a4d09c",
-    "458982c6cb3a347b13008f7d8130f633",  # 2025-04-08, odoo commit 3af276804101
+    "458982c6cb3a347b13008f7d8130f633",  # odoo commit 3af276804101
     "2e0c730330ae3bdf00c6bfd956b0527b",
+    "172e9656a6089deab948fe43d4bb31de",  # odoo commit 465f907835be
```
hbrunn, line 28: this commit is a change to `send`, not `_sent`: https://github.com/odoo/odoo/commit/465f907835be - this probably should point to some other commit?

hbrunn (changes requested):
still strikes me as odd that it should be really impossible to avoid this huge slap of copied code

PR #1828, opened 2026-03-17
[17.0] [ADD] mail_discuss_channel_hide_sidebar: New module

miquelalzanillas:
This module adds a "Hide from list" action in Discuss sidebar channels.

The action hides the channel for the current user without unsubscribing.
If a new message arrives in that channel, it appears again automatically.

@mpascuall @BernatObrador @ppyczko


ppyczko (approved):
LGTM! Tested in runboat.

BernatObrador:
Works as expected. The issue is that if you hide a channel and later want to send a message in it, you may run into a problem: if no one sends a message after you’ve hidden the channel, you won’t be able to make it visible again.

A possible solution could be to add a menu or section for hidden channels, or something similar.

miquelalzanillas:
> Works as expected. The issue is that if you hide a channel and later want to send a message in it, you may run into a problem: if no one sends a message after you’ve hidden the channel, you won’t be able to make it visible again.
> 
> A possible solution could be to add a menu or section for hidden channels, or something similar.

Thanks for the feedback. After hide a channel, you can access to the channel by browsing the channel in the search channel field or if the channel has a new message in the future it will appear again in the list.

miquelalzanillas:
Hi @hbrunn ,

What do you think about this module? Can we merge it?

Thanks,

mpascuall (approved):
LGTM!

PR #1830, opened 2026-03-18
[17.0][FIX] mass_mailing_partner: Allow multiple contacts with the same email address

victoralmau:
Allow multiple contacts with the same email address

There should be no restrictions (similar to what happens when multiple mailing contacts are created with the same email address without this module installed).

@Tecnativa TT61525


pedrobaeza (approved):
Yes, I think this is a historical debt when the model changed, allowing multiple mailing list per contact.

pilarvargas-tecnativa (approved):

victoralmau:
This is ready to merge.

pedrobaeza:
Expanded explanation of the why in https://github.com/OCA/mass-mailing/pull/17#issuecomment-4124660176

PR #1831, opened 2026-03-21
[17.0][FIX] mail_gateway_whatsapp: find existing template buttons 

bartacruz:
The method `mail.whatsapp.template._prepare_values_to_import()` is a model method (@api.model), so using **self** to search for existing buttons always returns an empty list.

When attempting to update templates from Meta, if a template with buttons already exists in the gateway configuration, the import process will fail because it cannot find the existing buttons and attempts to create them (again), causing the SQL constraint `unique_name_for_template` to fail.

This was resolved by searching for the buttons in the original template (using the UID).

The fix is based on 17.0 but it seems that it can be ported seamlessly to 18.0 also.

### DIFF
--- a/mail_gateway_whatsapp/models/mail_whatsapp_template.py
+++ b/mail_gateway_whatsapp/models/mail_whatsapp_template.py
@@ -327,7 +327,7 @@ class MailWhatsAppTemplate(models.Model):
             "gateway_id": gateway.id,
         }
         is_supported = True
-        template = self.search([("template_uid","=", json_data.get("id"))])
+        template = self.search([("template_uid", "=", json_data.get("id"))])
         for component in json_data.get("components", []):
             if component["type"] == "HEADER" and component["format"] == "TEXT":
                 vals["header"] = component["text"]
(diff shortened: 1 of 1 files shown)


pedrobaeza:
In fact, `self` contains the template. You just need to remove `@api.model`.

bartacruz:
> In fact, `self` contains the template. You just need to remove `@api.model`.

You're absolutely right. The problem wasn't  there, but in mail.gateway.button_import_whatsapp_template(), that calls the method using an empty recordset instead of the template...

```
def button_import_whatsapp_template(self):
    WhatsappTemplate = self.env["mail.whatsapp.template"]
[.....]
    ws_template = templates_by_id.get(template_data["id"])
    if ws_template:
        ws_template.write(
             ### HERE. It should be ws_template instead.
             WhatsappTemplate._prepare_values_to_import(self, template_data)
       )
```          
You don't even need to remove the `@api.model`  decorator.
I'll fix it later.
Should I ammend this PR or create a clean new one?

bartacruz:
Created a new PR with the (I hope) right fix to this.

PR #1833, opened 2026-03-25
[17.0][FIX] mail_gateway_whatsapp: Use the existing template to process imported data

bartacruz:
When attempting to update templates from Meta, if a template with buttons already exists in the gateway configuration, the import process will fail because it cannot find the existing buttons and attempts to create them (again), causing the SQL constraint unique_name_for_template to fail.
The reason being that even if the imported template exists, the code calls the update method using an empty recordset instead of the actual existing template.

I've previously proposed another solution to this in #1831 , but as @pedrobaeza noted, it was easier than that.

### DIFF
--- a/mail_gateway_whatsapp/models/mail_gateway.py
+++ b/mail_gateway_whatsapp/models/mail_gateway.py
@@ -56,7 +56,7 @@ class MailGateway(models.Model):
             ws_template = templates_by_id.get(template_data["id"])
             if ws_template:
                 ws_template.write(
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

luisDIXMIT (approved):
Simply change, LGTM!

pedrobaeza:
Can you please fw-port it to 18?

bartacruz:
> Can you please fw-port it to 18?

I will

PR #1835, opened 2026-03-25
[17.0][FIX] mass_mailing_partner: Add partner_unique option to lists

victoralmau:
Add partner_unique option to lists

The `partner_unique` field is added to the lists; the existing validation error (multiple contacts with the same partner) will only be displayed if this field is checked.

Related to https://github.com/OCA/social/commit/b769848f7704582cbb28b5c3f248d9bd388be426

@Tecnativa TT61525

### DIFF
--- a/mass_mailing_partner/models/mailing_list.py
+++ b/mass_mailing_partner/models/mailing_list.py
@@ -15,11 +15,16 @@ class MailingList(models.Model):
     partner_category = fields.Many2one(
         comodel_name="res.partner.category", string="Partner Tag"
     )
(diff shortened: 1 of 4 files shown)


pedrobaeza (commented):
Yes, if Odoo allows to have the same mail several times in the same list, we should allow to have the same partner in the same list. I'm now thinking that maybe we should provide this as a configuration option at list level.

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #1836, opened 2026-03-25
[18.0][FIX] mail_gateway_whatsapp: Use the existing template to process imported data.

bartacruz:
When attempting to update templates from Meta, if a template with buttons already exists in the gateway configuration, the import process will fail because it cannot find the existing buttons and attempts to create them (again), causing the SQL constraint unique_name_for_template to fail.
The reason being that even if the imported template exists, the code calls the update method using an empty recordset instead of the actual existing template.

It was fixed in 17.0 by #1833. This is a forward-port to 18.0


pedrobaeza (approved):
Thanks for the fw-port. Let's fast-track it as already merged in 17:

/ocabot merge patch

PR #1837, opened 2026-03-25
[17.0][IMP] mail_gateway_whatsapp: Add parent_id when message is a response

bartacruz:
When a message from whatsapp arrives and it's a response from a message sent from Odoo, it doesn't show as a response in the chat window. 
Basically, because the `parent_id` field is not set when posting the new message.

Code-wise, I've simply moved the `related_message` detection BEFORE calling message_post, and add the `parent_id` field to the method call.

### DIFF
--- a/mail_gateway_whatsapp/models/mail_gateway_whatsapp.py
+++ b/mail_gateway_whatsapp/models/mail_gateway_whatsapp.py
@@ -167,19 +167,9 @@ class MailGatewayWhatsappService(models.AbstractModel):
                 chat = chat.with_user(self.env.ref("base.public_user").id).with_context(
                     guest=author
                 )
-            # TODO: Check the sudo...
-            new_message = chat.sudo().message_post(
-                body=body,
-                author_id=author and author._name == "res.partner" and author.id,
-                gateway_type="whatsapp",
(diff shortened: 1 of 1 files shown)


bartacruz:
I've made a lot of small improvements to the whatsapp module, (audio files with player, image captions, actions for quickreply buttons, etc)  but I'm not sure if it's best to make lots of small PRs or just squash them and make one big PR.
@Tecnativa @pedrobaeza @percevaq @diegonc @carlos-lopez-tecnativa  would love your feedback!

pedrobaeza:
About this PR, is there a way to reduce the diff to achieve the same effect?

About the rest of the features, proposing them individually will help review and eventual merging them.

bartacruz:
Thanks for the feedback!

About this PR: The diff is bigger than the actual changes because I'm moving a block of code upwards and changing the indentation of another block. The only REAL change is adding the parent_id field to message_post()

I don't see a way of achieving the same result effectively. 

I mean, I can simply add the fetching of the related message before the call to message_post(), without touching anything else. It will work, the diff will be smaller, but it'll result in redundant code (the related message is fetched again afterwards...)

PR #1839, opened 2026-03-26
[18.0][FIX] mail_gateway: Stop using RecordCapturer

### LINKED ISSUE #1812: [18.0][mail_gateway]The module fails to install because it attempts to import the Odoo test.
Describe the bug
The module fails to install because it attempts to import the Odoo test framework within a business logic model. In Odoo, importing from odoo.tests in non-test mode is strictly forbidden to prevent security risks and memory issues.

The error specifically occurs in mail_message.py due to the import of RecordCapturer.

To Reproduce
Affected versions: 18.0

Use a clean Odoo 18.0 environment.

Install the mail_gateway_telegram module.

The installation will stop with an ImportError or a specific Odoo error log: "Importing test framework, avoid importing from business modules and when not running in test mode".

Expected behavior The module should install correctly without attempting to load the odoo.tests suite during standard execution.

Additional context
Log Traceback:

Python
File "/addons/social/mail_gateway/models/mail_message.py", line 6, in <module>
    from odoo.tests import RecordCapturer
File "/server/odoo/tests/common.py", line 99, in <module>
    _logger.error("Importing test framework, avoid importing from business modules...")
System info:

OS: Linux (Ubuntu 22.04)

Python version: 3.10

Odoo version: 18.0 Community

carlos-lopez-tecnativa:
RecordCapturer is only available in tests; it must be avoided in business models. Adapt the code to use a similar feature.

Closes https://github.com/OCA/social/issues/1812

@Tecnativa @pedrobaeza @percevaq @diegonc could you please review this?

### DIFF
--- a/mail_gateway/models/mail_message.py
+++ b/mail_gateway/models/mail_message.py
@@ -3,7 +3,6 @@
 
 
 from odoo import api, fields, models
-from odoo.tests import RecordCapturer
 
 from odoo.addons.mail.tools.discuss import Store
 
@@ -118,10 +117,10 @@ class MailMessage(models.Model):
             gateway_channel_id.gateway_token
         )
         channel = self.env["discuss.channel"].browse(chat_id)
-        with RecordCapturer(
-            self.env["mail.notification"], [("gateway_channel_id", "=", channel.id)]
-        ) as capt:
-            channel.message_post(**self._get_gateway_thread_message_vals())
+        messages = channel.message_post(**self._get_gateway_thread_message_vals())
+        notification = messages.notification_ids.filtered(
+            lambda n: n.gateway_channel_id == channel
+        )
         if not self.gateway_type:
             self.gateway_type = gateway_channel_id.gateway_id.gateway_type
         notification_vals = {
@@ -131,7 +130,6 @@ class MailMessage(models.Model):
             "notification_type": "gateway",
             "gateway_type": gateway_channel_id.gateway_id.gateway_type,
         }
-        notification = capt.records
         if notification:
             # Set the same gateway_message_id for both notifications.
             # When the webhook is received, both notifications must be updated.


pedrobaeza (approved):
I don't understand why such useful tools are not available outside tests...

carlos-lopez-tecnativa:
> I don't understand why such useful tools are not available outside tests...

According to this PR https://github.com/odoo/odoo/pull/177671 I understand the underlying reason, but in this case I think the `RecordCapturer` and other utility functions could be moved to another file. Can I try proposing a PR to Odoo?

pedrobaeza:
Ask first in that PR if they would accept it (obviously they will refuse it except maybe in master)

mail_gateway/models/mail_message.py
```diff
@@ -118,10 +117,9 @@ def _send_to_gateway_thread(self, gateway_channel_id):
             gateway_channel_id.gateway_token
         )
         channel = self.env["discuss.channel"].browse(chat_id)
-        with RecordCapturer(
-            self.env["mail.notification"], [("gateway_channel_id", "=", channel.id)]
-        ) as capt:
-            channel.message_post(**self._get_gateway_thread_message_vals())
+        domain = [("gateway_channel_id", "=", channel.id)]
+        current_records = self.env["mail.notification"].search(domain, order="id")
+        channel.message_post(**self._get_gateway_thread_message_vals())
```
Xavier-Do: I may be mistaken (didn't test it) but shouldn't something like that work ? 

```suggestion
        messages = channel.message_post(**self._get_gateway_thread_message_vals())
        notification = messages.notification_ids.filtered(lambda n: n.gateway_channel_id == channel)
```

I suppose that notifications are linked to the posted messages, so using a search diff to get them is maybe not needed.

carlos-lopez-tecnativa: Thanks so much for the suggestion. I applied it and adapted a few things, and it works."

PR #1842, opened 2026-03-28
[18.0][FIX] mail_gateway_whatsapp: Display an audio player on incoming audios 

bartacruz:
Right now, incoming whatsapp audios are shown as a regular attachment. That means that the only option is to download them and using an external app to reproduce them.

This small patch places a record (`voice: True`) in the attachment info so that, when processed by discuss, it recognizes it as a voice attachment and makes the necessary adjustments.

In short, this will convert this:
 
image

Into this:
image

### DIFF
--- a/mail_gateway_whatsapp/models/mail_gateway_whatsapp.py
+++ b/mail_gateway_whatsapp/models/mail_gateway_whatsapp.py
@@ -139,6 +139,10 @@ class MailGatewayWhatsappService(models.AbstractModel):
                     proxies=self._get_proxies(),
                 )
                 image_request.raise_for_status()
+                attachment_info = {}
+                if "audio/" in image_info["mime_type"]:
+                    # Tell discuss to treat this attachment as voice.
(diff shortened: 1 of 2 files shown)


bartacruz:
For reference, this is where the "voice" key is detected and the attachment is processed as such:

https://github.com/odoo/odoo/blob/3a54f1e4b706d6732d8dc16bc206f5fc4d6ab923/addons/mail/models/discuss/ir_attachment.py#L29-L32

pedrobaeza (commented):
Interesting improvement! Please squash all the commits into one, and you can include the explanation about the upstream reference where the audio is detected in the commit message itself. In general, the more explanation you give in the commit message, the better for our future we to find it and figure out the things.

bartacruz:
Done! 
Once it's approved I'll back-port it to 17.0

pedrobaeza (approved):

marcelsavegnago (approved):
❤️

PR #1846, opened 2026-03-31
[16.0][FIX] mail_gateway

etobella:
The goal of this PR is to add an index to make search faster and don't force partner, as the same id can be reused in several partners, as the id is the phone number and it is used for searching


luisDIXMIT (approved):
Code review, LGTM!

pedrobaeza (approved):
/ocabot merge patch

PR #1848, opened 2026-04-09
[18.0][FIX] mail_gateway_whatsapp: Add parent_id when message is a response.

bartacruz:
When a message from whatsapp arrives and it's a response from a message sent from Odoo, set the parent_id field.

In doing so, discuss correctly shows the message as a response, like this:
image

Also, having the parent_id set is very useful for detecting template's button responses.

### DIFF
--- a/mail_gateway_whatsapp/models/mail_gateway_whatsapp.py
+++ b/mail_gateway_whatsapp/models/mail_gateway_whatsapp.py
@@ -191,6 +191,7 @@ class MailGatewayWhatsappService(models.AbstractModel):
                     )
                     .mail_message_id
                 )
+                new_message.parent_id = related_message.id
                 if related_message and related_message.gateway_message_id:
                     new_related_message = (
                         self.env[related_message.gateway_message_id.model]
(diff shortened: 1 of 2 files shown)


bartacruz:
This is an enhancement of #1837.
As @pedrobaeza requested, I've found a way to reduce the diff, and provided a test case.
If/when this PR is approved, i'll do a 17.0 back-port.

mail_gateway_whatsapp/models/mail_gateway_whatsapp.py
```diff
@@ -191,6 +191,7 @@ def _process_update(self, chat, message, value):
                     )
                     .mail_message_id
                 )
+                new_message.parent_id = related_message and related_message.id
```
pedrobaeza:
```suggestion
                new_message.parent_id = related_message.id
```

pedrobaeza (commented):
Thanks. Please check pre-commit

pedrobaeza:
Please always squash these suggestions/changes into the existing commit.

bartacruz:
> Please always squash these suggestions/changes into the existing commit.
Yes, i'm on it

bartacruz:
Is there a way that I can run the pre-commit tests locally?
EDIT: I've found it. installed pre-commit  hooks.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #1849, opened 2026-04-13
[17.0][FIX] mail_gateway: Adapt onFocusIn signature

pedrobaeza:
Since https://github.com/odoo/odoo/commit/83c9c6168c7b40cee7329bf7c4cd0, Odoo has changed the signature of the JS method `onFocusIn`, provoking that the override on this module crashes.

Let's adapt the override to the new signature.

@Tecnativa TT61993


pedrobaeza:
Let's fast-track it, as this invalidates whole Odoo messaging system if the module is installed.

/ocabot merge patch

pedrobaeza:
Not yet on 18.0, but to watch https://github.com/odoo/odoo/pull/258713 when merged.

PR #1853, opened 2026-04-16
[ADD] mass_mailing_list_prune_blacklisted

SirPyTech:
Allow the user to massively remove blacklisted contacts from mailing lists.


mass_mailing_list_prune_blacklisted/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+Blacklisted contacts can be removed while creating a new mailing:
+
+1. Create a new mailing
+2. Select a miling list that contains blacklisted contacts: a button "Prune lists..." appears
```
quirino95, line 4: **Chore (non blocking)**: small typo here
```suggestion
2. Select a mailing list that contains blacklisted contacts: a button "Prune lists..." appears
```

quirino95 (approved):
Code review: LGTM.

Functional review: find a non-blocking issue. I've already approved but if you could, I'd like you to improve this thing:
When selecting a mailing list, I need to save before button "Prune lists" appears.
As consequence, if an user click on "SEND" button directly, won't never see "Prune lists".
Would it be possible to make the button appear when you choose the mailing list instead of waiting for the save?

mass_mailing_list_prune_blacklisted/readme/USAGE.rst
SirPyTech: :+1: thanks for catching this!

SirPyTech (commented):
> When selecting a mailing list, I need to save before button "Prune lists" appears.
As consequence, if an user click on "SEND" button directly, won't never see "Prune lists".
Would it be possible to make the button appear when you choose the mailing list instead of waiting for the save?

It needs a few more lines of code but I should have managed to show the button while creating/editing, please check.

quirino95:
Congratulations, it works fine and is now much more user-friendly IMO! 😄

mass_mailing_list_prune_blacklisted/wizards/prune_mailing_list.py
```diff
@@ -0,0 +1,52 @@
...
+        return default_values
+
+    def remove_blacklisted_contacts(self):
+        for mailing_list in self.list_ids:
+            list_blacklisted_contacts = self.blacklisted_contact_ids.filtered(
+                lambda c: mailing_list in c.list_ids
+            )
+            if list_blacklisted_contacts:
+                mailing_list.contact_ids = [
+                    fields.Command.unlink(list_blacklisted_contacts.id)
```
anusriNPS: **Issue:** Error seen due to typo when accessing black list contacts during unlink. Need to use `list_blacklisted_contact` otherwise below error seen when a mailing list having more than one blacklist contact. As single blacklist contact is used in testcases, this issue is not seen.

Image

mass_mailing_list_prune_blacklisted/tests/test_mailing.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2026 Simone Rubino - PyTech
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from .common import Common
+
+
+class TestMailing(Common):
+    def test_show_prune_wizard(self):
+        """Teh wizard created from a Mailing is correct."""
```
anusriNPS: **Typo:** Test

mass_mailing_list_prune_blacklisted/wizards/prune_mailing_list_views.xml
```diff
@@ -0,0 +1,63 @@
...
+    ~ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+-->
+<odoo>
+    <record id="prune_wizard_form_view" model="ir.ui.view">
+        <field name="name">Prune mailing list wizard Form view</field>
+        <field name="model">mass_mailing_list_prune_blacklisted.wizard</field>
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <span
```
anusriNPS, line 13: **Question:** Do we require this span block? As we don't display `Prune Lists` button when there are no blacklist contacts available as part of any selected mailing lists, user wouldn't be able to see this block displaying `The selected mailing lists have no blacklisted contacts`. So, it can be removed.

Similarly, `invisible` tag for other group also doesn't have any impact. Please correct me if my understanding is wrong.

anusriNPS (commented):
Code and Functional Review: Minor issue due to typo. Otherwise, PR looks great!!

mass_mailing_list_prune_blacklisted/wizards/prune_mailing_list_views.xml
anusriNPS: Understood this block would be required while executing action `Remove blacklisted Contacts`, so this thread can be resolved.

mass_mailing_list_prune_blacklisted/wizards/prune_mailing_list.py
SirPyTech: Nice catch, thanks! That was one `s` too many, I also updated the test, please check

mass_mailing_list_prune_blacklisted/tests/test_mailing.py
SirPyTech: :+1: thanks, please check the updated docstring

mass_mailing_list_prune_blacklisted/wizards/prune_mailing_list.py
anusriNPS: Work as expected and testcases are also intact.

anusriNPS (approved):
Code and Functional Review: LGTM

PR #1328, opened 2024-03-08
[16.0][MIG] mail_edit

ntsirintanis:

PR #1854, opened 2026-04-17
[MIG][16.0] mig mail_edit

NL66278:
Replaces https://github.com/OCA/social/pull/1328 as that one is based on an archived repository, and still needs fixes.


mail_edit/README.rst
```diff
@@ -0,0 +1,121 @@
...
+  user (the admin user has these rights by default)
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Go to *Message* *Chatter* in any model, click the Edit (pen) Icon to open up
+  the 'Edit or Move' wizard.
+* To edit, edit the Mail as necessary and click Save.
```
KKamaa, line 54: @NL66278 another thing I noticed while testing, it seems default v16.0 has Edit and Delete if you check this [line](https://github.com/odoo/odoo/blob/08786e318e28384ff81b20cad4fe32c4755751df/addons/mail/static/src/models/message_action_view.js#L165) and image below:

Feature:

image

Buttons:

Image

 maybe we can use the above?

mail_edit/views/compose_message.xml
```diff
@@ -0,0 +1,53 @@
...
+                                name="subject"
+                                groups="mail_edit.group_mail_edit_editor,mail_edit.group_mail_edit_superuser"
+                            />
+                        <field name="date" readonly="1" />
+                        <field name="message_type" readonly="1" />
+                        <field name="subtype_id" readonly="1" />
+                    </group>
+                    <group>
+                        <field
+                                name="destination_object_id"
```
KKamaa, line 23: @NL66278  I now see the move based on the destination_object_id and working with the function ``change_destination_object``

mail_edit/README.rst
KKamaa: I fixed this.

KKamaa:
Just a question on the security groups should they be applied to the buttons or how they were before is better? If they were to be applied we can use ``await this.env.services.user.hasGroup(...)`` just placing this for review to consider before merging.

NL66278:
> Just a question on the security groups should they be applied to the buttons or how they were before is better? If they were to be applied we can use `await this.env.services.user.hasGroup(...)` just placing this for review to consider before merging.
@KKamaa I think the functionality with the groups should remain the same, but I think you can best decide what the best implementation is.

Please smash all your commits into one.

KKamaa:
> > Just a question on the security groups should they be applied to the buttons or how they were before is better? If they were to be applied we can use `await this.env.services.user.hasGroup(...)` just placing this for review to consider before merging.
> > @KKamaa I think the functionality with the groups should remain the same, but I think you can best decide what the best implementation is.
> 
> Please smash all your commits into one.

I squashed the commits, I will leave the groups 'thing' for now to allow testing, and maybe later on we can check it out if need be

NL66278:
@KKamaa I installed this locally, but even after regenerating asset bundles I do not see any edit button (pencil) in messages. Is there anything I should do extra? I also checked my user had all the edit rights set.

KKamaa:
> @KKamaa I installed this locally, but even after regenerating asset bundles I do not see any edit button (pencil) in messages. Is there anything I should do extra? I also checked my user had all the edit rights set.

@NL66278 I tested it on a new local demo db and it works, see attached test below:

mail

Did you hover on the right side of the message when testing so that the action icons appear? Also maybe ran ./build on waft to pull latest Odoo v16.0 as you pull the branch latest.

NL66278:
> > @KKamaa I installed this locally, but even after regenerating asset bundles I do not see any edit button (pencil) in messages. Is there anything I should do extra? I also checked my user had all the edit rights set.
> 
> @NL66278 I tested it on a new local demo db and it works, see attached test below:
> mail
> 
> Did you hover on the right side of the message when testing so that the action icons appear? Also maybe ran ./build on waft to pull latest Odoo v16.0 as you pull the branch latest.

Yes, obviously I did a new build before testing. Also checked the commit hash to make sure I have the same code. If I move to the right of the message the two standard icons for react and mark todo appear. Several times build again, updated base, updated the module, nothing helps. Tried in Chromium and Librewolf... Racking my brain about what could be different, as obviously from your test the module works perfect.

NL66278:
@KKamaa I found it, the edit function only works on comment and email messages, not on notifications.

NL66278:
@KKamaa There where some formatting errors in the pot and po file. I will add a commit to solve that. This was also giving errors/warnings (not fatal) when loading the module.

NL66278 (commented):
👍 LGTM. Cannot approve as technically I submitted the PR, although @KKamaa really did all the work.

ntsirintanis (approved):

KKamaa (approved):

PR #1855, opened 2026-04-22
[15.0][IMP] mail_post_defer: consider force_send context

smorita7749:
Standard Odoo modules (e.g. sale.order) use with_context(force_send=True) to request immediate email delivery. However, mail_post_defer was not checking force_send in context, causing these emails to be deferred unexpectedly. This change adds the context check so that force_send=True in context is properly considered.

@qrtl QT6282

### DIFF
--- a/mail_post_defer/models/mail_thread.py
+++ b/mail_post_defer/models/mail_thread.py
@@ -12,8 +12,10 @@ class MailThread(models.AbstractModel):
     def message_post(self, **kwargs):
         """Post messages using queue by default."""
         _self = self
-        force_send = self.env.context.get("mail_notify_force_send") or kwargs.get(
-            "force_send", False
+        force_send = (
+            self.env.context.get("mail_notify_force_send")
+            or self.env.context.get("force_send")
+            or kwargs.get("force_send", False)
         )
         kwargs.setdefault("force_send", force_send)
         if not force_send:
--- a/mail_post_defer/tests/test_mail.py
+++ b/mail_post_defer/tests/test_mail.py
@@ -69,6 +69,30 @@ class MessagePostCase(MailCommon):
                 fields_values={"scheduled_date": False},
             )
 
+    def test_forced_context_with_template(self):
+        """A forced send via context is sent directly when using a template."""
+        with self.mock_mail_gateway():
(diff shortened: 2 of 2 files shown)


smorita7749:
Pre-commit error may be caused by https://github.com/OCA/social/pull/1856

AungKoKoLin1997 (approved):
Code Review: LG 👍

mail_post_defer/models/mail_thread.py
```diff
@@ -12,8 +12,10 @@ class MailThread(models.AbstractModel):
...
         _self = self
-        force_send = self.env.context.get("mail_notify_force_send") or kwargs.get(
-            "force_send", False
+        force_send = (
+            self.env.context.get("mail_notify_force_send")
+            or self.env.context.get("force_send")
+            or kwargs.get("force_send", False)
         )
         kwargs.setdefault("force_send", force_send)
         if not force_send:
```
yajo, line 21: suggestion: instead of the above change, just add this:

```suggestion
        if not force_send or self.env.context.get("force_send"):
```

Explanation: The context is already kept in sub-calls. We just need to honor the context if present, but we don't need to alter the kwarg.

yajo (changes requested):
Thanks, good catch!

mail_post_defer/models/mail_thread.py
smorita7749: Thanks for the review!

I believe this change is necessary.

`_notify_record_by_email` reads context only for
`mail_notify_force_send`, not `force_send`:

    force_send = self.env.context.get('mail_notify_force_send', force_send)

This PR reads `force_send` from context early so that
`kwargs.setdefault` propagates `True` to
`_notify_record_by_email`, and `emails.send()` is
actually called.

Please let me know if my understanding is correct.

yostashiro: I suppose @yajo meant `if not force_send and not self.env.context.get("force_send"):`, but it still doesn't look like it would work in `_notify_record_by_email`, since the `force_send` kwarg would be set to False and would prevail. I think the approach of this PR is correct.

yostashiro (approved):
LGTM.

PR #1857, opened 2026-04-27
[15.0][ADD] attachment_mimetype_restriction

AungKoKoLin1997:
This module blocks specific MIME types from being uploaded as attachments using content-based detection rather than filename extensions. This prevents uploading potentially dangerous files (executables, scripts, etc.).

@qrtl QT5832


attachment_mimetype_restriction/models/ir_attachment.py
```diff
@@ -0,0 +1,78 @@
...
+                    "mimetype": mimetype,
+                    "model": res_model or _("attachments"),
+                }
+            )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            self._validate_mimetype_from_vals(vals)
+        return super().create(vals_list)
```
yostashiro, line 70: What's the handling of JS and CSS files that come from module installation?

AungKoKoLin1997: I handled that.

NL66278 (approved):
👍 LGTM

SanamiK:
Here is a summary of what I tested and the issues I found.

What I tested
I tried to attach a non-allowed type file in Send Invoice dialog in the Invoicing module.

**Issue 1: Internal Odoo files are incorrectly blocked**
Before even attaching a file, internal Odoo operations were being blocked. Configuring the document layout (company logo, font, or colors) triggers a ValidationError for text/scss.
Invoicing_Send_Invoice
text


**Issue 2: Adding text/scss to the allowlist causes a different error**
As a workaround for Issue 1, I added text/scss to the allowlist. This time, an application/octet-stream error occurred. It appears that legitimate files(PDF) processed internally by Odoo are also being blocked.
application


**Issue 3: Users are not shown a meaningful error message in some cases**
When a user tries to attach a non-allowed file type, the generic message "Something horrible happened" is shown instead of a specific validation message. The user has no way to tell what went wrong.
スクリーンショット 2026-06-09 104009

SanamiK (approved):
Functional review. LGTM!

I checked the behavior with settings applied to all models and settings applied to individual models.

Image
Image

PR #1859, opened 2026-04-28
 [FIX][16.0] mail_gateway_whatsapp: Fixed webhook challenge 

rodmad85:
Fixed webhook challenge with META.


rvalyi (changes requested):
The diff is empty, it seems the second commit is canceling the 1st one... @rodmad85 can you check?

rodmad85:
The initial commit had the wrong message, so I ended up making another one, but from your review request I also saw that the comment in the code was in Portuguese and I've already corrected that as well.

PR #1719, opened 2025-09-04
[17.0][FIX] mail_gateway: notifications are added when linking a message from gateway discuss to a record

CarlosRoca13:
Before this changes when creating the new message when pressing link button, notifications where created too. So the message was notified to the record followers.

After this changes, the message is created but not notified to the followers.

cc @Tecnativa TT57889

ping @pedrobaeza @victoralmau

PR #1861, opened 2026-05-14
[18.0][FIX] mail_gateway: notifications are added when linking a message from gateway discuss to a record

CarlosRoca13:
Before this changes when creating the new message when pressing link button, notifications where created too. So the message was notified to the record followers.

After this changes, the message is created but not notified to the followers.

FWPORT of: https://github.com/OCA/social/pull/1719

cc @Tecnativa TT57889

ping @carlos-lopez-tecnativa @pedrobaeza


pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #1872, opened 2026-06-05
[15.0][ADD] mail_bulk_send

nobuQuartile:
@qrtl
QT6861


nobuQuartile:
I decided to use a standard function to add an E-mail template context action.

PR #1875, opened 2026-06-19
[17.0][FIX] mail_activity_board: fix computing related_model_instance

astirpe:
Issue: Loading the chatter crashes with a ValueError if a user lacks read access to an activity's related model. This happens because Odoo's ORM strictly validates assignments to Reference fields and rejects the restricted model.

Fix: Added a check inside _compute_related_model_instance. If the user cannot read the target model, the field safely defaults to False instead of crashing the entire view.

### DIFF
--- a/mail_activity_board/models/mail_activity.py
+++ b/mail_activity_board/models/mail_activity.py
@@ -2,6 +2,7 @@
 # Copyright 2018 ForgeFlow S.L.  <https://www.forgeflow.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import api, fields, models
+from odoo.exceptions import AccessError
 
 
 class MailActivity(models.Model):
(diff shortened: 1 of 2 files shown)


mail_activity_board/models/mail_activity.py
```diff
@@ -25,10 +25,22 @@ class MailActivity(models.Model):
...
+                and record.res_model
+                and record.res_model in allowed_models
+            ):
+                # Check the user read access before formatting the reference field
+                has_access = self.env["ir.model.access"].check(
+                    record.res_model, "read", raise_exception=False
+                )
+                if has_access:
+                    ref = f"{record.res_model},{record.res_id}"
             record.related_model_instance = ref
```
hbrunn: can't we catch an access error here instead?

hbrunn (changes requested):

mail_activity_board/models/mail_activity.py
astirpe: Yes indeed, I rewrote it, could you check again?

mail_activity_board/models/mail_activity.py
```diff
@@ -26,10 +27,16 @@ class MailActivity(models.Model):
...
-            ref = False
-            if record.res_id:
-                ref = f"{record.res_model},{record.res_id}"
-            record.related_model_instance = ref
+            try:
+                if record.res_id and record.res_model:
+                    self.env[record.res_model].check_access_rights("read")
+                    record.related_model_instance = (
+                        f"{record.res_model},{record.res_id}"
+                    )
```
hbrunn, line 35: shouldn't it be enough to wrap this?

astirpe: It's not enough, it would reintroduce the crash we are trying to fix.
The original crash didn't occur during the compute method, which didn't fail itself. The crash would simply happen a moment later: when the web client tries to render the view, it attempts to fetch the target record's display name.

hbrunn: but when will the else branch of this conditional throw any of the errors you catch here?

astirpe: you are right, I just moved the else branch to outside the try-catch, thanks!

hbrunn (approved):

PR #1881, opened 2026-07-09
[FIX] mail_gateway, mail_gateway_whatsapp: Fix channel member duplicates and template view crash

DevOpsMBAConsultings:
This PR fixes two critical issues with the `mail_gateway` and `mail_gateway_whatsapp` modules in Odoo 18.0.

### 1. UniqueConstraint Violation on `discuss.channel.member` Creation
**Module**: `mail_gateway`
When a webhook is received and the `author` (sender) happens to be one of the users already configured in `gateway.member_ids`, the previous logic resulted in the `author`'s `partner_id` being appended to the `members` list even though it was already iterated over from `gateway.member_ids.partner_id`. 
This caused Odoo's `discuss.channel.create` to fail with `psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "discuss_channel_member_partner_id_channel_id_uniq"`. 
**Fix**: Using a `set` for the member partner IDs and explicitly verifying that `author.id not in member_pids` before appending ensures `partner_id` is unique.

### 2. Owl Crash on Variables Notebook Page (`mail.whatsapp.template`)
**Module**: `mail_gateway_whatsapp`
When opening an approved template and clicking the "Variables" notebook page, the Odoo 18 Web Client completely crashes with:
`OwlError: An error occured in the owl lifecycle... Cannot read properties of undefined (reading '1') at get string ... at SelectionField.template`
**Cause**: The `variable_ids` list view has `model`, `line_type`, and `name` marked as `column_invisible="True"`. In Odoo 17/18, if a Selection field is `column_invisible`, the options are not sent to the frontend. Owl still tries to parse the `SelectionField` and crashes when the choices are empty.
**Fix**: Added `widget="char"` to the `column_invisible="True"` fields to bypass the `SelectionField` rendering entirely and resolve the Owl crash.

### DIFF
--- a/mail_gateway/models/mail_gateway_abstract.py
+++ b/mail_gateway/models/mail_gateway_abstract.py
@@ -47,16 +47,26 @@ class MailGatewayAbstract(models.AbstractModel):
 
     def _get_channel_vals(self, gateway, token, update):
         author = self._get_author(gateway, update)
+        member_pids = set(gateway.member_ids.mapped("partner_id").ids)
         members = [
-            Command.create({"partner_id": partner.id, "unpin_dt": False})
-            for partner in gateway.member_ids.partner_id
+            Command.create({"partner_id": partner_id, "unpin_dt": False})
+            for partner_id in member_pids
         ]
-        if author:
+        if author and author._name == "res.partner" and author.id not in member_pids:
             members.append(
                 Command.create(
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please one fix per PR.

PR #1882, opened 2026-07-09
[19.0][MIG] mail_history_mark_unread: Migration to 19.0

GuillermoNforgeflow:


mail_history_mark_unread/__manifest__.py
```diff
@@ -13,7 +13,7 @@
     "assets": {
         "web.assets_backend": [
             "mail_history_mark_unread/static/src/message_actions_patch.esm.js",
-            "mail_history_mark_unread/static/src/message_service_patch.esm.js",
+            "mail_history_mark_unread/static/src/message_model_patch.esm.js",
```
kalpeshgajera-uncannycs: In Odoo 18 & 19, all web client assets are native ES modules by default (which is why `/* @odoo-module */` is no longer needed). The `.esm.js` extension is obsolete legacy from 16/17; please rename the files to standard `*.js` and update `__manifest__.py`.

mail_history_mark_unread/static/src/mail_core_web_service_patch.esm.js
```diff
@@ -8,26 +6,19 @@ patch(MailCoreWeb.prototype, {
         super.setup();
         this.busService.subscribe("mail.message/mark_as_unread", (payload) => {
```
kalpeshgajera-uncannycs, line 7: In Odoo 19, bus notification handlers receive `(payload, { id: notifId })`. 
Directly incrementing `thread.message_needaction_counter` and overwriting `inbox.counter` without checking `notifId` can lead to desynced counts if bus notifications arrive out of order or across multiple browser tabs. 

https://github.com/odoo/odoo/blob/19.0/addons/mail/static/src/core/web/mail_core_web_service.js#L57-L89

mail_history_mark_unread/static/src/message_actions_patch.esm.js
```diff
@@ -1,30 +1,26 @@
...
-        if (message.isHistory) {
+        if (thread?.eq(store.history)) {
             return true;
         }
-        if (component.env.messageCard && !message.isNeedaction) {
+        if (owner.env.messageCard && !message.needaction) {
             return true;
         }
         return false;
     },
```
kalpeshgajera-uncannycs, line 17: In search mode, `owner.env.messageCard && !message.needaction` evaluates to `true` for almost every message (including messages in public channels or chatter where the user was never an inbox recipient).

When a user clicks "Mark as Unread" on such a message:
1. The backend `set_message_undone()` finds no `mail.notification` with `is_read = True` for the current partner and silently returns (see `models/mail_message.py:28`).

2. But the frontend still splices the card away from the search view, falsely indicating to the user that it was restored to their inbox.Notice that in core Odoo 19, message actions are scoped to the active mailbox thread rather than generic subEnv flags:

https://github.com/odoo/odoo/blob/19.0/addons/mail/static/src/core/common/message_actions.js#L113-L118

mail_history_mark_unread/models/mail_message.py
kalpeshgajera-uncannycs, line 21: In Odoo 19, `search_fetch()` is preferred here because we only need `mail_message_id`. 
Instead of loading the entire notification record from the database, `search_fetch` only grabs that one field in a single query, which is faster and uses less memory.

This matches how Odoo 19 core does it in `set_message_done()`: https://github.com/odoo/odoo/blob/19.0/addons/mail/models/mail_message.py#L901-L905

kalpeshgajera-uncannycs (changes requested):

PR #1883, opened 2026-07-09
[18.0][FIX] mail_gateway: Fix channel member duplicates

DevOpsMBAConsultings:
**Module**: `mail_gateway`

When a webhook is received and the `author` (sender) happens to be one of the users already configured in `gateway.member_ids`, the previous logic resulted in the `author`'s `partner_id` being appended to the `members` list even though it was already iterated over from `gateway.member_ids.partner_id`. 

This caused Odoo's `discuss.channel.create` to fail with `psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "discuss_channel_member_partner_id_channel_id_uniq"`. 

**Fix**: Using a `set` for the member partner IDs and explicitly verifying that `author.id not in member_pids` before appending ensures `partner_id` is unique.

### DIFF
--- a/mail_gateway/models/mail_gateway_abstract.py
+++ b/mail_gateway/models/mail_gateway_abstract.py
@@ -47,16 +47,26 @@ class MailGatewayAbstract(models.AbstractModel):
 
     def _get_channel_vals(self, gateway, token, update):
         author = self._get_author(gateway, update)
+        member_pids = set(gateway.member_ids.mapped("partner_id").ids)
         members = [
-            Command.create({"partner_id": partner.id, "unpin_dt": False})
-            for partner in gateway.member_ids.partner_id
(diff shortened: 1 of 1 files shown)


mail_gateway/models/mail_gateway_abstract.py
```diff
@@ -47,16 +47,26 @@ def _get_author(self, gateway, update):
...
-            for partner in gateway.member_ids.partner_id
+            Command.create({"partner_id": partner_id, "unpin_dt": False})
+            for partner_id in member_pids
         ]
-        if author:
+        if author and author._name == "res.partner" and author.id not in member_pids:
             members.append(
                 Command.create(
                     {
-                        "partner_id": author._name == "res.partner" and author.id,
```
pedrobaeza, line 58: Why are you removing the part that checks the model?

PR #1884, opened 2026-07-09
[18.0][FIX] mail_gateway_whatsapp: Fix Owl Crash on template view

DevOpsMBAConsultings:
**Module**: `mail_gateway_whatsapp`

When opening an approved template and clicking the "Variables" notebook page, the Odoo 18 Web Client completely crashes with:
`OwlError: An error occured in the owl lifecycle... Cannot read properties of undefined (reading '1') at get string ... at SelectionField.template`

**Cause**: The `variable_ids` list view has `model`, `line_type`, and `name` marked as `column_invisible="True"`. In Odoo 17/18, if a Selection field is `column_invisible`, the options are not sent to the frontend. Owl still tries to parse the `SelectionField` and crashes when the choices are empty.

**Fix**: Added `widget="char"` to the `column_invisible="True"` fields to bypass the `SelectionField` rendering entirely and resolve the Owl crash.

### DIFF
--- a/mail_gateway_whatsapp/views/mail_whatsapp_template_views.xml
+++ b/mail_gateway_whatsapp/views/mail_whatsapp_template_views.xml
@@ -1,23 +1,24 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-    <record id="view_mail_whatsapp_template_tree" model="ir.ui.view">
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please check CI.

PR #1889, opened 2026-07-28
[17.0][ADD] social_media_base

edescalona:
@BinhexTeam

T25184

This PR adds **Social Media Base**, the media-agnostic foundation for managing social media accounts and posts from Odoo.

It brings the models, the security, the scheduled actions and the common interface, but it does not connect to any network by itself: a connector module (*Social Media Linkedin*, *Social Media X*, ...) registers the media, implements the OAuth authorization and publishes for real.

### What it adds

- **Accounts** (`social.account`) with their OAuth credentials, visible only to administrators, and a responsible user that scopes what each user sees.
- **Posts** (`social.post`) with the state flow `draft → planned → publishing → published / cancelled / failed`, fanning out into one publication (`social.post.account`) per selected account.
- **Daily statistics** (`social.account.statistics`), one row per account and day, drawn with the native graph and pivot views. A cron rewrites the last days every two hours, because networks revise figures already past.
- **Per-network validation while the post is written**: each connector declares its limits once (message length, number of images and videos, formats, sizes) and the form shows what a network will publish differently and what it will not publish at all. Saving is never blocked; the publication is where an objection stops something, reading the very same rules.
- **Campaign integration**: a *Campaigns* menu inside the application, a UTM medium per network and a UTM source per publication, and the links of a published message routed through the Odoo link tracker, so a click from the network is counted in Odoo and attributed to the campaign and to the publication.
- **Security**: `group_social_media_user` (own accounts) and `group_social_media_manager` (all accounts).

### Design notes

- The cost line is deliberate: this module only ever spends a fixed number of calls per account, however much that account has published. Anything whose cost grows with the history of the account — importing the publications, their comments and reactions — belongs to *Social Media Sync*.
- Publishing is an irreversible external effect, so each publication runs inside its own savepoint: a failure on one account never rolls back the remote reference another one already obtained.
- `excludes: ["social"]`: these modules cannot coexist with Odoo's Social Marketing.

### Tests

Full test suite included, with no network calls — every HTTP call is patched.


jelenapoblet (approved):
@edescalona great job here. Splitting social_media and social_media_advertising was the right architectural decision.

jelenapoblet:
@ivs-cetmix ping

social_media_base/models/social_media.py
```diff
@@ -0,0 +1,54 @@
...
+    """Social media supported by an installed connector module."""
+
+    _name = "social.media"
+    _inherit = "social.media.base.mixin"
+    _description = "Social Media"
+
+    name = fields.Char()
+    description = fields.Text()
+    media_type = fields.Selection(
+        [],
```
ivs-cetmix, line 17: Not a bug but a hint: you can add selection as a lambda function - much easier to override and modify later.

```pyton 

media_type = fields.Selection(selection=lambda self: self._selection_media_type())

# And later

def _selection_media_type(self):
    return []
```

social_media_base/models/social_post_mixin.py
```diff
@@ -0,0 +1,65 @@
...
+# Sort key of a media that has no database identifier yet, so that it is
+# ordered after the stored ones instead of being compared with them.
+UNSAVED_MEDIA_ORDER = float("inf")
+
+
+class SocialPostMixin(models.AbstractModel):
+    _name = "social.post.mixin"
+    _description = "Social Media Post Mixin"
+
+    image_urls = fields.Char(compute="_compute_image_urls", store=True)
```
ivs-cetmix, line 46: Question: why not to inherit https://github.com/OCA/server-tools/tree/d6f73382749b8fc6be1311f6d566f1c48f0cbc34/base_multi_image for this?

social_media_base/models/social_media.py
edescalona: Thanks for the hint @ivs-cetmix. 

I kept the list form on purpose. A callable selection is incompatible with selection_add, which is how every connector plugs its own media in — Selection._setup_attrs in odoo/fields.py asserts values is not None and a callable leaves it None, so social_media_linkedin and social_media_x would fail at registry load and would each have to override the new _selection_media_type() instead.

A callable selection is also not reflected into ir.model.fields.selection (_reflect_selections only handles lists), so we would lose the ondelete cleanup policy on uninstall and the translation of the labels. The empty-list + selection_add pattern is what core uses.

social_media_base/models/social_post_mixin.py
edescalona: Good question. @ivs-cetmix 
I did look at it before going with attachments.

The module does not really fit what the posts need:

 - It covers images only (base_multi_image.image inherits image.mixin), while a post also carries video_ids — LinkedIn uploads videos through a multi-part flow. We would end up with two different media mechanisms in the same model.
 - It has no mimetype, only filename/extension. We validate the file kind before publishing (_check_media_kind) and each connector checks the mimetype against what the network accepts before uploading the bytes.
 - Its storage selection allows url and file: _get_image_from_url does an urlopen() inside a compute and _get_image_from_file reads an arbitrary server path. Reading a post's images would then be able to hit the network during the publishing cron, and its ACL grants full CRUD to base.group_user with no record rules, while our records are scoped by user_id.
 - UNIQUE(owner_id, owner_model, name) forbids two files with the same name on one record, and the copies we keep per account are named after the remote media.
 - The owner is a split many2one (owner_id Integer + owner_model Char), so there is no foreign key and the cascade is mimicked in Python. Our many2many to ir.attachment has a real FK with ondelete="restrict", and attachments carry res_model/res_id, which is what the sync-back path relies on when medias are downloaded from the network.

 Images and videos here are plain ir.attachment records: shared on the post, copied per account after publishing (_copy_medias_account), served as /web/image/{id}-{checksum} for immutable browser caching, and uploaded to the networks as raw bytes plus mimetype. Adding the dependency would not remove any of that code, so I would rather keep the module self-contained.

social_media_base/models/social_media.py
ivs-cetmix: Makes sense.

social_media_base/models/social_post.py
```diff
@@ -0,0 +1,773 @@
...
+        ondelete="restrict",
+        relation="social_post_image_rel",
+    )
+    video_ids = fields.Many2many(
+        "ir.attachment",
+        relation="social_post_video_rel",
+        column1="post_id",
+        column2="video_id",
+        ondelete="restrict",
+    )
```
ivs-cetmix, line 127: Storing media inside the attachments will grow the db dump. Especially taking into account that those images / videos have minumum value after the post is published. E.g. we are storing images in R2 in our custom addon that we are using for social media.
Possible solutions that I see:
- Ignore and use some custom modules to solve it. Pros - can merge faster, cons - potential mess in the future.
- Link to an intermediate model, eg `social.media.attachment` instead and than implement custom resolvers there (attachments, URLs etc).

jelenapoblet: I agree that storing all the media as attachments will increase the filestore size and the size and cost of backups.

System maintainers could manage the attachment lifecycle through a configurable cleanup policy. Of course, the storage approach could be improved in the future.

ivs-cetmix: Ok, let's have this merged asap and then will see how to improve it.

ivs-cetmix (approved):
LGTM

social_media_base/models/social_post_account.py
```diff
@@ -0,0 +1,737 @@
...
+from odoo.exceptions import ValidationError
+from odoo.service.model import PG_CONCURRENCY_ERRORS_TO_RETRY
+from odoo.tools import TEXT_URL_REGEX, plaintext2html
+
+from ..exceptions import SocialCredentialsError
+
+_logger = logging.getLogger(__name__)
+
+
+class SocialPostAccount(models.Model):
```
pedrobaeza, line 22: What is this model for? The description doesn't help.

social_media_base/models/social_post_mixin.py
```diff
@@ -0,0 +1,65 @@
...
+
+from odoo import api, fields, models
+
+# Sort key of a media that has no database identifier yet, so that it is
+# ordered after the stored ones instead of being compared with them.
+UNSAVED_MEDIA_ORDER = float("inf")
+
+
+class SocialPostMixin(models.AbstractModel):
+    _name = "social.post.mixin"
```
pedrobaeza, line 43: And this mixin? What is it for?

social_media_base/models/utm_campaign.py
```diff
@@ -0,0 +1,195 @@
...
+
+from odoo import _, api, fields, models
+
+_SOCIAL_GROUPS = "social_media_base.group_social_media_user"
+
+
+class UtmCampaign(models.Model):
+    """Social media publications promoting a marketing campaign."""
+
+    _inherit = "utm.campaign"
```
pedrobaeza, line 12: Shouldn't this go into `social_media_advertising`?

social_media_base/models/social_media_base_mixin.py
```diff
@@ -0,0 +1,110 @@
...
+
+from markupsafe import Markup, escape
+
+from odoo import _, models
+from odoo.http import request
+
+
+class SocialMediaBaseMixin(models.AbstractModel):
+    _name = "social.media.base.mixin"
+    _description = "Social Media Base Mixin"
```
pedrobaeza: Please explain how to use this mixin and which is their aim.

social_media_base/__manifest__.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2025 Binhex <https://www.binhex.cloud>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Social Media Base",
+    "summary": "Basic module for social media management.",
+    "version": "17.0.1.0.0",
+    "category": "Social Network",
+    "development_status": "Beta",
+    "license": "AGPL-3",
+    "author": "BinhexTeam,Odoo Community Association (OCA)",
```
pedrobaeza: Is this the author you want to put and that spelling?

pedrobaeza (commented):
On the previous PR, I expressed about using just `social_media`, without the `_base` suffix, and it's not adding value and just making the name longer.

Now there's no dependency over external graph libraries, but it still has a lot of JS/OWL around, which complicates the maintenance of the module. Why not simply putting the same aspect that the rest of Odoo? With 2 OWL migrations (2 and 3) pending, I think it's worth to reduce that layer.

social_media_base/social_utils.py
```diff
@@ -0,0 +1,261 @@
...
+from urllib.parse import quote, urlencode
+
+import pytz
+
+from odoo import _
+from odoo.exceptions import UserError
+from odoo.tools import DEFAULT_SERVER_DATE_FORMAT
+
+
+def convert_to_days(seconds=None, milliseconds=None):
```
pedrobaeza: This tools are too heavy IMO, and you should have already tools for this in Python standard/Odoo used libraries, and in any case, they should be IMO in the mixin.

social_media_base/social_utils.py
```diff
@@ -0,0 +1,261 @@
...
+                start = positions[int(repetition) - 1] - (
+                    (len(character_replace) - 1) * count
+                )
+                count += 1
+            fin = start + len(character_replace)
+            text_result[start:fin] = character_new
+    return "".join(text_result)
+
+
+def social_url_encode(
```
pedrobaeza: I feel that these methods are very platform dependent, so putting in the base are useless for the rest of the platforms.

edescalona:
Hi @pedrobaeza , thank you for your comments, I'll review them and get back to you as soon as possible.

pedrobaeza:
Coordinate with @jelenapoblet. I have been commenting with him some things.

social_media_base/models/social_post_account.py
edescalona: This `social.post.account` model stores the posts made on the social network by each selected account.

Example:

We have a `social.post` record with all its data, as shown in the image.

image

Then, when the "Publish" button is pressed, the post is sent to the social media for each selected account (4 in this case). This triggers the creation of 4 `social.post.account` records, each with its own statistics, social network identifier, etc. These posts are displayed in a tab below each account.

image

I will document it in the file for better clarity.

social_media_base/models/social_post_mixin.py
edescalona: This mixin became necessary to improve image loading and maintain the order of published images.

**`_compute_image_urls`**
It allows you to add the checksum to the image URL, so Odoo treats it as immutable, and the browser caches it and doesn't send a conditional request. If the checksum changes, then the request is made. This might seem trivial if there are only a few posts.

On a dashboard with dozens of posts, this makes the difference between over a hundred conditional requests per render and none at all. It's especially noticeable when clicking "Update": only the images from new posts brought in by the API are requested; existing images don't generate traffic.

The benefit is in the browser, not in the database. The `image_urls` field is `store=True`, so the URL travels with the record in the same way as always. What's saved are the HTTP requests to `/web/image/...`.


`**_sorted_medias**`
This ensures that the images are published on the social network in the same order in which the user uploaded them to the post. Without this treatment, Odoo reverses the order, and they wouldn't be displayed as the user intended. Adding `_order = "id asc"` would be possible, but it's not practical: simply installing this module would change the behavior across the entire system, and no core module redefines this attribute.

Example without this treatment:
Creating the post
social_post

Post published
social_post_account

Social media linkedin
linkedin

I will document it in the file for better clarity.

social_media_base/models/utm_campaign.py
edescalona: The relationship with the campaigns was added directly to the base module so as not to depend on another module that creates new tables that will not necessarily be used: a user who has no interest in configuring their own social network campaigns would end up installing structures that they never use.

social_media_base/models/social_media_base_mixin.py
edescalona: A mix that allows you to inform/notify the user about what's happening on the network. Inherited by `social.media`, `social.account`, and the account association wizards that each connector extends.

`_notify_user_client()` communicates via `bus.bus` for any event triggered from the web client;

`_notify_user_session()` stores the message in the session (delivered by `ir.http.session_info`) because OAuth callbacks respond with a redirect that would conflict with a message from the bus.

social_media_base/social_utils.py
edescalona: Agreed.

No existing tool covers it — `odoo.tools.date_utils` has no epoch conversion, and bare
`datetime.timestamp()` reads the ORM's naive UTC value as process-local time, so the
window shifts with the server TZ.

social_media_base/social_utils.py
edescalona: Correct. Those auxiliary signals now reside in their connector.

pedrobaeza:
Aborting, there is still this thing pending: https://github.com/OCA/social/pull/1891#issuecomment-5406495123

And I haven't checked the answers. Why are they marked as resolved?

social_media_base/models/social_media_base_mixin.py
pedrobaeza: You should reflect it in the class itself as comments, and the `_description` should reflect the summary, not simply transcribing the model name without dots. The same for the rest of the mixins.

pedrobaeza (changes requested):
Check my comments inline.

social_media_base/models/social_post_account.py
pedrobaeza: Please document it inline and reflect a proper `_description`.

social_media_base/models/social_post_mixin.py
pedrobaeza: Sorry, I don't get it, and why it cannot be in `social.post` itself.

social_media_base/social_utils.py
pedrobaeza: But epoch conversion is a need for a determined network, not in base, so you should add that tooling in the corresponding module that needs it.

pedrobaeza:
Complete about the mechanism (with the banner) with the trigger for the checks, the descriptions, and to move specific network things, and I won't oppose more to the merging :)

We can do it today!

social_media_base/social_utils.py
edescalona: I understand your comment. The idea behind including it in the base module is precisely to anticipate the possibility that another social network might need it in the future, thus avoiding duplication in each module. Therefore, I don't believe keeping it in the base module presents any problem.

social_media_base/models/social_media_base_mixin.py
edescalona: There's already a docstring in the class with an explanation, and the same goes for the other mixins. I'll adjust the description when I upload the improved checks.

social_media_base/social_utils.py
pedrobaeza: Yes, it does, because several things:

- It increases the size and maintenance of the base module while you may not use the network with such need. 
- You may not be able to check is correctness, apart from a theoretical usage, as the code from the network will come in another module.

Thus, I insist to put it in the specific network module. If several of these networks require this conversion, which I doubt, repeating it is not so bad, taking into account that sometimes the incoming format is slightly different and the amount of code is low. Respect the encapsulation principle.

social_media_base/models/social_post_mixin.py
edescalona: The logic cannot be limited to social.post, as social.post.account also inherits it. Both models contain the post images and must render them in the same order, so `_sorted_medias` and `image_urls` are shared: social.post uses them for previewing, and the connectors use them when publishing. If they were defined only in social.post, social.post.account would have to duplicate them.

social_media_base/social_utils.py
edescalona: Resolved

social_media_base/models/social_post_mixin.py
pedrobaeza: But why having the same images in both places? Shouldn't the message content be unique?

edescalona: The mechanism is that the social.post model is a 'template' for the posts, since from it as many posts as selected accounts will be generated; that is, the social.post.accounts are generated and these will have their own data

edescalona: Furthermore, if you link the account and import your posts from the social network, then social.post.account files are generated for each post, containing your data, and in that case, social.post is not created.

pedrobaeza: Yeah, I understand that, but the content of the template is the same, so why duplicating it?

edescalona: The X rows in ir.attachment need to be created because when an image is published, each social network returns an image identifier. The same image for the same social network returns different identifiers for different accounts, and these identifiers are added in Odoo as the 'name' of the row. This allows for future comparisons to determine if a new image was uploaded and avoids data duplication. Furthermore, at the local storage level, it points to the same filestore, thus preventing multiple copies of the same file.

Same image repeated 5 times in the attachment and the same store_fname:

image

Does this explain why the same file appears N times in ir.attachment?

social_media_base/models/social_post_account.py
edescalona: The _description and the Docstrings are already there

social_media_base/models/social_post_mixin.py
pedrobaeza: Yeah, I know the space is not duplicated due to the hash filename technique in the attachments, but it's more a question of having the ideal database structure, and  repeating the same content several times is not ideal. You are mentioning an identifier for the images, but I don't get why is that important. If the same image is shared across multiple posts, just upload it again and let the network to manage that. Or is there any other reason?

pedrobaeza: I'm using [DRY](https://es.wikipedia.org/wiki/No_te_repitas) and [KISS](https://es.wikipedia.org/wiki/Principio_KISS) principles for stating these things.

edescalona: The identifier has nothing to do with uploading — it is for the inbound direction, the synchronization.

Every publication in Odoo is kept in sync with what is on the network: a check every 2 hours and a full resync every week, per account. On each pass the API returns the posts with the ids of their medias in the payload. Matching those ids against the name of the attachments tells us, without a single HTTP call, which medias we already have locally. Drop the identifier and there is no cheap way left to answer that question: the only alternative is to download every image of every post on every pass and compare the bytes, so the weekly resync would re-download the whole timeline of every connected account. That is API quota and traffic, not disk.

Record 136 above has post_id empty: it was never written in Odoo, it was imported from LinkedIn. There is no local image to upload — the image came from the network, and its LinkedIn URN is the only stable handle we have on it.

pedrobaeza: So you are talking about posts not done on Odoo, but done directly on the social network? I think that resync is too many effort and have a lot of drawbacks. It's the same as in an e-commerce connector, you create both in the e-commerce and in Odoo the products. You should choose one side (Odoo in this case), and manage everything in that side. If not, a lot of effort, problems, and maintenance (both in code and in the day to day) needs to be put.

And anyway, even if you keep sync (which as said, I think it should be dropped and kept apart from this base), you can just store that attachments with their IDs, not the whole message content. And I suppose there's a message identifier that you can just store without storing the whole message.

I think this overcomplicates too much everything.

edescalona: If I'm referring to synchronization from the social network, the important thing here is that while the publication isn't on the social network, no impressions, likes, comments, etc. are generated. Therefore, in order to review them in Odoo, it's mandatory at some point to synchronize that data from the social network. Similarly, cron jobs can be deactivated, and the user can synchronize them when they deem it convenient; it's their decision.

edescalona: Where I do agree: the base must not grow network-specific behaviour. It does not — it defines the models and the hooks, and each connector fills them.

edescalona: Another way is to synchronize posts without the default media (by creating a configuration) that are brought from the social network and were not created in Odoo, but the user will not see the post with the correct data instantly, they will have to check it on the social network, a link that already exists today when pressing a post in the Dashboard.

jelenapoblet: As agreed we will move the synchronization of posts created outside odoo to dedicated modules:

- social_media_synchronization
- social_media_synchronization_linkedin
- social_media_synchronization_x

This will keep social_media_base focused on publications managed from odoo, while inbound synchronization remains an optional feature.

pedrobaeza: Thank you! Waiting for the change for the latest review and merge!

edescalona: Hi @jelenapoblet @pedrobaeza, maybe we can shorten the names a bit and leave them as:

- social_media_sync
- social_media_sync_linkedin
- social_media_sync_x

What do you think?

jelenapoblet: For me OK

pedrobaeza: Or maybe call it `import` instead of `sync`/`synchronization`

pedrobaeza: What you prefer

edescalona: I think sync is more appropriate, because it not only imports posts, but also synchronizes comments, statistics, etc.

jelenapoblet: Sounds good, go for it. Initially I would remove it from base so we can start merging this, and then create the 3 modules for sync

social_media_base/__manifest__.py
```diff
@@ -0,0 +1,51 @@
...
+    "name": "Social Media Base",
+    "summary": "Accounts, posts and a daily statistics series for the connectors",
+    "version": "17.0.1.0.0",
+    "category": "Social Network",
+    "development_status": "Beta",
+    "license": "AGPL-3",
+    "author": "Binhex, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/social",
+    "maintainers": ["edescalona"],
+    "depends": ["base", "web", "mail", "utm", "link_tracker"],
```
ivs-cetmix, line 13: why to depend on web, mail and base if the other modules already depend on them?

ivs-cetmix (changes requested):

social_media_base/__manifest__.py
edescalona: Hi @ivs-cetmix , I'm making changes to this PR, with all these adjustments, regards

ivs-cetmix (approved):
LGTM

edescalona:
Hello @pedrobaeza , the adjustments discussed in the comment thread have been applied, along with some additional improvements. I would appreciate your review whenever convenient. The remaining modules will be pushed shortly and I will keep you informed.

jelenapoblet:
Thanks, this now follows the agreed separation. social_media_base remains focused on publishing from Odoo and refreshing the related statistics, while posts created outside Odoo will be handled by the optional social_media_sync module.

This keeps the base simpler while preserving synchronization as an optional feature. Great work!

pedrobaeza:
@sbidoul and this name to be reserved too, please.

social_media_base/readme/CONFIGURE.md
```diff
@@ -0,0 +1,18 @@
+Tracked links.
+---------------
```
pedrobaeza: Is the underlining working without being the same length of the text?

```suggestion
Tracked links
-------------
```

social_media_base/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,2 @@
+- [Binhex](https://www.binhex.cloud):
+  - Edilio Escalona Almira <e.escalona@binhex.cloud>
```
pedrobaeza, line 2: No more persons have collaborated? I think that at least Jorge should be here (and me if you consider my design comments as relevant enough, but not required).

pedrobaeza (approved):

social_media_base/readme/CONTRIBUTORS.md
edescalona: Agreed, their contributions were important; I'll add them now.

edescalona:
ready @pedrobaeza

jelenapoblet:
I think we can try to merge

pedrobaeza:
@jelenapoblet @edescalona I'm now seeing that there's one missing comment in CREDITS.rst: that the initial work was funded by AEOdoo.

edescalona:
OK @pedrobaeza , should I create a PR for that?

pedrobaeza:
Yes, please

PR #1890, opened 2026-07-28
[17.0][ADD] social_media_linkedin

edescalona:
@BinhexTeam

T25184

### What it does

Adds `social_media_linkedin`, the LinkedIn connector of the `social_media_*` set. It plugs into `social_media_base` and covers everything whose cost against the API is fixed per account or per publication: association, publishing and
the daily figures of the page.

Reading back what the page already published — importing the publications, their comments and reactions — grows with the history of the account and is therefore *not* here: it lives in`social_media_linkedin_sync`.

### Scope

- **Accounts.** Only company pages (organizations) where the authorizing member   is an `ADMINISTRATOR` with the invitation already `APPROVED`. Personal profiles are not supported. Association and re-authorization go through the wizard inherited from `wizard.social.account` plus the `/linkedin/callback` controller.
- **Publishing.** Text, images or one video. LinkedIn takes either the images or the video, never both: a post carrying a video is published without them, and both the preview and a banner on the form say so before it is sent. Limits checked on the form and again at publish time: 3000 characters, 20 images of at most 10 MB (JPG/PNG/GIF), one MP4 of at most 500 MB.
- **Statistics.** The figures LinkedIn reports by day are written as a time series, with graph and pivot views over them. The query is asked for the organization as a whole and never carries the URNs of the publications, so the number of calls depends on the width of the window, not on how much the page published.
- **Existence check.** Opening a publication reads it on LinkedIn first. Only a `404` marks it `deleted`; a lost role or a throttled application leaves the line untouched — a publication is not gone just because Odoo could not read it.

### Notes for the reviewer

- All HTTP goes through `social.account._request_linkedin()`, the only place an authentication failure becomes `SocialCredentialsError` and the seam the tests patch.
- Endpoints, scopes, URNs, mimetypes and limits are in `social_linkedin_utils.py`. The API version is the `LinkedIn-Version` header (`_VERSION_STRING_LINKEDIN`), not the URL.
- OAuth scopes (`_SCOPE_LINKEDIN`) are all-or-nothing on the LinkedIn side: a scope the App was not granted fails the whole consent redirect, so the list is also the set of products the customer's App must hold.
- `uninstall_hook` calls `remove_social_media()` so the `social.media` record does not survive with a `media_type` no longer in the selection.
- Tests are `post_install` (the connector's `media_type` is not a legal value of the selection at install time) and make no network call.
- No new Python dependency: plain `requests`.

### Known limitations (also in ROADMAP)

- Visibility, feed distribution, targeting and third-party distribution are fixed in the code.
- Video duration, codecs, dimensions and aspect ratio are not validated before upload; LinkedIn rejects them during processing.
- Rate limits are not handled: a throttle error is recorded as a failure and has to be retried by hand.


edescalona:
ping @ivs-cetmix @jelenapoblet @antoniodavid

edescalona:
Hi @pedrobaeza @ivs-cetmix, this PR adds a new module: social_media_linkedin. It is ready for review whenever you have time. I look forward to any questions or changes you deem necessary.

jelenapoblet (approved):
Functional review done.

I tested the LinkedIn integration, including the account connection and publishing flow, and everything worked as expected on my side.

From a functional perspective, LGTM.

PR #1891, opened 2026-07-28
[17.0][ADD] social_media_x

edescalona:
@BinhexTeam

T25184

Integration of the X social media.

### What it adds

- **Account integration** through the X OAuth flow, with the credentials stored on `social.account`.
- **Publishing** on X from the standard post form, with images or a video.
- **Statistics on the card of the account**: the publications of the last 30 days are refreshed once a day and the card adds up the counters its publications hold. *Social Media X Sync* widens what is known (the publications Odoo did not send, and the older history), but the card is not empty without it.
- **Reposts and quotes**, the two figures only X reports, shown in the *Statistics* dialog of a publication and counted towards its *Interactions* and *Engagement*.
- **Limits checked while the post is written**: **280** characters without X Premium and **25 000** with it, **4 images** of at most **5 MB** each (**15 MB** for a GIF) in JPG, PNG, WEBP or GIF, or **one video** of at most **512 MB** in MP4. X takes images or a video, never both, so a post mixing them is refused instead of warned about. The same rules refuse the publication if the post reaches it through an import or an RPC call.

### Design notes

- **An X account never draws a time series in *Social Media* > Statistics.** That screen reads a history per day, and the [metrics](https://docs.x.com/x-api/fundamentals/metrics) the API v2 answers are the current lifetime counters of each post, with no way of asking for them grouped by day. Repeating the lifetime counter per day draws a cumulative curve of a different shape than the other networks on the same graph, and storing the delta between two readings is false for any day the scheduled action did not run. Neither is a measurement, so nothing is written and the figures live on the card of the account.
- The **X Premium** switch is declared by hand: nothing reads the plan back from X, so an account wrongly marked as Premium sends a post X refuses and the publication is left as *Failed* with the reason X gives.

### Note on the X API

This module was designed and developed **before** X restricted its API access policy. Back then, a free developer account was enough to link an account, publish posts and read back their statistics.

As of today the **Free** access level only grants a very limited set of write endpoints and no read access to posts/metrics, so reviewing/testing this module end to end requires at least a **paid (Basic) X API plan**. This is a platform restriction, not a limitation of the module.

References:
- Access levels and limits: https://docs.x.com/x-api/introduction
- Current plans: https://developer.x.com/en/portal/products

image


pedrobaeza:
How do you deal with the number of characters limitation in a multi-network post? Is there any generic mechanism for warning that in a specific network, the post is not going to be preserved (or directly avoid to post)

edescalona:
Hi @pedrobaeza 

Currently, no such mechanism exists, neither generic nor specific: the module
does not check the message length before posting. If the text exceeds
a network's limit, X rejects the request, and the failure is logged only on
that post line (`social.post.account` changes to `failed`).

If deemed appropriate, a field could be added to `social.media` that defines the character limit per network, and this field could be registered in the base module, with each social network-specific module also registering it.

Similarly, there is currently a limitation: it must be a paid version to use it (details in the PR description).

pedrobaeza:
I was thinking more on a generic "check" method that each specific account module overrides and add their rules, and if failed, when redacting the post, the warning/error appears and you decide before clicking on "Post".

edescalona:
@pedrobaeza 

Yes, that's also valid and remains open to future restrictions.

Should I add that option now or in the future?

pedrobaeza:
Given the limitation in this network, I would implement it already, as it will be a constant pain if not.

edescalona:
@pedrobaeza 

Don't you think it's better to display a banner message instead of a dialog box every time there's a problem with the edit?

image

pedrobaeza:
Yes, that's what I was saying... when I mentioned a warning or an error, I didn't talk about a dialog, but a warning/error banner.

edescalona:
Hi @pedrobaeza @ivs-cetmix, this PR adds a new module: social_media_x. It is ready for review whenever you have time. I look forward to any questions or changes you deem necessary.

social_media_x/readme/CONFIGURE.md
```diff
@@ -0,0 +1,139 @@
+To configure this module, you need to:
+---------------
+Please note that you must have a developer account.
```
pedrobaeza, line 3: Isn't there any option to not use a developer account? Anyway, how an account becomes a developer account?

And only paid access to API?

edescalona: No, there is no way around it, and I have documented these three points in CONFIGURE.md.

A developer account is required. It is the only place where API keys are generated. While third-party services exist, they are not recommended: read-only services (like TwitterAPI.io or SocialData) scrape X and cannot post, whereas services that do post (such as Ayrshare, Postproxy, and standard schedulers like Buffer or Hootsuite) pay for the same API and resell access, meaning the cost remains and includes their markup. In any case, a third party would stand between Odoo and the customer's data—something this module should not impose.

Access is paid, but not via subscription. X eliminated the free tier and the old Basic/Pro plans; what remains is a pay-as-you-go model with no monthly fee or minimum purchase requirement—the application simply requires a positive credit balance. A database that only posts content incurs costs of just cents per month; the factor that drives up costs based on account history is the data read by `social_media_sync`, which is a separate module.

Cost details based on the specific endpoint have been added to the CONFIGURE.md file.

jelenapoblet: I was planning to do a functional review of this PR, but since testing the X integration requires paid API access, I won't be able to complete the test on my side.

PR #1895, opened 2026-07-30
[17.0][ADD] social_media_advertising_linkedin

edescalona:
@BinhexTeam

T23184

Implementation of `social_media_advertising` for LinkedIn Ads. Depends on `social_media_advertising` and `social_media_linkedin`.

### What it does

- **Advertising accounts.** *Fetch advertising accounts* on the LinkedIn account brings the ones the authorized member holds a role on, with their LinkedIn status, serving status, type, owner and `test` flag, plus an *Open advertising account* button to the Campaign Manager.
- **Campaign groups and campaigns.** Created on LinkedIn in `DRAFT` from Odoo (*Create in LinkedIn*), updated (*Update in LinkedIn*) and ended (*Archive in LinkedIn* — LinkedIn's API offers no deletion), or imported with
  *Fetch campaigns*. Values are validated together before the call, so the errors come out at once instead of one round trip at a time.
- **Sponsored creatives.** Publishing a post linked to a campaign attaches it to that campaign as a creative. The import matches every creative with the publication it promotes and resolves the campaign of the publications brought
  from the wall, which the Posts API does not return.
- **Ads.** *Delete ad* requests the deletion and reads the creative back, so the status shown is the one LinkedIn reports and not the one that was asked for.

### Design notes

- LinkedIn statuses (Draft, Active, Paused, Archived, Canceled, Completed, Pending deletion, Removed) are `social.stage` **data** of this module, as the generic layer requires, not a hardcoded selection. They are `noupdate`.
- Advertising mirrors the network: records LinkedIn no longer accepts changes on become read-only in Odoo, deleted campaigns and ads are kept as history (the API keeps answering them because of their performance data), and figures carry their statistics window.
- Every call goes through `social.account._request_linkedin`; endpoints, versions, page sizes and URN helpers live in
  `social_advertising_linkedin_utils.py`.
- Import priority is hybrid: LinkedIn-only fields (stage, test flag) are always refreshed, while local pending changes on the editable ones are preserved and the LinkedIn values logged in the chatter.

### Upgrade note

The module adds the `r_ads`, `rw_ads` and `r_ads_reporting` scopes to the inkedIn authorization, which belong to the **Advertising API** product of the LinkedIn application. A token keeps the scopes it was issued with, so accounts
associated before installing this module must be authorized again (*Update account* with *Update keys*). `post_init_hook` posts that on their chatter and the account form shows a banner with the missing scopes.

### Not covered (see ROADMAP)

Carousel ads (so a multi-image post cannot be sponsored), targeting and bidding (defined in the Campaign Manager), and three of the seven LinkedIn objectives (lead generation, website conversions, job applicants) which need configuration living on LinkedIn.

### Tests

`social_media_advertising_linkedin/tests/`, all `post_install` and fully patched — no network call.


edescalona:
ping @ivs-cetmix @jelenapoblet @antoniodavid

edescalona:
Hi @pedrobaeza @ivs-cetmix, this PR adds a new module: social_media_advertising_linkedin. It is ready for review whenever you have time. I look forward to any questions or changes you deem necessary.

social_media_advertising_linkedin/models/social_advertising_campaign.py
```diff
@@ -0,0 +1,627 @@
...
+        """Check that the stored campaign URN still exists on LinkedIn.
+
+        :param account: The social.account used to call the LinkedIn API.
+        :return: The campaign URN if it exists, False otherwise.
+        :rtype: str | bool
+        """
+        self.ensure_one()
+        if not self.remote_ref:
+            return False
+        ad_account_id = account._require_linkedin_ad_account_id()
```
szalatyzuzanna, line 301: The advertising account here is resolved from the one currently in use rather than from the advertising_account_id stored on the record, unlike _linkedin_creative_endpoint, which documents the opposite rule and prefers the stored one — so once the user switches the advertising account in use, this lookup hits the wrong advertiser and the 404 is read as "the campaign does not exist", creating a duplicate; same in action_update_linkedin (:543) and action_archive_linkedin (:594) here, and in both of them on the campaign group (social_advertising_campaign_group.py:309 and :359), where it surfaces as a generic update/archive failure instead.

social_media_advertising_linkedin/tests/test_social_account_advertising_linkedin.py
```diff
@@ -0,0 +1,1173 @@
...
+                ads_ids, self.start_datetime, self.end_datetime
+            )
+            self.assertEqual(result, expected_result)
+            mock_get_linkedin_statistics.assert_called_once()
+
+    @patch(PATCH_ACCOUNT_LINKEDIN.format("_request_linkedin"))
+    @patch(
+        PATCH_ADVERTISING_ACCOUNT_LINKEDIN.format("_get_linkedin_advertising_account")
+    )
+    def test_action_import_campaigns(self, mock_advertising, mock_request_linkedin):
```
szalatyzuzanna, line 254: This test only ever imports one campaign group, so it would not catch a regression in the positional pairing between the records returned by create() and their URNs in _upsert_linkedin_campaign_groups — worth importing several at once, with names whose alphabetical order differs from the order they arrive in.

social_media_advertising_linkedin/models/social_advertising_campaign.py
```diff
@@ -0,0 +1,627 @@
...
+        :param account: The social.account used to call the LinkedIn API.
+        :param advertising_account_urn: The advertising account URN.
+        :return: The campaign group LinkedIn URN.
+        :rtype: str
+        """
+        self.ensure_one()
+        if not self.campaign_group_id:
+            raise UserError(
+                _(
+                    "The campaign %(campaign)s does not belong to a campaign " "group.",
```
szalatyzuzanna, line 257: This translatable string is split into two adjacent literals with a stray space between them, which leaves an odd entry in the .pot — same at :283.

szalatyzuzanna (commented):
Thanks for this contribution. Campaigns and campaign groups resolve the advertising account from the one currently in use rather than from the advertising_account_id stored on the record, so they address the wrong advertiser once the user switches accounts — and in _linkedin_verify_campaign the resulting 404 is read as "the campaign does not exist", creating a duplicate; the ads already do this right and document the rule, so the fix is to follow it in the other two models, in the five places I marked. I could only review the code, so I may be wrong about how the API responds here. The other two comments are minor.

edescalona:
Hi @szalatyzuzanna , thanks for your comments; I'll review them and get back to you.

edescalona:
Hi @szalatyzuzanna , the review comments are addressed. Could you take another look at the changes when you have a moment?

The failing test checks are not caused by this module.

szalatyzuzanna (approved):
Thanks @edescalona, all three points are addressed. Code review only, LGTM.

jelenapoblet (approved):
Tested, functional review ok

PR #1896, opened 2026-08-04
[15.0][FIX] attachment_mimetype_restriction: skip binary field storage

AungKoKoLin1997:
Binary/image fields default to `attachment=True`, so the ORM stores their values as `ir.attachment` records with `res_field` set — `image_1920` and its resized variants on partners/products, the company logo, the company favicon. Nothing in `_validate_mimetype_from_vals` exempted those, so they were validated against the allowlist even though they are field storage rather than user-facing uploads, and their content type is chosen by Odoo rather than by the uploader.

The effect is that a configured allowlist which does not happen to cover those content types breaks ordinary record saves. Creating a company fails unconditionally, because `res.company.create()` injects a default favicon (`res_company.py`) whose mimetype no allowlist would reasonably include:

```python
if not vals.get('favicon'):
    vals['favicon'] = self._get_default_favicon()
```

This PR returns early when `res_field` is set, and adds `res_field` to the vals dict rebuilt in `write()` so the update path is covered too — replacing an existing image goes through `atts.write({'datas': value})` in `fields.Binary.write()`, not through `create()`.

Added a regression test that sets `res.partner.image_1920` under a `text/plain`-only allowlist. It fails on 15.0 today with `ValidationError: File type 'image/png' is not allowed.` and passes with this change.

Restriction of user-uploaded attachments (chatter uploads, incoming mail attachments) is unchanged.

Three smaller fixes on the same validation path are included:

- `write()` validated the mimetype stored on the record instead of the content being written, so a replacement file was checked against the old type. The name is now passed and the stale mimetype dropped when `datas`/`raw` is present, so `_compute_mimetype` works on the incoming bytes.
- The `upload_attachment` controller raised `AttributeError` on every blocked upload, since the core method returns a `str` and not a `Response`.
- `test_message_post_filters_blocked_attachment_ids` looked the kept attachment up by `res_model`, which `message_post()` only sets for pending attachments, so the assertion never found it. It now asserts on the attachments of the posted message.

Assisted-by: Claude Opus 5

@qrtl QT5832

### DIFF
--- a/attachment_mimetype_restriction/models/ir_attachment.py
+++ b/attachment_mimetype_restriction/models/ir_attachment.py
@@ -43,6 +43,8 @@ class IrAttachment(models.Model):
     def _validate_mimetype_from_vals(self, vals):
         if self.env.context.get("install_mode"):
             return
+        if vals.get("res_field"):
+            return
         # Skip framework-generated assets: compiled bundles (detected at create
         # by res_model='ir.ui.view' + public=True, since their /web/assets/ url
         # is only set in a later write) and customized scss/js overrides (which
@@ -51,13 +53,13 @@ class IrAttachment(models.Model):
             "url"
         ):
             return
-        mimetype = self._compute_mimetype(vals)
         res_model = vals.get("res_model")
         company_id = self._resolve_attachment_company_id(vals)
         allowed_mimetypes = self._get_allowed_mimetypes(company_id, res_model)
         if not allowed_mimetypes:
             return
-        if mimetype.lower() not in allowed_mimetypes:
+        mimetype = self._compute_mimetype(vals)
+        if mimetype not in allowed_mimetypes:
             message = _("File type '%s' is not allowed.") % mimetype
             if request:
                 request.mimetype_error = message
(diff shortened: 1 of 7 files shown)


attachment_mimetype_restriction/tests/test_attachment_mimetype_restriction.py
```diff
@@ -63,6 +68,20 @@ def test_write_revalidates_on_datas_change(self):
...
+        self.partner.image_1920 = PNG_DATA
+        attachment = self.Attachment.sudo().search(
+            [
+                ("res_model", "=", "res.partner"),
+                ("res_id", "=", self.partner.id),
+                ("res_field", "=", "image_1920"),
+            ]
+        )
+        self.assertEqual(attachment.mimetype, "image/png")
+        self.partner.image_1920 = PNG_DATA
```
yostashiro: Please use different data to confirm the write path works. Using the same data ends up testing nothing as it won't reach ir.attachment.write().

yostashiro (changes requested):

attachment_mimetype_restriction/models/ir_attachment.py
```diff
@@ -28,62 +60,199 @@ def _get_allowed_mimetypes(self, company_id, res_model=None):
...
-        res_model = vals.get("res_model")
-        res_id = vals.get("res_id")
+    def _resolve_attachment_company_id(self, res_model=None, res_id=None):
         if res_model and res_id and res_model in self.env:
             record = self.env[res_model].sudo().browse(res_id).exists()
             if record and "company_id" in record._fields and record.company_id:
                 return record.company_id.id
         return self.env.company.id
 
-    def _validate_mimetype_from_vals(self, vals):
```
yostashiro, line 43: We can't drop a method like this. Please keep it and let it fall back to the successor method.

yostashiro (changes requested):
This PR looks a bit bloated and seems to go beyond the PR description. Can we make this PR more focused on concise?

PR #1898, opened 2026-08-05
[19.0][MIG] fetchmail_thread_default

KKamaa:


Kiplangatdan (approved):

thomaspaulb (approved):

thomaspaulb:
@OCA/social-maintainers

fetchmail_thread_default/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2017 Tecnativa - Jairo Llopis <jairo.llopis@tecnativa.com>
+# Copyright 2023-24 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Default Thread For Unbounded Emails",
+    "summary": "Post unkonwn messages to an existing thread",
```
NL66278, line 6: `s/unkonw/unknown/`

fetchmail_thread_default/models/mail_thread.py
```diff
@@ -0,0 +1,34 @@
...
+    def message_process(
+        self,
+        model,
+        message,
+        custom_values=None,
+        save_original=False,
+        strip_attachments=False,
+        thread_id=None,
+    ):
+        server = self.env["fetchmail.server"].browse(
```
NL66278, line 20: This is only needed when either model or thread_id are not set. So in many cases the browse action will be done for no purpose. Better to split the if condition like so:
```
if not (model or thread_id):
    server = self.env["fetchmail.server"].browse(
        self.env.context.get("default_fetchmail_server_id")
    )
    if server.default_thread_id:
            model = server.default_thread_id._name
            thread_id = server.default_thread_id.id
```
I kept the `not (model or thread_id)` but I wonder whether this is correct. A thread_id with no model does not make sense. But a model with thread_id filled will lead to creation of a new record in the model. We can assume no thread_id is passed without a model, but the following condition should be enough: `if not model`.

NL66278 (commented):
👍 No really blocking issues, but if you could take up my comments, that would be nice.

PR #1902, opened 2026-08-19
[18.0][FIX] mail_gateway_whatsapp: preserve incoming text layout

Adridot:
## Problem

`_process_update` passes the incoming WhatsApp text straight to `message_post`:

    new_message = chat.sudo().message_post(body=body, ...)   # body is a plain str

The body it assembles only ever comes from two Cloud API fields, `messages[].text.body` and the media `caption`, both plain text. `mail_thread._message_post` only escapes a `str` body (`'body': escape(body)`), and `escape()` does not convert `\n`. A multi-line message is therefore stored as `<p>line1\nline2…</p>`, and in HTML a newline is just a space: the message shows up as a single block, the blank lines that separated its sections are gone, and URLs are not clickable.

The same root cause affects the `location` branch of the same function: the `<a href=…>` is built as a `str`, so `escape()` renders it as literal HTML — a shared location shows the source of the link instead of the link.

## Fix

Convert the assembled body (text + media captions) with `plaintext2html` once, before the `location` branch, so that both `message_post` calls of the function receive `Markup`. This mirrors the outgoing path of the same model, which already sends `html2plaintext(body)`. The location anchor is rebuilt as `Markup` with `%` interpolation, which keeps the coordinates coming from the payload escaped.

`plaintext2html` escapes its input, so the protection `escape()` gave is not lost, only moved: text that looks like markup (`<b>`, `<script>`) still shows literally.

No version bump, left to the OCA tooling.

## Tests

One test added to the module's existing suite, on a multi-line incoming message carrying a location: line breaks rendered, blank line kept as a paragraph break, URL turned into a link, markup-looking text left literal, location anchor kept as a link.

Checked on an 18.0 production instance: incoming messages keep their layout and their links, and previously archived ones were reformatted with the same result.

### DIFF
--- a/mail_gateway_whatsapp/tests/test_mail_gateway_whatsapp.py
+++ b/mail_gateway_whatsapp/tests/test_mail_gateway_whatsapp.py
@@ -1,6 +1,7 @@
 # Copyright 2022 CreuBlanca
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import copy
 import hashlib
 import hmac
 import json
@@ -317,6 +318,23 @@ class TestMailGatewayWhatsApp(MailGatewayTestCase):
             message = self.receive_message(self.audio_message)
             # Check that the attachment is marked as voice
             self.assertTrue(message.attachment_ids.voice_ids)
+
+    def test_receive_message_keeps_layout(self):
+        """Text keeps its layout, markup stays literal, location stays a link."""
+        message = copy.deepcopy(self.message_01)
+        content = message["entry"][0]["changes"][0]["value"]["messages"][0]
+        # Plain text, brackets included: what a sender typing "<b>" really sends.
+        content["text"] = {
+            "body": "Line 1\n<b>bold</b>\n\nSection 2 https://example.com/track"
+        }
+        content["location"] = {"latitude": "41.3851", "longitude": "2.1734"}
+        body = self.receive_message(message).body
+        self.assertIn("Line 1<br", body)
+        self.assertIn("</p><p>", body)
+        self.assertIn('href="https://example.com/track"', body)
+        self.assertIn("&lt;b&gt;bold&lt;/b&gt;", body)
+        self.assertIn("41.3851,2.1734", body)
+        self.assertNotIn("&lt;a", body)
 
     @mute_logger("odoo.addons.mail_gateway.controllers.gateway")
     def test_post_no_signature_no_message(self):
--- a/mail_gateway_whatsapp/models/mail_gateway_whatsapp.py
+++ b/mail_gateway_whatsapp/models/mail_gateway_whatsapp.py
@@ -10,11 +10,12 @@ from io import StringIO
 
 import requests
 import requests_toolbelt
+from markupsafe import Markup
 
 from odoo import models
 from odoo.exceptions import UserError
 from odoo.http import request
-from odoo.tools import html2plaintext
+from odoo.tools import html2plaintext, plaintext2html
 
 from odoo.addons.base.models.ir_mail_server import MailDeliveryException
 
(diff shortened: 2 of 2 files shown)


mail_gateway_whatsapp/tests/test_mail_gateway_whatsapp.py
```diff
@@ -318,6 +319,22 @@ def json(self):
             # Check that the attachment is marked as voice
             self.assertTrue(message.attachment_ids.voice_ids)
 
+    def test_receive_message_keeps_layout(self):
+        """Text keeps its layout, markup stays literal, location stays a link."""
+        message = copy.deepcopy(self.message_01)
+        content = message["entry"][0]["changes"][0]["value"]["messages"][0]
+        content["text"] = {
+            "body": "Line 1\n<b>bold</b>\n\nSection 2 https://example.com/track"
```
pedrobaeza, line 328: Isn't this mixed between HTML and non HTML?

pedrobaeza (commented):
Are you sure the message is not HTML and `\n` are just line feeds of the HTML?

What is `mail_gateway_bailey`?

mail_gateway_whatsapp/tests/test_mail_gateway_whatsapp.py
Adridot: The string is plain text on purpose, brackets included: it is what a sender typing `<b>bold</b>` in WhatsApp actually sends — the app has no HTML, its own formatting is `*bold*`/`_italic_`. The assertion on `&lt;b&gt;bold&lt;/b&gt;` is there to show the change does not lose the escaping `message_post` gave, since `plaintext2html` escapes before converting. Comment added on that line to make the intent readable.

Adridot:
@pedrobaeza 

The body this function builds is never HTML: it only ever comes from two Cloud API fields, `messages[].text.body` and the media `caption`, both plain UTF-8 text. WhatsApp has no markup of its own beyond `*bold*`/`_italic_`, so the `\n` are real line feeds typed by the sender, not whitespace inside HTML. The module already relies on that on the way out — `_send_message` posts `html2plaintext(body)` — so `plaintext2html` on the way in is the same assumption, mirrored.

`mail_gateway_baileys` is a module of ours, not one of this repository: it plugs `mail_gateway` into a local Baileys bridge (WhatsApp multi-device, our own number) so that **group** conversations reach Discuss, which the official Cloud API does not deliver. Its incoming path already went through `plaintext2html`, which is how the difference showed up here — but it is not a reference anyone can check from this repo, so the comparison table has been removed from the description and from the commit message. The fix stands on the Cloud API payload alone.

Also fixed in the description: it announced two test cases where there is a single one (a multi-line message carrying a location).

pedrobaeza (approved):
Another doubt: are bold and italic converted to/from? It should be interesting...

Adridot:
@pedrobaeza Neither direction converts them today, and this PR changes nothing on that front — but the outgoing side happens to get bold right by accident.

**Odoo → WhatsApp** (`_send_payload` → `html2plaintext`, untouched here). Odoo's own `html2plaintext` already emits plain-text markers: `<strong>`/`<b>` → `*…*`, `<em>` → `/…/`, `<h1>`/`<h2>` → `**…**`. Running 18.0's helper on a few bodies:

```
'<p>Hello <strong>bold</strong> and <em>italic</em></p>'  -> 'Hello *bold* and /italic/'
'<p>Hello <b>bold</b> and <i>italic</i></p>'              -> 'Hello *bold* and italic'
'<p><s>struck</s></p>'                                    -> 'struck'
'<p><code>mono</code></p>'                                -> 'mono'
'<p><strong style="color:red">attr bold</strong></p>'     -> 'attr bold*'
```

So bold survives, because `*bold*` is exactly WhatsApp's bold syntax. Italic does not: WhatsApp expects `_italic_` and receives `/italic/`, which it shows literally; `<i>`/`<s>`/`<code>` lose their formatting silently. And since the replacement is a plain string match, a tag carrying attributes only matches on the closing side and leaves a stray `*`.

**WhatsApp → Odoo** (what this PR touches): `*bold*`, `_italic_`, `~strike~`, ```` ```mono``` ```` are just characters inside `messages[].text.body`. `plaintext2html` escapes, turns line feeds into `<br/>`/`<p>` and linkifies URLs — nothing else — so they stay literal in Discuss:

```
'*bold* and _italic_ here' -> '<p>*bold* and _italic_ here</p>'
```

That is what `escape()` did before the PR too, so no regression either way.

Making it symmetric would be a small self-contained follow-up: a WhatsApp-markup ↔ HTML pair in the module (`*`↔`<strong>`, `_`↔`<em>`, `~`↔`<s>`, ` ``` `↔`<code>`), applied after `plaintext2html` on the way in and in place of the bare `html2plaintext` on the way out. Happy to open it as a separate PR if you find it worth having — I'd rather keep this one to the layout fix.

pedrobaeza:
Thanks for the analysis. If you can do the followup PR, it would be great.

BhaveshHeliconia (approved):

PR #1908, opened 2026-08-28
[18.0][ADD] mail_gateway_instagram: add DM gateway

ivs-cetmix:
Instagram messages in your chatter.
mail_gateway already routes public webhooks into Discuss channels, but only Telegram and WhatsApp are implemented. This module adds an Instagram Login connector so professional-account DMs become Discuss threads and replies are sent back as Instagram text.

NB: AI assistance was used for development  (why not? 😄 ) however the code was checked manually to ensure it meets the OCA code quality standards.

Task 5591


ivs-cetmix:
@etobella @josep-tecnativa  @jelenapoblet (why not 😄 )

etobella (approved):
Nice work!

I cannot check it functionally, but technically looks Ok. About the limit of 24 hours, something similar applies to whatsapp, so it should make sense to add the functionality on general.

Also, the code is quite similar to whatsapp, it is clear that both of them belong to meta.

ivs-cetmix:
> Also, the code is quite similar to whatsapp, it is clear that both of them belong to meta.
Thank you! I was actually thinking about extracting shared method to a special `mail_gateway_meta` module so it could be used as a shared foundation for all the Meta service implementations. However this is not for now, can be done later.

etobella:
> > Also, the code is quite similar to whatsapp, it is clear that both of them belong to meta.
> > Thank you! I was actually thinking about extracting shared method to a special `mail_gateway_meta` module so it could be used as a shared foundation for all the Meta service implementations. However this is not for now, can be done later.

No need to do this for now.  I just shared my surprise, but for now, I think that the effort is not necessary.

Aldeigja (approved):
Functional LGTM.

ivs-cetmix:
hey @OCA/social-maintainers would appreciate this module review and merge - this will unblock some new features following up 😄

pedrobaeza:
Uhm, curious that if we didn't push anything new during last day, the rate limit in PyPi is exhausted... @sbidoul can you please reserve the name?

sbidoul:
Name reserved.

PR #1909, opened 2026-09-02
[18.0][IMP] mail_gateway_whatsapp: convert the text formatting both ways

Adridot:
Follow-up of #1902, as asked there. It builds on the `plaintext2html` call of that PR, so its commit is part of this branch until it is merged.

WhatsApp clients format text with `*bold*`, `_italic_`, `~strikethrough~` and triple backticks for monospace. Incoming, those markers reached Discuss as literal characters. Outgoing, `html2plaintext` sent `<strong>` as `*bold*` — which WhatsApp reads as bold, by chance — but `<em>` as `/italic/`, shown as is, and dropped `<i>`, `<s>`, `<del>` and `<code>`.

Both ways now share one table. Incoming, markers become html tags after `plaintext2html`, leaving the tags and links it just produced alone, so a url holding `_` or `*` is untouched. Outgoing, the formatting tags become markers again before `html2plaintext`.

A marker only counts when it wraps a non-space run and is not glued to a word, so `snake_case`, `2*3*4` and unclosed markers stay as typed.

One test per direction, through the webhook and through `message_post`.

### DIFF
--- a/mail_gateway_whatsapp/models/mail_gateway_whatsapp.py
+++ b/mail_gateway_whatsapp/models/mail_gateway_whatsapp.py
@@ -10,11 +10,12 @@ from io import StringIO
 
 import requests
 import requests_toolbelt
+from markupsafe import Markup
 
 from odoo import models
 from odoo.exceptions import UserError
 from odoo.http import request
-from odoo.tools import html2plaintext
+from odoo.tools import html2plaintext, plaintext2html
 
 from odoo.addons.base.models.ir_mail_server import MailDeliveryException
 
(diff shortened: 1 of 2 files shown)


Adridot:
One limitation, for the record: a marker does not span a line break, so `*multi line*` stays literal. The conversion runs after `plaintext2html`, on the text between the tags it produced, and a run crossing a `<br/>` or a paragraph is left alone rather than closed over a tag boundary. Single-line runs are the common case; multi-line ones can be added later if it comes up.

pedrobaeza:
Please remove the extra commit belonging to the other PR

Adridot:
Done — rebased on 18.0 now that #1902 is merged; the PR is down to its own commit.

pedrobaeza (approved):

PR #1910, opened 2026-09-03
[17.0][ADD] social_media_sync

edescalona:
@BinhexTeam

T25184

Brings back into Odoo what an account already published on its social media: the posts themselves, the figures each of them collected, and their comments and reactions.

The split from `social_media_base` is a cost line, not a feature line. Base spends a fixed number of calls per account however much that account has published (publishing, account figures, the daily series, and the single call that answers whether the one publication the user is opening is still there). Everything whose cost grows with the history of the account lives here: one call per page of posts, one per publication to check it is still online, one per comment thread. An installation that only writes and publishes never pays for any of it.

`social_media_base` never depends on this module nor calls into it: where base needs something only the synchronization knows how to do, it declares an empty hook and works installed alone.

### What it adds

- **Initial import** of what an account already published, triggered on the spot when the account is linked; the scheduled action *Social: Initial sync of the new accounts* (monthly) is the safety net for the ones still waiting for it.
- **Weekly full resync** (*Social: Full resync of the accounts*), the only pass that notices a publication deleted on the social media, and the most expensive one.
- **Views over the whole history**: list of publications with its search filters and form, the imported figures (impressions, clicks, shares, likes, comments, interactions, engagement) on publications and on the post totals, and the likes and comments the footer of a dashboard card draws.
- **Comment threads read live from the network**, never mirrored into Odoo: writing, answering, deleting a comment, and *Recommend* as a toggle on the publication and on each comment. `/mail/message/post` is overridden for `social.post.account` so a message is published through the connector instead of being stored as a `mail.message`, and the result is pushed to the author's partner channel.
- **Media retention**: the import downloads the medias of the publications it brings in. `social_media_sync.media_max_age_days` (system parameter, `0` = no policy) releases the medias of imported publications older than that; the deletion belongs to the vacuum, a day later. Medias of posts published from Odoo are editorial content and are never aged out.

### Connector contract

This module implements the API of no social media in particular: scheduled actions, frontend and the common interface. A synchronization connector (`social_media_linkedin_sync`, `social_media_x_sync`) fills the hooks on `social.post.account`: `get_comments`, `get_comment_replies`, `create_comment`, `delete_comment`, `action_like_post` / `action_unlike_post`, `action_like_comment` /`action_unlike_comment`. A connector whose API returns the whole thread nested leaves `get_comment_replies` unimplemented, and only what a connector declares is offered in the UI.

A *not found* on a reaction or a comment does not mark the publication deleted on its own: the publication is asked about first, because a network answers the same for an unknown reference and for a lost permission.

Installed without any connector, the scheduled actions run, find nothing to ask anybody, and leave everything as it is.


edescalona:
Hi @pedrobaeza @ivs-cetmix , this PR adds a new module: `social_media_sync`. It is ready for review whenever you have time. I look forward to any questions or changes you deem necessary.

jelenapoblet:
@edescalona I think the issue with the dashboard not updating is coming from this module. I’m sure you already know it, but since the Campaign card/footer is handled here, maybe the issue is somewhere around that part.

jelenapoblet (approved):
LGTM

social_media_sync/controllers/thread.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        partner = request.env.user.partner_id
+        return {
+            "author": partner.mail_partner_format(
+                {"id": True, "name": True, "is_company": True, "user": {}}
+            )[partner]
+        }
+
+    @route("/mail/message/post", methods=["POST"], type="json", auth="public")
+    @add_guest_to_context
```
rrebollo:
```suggestion
    @route()
```

social_media_sync/models/social_account.py
```diff
@@ -0,0 +1,533 @@
...
+        if not self:
+            accounts = accounts._accounts_to_import()
+            if not accounts:
+                return []
+        imported = set()
+        statistics = accounts._update_posts_statistics(post_id, domain, imported)
+        imported_accounts = accounts.filtered(lambda account: account.id in imported)
+        pending = imported_accounts.filtered("pending_initial_sync")
+        if pending:
+            pending.sudo().write({"pending_initial_sync": False})
```
rrebollo:
```suggestion
            pending.sudo().pending_initial_sync = False
```

social_media_sync/models/social_account.py
```diff
@@ -0,0 +1,533 @@
...
+        The cron is asked for a moment ahead instead of for right now. The
+        caller is in the middle of the association, which writes the very row
+        the import writes its statistics on, and a cron woken inside that
+        transaction starts on a snapshot older than its commit: it reaches the
+        account, loses the race and leaves the dashboard announcing an import
+        that only the retry a few minutes later brings in.
+        """
+        if not self:
+            return
+        self.sudo().write({"pending_initial_sync": True})
```
rrebollo:
```suggestion
        self.sudo().pending_initial_sync = True
```

social_media_sync/models/social_post_account.py
```diff
@@ -0,0 +1,445 @@
...
+        ``media_refs`` is kept: it is what tells the next synchronization
+        pass that this publication already had that media, and dropping it
+        would only buy the same download again.
+
+        :param limit: how many publications one pass reaches at most.
+        """
+        days = self._media_retention_days()
+        if days <= 0:
+            return
+        lines = self.sudo().search(self._media_retention_domain(days), limit=limit)
```
szalatyzuzanna: The retention domain keeps matching publications whose medias were already released (media_refs is kept and nothing filters on the remaining attachments), and with no _order the search is by id, so once more than limit publications are past the policy every pass picks the same first ones, skips them all and never reaches the rest. Could the domain only return lines that still hold downloaded medias?

social_media_sync/models/social_post_account.py
```diff
@@ -0,0 +1,445 @@
...
+        call that should never have been made gets.
+
+        :param comment_ref: reference of the comment on the social media, as
+            ``get_comments`` answered it.
+        :return: ``success`` and, on a failure, the ``message`` to show.
+        :rtype: dict
+        """
+        return {"success": False}
+
+    def create_comment(self, post_data, context=None):
```
szalatyzuzanna: Unlike the other hooks, create_comment returns None, so the controller broadcasts a payload without success or message and the dialog shows "Comment created" as a danger notification; could it return the same failure shape as its siblings, and the client avoid that default text when success is not true?

social_media_sync/models/social_post_account.py
```diff
@@ -0,0 +1,445 @@
...
+        had run, and the reference would be lost.
+
+        The ``url`` the media is downloaded from is required: the bridges
+        skip a media the social media reported without one instead of asking
+        for it here.
+
+        :return: the attachment created, or an empty recordset on failure.
+        :rtype: odoo.models.Model
+        """
+        Attachment = self.env["ir.attachment"]
```
szalatyzuzanna: The url key is passed straight to ir.attachment.create(), so a binary attachment ends up carrying the remote (often signed) CDN URL in its url field; could it be used only for the download? It would also be worth guarding the size of what is downloaded, since the whole content is loaded in memory inside the import transaction.

social_media_sync/models/social_account.py
```diff
@@ -0,0 +1,533 @@
...
+
+        The cron record does not set a user, so the search needs ``sudo()``
+        to reach the pending accounts of every responsible, like
+        :meth:`_run_check_media_updates` does.
+        """
+        pending = self.sudo().search([("pending_initial_sync", "=", True)])
+        postponed = self.browse()
+        for account in pending:
+            error = None
+            try:
```
szalatyzuzanna: Both savepoints share one try, so when the import succeeded and only _backfill_statistics() fails, the account gets the "posts could not be imported" note although they were imported, and on a concurrency error it is postponed without _close_initial_sync() while its pending_initial_sync is already cleared, so the retry never runs the backfill again. Could the backfill failure be handled on its own?

szalatyzuzanna (commented):
Thanks a lot for this contribution, @edescalona. Code review only, a few points below.

PR #1911, opened 2026-09-03
[18.0][IMP] mail_gateway_whatsapp: store messages answered outside Odoo

Neitherkx:
When using WhatsApp Business App alongside the WhatsApp Cloud API, messages sent by business users from the WhatsApp Business App or linked devices are delivered by Meta through the smb_message_echoes webhook.

Currently, mail_gateway_whatsapp only processes messages webhooks, so messages answered outside Odoo are not stored in the corresponding Odoo channel.

This change adds support for smb_message_echoes webhooks:
- Process messages received through smb_message_echoes.
- Resolve the WhatsApp channel using the echo recipient (to).
- Store echoed messages as outgoing messages in the Odoo channel.
- Use the gateway company partner as the author, since Meta does not provide information about which business user sent the message.
- Reuse the existing message content handling for text, attachments and locations.
- Store the WhatsApp message ID (wamid) in the notification to prevent duplicate messages and allow proper message threading.
- Create a guest contact from the phone number when the echo does not contain contact information.

Messages originating from Odoo or re-delivered by Meta are ignored when their wamid has already been stored, preventing duplicate messages.

No outgoing WhatsApp messages are sent as a result of processing an echo.


pedrobaeza (changes requested):
For me this should be a separate module for those wanting this, as this complicates the base module for no gain for those not wanting the feature.

Neitherkx:
> For me this should be a separate module for those wanting this, as this complicates the base module for no gain for those not wanting the feature.

 Move to a new standalone module #1912

PR #1914, opened 2026-09-07
[18.0][REF] mail_gateway: send gateway channels from _to_store()

JuanJaraSandoval:
Store.one_id() builds the reference to a record rather than its content, and its docstring asks callers to avoid it. Sending the gateway channels from a res.partner._to_store() override instead, the way hr_holidays and im_livechat extend partner store data, removes the need for the registry guard added in #1913: a model override only exists in the registries where its module is installed.

_thread_to_store() read the channels back from the follower references and passed them on as gateway_followers without ever adding the partners to the store, so it relates them with Store.many() now.


JuanJaraSandoval:
@pedrobaeza @CarlosRoca13 this is the follow-up to #1913 you suggested there: the monkey-patch is replaced with a res.partner._to_store() override, so the guard added in #1913 is no longer needed.

pedrobaeza:
When @CarlosRoca13 comes from holidays, will check this.

CarlosRoca13 (approved):

pedrobaeza (approved):
/ocabot merge patch

JuanJaraSandoval:
@pedrobaeza He visto que el bot en la primera pasada abortó el merge por errores de pre-commit. ¿Qué ha pasado? ¿Puedes indicármelo por si es error mío solucionarlo para futuras pull request?
Gracias de antemano

pedrobaeza:
In this repo, we should talk in English. You did the PR over an outdated branch, so pre-commit was incorrect, as other PR added the same import. I have rebase and forced push over your branch to fix it.

JuanJaraSandoval:
Thanks for the information.
I'll keep it in mind for future pull requests.

PR #1916, opened 2026-09-10
[17.0][ADD] social_media_linkedin_sync

edescalona:
@BinhexTeam

T25184

## What it does

LinkedIn half of `social_media_sync`. It brings back into Odoo what a LinkedIn page already published: the posts, the figures each of them collected, and their comments and reactions.

The split against `social_media_linkedin` is a cost line, not a feature line. The connector asks LinkedIn for a fixed number of things per account — publish, delete, the daily figures of the whole page — and that number does not change whether the page published once or ten thousand times. Everything whose cost grows with the history of the page lives here: reading the feed one page at a time, asking about missing publications a hundred URNs at a time, one call per comment thread. An Odoo that only publishes installs `social_media_linkedin` alone and pays for none of it.

`auto_install`: the choice to synchronize was already made by whoever installed `social_media_sync`; this only keeps a LinkedIn page from being left without the half that belongs to it.

## Features

- **Import** of publications and their statistics, on demand and from   the scheduled actions of `social_media_sync`. Three discovery paths: one publication by reference, the whole feed, or the first page sorted by last modification (the ordinary one). Medias created outside Odoo are downloaded; medias removed on LinkedIn are dropped from the card, and files attached by hand in Odoo are never touched.
- **Full resync**, the only pass that reports a publication deleted on LinkedIn. Missing from the feed is not proof: each suspect is asked for by URN, one call per hundred, and only a `404` writes the deletion. The Posts API publishes asynchronously, answers the author finder in reader context and documents that a short page is not the end of the feed, so a live publication can be absent from a listing read whole.
- **Deletions caught from the dashboard**: a `404` on a reaction or a comment makes Odoo ask about the publication itself before marking the line `deleted`.
- **Comments and reactions**: threads with their replies, comment creation and deletion under the organization, and *Recommend* on the publication and on each comment, which also withdraws the reaction.
- **Update check** every two hours: at most two calls per account, comparing the daily figures of the page against the checkpoint left by the last import, reusing the buckets the connector read on the same pass. It tells that something moved, never which publication.

## Quota

LinkedIn throttles per day, per application and per endpoint, so what matters is the spread:

| Pass | Cost |
| --- | --- |
| Update check (2 h) | 2 calls per account and run, whatever the number of publications |
| *Update* button | 1 feed call + the statistics reads of the connector |
| *Full resync* / weekly cron | 1 call per 100 publications, per account |

`social_media_linkedin_sync.posts_max_pages` (default 50, bounded 1..500) is how far one pass reads into the feed. A page with a longer history raises it; a feed read short skips the deletion sweep for that run rather than marking publications that are still online.

## Scopes

Nothing new is requested: reading the feed, its comments and its reactions travels on `r_organization_social`, which
`social_media_linkedin` already asks for. But a token keeps the scopes it was issued with, so an account associated before that permission was requested has to be authorized again (*Update account* with *Update keys*). Those accounts say so themselves: warning on the form, notification from the bihourly check, and an import that refuses naming the permission.

## Known limits

Written up in `readme/ROADMAP.md`. The ones worth knowing before reviewing:

- **A commenter cannot be named.** The comments endpoint identifies the author by URN and no public endpoint turns a `urn:li:person:` into a name. Organizations are resolved with `GET /v2/organizations/{id}`; everything else is drawn as *LinkedIn member* / *LinkedIn page*. A comment is never signed with the name of the page that did not write
  it.
- **No media in comments.** The versioned endpoint that would take it answers `403 partnerApiSocialActions.CREATE`; the `/v2` one the connector uses answers `500` with `content` and publishes without it.
- **Only `LIKE`.** The Reactions API also takes `PRAISE`, `EMPATHY` and the rest, with the same permissions; offering them is a UI decision, not an API one.
- **No incremental statistics.** The `/posts` finder has no `since`, per-share statistics cannot be time-bound, and the daily figures of the page do not say which publication moved. The engagement webhook would, but it reports no impressions or clicks, only the webhook payload carries the matching `urn:li:share:`, and it needs a public HTTPS endpoint plus a subscription per organization.
- The weekly full resync walks every account in the same run instead of spreading them over the days of the week.

## Tests

`social_media_linkedin_sync/tests/` — account sync, actor resolution, update check, post import, scopes, comment dialog, post reaction and the card footer, this last one driving the JS tour `social_media_linkedin_sync.card_footer`. No network calls: every HTTP call is patched through `_request_linkedin`.

### Depends on

#1910 `social_media_sync`
#1890 `social_media_linkedin`


edescalona:
Hi @pedrobaeza @ivs-cetmix, this PR adds a new module: social_media_linkedin_sync. It is ready for review whenever you have time. I look forward to any questions or changes you deem necessary.

jelenapoblet:
I found one issue during the functional review.

I published a post and then interacted with it directly on LinkedIn with a like and a comment. After syncing, the new engagement is correctly reflected in the dashboard and in the Statistics view.

However, the Campaign view is not updated and still shows Engagement: 0.00 for the same post.

I would expect the campaign to reflect the synchronized engagement as well, so the metrics remain consistent across the different views.

edescalona:
Hello @jelenapoblet , thanks for your comment; I'll take a look and get back to you.

edescalona:
Hi @jelenapoblet, thanks for the review.

**What I notice is how LinkedIn defines the metric.** Its documentation
describes the field as follows:

> `engagement` | double | Number of organic clicks, likes, comments, and shares
> divided by impressions.

It is a rate calculated *based on impressions*. Your "like" and comment were recorded, but the post didn't have any impressions yet; therefore, there is no denominator, and LinkedIn returns zero.

However, posts store *engagement* to four decimal places, whereas the aggregated campaign fields (`social_engagement` and `count_post_engagement`) are simple floating-point numbers; consequently, an actual rate of 0.0303 would appear as 0.03, and any value below 0.005 would appear as 0.00. That issue lies in `social_media_base`, not in this module.

I am going to open a separate Pull Request (PR) to fix it and will link it here. For clarity, that fix won't change the figure you saw, since in your case the value was indeed zero. It is relevant for posts that do generate impressions.

edescalona:
Hi @jelenapoblet , this PR (https://github.com/OCA/social/pull/1924) contains the change to display engagement values ​​with 4 decimal places. Could you please review it? Thanks.

jelenapoblet (approved):
Now its working!

PR #1917, opened 2026-09-10
[17.0][ADD] social_media_x_sync

edescalona:
@BinhexTeam

T25184

Adds `social_media_x_sync`, the X half of `social_media_sync`. Auto-installed when `social_media_sync` and `social_media_x` are both present, and it adds no Python dependency of its own — it asks the connector for the tweepy client.

### Why a separate module

The split from `social_media_x` is a cost line, not a feature line. The connector spends a fixed number of calls per account however much that account has published: link it, refresh its card, publish, delete, read the figures of
the publications of the last 30 days, read the single publication being opened.

Everything whose cost grows with the history of the account lives here — the timeline, one download per media not stored yet, and up to five calls per publication whose comments are read. An Odoo that only publishes on X installs
`social_media_x` alone and pays for none of it.

### What it adds

- **Timeline import** with [user posts](https://docs.x.com/x-api/posts/user-posts-timeline), up to the 100 publications the endpoint answers in one request (`next_token` is not followed). Retweets, replies and quotes are discarded; media with a direct address are downloaded once and stored as attachments.
- **Account statistics** as the sum of every publication stored for the account, with no window and no ceiling. Engagement is derived from impressions and interactions, since X reports no rate of its own. A publication deleted on X is marked *Deleted* and keeps the figures of its last read.
- **Comments** with [recent search](https://docs.x.com/x-api/posts/recent-search), which only covers the last 7 days. The conversation is walked a hundred replies at a time up to five pages; read to the end, the reply count of each
  comment is computed in Odoo, and cut short by the ceiling no number is stated. Replies may carry images.
- **Enable since / Post since**, so an account does not have its whole history read again on every import.
- **With Reposts / With Quotes** filters on the dashboard search panel.
- Per-endpoint **rate-limit windows** on the account. While a window lasts, reading the conversation shows a notice instead of an empty thread, and publishing a comment reports the limit instead of a reply X never received.

### Scheduled actions

The passes of `social_media_sync` plus two of `social_media_base`:

| Action | Schedule | On X |
| --- | --- | --- |
| Checking social media updates | 2 h | imports the timeline — the only endpoint that knows whether anything moved is the import itself |
| Initial sync of the new accounts | monthly (and on linking) | first import of the timeline |
| Full resync | weekly | same read as the ordinary import, so it notices no remote deletion |
| Refresh statistics of recent publications | daily | call of `social_media_x`, not of this module |

An account whose first import has not run yet is skipped by the bihourly check: that check and the initial import write the same row from two threads.

### Not implemented

- Recommending a publication or a comment. X serves `POST /2/users/:id/likes` and `DELETE /2/users/:id/likes/:tweet_id` and a comment is a post like any other, so both would be the same call; the button is not rendered for X publications until they are implemented.
- Deleting a comment, although it is the same `DELETE /2/tweets/:id` already used for publications.
- Pagination of the timeline beyond the first page, and conversations longer than 500 replies.
- No migration script: `last_post_ref`, `enable_since` and `post_since_id` move from `social_media_x` to this module, so an existing database needs `ir_model_fields` updated before uninstalling the connector, or a recreation.

### Requirements

A paid X API plan, the same one the connector needs: the user timeline and the recent search are API v2 endpoints and answer `403` outside a Project.

### Tests

`social_media_x_sync/tests/` — `test_sync_account_x.py`, `test_sync_post_account_x.py` and `test_card_footer_x.py` (driving the `social_media_x_sync.card_footer` tour). 
Every HTTP call is patched; no network access.

## Depends on

#1910  `social_media_sync`
#1891 `social_media_x`


edescalona:
Hi @pedrobaeza @ivs-cetmix, this PR adds a new module: social_media_x_sync. It is ready for review whenever you have time. I look forward to any questions or changes you deem necessary.

PR #1922, opened 2026-09-14
[16.0][FIX] mail_activity_team: do not refetch activities on every document click

baguenth:
## Problem

The core `ActivityMenuView` registers a capture-phase click listener on the
document that calls `close()` for every click outside the menu, whether the
dropdown is open or not.

The `close()` override of this module always re-activates the *my activities*
filter, which ends in `fetchData()` and therefore issues a
`systray_get_activities` RPC on **every click anywhere in the web client**.
In a normal working session this adds up to thousands of useless requests,
each running the activity aggregation query on the server.

## Fix

Only reset the filter (and refetch) when the menu was actually open. Behaviour
of the menu itself is unchanged.


baguenth:
Hey @StefanRijnhart, @hbrunn, @simahawk,

kindly asking for a review :)

QuatschSalat (approved):

sebinthomasst (approved):

BhaveshHeliconia (approved):

baguenth:
Hey @StefanRijnhart, @hbrunn, @simahawk,

since this is a straight forward fix, any chance that someone can take a look shortterm?

StefanRijnhart (approved):
Thanks for the fix! Functional review: before the fix, systray_get_activities gets called whenever clicking anywhere on the page. After the fix, this only happens when opening the Activities widget, switching to team activities or closing the widget. 17.0 does not seem to exhibit the same issue so no forward ports needed.

/ocabot merge minor

PR #1923, opened 2026-09-14
[16.0][FIX] mail_activity_board: cache the related model selection

baguenth:
## Problem

`_selection_related_model_instance` searches `ir.model` and reads the name of
every activity-capable model. The core `Reference` field calls the selection
function **once per record** while validating the computed value, so every
page of the activity list ran this search once per row.

Reading `related_model_instance` for 80 activities:

|        | queries | time  |
|--------|--------:|------:|
| before |     162 | 31 ms |
| after  |      81 |  7 ms |

## Fix

The result only depends on the registry and on the language of the model
names (`ir.model.name` is translatable), so cache it with `ormcache` keyed by
`env.lang`.

Cache invalidation is covered by the framework: every path that changes the
set of activity-capable models (module install/upgrade/uninstall, `ir.model`
create/write/unlink for custom models) goes through a registry reload or
`setup_models()`, both of which clear the ormcache. Odoo does not allow
switching `is_mail_activity` off on a custom model, so a model never drops out
of the list without such a reload.


baguenth:
Hey @StefanRijnhart, @hbrunn, @simahawk,

kindly asking for a review :)

QuatschSalat (approved):

sebinthomasst (approved):

BhaveshHeliconia (approved):

simahawk (approved):
LG

PR #1924, opened 2026-09-15
[17.0][FIX] social_media_base

edescalona:
@BinhexTeam

The engagement of a publication is whatever the social media reports for it. A social media expressing it as a ratio over the impressions, as LinkedIn does, answers figures of four decimals for an account of ordinary size, and `social.post.account.engagement` holds them with that precision:

```python
engagement = fields.Float(default=0, digits=(16, 4))
```

The two fields that aggregate it do not declare any precision, so they fall back to two decimals:

- utm.campaign.social_engagement, shown on the Social Media page of a campaign
- social.post.count_post_engagement, shown on the card of a post right next to it

A genuine rate of 0.0303 therefore reads as 0.03 on both, and anything below 0.005 as no engagement at all, even though the figure behind it is not zero.

### What this does

Declares `digits=(16, 4)` on both aggregated fields, so they keep the precision of the publications they add up.

Nothing else changes: the engagement is still totalled and not averaged, which is the existing behaviour and is covered by
test_social_engagement_is_a_total_and_not_an_average.

A test covering this is added to social_media_base/tests/test_utm_campaign.py.


BhaveshHeliconia (approved):

jelenapoblet (approved):
Tested again and the engagement is now working in the campaign view.

